PR #1, opened 2020-06-04
Add Velero supporting CRDs

shubham-pampattiwar:
Added setup task for addition of supporting velero CRDs


dymurray:
We need to figure out travis running `kind` so not exposing OCP APIs.

dymurray:
Will follow up in a different PR.

PR #2, opened 2020-06-05
Add setup velero controller task

shubham-pampattiwar:
Added the setup velero controller task using the template velero.yml.j2
Added cloud secret creation tasks for AWS, GCP and Azure cloud credentials
Added default vars used in the template velero.yml.j2

### DIFF
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -42,7 +42,7 @@ spec:
             - name: VELERO_REPO
               value: velero
             - name: VELERO_PLUGIN_REPO
-              value: migration-plugin   #TODO: Ask if this needs to be changed or not
+              value: openshift-velero-plugin   #TODO: Ask if this needs to be changed or not
             - name: VELERO_RESTIC_RESTORE_HELPER_REPO
               value: velero-restic-restore-helper
             - name: VELERO_AWS_PLUGIN_REPO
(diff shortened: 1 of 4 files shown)


deploy/operator.yaml
```diff
@@ -35,6 +35,34 @@ spec:
...
             - name: ANSIBLE_GATHERING
               value: explicit
+            - name: REGISTRY
+              value: quay.io
+            - name: PROJECT
+              value: konveyor
+            - name: VELERO_REPO
+              value: velero
+            - name: VELERO_PLUGIN_REPO
+              value: migration-plugin   #TODO: Ask if this needs to be changed or not
```
dymurray, line 45: Let's leave this here for now, and remove it as an initcontainer since this will change.

roles/velero/templates/velero.yml.j2
```diff
@@ -0,0 +1,278 @@
...
+          secret:
+            secretName: {{ velero_azure_secret_name }}
+        - name: plugins
+          emptyDir: {}
+        - name: scratch
+          emptyDir: {}
+        - name: certs
+          emptyDir: {}
+      initContainers:
+        - image: {{ velero_plugin_image }}:{{ velero_plugin_version }}
```
dymurray, line 108: Take this initContainer out.

deploy/operator.yaml
shubham-pampattiwar: Updated the value !

PR #4, opened 2020-06-09
enable configuration for velero plugins

shubham-pampattiwar:
Add two specs `default_velero_plugins` and `custom_velero_plugins` in Velero CR spec to facilitate plugin customization and configuration.

### DIFF
--- a/deploy/crds/konveyor.openshift.io_v1alpha1_velero_cr.yaml
+++ b/deploy/crds/konveyor.openshift.io_v1alpha1_velero_cr.yaml
@@ -8,6 +8,3 @@ spec:
   - azure
   - gcp
   - aws
-  custom_velero_plugins:
-  - name: custom-shubham-plugin
-    image: quay.io/spampatt/custom-velero-plugin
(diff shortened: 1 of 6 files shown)


deploy/crds/konveyor.openshift.io_v1alpha1_velero_cr.yaml
```diff
@@ -4,3 +4,10 @@ metadata:
   name: example-velero
 spec:
   # Add fields here
+  default_velero_plugins:
+  - azure
+  - gcp
+  - aws
+  custom_velero_plugins:
+  - name: custom-shubham-plugin
```
dymurray, line 12: Let's remove this for now from the default CR.

deploy/operator.yaml
```diff
@@ -17,7 +17,7 @@ spec:
       containers:
         - name: oadp-operator
           # Replace this with the built image name
-          image: "REPLACE_IMAGE"
+          image: "quay.io/spampatt/oadp-operator"
```
dymurray, line 20: We can leave this for now. I will work to get us autobuilds in quay tied to github and we can use that image here.

deploy/crds/konveyor.openshift.io_v1alpha1_velero_cr.yaml
shubham-pampattiwar: Updated the code.

deploy/operator.yaml
shubham-pampattiwar: Done

PR #10, opened 2020-06-15
expose bsl for aws and gcp

shubham-pampattiwar:
- Added BSL templates for Azure and GCP
- Added tasks to create and validate BSLs for Azure and GCP
- Updated readme

### DIFF
--- a/roles/velero/tasks/main.yml
+++ b/roles/velero/tasks/main.yml
@@ -78,19 +78,19 @@
     state: "{{ velero_state }}"
     definition: "{{ lookup('template', 'velero.yml.j2') }}"
 
-- name: "Create BackupStorageLocations if aws is configured for velero"
+- name: "Create AWS BackupStorageLocation"
   k8s:
     state: "{{ velero_state }}"
     definition: "{{ lookup('template', 'velero_bsl_aws.yml.j2') }}"
   when: velero_aws_bsl_configs | int != 0
 
-- name: "Create BackupStorageLocations if gcp is configured for velero"
+- name: "Create GCP BackupStorageLocation"
   k8s:
     state: "{{ velero_state }}"
     definition: "{{ lookup('template', 'velero_bsl_gcp.yml.j2') }}"
   when: velero_gcp_bsl_configs | int != 0
 
-- name: "Create BackupStorageLocations if azure is configured for velero"
+- name: "Create Azure BackupStorageLocation"
   k8s:
     state: "{{ velero_state }}"
     definition: "{{ lookup('template', 'velero_bsl_azure.yml.j2') }}"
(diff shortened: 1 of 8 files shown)


roles/velero/tasks/main.yml
```diff
@@ -78,11 +78,24 @@
     state: "{{ velero_state }}"
     definition: "{{ lookup('template', 'velero.yml.j2') }}"
 
-- name: "Create BackupStorageLocations"
+- name: "Create BackupStorageLocations if aws is configured for velero"
```
dymurray, line 81: NIT: change log `Create AWS BackupStorageLocation`

roles/velero/tasks/main.yml
```diff
@@ -78,11 +78,24 @@
...
   k8s:
     state: "{{ velero_state }}"
-    definition: "{{ lookup('template', 'velero_bsl.yml.j2') }}"
+    definition: "{{ lookup('template', 'velero_bsl_aws.yml.j2') }}"
+  when: velero_aws_bsl_configs | int != 0
 
+- name: "Create BackupStorageLocations if gcp is configured for velero"
```
dymurray, line 87: NIT: Create GCP BSL...

roles/velero/templates/velero_bsl_gcp.yml.j2
```diff
@@ -0,0 +1,33 @@
...
+apiVersion: velero.io/v1
+kind: BackupStorageLocation
+metadata:
+  name: {{ bsl.name }}
+  namespace: {{ velero_namespace }}
+spec:
+  provider: {{ bsl.provider }}
+  objectStorage:
+    bucket: {{ bsl.object_storage.bucket }}
+{% if bsl.object_storage.prefix is defined %}
```
dymurray, line 13: Just curious, do we need all of these `if` checks? What happens if we just specify them and the fields don't exist? Does it break or do we just set empty values?

shubham-pampattiwar: The creation of the resource using the template breaks. For instance, if you are configuring BSL via AWS and lets assume that the end user did not give any `config` related props in the CR as well as there are no `if defined` checks for around `config` in the template, the resource creation will fail, it will state that the `dict object` has no attribute `config` and this would result in failure of the ansible task.

roles/velero/tasks/main.yml
shubham-pampattiwar: Done, Updated the log.

roles/velero/tasks/main.yml
shubham-pampattiwar: Done.

roles/velero/templates/velero_bsl_gcp.yml.j2
dymurray: Awesome, thanks for the info.

PR #16, opened 2020-06-16
enable usage of upstream images

shubham-pampattiwar:
- Added changes to enable usage of upstream images for velero and its plugins
- documented changes in readme


shubham-pampattiwar:
Issue fix https://github.com/konveyor/oadp-operator/issues/15

PR #19, opened 2020-06-17
enable customisation of cpu and mem res for velero pod

shubham-pampattiwar:
- Added changes to enable customization of cpu and memory requests and limits for the Velero pod
- Updated readme for the same


shubham-pampattiwar:
task -Customize resource requests + limits

shubham-pampattiwar:
Added res allocation changes for restic pods as well.

PR #20, opened 2020-06-26
Enable usage of self-signed certs

shubham-pampattiwar:
- changes to enable the usage of self-signed certs
- added validation to check whether the feeded `ca_cert` is in base64 format
- Modifications to the bsl templates to respect `ca_cert` spec
- Modifications to the bsl CRD to respect `ca_cert` spec
- `registry` and `project` path changes for `openshift-velero-plugin`
- Updated Readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -269,6 +269,11 @@ restic_resource_allocation:
       memory: 128Mi
   ```
 
+### Use Velero with a storage provider secured by a self-signed certificate
+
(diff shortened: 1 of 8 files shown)


shubham-pampattiwar:
Use storage provider with self-signed certificate issue fix  https://github.com/konveyor/oadp-operator/issues/12

PR #22, opened 2020-07-06
add changes to deploy aws registry

shubham-pampattiwar:
- Added template to create the aws registry deployment, service and route
- Added tasks to discover, decode and store aws creds
- Updated the operator tasks to include creation of aws image registry using template
- Added template for config map 
- Added task to create the registry configMap resource
- Added registry templates for gcp and azure
- Updated the deprecated k8s_facts ansible module to k8s_info module

### DIFF
--- a/roles/velero/defaults/main.yml
+++ b/roles/velero/defaults/main.yml
@@ -2,6 +2,8 @@
 aws_secret_split: []
 aws_access_key_id: ''
 aws_secret_access_key: ''
+azure_account_key: ''
+gcp_secret_key_file: ''
 backup_storage_locations: []
 custom_velero_plugins: []
 http_proxy: "{{ lookup( 'env', 'HTTP_PROXY') }}"
--- a/roles/velero/defaults/main.yml
+++ b/roles/velero/defaults/main.yml
@@ -1,9 +1,4 @@
 ---
-aws_secret_split: []
-aws_access_key_id: ''
-aws_secret_access_key: ''
-azure_account_key: ''
-gcp_secret_key_file: ''
 backup_storage_locations: []
 custom_velero_plugins: []
 http_proxy: "{{ lookup( 'env', 'HTTP_PROXY') }}"
--- a/roles/velero/defaults/main.yml
+++ b/roles/velero/defaults/main.yml
@@ -1,4 +1,7 @@
 ---
+aws_secret_split: []
+aws_access_key_id: ''
+aws_secret_access_key: ''
 backup_storage_locations: []
 custom_velero_plugins: []
 http_proxy: "{{ lookup( 'env', 'HTTP_PROXY') }}"
--- a/roles/velero/templates/registry_aws.yml.j2
+++ b/roles/velero/templates/registry_aws.yml.j2
@@ -1,0 +1,84 @@
+---
+{% for bsl in backup_storage_locations %}
+{% if bsl.provider == 'aws' %}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: {{ velero_namespace }}
+  name: aws-registry
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      component: registry
+  template:
+    metadata:
+      labels:
+        component: registry
+    spec:
+      restartPolicy: Always
+      containers:
+      - image: quay.io/konveyor/registry:latest
+        name: aws-registry-container
+        ports:
+        - containerPort: 5000
+          protocol: TCP
+        env:
+          - name: REGISTRY_STORAGE
+            value: s3
+          - name: REGISTRY_STORAGE_S3_ACCESSKEY
+            value: "{{ aws_access_key_id }}"
+          - name: REGISTRY_STORAGE_S3_BUCKET
+            value: "{{ bsl.object_storage.bucket }}"
+          - name: REGISTRY_STORAGE_S3_REGION
+            value: "{{ bsl.config.region }}"
+          - name: REGISTRY_STORAGE_S3_SECRETKEY
+            value: "{{ aws_secret_access_key }}"
+{% if bsl.config.s_3__url is defined %}
+          - name: REGISTRY_STORAGE_S3_REGIONENDPOINT
+            value: {{ bsl.config.s_3__url }}
+{% endif %}
+{% if bsl.config.root_directory is defined %}
+          - name: REGISTRY_STORAGE_S3_ROOTDIRECTORY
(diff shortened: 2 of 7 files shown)


roles/velero/defaults/main.yml
```diff
@@ -1,4 +1,9 @@
 ---
+aws_secret_split: []
+aws_access_key_id: ''
+aws_secret_access_key: ''
```
dymurray, line 4: Not sure I would even bother putting these in defaults. We set them with `set_fact` so don't need a default.

roles/velero/templates/registry_aws.yml.j2
```diff
@@ -0,0 +1,84 @@
...
+            value: "{{ aws_access_key_id }}"
+          - name: REGISTRY_STORAGE_S3_BUCKET
+            value: "{{ bsl.object_storage.bucket }}"
+          - name: REGISTRY_STORAGE_S3_REGION
+            value: "{{ bsl.config.region }}"
+          - name: REGISTRY_STORAGE_S3_SECRETKEY
+            value: "{{ aws_secret_access_key }}"
+{% if bsl.config.s_3__url is defined %}
+          - name: REGISTRY_STORAGE_S3_REGIONENDPOINT
+            value: {{ bsl.config.s_3__url }}
```
dymurray, line 39: Is this right?

roles/velero/templates/registry_aws.yml.j2
```diff
@@ -0,0 +1,84 @@
...
+          - name: REGISTRY_STORAGE_S3_SKIPVERIFY
+            value: "{{ bsl.config.insecure_skip_tls_verify | string }}"
+{% endif %}
+
+---
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    component: aws-registry
```
dymurray, line 55: NIT: probably good to rename this `oadp-aws-registry`

roles/velero/templates/registry_azure.yml.j2
```diff
@@ -0,0 +1,70 @@
+---
+{% for bsl in backup_storage_locations %}
+{% if bsl.provider == 'azure' %}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  namespace: {{ velero_namespace }}
+  name: azure-registry
```
dymurray, line 8: `oadp-azure-registry`

roles/velero/templates/registry_gcp.yml.j2
```diff
@@ -0,0 +1,72 @@
...
+    port: 5000
+    targetPort: 5000
+    protocol: TCP
+  selector:
+    component: gcp-registry
+  type: ClusterIP
+
+---
+apiVersion: route.openshift.io/v1
+kind: Route
```
dymurray, line 58: I'm not sure we need new route definitions for AWS+GCP+Azure. The svc + route objects are identical no matter which cloud provider is being used. We should just rename these into separate templates `registry_svc_and_route.yml.j2`. This way if we ever have to change them we only have to update one spot.

roles/velero/templates/registry_gcp.yml.j2
```diff
@@ -0,0 +1,72 @@
...
+    spec:
+      restartPolicy: Always
+      containers:
+      - image: quay.io/konveyor/registry:latest
+        name: gcp-registry-container
+        ports:
+        - containerPort: 5000
+          protocol: TCP
+        env:
+          - name: REGISTRY_STORAGE
```
dymurray, line 27: Same comment as below. Let's use one template and then only set the GCP env vars if we determine its a GCP registry.

roles/velero/templates/registry_config_map.yml.j2
```diff
@@ -0,0 +1,18 @@
+---
+{% for bsl in backup_storage_locations %}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: registry-config
+  namespace: {{ velero_namespace }}
+data:
+{% if bsl.provider == 'aws' and velero_aws_bsl_configs | int != 0 %}
```
dymurray, line 9: I'm worried this won't be enough to support multiple BSLs that use AWS. We cannot use a hardcoded name for the route, will likely need to let Kubernetes generate the route name and register it to be stored in the configmap.

roles/velero/templates/registry_azure.yml.j2
```diff
@@ -0,0 +1,70 @@
...
+    protocol: TCP
+  selector:
+    component: azure-registry
+  type: ClusterIP
+
+---
+apiVersion: route.openshift.io/v1
+kind: Route
+metadata:
+  name: azure-registry-route
```
dymurray, line 58: `generateName: oadp-<bsl_name>-registry-`

dymurray: In ansible task, `register: registry_route`

Pass this name into the confiigmap

roles/velero/templates/registry_aws.yml.j2
dymurray: This is because of the auto-conversion from camelCase to snake_case.

roles/velero/defaults/main.yml
shubham-pampattiwar: Done

roles/velero/templates/registry_aws.yml.j2
shubham-pampattiwar: Updated it to `oadp-<bls.name>-<bsl.provider>-registry`

roles/velero/templates/registry_azure.yml.j2
shubham-pampattiwar: Done `oadp-<bls.name>-<bsl.provider>-registry`

roles/velero/templates/registry_gcp.yml.j2
shubham-pampattiwar: Moved the svc + route from each provider template to an unified one. Updated the changes

roles/velero/templates/registry_gcp.yml.j2
shubham-pampattiwar: Kept deployment templates different for each provider as discussed and moved out the svc + route creation.

roles/velero/templates/registry_config_map.yml.j2
shubham-pampattiwar: Used the combination of `<bls.name>-<bsl.provider>`  for creation of registry, svc, routes as well as the component label so that the naming remains unique.

PR #65, opened 2020-11-03
Set LD_LIBRARY_PATH env var

### LINKED ISSUE #64: Missing LD_LIBRARY_PATH in Velero pod deployed by OADP operator
Velero's LB_LIBRARY_PATH is set to "/plugins" directory so if any plugin binary needs to load additional library during execution, it can put the library there.  However, Velero installed by OADP missing this environment variable and therefore fail the plugins that needs to load libraries even though the library binaries are there in the /plugins directory.

Reproduce steps:
- install OADP operator
- create simple Velero with default csi, openshift plugins (see config below)
apiVersion: konveyor.openshift.io/v1alpha1
kind: Velero
metadata:
  name: example-velero
  namespace: oadp-operator
spec:
  olm_managed: true
  use_upstream_images: true
  default_velero_plugins:
    - csi
    - openshift
  enable_restic: true
  
 then use "exec -it " to log into Velero pod and print out all evn

#oc get pod velero-7d9cd8489b-7v5cx -n oadp-operator bash
#env | grep LD_LIBRARY_PATH 
==> found nothing

dymurray:
Fixes #64 

See: https://github.com/vmware-tanzu/helm-charts/blob/main/charts/velero/templates/deployment.yaml#L117

### DIFF
--- a/roles/velero/templates/velero.yml.j2
+++ b/roles/velero/templates/velero.yml.j2
@@ -93,6 +93,8 @@ spec:
             - name: certs
               mountPath: /etc/ssl/certs
           env:
+            - name: LD_LIBRARY_PATH
+              value: /plugins
 {% if 'aws' in default_velero_plugins %}
             - name: AWS_SHARED_CREDENTIALS_FILE
               value: /credentials/cloud


shubham-pampattiwar (approved):
ACK

PR #67, opened 2020-11-17
Update to v0.1.3

dymurray:


deploy/olm-catalog/bundle/manifests/oadp-operator.v0.1.3.clusterserviceversion.yaml
```diff
@@ -211,7 +210,7 @@ spec:
                 - name: VELERO_OPENSHIFT_PLUGIN_REPO
                   value: openshift-velero-plugin
                 - name: VELERO_OPENSHIFT_PLUGIN_TAG
-                  value: unified-repo
+                  value: latest
```
shubham-pampattiwar, line 213: Should we not change this in the `operator.yaml` as well ?

shubham-pampattiwar (approved):
Just one question, apart from that looks good !

deploy/olm-catalog/bundle/manifests/oadp-operator.v0.1.3.clusterserviceversion.yaml
dymurray: Oh yes thanks!

PR #87, opened 2021-03-31
Doc fix for BSL/VSL configuration

### LINKED ISSUE #80: Problems with region endpoint URL
We use oadp-oeprator with ROKS (managed Openshift on IBMCloud).
We failed to set-up oadp-operator with ICOS(S3 compatible IBM Cloud Object Storage)endpoint due to the following error,
- panic: invalid region provided: jp-tok

This is because jp-tok is an IBM region name that AWS does not have.
I know that  we can bypass region check with env "REGISTRY_STORAGE_S3_REGIONENDPOINT",
but the value of s3url has been ignored.
veloro setup YAML:
```
apiVersion: konveyor.openshift.io/v1alpha1
kind: Velero
metadata:
  name: velero
  namespace: oadp-operator
spec:
  olm_managed: true
  backup_storage_locations:
  - config: 
      region: us-east-1
      s3ForcePathStyle: true
      s3Url: 'https://s3.private.jp-tok.cloud-object-storage.appdomain.cloud'
...
```
oadp-default-aws-registry env:
```
...
          env:
            - name: REGISTRY_STORAGE
              value: s3
            - name: REGISTRY_STORAGE_S3_ACCESSKEY
              value: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
            - name: REGISTRY_STORAGE_S3_BUCKET
              value: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
            - name: REGISTRY_STORAGE_S3_REGION
              value: jp-tok
            - name: REGISTRY_STORAGE_S3_SECRETKEY
              value: XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
...
```
We can work around this by manually adding env REGISTRY_STORAGE_S3_REGIONENDPOINT to oadp-default-aws-registry  deployment, but I don't think this is recommended since this deployment controlled by an operator.
How should I handle this?

sseago:
This makes it clear that BSL/VSL configuration in the OADP Velero CR
must specify Velero parameter names using snake_case rather than
camelCase.

Fixes #80

### DIFF
--- a/docs/bsl_and_vsl.md
+++ b/docs/bsl_and_vsl.md
@@ -38,5 +38,7 @@ spec:
 <b>Note:</b> 
 - Be sure to use the same `secret` name you used while creating the cloud credentials secret in step 3 of Operator   installation section.
 - Another thing to consider are the CR file specs, they should be tailored in accordance to your own cloud provider accounts, for instance `bucket` spec value should be according to your own bucket name and so on.
-- Do not configure more than one `backupStorageLocations` per cloud provider, the velero installation will fail.  
+- Do not configure more than one `backupStorageLocations` per cloud provider, the velero installation will fail.
+- bsl/vsl parameters in the OADP Velero CR must be specified using `snake_case` rather than `camelCase`.
+  Keep this in mind when using parameters from the below parameter reference. For example, use `s3_url` rather than `s3Url`.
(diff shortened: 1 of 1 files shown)


GMerigo:
I can confirm point 2, I cannot verify point 1 as our test cluster is currently unavailable, I waited until now to see if we could make any test but I should replicate this on another cluster, unfortunately there we don't have ebs-csi installed so the test would be very different.

sseago:
@GMerigo Thanks for the update. I'm going to go ahead and merge this fix. If you try this change in the future and it still fails for you, feel free to open another issue.

PR #121, opened 2021-06-25
Upgrade ansible-operator base image to v1.5.0

### LINKED ISSUE #102: Thousands of "Skipping cache lookup" messages
Hi.

We've deployed OADP operator on OCP v4.6 via this procedure: https://github.com/konveyor/oadp-operator#getting-started-with-basic-install-non---olm

We have observed an immense amount of logs being emitted by the oadp-operator pod:

image
  
```
  ...
{"level":"info","ts":1620604499.8003464,"logger":"proxy","msg":"Skipping cache lookup","resource":{"IsResourceRequest":false,"Path":"/apis","Verb":"get","APIPrefix":"","APIGroup":"","APIVersion":"","Namespace":"","Resource":"","Subresource":"","Name":"","Parts":null}}
...
--------------------------- Ansible Task StdOut -------------------------------
...
TASK [velero : Set up velero controller] ***************************************
...
```

and so on.

The result is that our Cluster Logging Elasticsearch instance is being filled completely in 2-3 days.

How do we disable this excessive logging? I have found that the ansible-operator base image used by the oadp-opearator is still on [v0.17.0](https://github.com/operator-framework/operator-sdk/releases/tag/v0.17.0). There was a [fix](https://github.com/operator-framework/operator-sdk/pull/4511) that was released a few months ago and is not included in that version. It is included in [v1.5.0](https://github.com/operator-framework/operator-sdk/releases/tag/v1.5.0) and later.

Refs:
https://github.com/openshift/openshift-restclient-python/issues/384
https://github.com/operator-framework/operator-sdk/pull/3933
https://stackoverflow.com/questions/67415884/how-to-reduce-the-log-verbosity-of-ansible-based-kubernetes-operator-deployment
https://github.com/operator-framework/operator-sdk/issues/4519
https://github.com/operator-framework/operator-sdk/pull/4511

**Edit:**
We built the oadp-operator using the ansible-operator:v1.5.0 as base image and the excessive logging has stopped.

dymurray:
Fixes #102

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/operator-framework/ansible-operator:v0.17.0
+FROM quay.io/operator-framework/ansible-operator:v1.5.0
 
 COPY requirements.yml ${HOME}/requirements.yml
 RUN ansible-galaxy collection install -r ${HOME}/requirements.yml \


shubham-pampattiwar (approved):

build/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM quay.io/operator-framework/ansible-operator:v0.17.0
+FROM quay.io/operator-framework/ansible-operator:v1.5.0
```
alaypatel07, line 1: wonder if this should point to latest? we do that for mig-operator

https://github.com/konveyor/mig-operator/blob/master/build/Dockerfile#L1

jwmatthews: Agreed, we've been actively moving to use :latest unless we know of a reason we can't.

dymurray: I was actually going to say we should probably pin mig-operator to a stable release :)

jwmatthews: We had an issue with pinning in past of forgetting to update and then not picking up newer versions that addressed CVEs.  The CVEs were the main driver to move us to consume :latest.  There was also an impression that ansible-operator has matured so less concern of breaking, and if we did break on a new update good to learn ASAP and catch the issue.

dymurray: Thanks, I will move to latest

PR #123, opened 2021-06-28
add dockerfiles to support openshift CI

hhpatel14:


build/Dockerfile.bundle
```diff
@@ -0,0 +1,16 @@
+FROM scratch
+
+LABEL operators.operatorframework.io.bundle.mediatype.v1=registry+v1
+LABEL operators.operatorframework.io.bundle.manifests.v1=manifests/
+LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
+LABEL operators.operatorframework.io.bundle.package.v1=crane-operator
```
dymurray, line 6: This needs to be `oadp-operator`

build/custom-ci-Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM openshift/origin-release:golang-1.14
+RUN yum -y install epel-release --disablerepo=epel && yum clean all
+RUN yum -y install python3-pip && yum clean all
+RUN pip3 install -U pip
+RUN pip3 install ansible jmespath molecule more-itertools openshift yamllint
```
shawn-hurley, line 5: Are we using molecule tests for oadp-operator?

hhpatel14: Yes, based on what I discussed with Dylan. But I am always open to suggestions.

dymurray: @shawn-hurley I asked Hiteshwari to integrate openshift CI with calling out to molecule so that we could leverage it like mig-operator if desired.

It is likely that we change this to simply be an e2e test suite written in golang... so I will hold off on merging until we've made that decision.

dymurray, line 5: remove these, add `yum -y install make`

dymurray:
```
go get -u github.com/onsi/ginkgo/ginkgo
```

shawn-hurley: There may already be a testing base image that we could use for ginkgo tests

dymurray: Good idea Shawn

build/custom-ci-Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM openshift/origin-release:golang-1.14
+RUN yum -y install epel-release --disablerepo=epel && yum clean all
+RUN yum -y install make
+RUN go get -u github.com/onsi/ginkgo/ginkgo
```
dymurray, line 4: According to https://onsi.github.io/ginkgo/ we may also need:
```
go get github.com/onsi/gomega/...
```

I would add it to be safe

dymurray (commented):
LGTM once we add the last `go get` command

dymurray (approved):

Issue #117, opened 2021-06-22
Support different cloud providers between BSL + VSL


dymurray:
Today OADP requires that the same cloud provider is used for BSL+VSL. This is not desired, as someone may want to use AWS for a BSL and Azure for a VSL or vice versa.

There are a few things to fix here:

- [ ] OADP operator should allow you to specify different provider details between BSL/VSL, rather than sharing the same cloud provider
- [ ] Test OADP with GCP for backup storage and volume snapshots and identify bugs
- [ ] Test OADP with Azure for backup storage and volume snapshots and identify bugs
- [ ] Negative testing against azure/gcp configuration params

PR #124, opened 2021-06-29
Add different cloud provider support between BSL and VSL

shubham-pampattiwar:
This PR does the following:
- Supports different cloud provider between BSL and  VSL at the same time.
- Adds checks in preflight tasks for VSL as well as updates operator tasks
- Updates default Velero CR accommodating VSL credentials reference 
- Fixes https://github.com/openshift/oadp-operator/issues/117

Changes tested for:
- AWS BSL + AWS VSL
- Azure BSL + AWS VSL


dymurray (commented):
This looks good. Just need to test it on my end. I would like to merge this before cutting a new release

deploy/crds/konveyor.openshift.io_v1alpha1_velero_cr.yaml
```diff
@@ -7,7 +7,6 @@ spec:
   default_velero_plugins:
   - aws
   - openshift
-  - csi
```
dymurray, line 10: Is this intentional? I think we want CSI plugin by default

dymurray: The feature flag may no longer be needed... unsure

roles/velero/defaults/main.yml
```diff
@@ -12,9 +12,18 @@ velero_state: "present"
 velero_aws_bsl_configs: 0
 velero_gcp_bsl_configs: 0
 velero_azure_bsl_configs: 0
+velero_aws_vsl_configs: 0
```
dymurray, line 15: TODO: We should come back and relook at this defaults yml file. These really belong in `vars.yml` since they aren't default values we expect the user to change.

This is fine for now though

roles/velero/defaults/main.yml
```diff
@@ -12,9 +12,18 @@ velero_state: "present"
...
+velero_gcp_vsl_configs: 0
+velero_azure_vsl_configs: 0
 velero_aws_secret_name: cloud-credentials
 velero_gcp_secret_name: gcp-cloud-credentials
 velero_azure_secret_name: azure-cloud-credentials
+velero_aws_bsl: false
```
dymurray, line 21: Do we need to define these vars though? It feels to me like we can just wait for `set_fact` to set them. There's no world where the user sets these themselves right?

dymurray (commented):
overall the functionality LGTM and it ran in my env fine.

I have some questions about the ansible logic that I think is just my peeve of doing this in ansible (and we did this before so I'm probably overthinking it).

I'm fine to merge this but if you want to wait until next week I'd be interested in talking through some ideas I have on changing this slightly.

PR #125, opened 2021-06-30
Added a Ginkgo starting test

eemcmullan:
This is to help get a setup started in order to proceed with the CI.


dymurray (approved):
This looks good to me. Would be cool to also add a `Makefile` so that the CI system can just call `make run` or whatever in this directory. This way the CI invocation stays the same and gives us flexibility to tweak the test runs in the future.

eemcmullan:
Sounds good-will do!

tests/e2e/go.mod
```diff
@@ -0,0 +1,8 @@
+module add_nums
```
dymurray, line 1: @eemcmullan Actually lets change this please

PR #128, opened 2021-07-06
Main file for e2e tests

eemcmullan:
Created main file with functions needed to run future OADP tests


tests/e2e/main.go
```diff
@@ -1,12 +1,149 @@
...
+      profile: "default"
+  enable_restic: true
+  velero_feature_flags: EnableCSI`
 
 func main() {
-	fmt.Println(add_nums(3, 6))
+	installVelero()
+
+	// get Velero pod status
+	veleroStatus := isPodRunning()
```
dymurray, line 54: You will want to put this in a loop that checks every ~5 seconds or so

PR #132, opened 2021-07-08
Update makefile to ignore the vendor directory for CI

hhpatel14:


dymurray:
Closing because I vendored the deps

PR #142, opened 2021-07-22
Adapt Test Suite for CI 

deepakraj1997:
Summary of changes: 

- Parameterized namespaces, instance names, cloud config etc.,
- Converted the Yaml to Unstructured to facilitate the above.
- Initialize few common things in test init.

To do: (To be done after raising e2e-tests against master)

- Test it with Openshift CI / Prow.
- Modify Makefile to make use of Build Env. used in Prow.

### DIFF
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -12,7 +12,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
@@ -73,39 +72,21 @@ func getDefaultVeleroConfig(namespace string, s3Bucket string, credSecretRef str
 	return &veleroSpec
 }
 
-func decodeYaml(defaultVeleroConfigYAML string) *unstructured.Unstructured {
-	// set new unstructured type for Velero CR
-	unstructVelero := &unstructured.Unstructured{}
-
-	// decode yaml into unstructured type
-	dec := yaml.NewDecodingSerializer(unstructured.UnstructuredJSONScheme)
-	_, _, err := dec.Decode([]byte(defaultVeleroConfigYAML), nil, unstructVelero)
-	if err != nil {
-		panic(err)
-	}
-	return unstructVelero
-}
-
-func getJsonData(path string) []byte {
+func getJsonData(path string) ([]byte, error) {
 	// Return buffer data for json
 	jsonData, err := ioutil.ReadFile(path)
-	if err != nil {
-		panic(err)
-	}
-	return jsonData
+	return jsonData, err
 }
 
-func decodeJson(data []byte) map[string]interface{} {
+func decodeJson(data []byte) (map[string]interface{}, error) {
 	// Return JSON from buffer data
 	var jsonData map[string]interface{}
 
 	err := json.Unmarshal(data, &jsonData)
-	if err != nil {
-		panic(err)
-	}
-	return jsonData
+	return jsonData, err
 }
 
+// Keeping it for now.
 func createOADPTestNamespace(namespace string) error {
 	// default OADP Namespace
 	kubeConf := getKubeConfig()
@@ -126,6 +107,7 @@ func createOADPTestNamespace(namespace string) error {
 	return err
 }
 
+// Keeping it for now.
 func deleteOADPTestNamespace(namespace string) error {
 	// default OADP Namespace
 	kubeConf := getKubeConfig()
@@ -175,6 +157,7 @@ func deleteVeleroCR(client dynamic.Interface, instanceName string, namespace str
 	return veleroClient.Delete(context.Background(), instanceName, metav1.DeleteOptions{})
 }
 
+// Keeping it for now.
 func isNamespaceDeleted(namespace string) wait.ConditionFunc {
 	return func() (bool, error) {
 		kubeConf := getKubeConfig()
@@ -182,7 +165,7 @@ func isNamespaceDeleted(namespace string) wait.ConditionFunc {
 		// create client for pod
 		clientset, err := kubernetes.NewForConfig(kubeConf)
 		if err != nil {
-			panic(err)
+			return true, err
 		}
 		_, exists := clientset.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
 		if exists == nil {
@@ -192,7 +175,19 @@ func isNamespaceDeleted(namespace string) wait.ConditionFunc {
 	}
 }
 
+// Keeping it for now.
 func waitForNamespaceDeletion(namespace string) error {
 	// poll pod every 5 secs for 2 mins until it's running or timeout occurs
 	return wait.PollImmediate(time.Second*5, time.Minute*2, isNamespaceDeleted(namespace))
+}
+
+func isNamespaceExists(namespace string) error {
+	kubeConf := getKubeConfig()
+	// create client for pod
+	clientset, err := kubernetes.NewForConfig(kubeConf)
+	if err != nil {
+		return err
+	}
+	_, exists := clientset.CoreV1().Namespaces().Get(context.Background(), namespace, metav1.GetOptions{})
+	return exists
 }
--- a/tests/e2e/common.go
+++ b/tests/e2e/common.go
@@ -72,13 +72,13 @@ func getDefaultVeleroConfig(namespace string, s3Bucket string, credSecretRef str
 	return &veleroSpec
 }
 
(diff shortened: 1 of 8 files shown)


tests/e2e/common.go
```diff
@@ -2,74 +2,121 @@ package e2e
...
-  - name: default
-    provider: aws
-    config:
-      region: us-west-2
-      profile: "default"
-  enable_restic: true
-  velero_feature_flags: EnableCSI`
-
-func decodeYaml() *unstructured.Unstructured {
+func getDefaultVeleroConfig(namespace string, s3Bucket string, credSecretRef string, instanceName string) *unstructured.Unstructured {
```
dymurray, line 22: +1 this is better than just hardcoding the yaml

tests/e2e/common.go
```diff
@@ -2,74 +2,121 @@ package e2e
...
 	}
 	return unstructVelero
 }
 
-func createOADPTestNamespace() error {
+func getJsonData(path string) []byte {
+	// Return buffer data for json
+	jsonData, err := ioutil.ReadFile(path)
+	if err != nil {
+		panic(err)
```
dymurray, line 92: Rather than panicking, we should simply return an error and log it properly

tests/e2e/common.go
```diff
@@ -2,74 +2,121 @@ package e2e
...
+	return jsonData
+}
+
+func decodeJson(data []byte) map[string]interface{} {
+	// Return JSON from buffer data
+	var jsonData map[string]interface{}
+
+	err := json.Unmarshal(data, &jsonData)
+	if err != nil {
+		panic(err)
```
dymurray, line 103: Same here

tests/e2e/common.go
```diff
@@ -2,74 +2,121 @@ package e2e
...
+
+func createOADPTestNamespace(namespace string) error {
+	// default OADP Namespace
 	kubeConf := getKubeConfig()
 	clientset, err := kubernetes.NewForConfig(kubeConf)
 	if err != nil {
 		return err
 	}
+	// veleroNamespace := flag.String("velero-namespace", "", "Velero Namespace")
+	flag.Parse()
```
dymurray, line 116: Is this being used at all?

tests/e2e/e2e_suite_test.go
```diff
@@ -8,10 +8,15 @@ import (
...
 )
 
-var cloud string
+// Common vars obtained from flags passed in ginkgo.
+var cloud, namespace, s3Bucket, s3BuckerFilePath, credSecretRef, instanceName string
 
 func init() {
-	flag.StringVar(&cloud, "cloud", "", "Credentials file path location")
+	flag.StringVar(&cloud, "cloud", "", "Cloud Credentials file path location")
+	flag.StringVar(&s3BuckerFilePath, "s3_bucket", "myBucket", "AWS S3 data file path location")
```
dymurray, line 16: Typo: `s3BucketFilePath`

tests/e2e/install_suite_test.go
```diff
@@ -1,37 +1,65 @@
...
-//})
+// })
 
 var _ = Describe("The default Velero custom resource", func() {
 	var _ = BeforeEach(func() {
-		err := installDefaultVelero()
+		prefix := "ts-" // To create individual instance per spec
+		flag.Parse()
+		s3Data := decodeJson(getJsonData(s3BuckerFilePath)) // Might need to change this later on to create s3 for each tests
+		s3Bucket = s3Data["velero-bucket-name"].(string)
```
dymurray, line 45: I'm curious is this actually going to work? My understanding was our latest version of the secret encodes the JSON in a file under `velero-credentials`. So to load the bucket, I thought we had to do something like:
```
data := decodeJson(getJsonData("/var/run/velero-credentials"))
s3Bucket := data["velero-bucket-name"]
```

Not totally sure though
```

Comment (deepakraj1997) [reply tests/e2e/install_suite_test.go:44]: So, i have divided the logic for getting the aws credentials. now two files will be created , one containing s3 bucket and other necessary details. and the other aws credentials created just like how we want to pass when installing / creating oc secret.

File: tests/e2e/common.go (hunk 2 above)
Comment (JaydipGabani) [reply tests/e2e/common.go:93]: +1, `Expect().NotTo()` I think provides the trace log as well. So we might not need to use a logger also.

Comment (deepakraj1997) [reply tests/e2e/common.go:93]: Got it. Will return the error back and clean the code.

File: tests/e2e/common.go [hunk 7]
```diff
@@ -2,74 +2,120 @@ package e2e
...
+						"provider": "aws",
+					},
+				},
+			},
+		},
+	}
+	return &veleroSpec
+}
+
+func decodeYaml(defaultVeleroConfigYAML string) *unstructured.Unstructured {
```
JaydipGabani, line 76: Is this an unused function? are we planning to use it later? or removing it later?

tests/e2e/restic_helpers.go
```diff
@@ -86,23 +110,23 @@ func areResticPodsRunning() wait.ConditionFunc {
 	}
 }
 
-func waitForResticPods() error {
+func waitForResticPods(namespace string) error {
 	// poll pod every 5 secs for 2 mins until it's running or timeout occurs
-	return wait.PollImmediate(time.Second*5, time.Minute*2, areResticPodsRunning())
+	return wait.PollImmediate(time.Second*5, time.Minute*2, areResticPodsRunning(namespace))
```
JaydipGabani, line 115: Gomega provides [Eventually](https://onsi.github.io/gomega/#eventually) block out of the box to poll for these kinds of situations. We might be able to use this here and remove this function altogether :thinking:. Maybe call `areResticPodsRunning` from an `Eventually` block. Ignore if this is implemented for a specific reason.

tests/e2e/common.go
deepakraj1997: Yeh, this was used in previous testing when the YAML was hardcoded. Yet to remove this and clean the code as a whole. Will remove it since I have converted everything to object.

tests/e2e/restic_helpers.go
deepakraj1997: Good suggestion! I was wondering if there would be a callback for this, eventually fits the case. I will try it out.

deepakraj1997:
We might need to raise this against master to actually check run the tests. As I think the hook / app maybe configured to run tests for PRs only against master brancg

tests/e2e/install_suite_test.go
dymurray: Ahh I see. Nice makes sense, thanks!

dymurray:
@deepak1725 +1 I think so too. I would be in favor of merging this to e2e-tests and opening a WIP PR of e2e-tests against master. This way we can push changes up to the shared branch and also get CI running against it. Thoughts?

deepak1725:
> @deepak1725 +1 I think so too. I would be in favor of merging this to e2e-tests and opening a WIP PR of e2e-tests against master. This way we can push changes up to the shared branch and also get CI running against it. Thoughts?

Hey @dymurray, you need to be careful while tagging. 

Thanks

dymurray:
@deepakraj1997 ^. Sorry Deepak1725 :)

tests/e2e/common.go
deepakraj1997: Removed the function.

tests/e2e/common.go
deepakraj1997: Returning errors back for every utility function.

tests/e2e/e2e_suite_test.go
deepakraj1997: Fixed

tests/e2e/common.go
deepakraj1997: Removed.

tests/e2e/restic_suite_test.go
```diff
@@ -1,42 +1,45 @@
...
 
 	Context("When the value of 'enable_restic' is changed to false", func() {
 		It("Should delete the Restic daemonset", func() {
-			errs := installDefaultVelero()
-			Expect(errs).NotTo(HaveOccurred())
-
-			err := waitForDeletedRestic()
-			Expect(err).NotTo(HaveOccurred())
+			fmt.Printf("here: 123, " + testSuiteInstanceName)
+			Eventually(isResticDaemonsetDeleted(namespace, testSuiteInstanceName, "restic"), time.Minute*2, time.Second*5).Should(BeTrue())
```
eemcmullan, line 42: Good idea moving the wait functions here!

tests/e2e/common.go
eemcmullan: +1 I agree

tests/e2e/restic_suite_test.go
```diff
@@ -1,42 +1,45 @@
...
 	})
 
 	Context("When the value of 'enable_restic' is changed to false", func() {
 		It("Should delete the Restic daemonset", func() {
-			errs := installDefaultVelero()
-			Expect(errs).NotTo(HaveOccurred())
-
-			err := waitForDeletedRestic()
-			Expect(err).NotTo(HaveOccurred())
+			fmt.Printf("here: 123, " + testSuiteInstanceName)
```
dymurray, line 41: Remove this

dymurray (approved):

PR #159, opened 2021-07-30
add units for ValidateBackupStorageLocations

alaypatel07:


oadp-operator/controllers/bsl_test.go
```diff
@@ -0,0 +1,251 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/record"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+)
+
+func getSchemeForFakeClient() (*runtime.Scheme, error) {
```
dymurray, line 20: Just a NIT, do you think we should make a fake client pkg separately? To make it easier for other tests to import the fake client usage?

dymurray (approved):
Thanks a lot Alay, tests look great. We can refactor the fake client in a separate PR

dymurray (approved):

PR #160, opened 2021-07-30
Add velero deployment controller

shubham-pampattiwar:
This PR adds the Velero deployment controller

### DIFF
--- a/oadp-operator/controllers/velero.go
+++ b/oadp-operator/controllers/velero.go
@@ -14,48 +14,48 @@ import (
 )
 
 const (
-	VELERO        = "velero"
-	VELERO_NS     = "oadp-operator"
-	OADP_OPERATOR = "oadp-operator"
-	SERVER        = "server"
+	Velero         = "velero"
+	VeleoNamespace = "oadp-operator"
+	OADPOperator   = "oadp-operator"
+	Server         = "server"
 	//TODO: Check for default secret names
-	VELERO_AWS_SECRET_NAME   = "cloud-credentials"
-	VELERO_AZURE_SECRET_NAME = "cloud-credentials-azure"
-	VELERO_GCP_SECRET_NAME   = "cloud-credentials-gcp"
+	VeleroAWSSecretName   = "cloud-credentials"
+	VeleroAzureSecretName = "cloud-credentials-azure"
+	VeleroGCPSecretName   = "cloud-credentials-gcp"
 )
 
 // Environment Vars keys
 const (
-	LD_LIBRARY_PATH_ENV_KEY               = "LD_LIBRARY_PATH_ENV_KEY"
-	VELERO_NAMESPACE_ENV_KEY              = "VELERO_NAMESPACE_ENV_KEY"
-	VELERO_SCRATCH_DIR_ENV_KEY            = "VELERO_SCRATCH_DIR_ENV_KEY"
-	AWS_SHARED_CREDENTIALS_FILE_ENV_KEY   = "AWS_SHARED_CREDENTIALS_FILE_ENV_KEY"
-	AZURE_SHARED_CREDENTIALS_FILE_ENV_KEY = "AZURE_SHARED_CREDENTIALS_FILE_ENV_KEY"
-	GCP_SHARED_CREDENTIALS_FILE_ENV_KEY   = "GCP_SHARED_CREDENTIALS_FILE_ENV_KEY"
+	LDLibraryPathEnvKey              = "LD_LIBRARY_PATH"
+	VeleroNamespaceEnvKey            = "VELERO_NAMESPACE"
+	VeleroScratchDirEnvKey           = "VELERO_SCRATCH_DIR"
+	AWSSharedCredentialsFileEnvKey   = "AWS_SHARED_CREDENTIALS_FILE"
+	AzureSharedCredentialsFileEnvKey = "AZURE_SHARED_CREDENTIALS_FILE"
+	GCPSharedCredentialsFileEnvKey   = "GCP_SHARED_CREDENTIALS_FILE"
 	//TODO: Add Proxy env vars
-	HTTP_PROXY  = "HTTP_PROXY"
-	HTTPS_PROXY = "HTTPS_PROXY"
-	NO_PROXY    = "NO_PROXY"
+	HTTPProxyEnvVar  = "HTTP_PROXY"
+	HTTPSProxyEnvVar = "HTTPS_PROXY"
+	NoProxyEnvVar    = "NO_PROXY"
 )
 
 //TODO: Add Image customization options
 // Images
 const (
-	VELERO_IMAGE           = "quay.io/konveyor/velero:latest"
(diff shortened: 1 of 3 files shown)


oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,469 @@
...
 	}
+
+	// Build Velero Deployment
+	veleroDeployment := r.buildVeleroDeployment(&velero)
+
+	// Create or Update Velero Deployment
+	op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, &veleroDeployment, func() error {
+
+		// Set controller reference to Velero controller
+		err := controllerutil.SetControllerReference(&velero, &veleroDeployment, r.Scheme)
```
shawn-hurley, line 72: I don't think that this is changing anything in the cluster just on the object reference

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,469 @@
...
+			corev1.EventTypeNormal,
+			"VeleroDeploymentReconciled",
+			fmt.Sprintf("performed %s on velero deployment %s/%s", op, veleroDeployment.Namespace, veleroDeployment.Name),
+		)
+	}
 	return true, nil
 }
+
+// Build Velero Deployment
+func (r *VeleroReconciler) buildVeleroDeployment(velero *oadpv1alpha1.Velero) appsv1.Deployment {
```
alaypatel07, line 99: return a pointer? `*appsv1.Deployment`

dymurray: +1

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,469 @@
...
+	veleroResourceReqs := r.getVeleroResourceReqs(velero)
+
+	deployment := appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      Velero,
+			Namespace: VeleroNamespace,
+		},
+		Spec: appsv1.DeploymentSpec{
+			//TODO: add velero nodeselector, needs to be added to the Velero CR first
+			Selector: &metav1.LabelSelector{
```
alaypatel07, line 114: https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#label-selector-updates

I think you will have to be careful here. You might want to accept the pointer to deployment here and change the parts that you need based on whether it's a create or update.

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,469 @@
...
+	VeleroNamespace = "oadp-operator"
+	//TODO: Check for default secret names
+	VELERO_AWS_SECRET_NAME   = "cloud-credentials"
+	VELERO_AZURE_SECRET_NAME = "cloud-credentials-azure"
+	VELERO_GCP_SECRET_NAME   = "cloud-credentials-gcp"
+)
+
+// Environment Vars
+const (
+	LD_LIBRARY_PATH               = "LD_LIBRARY_PATH"
```
dymurray, line 27: +1 good to make these constants

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,469 @@
...
+	GCP_SHARED_CREDENTIALS_FILE   = "GCP_SHARED_CREDENTIALS_FILE"
+	//TODO: Add Proxy env vars
+	HTTP_PROXY  = "HTTP_PROXY"
+	HTTPS_PROXY = "HTTPS_PROXY"
+	NO_PROXY    = "NO_PROXY"
+)
+
+//TODO: Add Image customization options
+// Images
+const (
```
dymurray, line 41: NIT: you will need to make the tags and images constants we can override with env vars

dymurray: We can do this in a future update though

oadp-operator/controllers/velero.go
dymurray: +1 you could be deleting labels by doing this

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,464 @@
...
 		return false, err
 	}
+
+	veleroVolumeMounts := r.getVeleroVolumeMounts(&velero)
+	veleroVolumes := r.getVeleroVolumes(&velero)
+	veleroEnv := r.getVeleroEnv(&velero)
+	veleroInitContainers := r.getVeleroInitContainers(&velero)
+	veleroResourceReqs := r.getVeleroResourceReqs(&velero)
+	veleroTolerations := velero.Spec.VeleroTolerations
+	veleroDeployment := &appsv1.Deployment{}
```
shawn-hurley, line 71: I think this may need the namespace and name

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,486 @@
...
+	//TODO: Check for default secret names
+	VELERO_AWS_SECRET_NAME   = "cloud-credentials"
+	VELERO_AZURE_SECRET_NAME = "cloud-credentials-azure"
+	VELERO_GCP_SECRET_NAME   = "cloud-credentials-gcp"
+)
+
+// Environment Vars
+const (
+	LD_LIBRARY_PATH               = "LD_LIBRARY_PATH"
+	VELERO_NAMESPACE              = "VELERO_NAMESPACE"
```
shawn-hurley, line 30: shouldn't this be "velero"?

oadp-operator/controllers/velero.go
```diff
@@ -1,14 +1,486 @@
...
+		}
+
+		// Setting controller owner reference on the velero deployment
+		err := controllerutil.SetControllerReference(&velero, veleroDeployment, r.Scheme)
+		if err != nil {
+			return err
+		}
+
+		// update the Deployment template
+		veleroDeployment = r.buildVeleroDeployment(veleroDeployment, veleroLabels, veleroVolumeMounts, veleroVolumes, veleroEnv, veleroInitContainers, veleroResourceReqs, veleroTolerations)
```
alaypatel07, line 99: an opinionated suggestion: this function call is not providing a lot of values, you are sending the arg list into the function and it is just setting it to the right place in deployment. 

Why don't we change this to `r.buildVeleroDeployment(veleroDeployment)`? then the function will call all the getters and build the right spec there. Then we can add a unit test for `buildVeleroDeployment`. 

Not asking for the units in this PR but let's start with refactoring this function so it more makes

oadp-operator/controllers/velero.go
dymurray: This is the env var key used on the container. The var name may want to communicate that

oadp-operator/controllers/velero.go
dymurray: Agree with this. Generally my rule of thumb is anything over 3 args means we can refactor the function call in some manner. I think Alay's suggestion makes sense

dymurray (approved):
Other than the NITS I left, this LGTM

alaypatel07 (approved):
Looks great.  Thanks!

dymurray (approved):

PR #161, opened 2021-07-30
Restic Go Controller and pkg/credentials, pkg/common

kaovilai:


oadp-operator/controllers/restic.go
```diff
@@ -230,6 +230,12 @@ func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, erro
 		},
 	}
 
+	if velero.Spec.EnableRestic != nil && !*velero.Spec.EnableRestic {
```
shawn-hurley, line 233: @dymurray do we have the ability to see if restic is in use? I could see a case where a backup is running, we probably don't want to blindly delete this unlesss there is a finalizer that velero uses

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,321 @@
...
 func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error) {
 	velero := oadpv1alpha1.Velero{}
 	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
 		return false, err
 	}
+
+	var mountPropagationToHostContainer = v1.MountPropagationHostToContainer
+
+	// Define "static" portion of daemonset
+	ds := appsv1.DaemonSet{
```
shawn-hurley, line 49: Should we consider that this is in a function?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,321 @@
...
+		// Daemonset found, check if equal.
+		if err := r.Update(r.Context, &ds); err != nil { // Update daemonset
+			return false, err
+		}
+	}
+
 	return true, nil
 }
+
+func getVeleroResourceReqs(velero *oadpv1alpha1.Velero) v1.ResourceRequirements {
```
shawn-hurley, line 275: We will need to handle the merge here as I think this will collide with @shubham-pampattiwar PR as well

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,321 @@
...
+		r.Delete(r.Context, &ds) //TODO: delete fail logic?
+		return true, nil
+	}
+	// Dynamically add to daemonset definition
+	// If the default velero plugins contains cloud provider, attach VolumeSource
+	for provider, providerSecretName := range cloudProviderSecretNames {
+		if contains(provider, velero.Spec.DefaultVeleroPlugins) {
+			ds.Spec.Template.Spec.Volumes = append(
+				ds.Spec.Template.Spec.Volumes,
+				v1.Volume{
```
shawn-hurley, line 244: Do we also need volume mounts?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,321 @@
...
+									Name: "AWS_SHARED_CREDENTIALS_FILE",
+									ValueFrom: &v1.EnvVarSource{
+										FieldRef: &v1.ObjectFieldSelector{
+											FieldPath: "/credentials/cloud",
+										},
+									},
+								},
+								{
+									Name: "GOOGLE_APPLICATION_CREDENTIALS",
+									ValueFrom: &v1.EnvVarSource{
```
shawn-hurley, line 178: Do we always need this Env? should we selectively add or remove it?

shawn-hurley (commented):
Looking really good just some high level notes and potential cautions

oadp-operator/controllers/restic.go
dymurray: @shawn-hurley the only way to do this would be to loop through all PVBs and PVRs and check if any are in progress. You bring up a good point. I'm unaware of any finalizers that Velero uses to stop this from happening

shawn-hurley: maybe just something to think about for the furture?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
+	"github.com/openshift/oadp-operator/pkg/credentials"
+	appsv1 "k8s.io/api/apps/v1"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+)
+
+/**
```
dymurray, line 17: can we remove this?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+- name: "Enable restic"
+  k8s:
+    state: "{{ velero_state }}"
+    definition: "{{ lookup('template', 'restic.yml.j2')}}"
+  when: enable_restic == true
+*/
+
+const (
+	veleroSAName     = "velero"
+	resticPvHostPath = "/var/lib/kubelet/pods"
```
dymurray, line 27: TODO: We need to make this configurable with an env var

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+/**
+- name: "Enable restic"
+  k8s:
+    state: "{{ velero_state }}"
+    definition: "{{ lookup('template', 'restic.yml.j2')}}"
+  when: enable_restic == true
+*/
+
+const (
+	veleroSAName     = "velero"
```
dymurray, line 26: NIT: Make constants uppercase

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+	resticPvHostPath = "/var/lib/kubelet/pods"
+)
+
+// const mountPropagationMode = v1.MountPropagationMode
```
dymurray, line 30: Remove this?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+var (
+	mountPropagationToHostContainer = v1.MountPropagationHostToContainer
```
dymurray, line 32: Any reason we need this? Just redeclaring a variable no?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
 func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error) {
 	velero := oadpv1alpha1.Velero{}
 	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
 		return false, err
 	}
+
+	// Define "static" portion of daemonset
+	ds := &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      resticLabelMap["name"],
```
dymurray, line 47: Not sure I understand why we are using a map for this. Can't we just declare this as a constant?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+					},
+				},
+				Tolerations: velero.Spec.ResticTolerations,
+				Containers: []v1.Container{
+					{
+						Name: "velero",
+						SecurityContext: &v1.SecurityContext{
+							Privileged: pointer.Bool(true),
+						},
+						Image: fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_REPO"), os.Getenv("VELERO_TAG")),
```
dymurray, line 156: Let's create a helper image `getResticImage()` that builds this out of the env vars rather than putting it all on one line

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+						// velero_image_fqin: "{{ velero_image }}:{{ velero_version }}"
```
dymurray, line 157: I'd remove this

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+								ValueFrom: &v1.EnvVarSource{
+									FieldRef: &v1.ObjectFieldSelector{
+										FieldPath: "/scratch",
+									},
+								},
+							},
+						},
+					},
+				},
+				InitContainers: []v1.Container{
```
dymurray, line 231: When we looked at this last week the conclusion we came to was this is no longerr needed, so we can remove this initContainer.

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,263 @@
...
+								Name:      string(oadpv1alpha1.DefaultPluginAWS),
+								MountPath: "/credentials",
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	if _, err := credentials.AppendCloudProviderVolumes(velero, ds); err != nil {
```
dymurray, line 259: Since `ds` is a pointer, we don't need it as a return

oadp-operator/pkg/credentials/credentials.go
```diff
@@ -0,0 +1,85 @@
...
+		},
+		oadpv1alpha1.DefaultPluginMicrosoftAzure: {
+			secretName:         "cloud-credentials-azure",
+			mountPath:          "/credentials-azure",
+			envCredentialsFile: "AZURE_CREDENTIALS_FILE",
+		},
+	}
+)
+
+func AppendCloudProviderVolumes(velero *oadpv1alpha1.Velero, ds *appsv1.DaemonSet) (*appsv1.DaemonSet, error) {
```
dymurray, line 42: Remove the `DaemonSet` return value since the input is a pointer

oadp-operator/pkg/credentials/credentials.go
```diff
@@ -0,0 +1,85 @@
...
+	var veleroContainer *corev1.Container
+	// Find Velero container
+	for _, container := range ds.Spec.Template.Spec.Containers {
+		if container.Name == "velero" {
+			veleroContainer = &container
+		}
+	}
+	for _, plugin := range velero.Spec.DefaultVeleroPlugins {
```
dymurray, line 50: +1 this is a good way to do this

oadp-operator/pkg/credentials/credentials.go
```diff
@@ -0,0 +1,85 @@
...
+)
+
+type CloudProviderFields struct {
+	secretName         string
+	mountPath          string
+	envCredentialsFile string
+}
+
+const (
+	veleroSAName     = "velero"
```
dymurray, line 16: We are redeclaring these constants even though you use them in the other pkg. i would move these to a `common` pkg or something. Or import them from the other pkg

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
+				Labels: map[string]string{
+					"name": Restic,
+				},
+			},
+			Spec: v1.PodSpec{
+				NodeSelector:       velero.Spec.ResticNodeSelector,
+				ServiceAccountName: common.Velero,
+				SecurityContext: &v1.PodSecurityContext{
+					RunAsUser:          pointer.Int64(0),
+					SupplementalGroups: []int64{},
```
dymurray, line 117: TODO: Update this to take in `velero.Spec.SupplementalGroups`

oadp-operator/controllers/restic.go
dymurray: Same question, think we can remove this.

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
 	velero := oadpv1alpha1.Velero{}
 	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
 		return false, err
 	}
+
+	// Define "static" portion of daemonset
+	ds := &appsv1.DaemonSet{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      Restic,
+			Namespace: r.NamespacedName.Namespace,
```
dymurray, line 37: Cool, this works. I think in other places I used `velero.Namespace` but using the reconciler namespace makes sense too

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
+		},
+	}
+
+	if velero.Spec.EnableRestic != nil && !*velero.Spec.EnableRestic {
+		// If velero Spec enableRestic exists and is false, attempt to delete.
+		r.Delete(r.Context, ds) //TODO: delete fail logic?
```
dymurray, line 43: Should check error here

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
+ */
+func (r *VeleroReconciler) buildResticDaemonset(velero *oadpv1alpha1.Velero, ds *appsv1.DaemonSet) (*appsv1.DaemonSet, error) {
+	if velero == nil {
+		return nil, fmt.Errorf("velero cannot be nil")
+	}
+	if ds == nil {
+		return nil, fmt.Errorf("ds cannot be nil")
+	}
+	ds.Spec = appsv1.DaemonSetSpec{
+		Selector: &metav1.LabelSelector{
```
dymurray, line 98: Better to grab this value from the existing DS. So:
```
Selector: ds.Spec.Selector
```

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
+		return true, nil
+	}
+
+	op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, ds, func() error {
+		// Deployment selector is immutable so we set this value only if
+		// a new object is going to be created
+		if ds.ObjectMeta.CreationTimestamp.IsZero() {
+			ds.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"component": common.Velero,
```
dymurray, line 53: Should this be `restic` instead of `velero`?

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,232 @@
...
+				},
+				Tolerations: velero.Spec.ResticTolerations,
+				Containers: []v1.Container{
+					{
+						Name: common.Velero,
+						SecurityContext: &v1.SecurityContext{
+							Privileged: pointer.Bool(true),
+						},
+						Image:           getResticImage(),
+						ImagePullPolicy: "Always",
```
dymurray, line 150: NIT: Use the k8s constant for `Always`. Think its in the core v1 package

oadp-operator/controllers/restic_test.go
```diff
@@ -0,0 +1,118 @@
...
+		Log            logr.Logger
+		Context        context.Context
+		NamespacedName types.NamespacedName
+		EventRecorder  record.EventRecorder
+	}
+	type args struct {
+		velero *oadpv1alpha1.Velero
+		ds     *appsv1.DaemonSet
+	}
+	tests := []struct {
```
dymurray, line 73: If you can add just one test that is a valid `velero` and valid `daemonset` object and verify the values match that would be good too

oadp-operator/controllers/restic.go
```diff
@@ -1,14 +1,250 @@
...
+)
+
+const (
+	Restic = "restic"
 )
 
+var (
+	resticPvHostPath string = getResticPvHostPath()
+
+	// v1.MountPropagationHostToContainer is a const. Const cannot be pointed to.
```
dymurray, line 25: +1 thanks

dymurray (approved):

PR #162, opened 2021-08-02
.gitignore vendor/**

kaovilai:


sseago:
I'm a little unclear on what's going on here. First, we actually have files under vendor/ in the repo. I'm not sure how that works with having that under gitignore. Second, we've moved all of the other go projects away from using vendor. Shouldn't we do the same here, and once that's done, we won't need to gitignore it either, since we won't be creating the directory locally.

dymurray:
Yeah I think we will want to resolve the issue with using vendor in CI before doing this. And additionally, I agree with Scott, we don't want to add it to gitignore, we'd rather just delete it entirely

PR #165, opened 2021-08-03
Add registry controller

shubham-pampattiwar:
This PR adds the registry controller.

### DIFF
--- a/oadp-operator/controllers/registry.go
+++ b/oadp-operator/controllers/registry.go
@@ -267,23 +267,21 @@ func (r *VeleroReconciler) buildRegistryContainer(bsl *velerov1.BackupStorageLoc
 
 func getRegistryEnvVars(bsl *velerov1.BackupStorageLocation) []corev1.EnvVar {
 	envVar := []corev1.EnvVar{}
-	if bsl.Spec.Provider == AWSProvider {
+	provider := bsl.Spec.Provider
+	switch provider {
+	case AWSProvider:
 		envVar = getAWSRegistryEnvVars(bsl, cloudProviderEnvVarMap[AWSProvider])
-	}
 
-	if bsl.Spec.Provider == AzureProvider {
+	case AzureProvider:
 		envVar = getAzureRegistryEnvVars(bsl, cloudProviderEnvVarMap[AzureProvider])
-	}
 
-	if bsl.Spec.Provider == GCPProvider {
+	case GCPProvider:
 		envVar = getGCPRegistryEnvVars(bsl, cloudProviderEnvVarMap[GCPProvider])
 	}
-
 	return envVar
 }
 
 func getAWSRegistryEnvVars(bsl *velerov1.BackupStorageLocation, awsEnvVars []corev1.EnvVar) []corev1.EnvVar {
-
 	for i, _ := range awsEnvVars {
 		//TODO: This needs to be fetched from the provider secret
 		if awsEnvVars[i].Name == RegistryStorageS3AccesskeyEnvVarKey {
@@ -318,7 +316,6 @@ func getAWSRegistryEnvVars(bsl *velerov1.BackupStorageLocation, awsEnvVars []cor
 }
 
 func getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocation, azureEnvVars []corev1.EnvVar) []corev1.EnvVar {
-
 	for i, _ := range azureEnvVars {
 		if azureEnvVars[i].Name == RegistryStorageAzureContainerEnvVarKey {
 			azureEnvVars[i].Value = bsl.Spec.StorageType.ObjectStorage.Bucket
@@ -336,7 +333,6 @@ func getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocation, azureEnvVars [
 }
 
 func getGCPRegistryEnvVars(bsl *velerov1.BackupStorageLocation, gcpEnvVars []corev1.EnvVar) []corev1.EnvVar {
-
 	for i, _ := range gcpEnvVars {
 		if gcpEnvVars[i].Name == RegistryStorageGCSBucket {
 			gcpEnvVars[i].Value = bsl.Spec.StorageType.ObjectStorage.Bucket
--- a/oadp-operator/controllers/registry.go
+++ b/oadp-operator/controllers/registry.go
@@ -159,15 +159,9 @@ func (r *VeleroReconciler) ReconcileRegistries(log logr.Logger) (bool, error) {
 				}
 			}
 
-			// Setting controller owner reference on the registry deployment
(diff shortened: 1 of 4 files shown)


oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,356 @@
...
+
+	return envVar
+}
+
+func buildRegistryEnvVars(bsl *velerov1.BackupStorageLocation, providerEnvVar []corev1.EnvVar, provider string) []corev1.EnvVar {
+
+	updatedEnvVar := providerEnvVar
+	// build AWS provider env vars
+	if provider == AWSProvider {
+		for _, key := range updatedEnvVar {
```
dymurray, line 290: I don't think this will work. You are not modifying the original array, you are modifying a copy of the element in the array. I would rework this to instead have functions like `getAwsEnvVars()` that returns `[]corev1.EnvVar` and then return it here.

You can see an example of this not working here: https://play.golang.org/p/HyEisqurot3

dymurray (changes requested):

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,356 @@
...
+			Name:  RegistryStorageGCSRootdirectory,
+			Value: "",
+		},
+	},
+}
+
+func (r *VeleroReconciler) ReconcileRegistries(log logr.Logger) (bool, error) {
+
+	bslLabels := map[string]string{
+		"app.kubernetes.io/name":       "oadp-operator-velero",
```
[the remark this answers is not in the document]
shubham-pampattiwar: I think this signifies the application name https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#applications-and-instances-of-applications And this is the same label used in BSL controller

oadp-operator/controllers/registry.go
shubham-pampattiwar: Thanks for the catch @dymurray Updated the PR.

oadp-operator/controllers/registry.go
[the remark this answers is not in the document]
dymurray: ah okay, ignore

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,353 @@
...
+			},
+		},
+	}
+
+	return containers
+}
+
+func getRegistryEnvVars(bsl *velerov1.BackupStorageLocation) []corev1.EnvVar {
+	envVar := []corev1.EnvVar{}
+	if bsl.Spec.Provider == AWSProvider {
```
dymurray, line 270: NIT: maybe make this a switch statement. Provider can't be all of these, so it should technically be `if; else` or switch

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,353 @@
...
+		//TODO: This needs to be fetched from the provider secret
+		if azureEnvVars[i].Name == RegistryStorageAzureAccountkeyEnvVarKey {
+			azureEnvVars[i].Value = ""
+		}
+	}
+	return azureEnvVars
+}
+
+func getGCPRegistryEnvVars(bsl *velerov1.BackupStorageLocation, gcpEnvVars []corev1.EnvVar) []corev1.EnvVar {
+
```
dymurray, line 339: NIT: no new line

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,353 @@
...
+		if awsEnvVars[i].Name == RegistryStorageS3SkipverifyEnvVarKey && bsl.Spec.Config[InsecureSkipTLSVerify] != "" {
+			awsEnvVars[i].Value = bsl.Spec.Config[InsecureSkipTLSVerify]
+		}
+	}
+	return awsEnvVars
+}
+
+func getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocation, azureEnvVars []corev1.EnvVar) []corev1.EnvVar {
+
```
dymurray, line 321: NIT: no newline

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,353 @@
...
+	if bsl.Spec.Provider == GCPProvider {
+		envVar = getGCPRegistryEnvVars(bsl, cloudProviderEnvVarMap[GCPProvider])
+	}
+
+	return envVar
+}
+
+func getAWSRegistryEnvVars(bsl *velerov1.BackupStorageLocation, awsEnvVars []corev1.EnvVar) []corev1.EnvVar {
+
```
dymurray, line 286: NIT: no newline

oadp-operator/controllers/registry_test.go
```diff
@@ -0,0 +1,412 @@
...
+	"reflect"
+	"testing"
+)
+
+func TestVeleroReconciler_buildRegistryDeployment(t *testing.T) {
+	tests := []struct {
+		name               string
+		registryDeployment *appsv1.Deployment
+		bsl                *velerov1.BackupStorageLocation
+		wantErr            bool
```
alaypatel07, line 19: do you even plan to test the error? since it is returned from controller-runtime, you can ignore it and remove this field from the struct totally.

oadp-operator/controllers/registry.go
```diff
@@ -0,0 +1,350 @@
...
+	labels := map[string]string{
+		"app.kubernetes.io/name":       OADPOperatorVelero,
+		"app.kubernetes.io/instance":   constructBSLRegistryName(bsl),
+		"app.kubernetes.io/managed-by": OADPOperator,
+		"app.kubernetes.io/component":  Registry,
+	}
+	return labels
+}
+
+func constructBSLRegistryName(bsl *velerov1.BackupStorageLocation) string {
```
alaypatel07, line 225: rename this to registryName()? 

Since it takes BSL as input, we know that it is fro BSL registry, since it returns string, we know it constructs something so dropping those two from name will not hurt IMO

oadp-operator/controllers/registry_test.go
```diff
@@ -0,0 +1,412 @@
...
+}
+
+func TestVeleroReconciler_buildRegistryContainer(t *testing.T) {
+	tests := []struct {
+		name    string
+		bsl     *velerov1.BackupStorageLocation
+		wantErr bool
+	}{
+		{
+			name: "given bsl appropriate container is built or not",
```
dymurray, line 128: NIT: I would explicitly set `wantErr` to `false` just for reading sake

oadp-operator/controllers/registry_test.go
dymurray: May be we want this in future tests, but yeah I would say remove it if it's not being used

shubham-pampattiwar: Removed unused yest args and updated them for all the unit tests.

dymurray (approved):

PR #170, opened 2021-08-09
Fix velero deployment

shubham-pampattiwar:
This PR does the following:
- Refactors Velero deployment controller
- Installs CRDs for non-olm installs
- Fixes registry tests
- Fixes Restic plugin specific update bug
- adds missing apis to scheme

### DIFF
--- a/oadp-operator/pkg/credentials/credentials.go
+++ b/oadp-operator/pkg/credentials/credentials.go
@@ -1,16 +1,19 @@
 package credentials
 
 import (
+	"fmt"
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
 	"github.com/openshift/oadp-operator/pkg/common"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	"os"
 )
 
-type CloudProviderFields struct {
+type DefaultPluginFields struct {
 	secretName         string
 	mountPath          string
 	envCredentialsFile string
+	pluginImage        string
 }
 
 const (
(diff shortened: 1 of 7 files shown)


oadp-operator/pkg/credentials/credentials.go
```diff
@@ -19,43 +22,53 @@ const (
...
 )
 
 func AppendCloudProviderVolumes(velero *oadpv1alpha1.Velero, ds *appsv1.DaemonSet) error {
 	var veleroContainer *corev1.Container
 	// Find Velero container
-	for _, container := range ds.Spec.Template.Spec.Containers {
+	for i, container := range ds.Spec.Template.Spec.Containers {
 		if container.Name == common.Velero {
-			veleroContainer = &container
+			veleroContainer = &ds.Spec.Template.Spec.Containers[i]
```
kaovilai, line 59: How is this different?
`container` is equivalent to `ds.Spec.Template.Spec.Containers[i]`? Or is it just a copy?

shubham-pampattiwar: @kaovilai yes `container` is different than `ds.Spec.Template.Spec.Containers[i]`, `container` is just mutated representation of the actual object in a temp variable accessible inside the `for` loop.

dymurray: Yeah inside a loop that var is just a copy. So this change looks correct

sseago: They're certainly equivalent in content. But since we need a pointer, we need the original. The `range` operator assigns the value of `Containers[i]` to a copy -- the local variable `container`, so `&container` returns the address of the local variable rather than the original:
https://medium.com/@betable/3-go-gotchas-590b8c014e0a

dymurray (approved):

PR #198, opened 2021-08-16
Change parameter scc instead of pointing to new scc object

### LINKED ISSUE #194: ReconcileVeleroSecurityContextContraints is not mutating inside CreateOrUpdate
See: https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L231-L240

Apart from the hilarious comments, `CreateOrUpdate` function expects (see [example](https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#example-CreateOrUpdate)) an actual mutation of the same object instead of pointing to a different scc object.

This is currently blocking successful reconcile of Velero

#### Conditions

| Type | Status | Reason | Message |
| ----- | ------ | -------- | --------- |
| Reconciled | False | Error | ErrorSecurityContextConstraints.security.openshift.io "velero-privileged" is invalid: [runAsUser.type: Invalid value: "": invalid strategy type. Valid values are MustRunAs, MustRunAsNonRoot, MustRunAsRange, RunAsAny, seLinuxContext.type: Invalid value: "": invalid strategy type. Valid values are MustRunAs, RunAsAny]

kaovilai:


kaovilai:
Fixes #194

Issue #223, opened 2021-08-20
travis-ci fail due to missing ansible in golang repo


kaovilai:
When making PR between forks outside openshift org, it can trigger travis-ci builds which currently fails due to lack of ansible file or directory
https://app.travis-ci.com/github/jaidevmane/oadp-operator/builds/235838787
```
Worker information
0.14s0.01s0.00s0.01s
system_info
Build system information
0.00s0.00s0.02s0.00s0.01s0.32s0.00s0.00s0.00s0.01s0.00s0.09stravis0.01s0.77s0.00s0.00s6.03s0.00s2.71s0.00s2.64s
docker_mtu_and_registry_mirrors
resolvconf
services
3.01s$ sudo systemctl start docker
git.checkout
2.13s$ git clone --depth=50 https://github.com/jaidevmane/oadp-operator.git jaidevmane/oadp-operator
0.01stravis_
$ source ~/virtualenv/python3.6/bin/activate
$ python --version
Python 3.6.7
$ pip --version
pip 20.1.1 from /home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pip (python 3.6)
install
15.08s$ pip3 install docker molecule ansible-lint yamllint flake8 openshift jmespath
0.26s$ molecule test -s test-local
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.6.7/bin/molecule", line 5, in <module>
    from molecule.__main__ import main
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/molecule/__main__.py", line 22, in <module>
    from molecule.shell import main
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/molecule/shell.py", line 29, in <module>
    from molecule import command, logger
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/molecule/command/__init__.py", line 27, in <module>
    from molecule.command import base  # noqa
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/molecule/command/base.py", line 31, in <module>
    from ansiblelint.prerun import prepare_environment
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/ansiblelint/prerun.py", line 16, in <module>
    from ansiblelint.config import (
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/ansiblelint/config.py", line 167, in <module>
[...]

PR #225, opened 2021-08-20
GitHub and Travis CI update

kaovilai:
Every commits on public github repo will have tests run so we can double check make test passes on those commits before prow CI finishes.


kaovilai:
See runs example
https://github.com/kaovilai/oadp-operator/actions/workflows/go-tests.yml

shubham-pampattiwar (approved):

kaovilai:
closes #223

PR #227, opened 2021-08-20
Fix velero resource requirements and add velero deployment unit tests

shubham-pampattiwar:


controllers/velero.go
```diff
@@ -537,27 +537,6 @@ func (r *VeleroReconciler) getAppLabels(velero *oadpv1alpha1.Velero) map[string]
 func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) corev1.ResourceRequirements {
 
 	ResourcesReqs := corev1.ResourceRequirements{}
```
kaovilai, line 539: Why not initilize line 539 with line 542

kaovilai (changes requested):

controllers/velero.go
shubham-pampattiwar: Done

kaovilai (approved):

PR #231, opened 2021-08-22
Use Restic DaemonSet name to get object for deletion.  Predicate logic update Closes #218

### LINKED ISSUE #218: `enableRestic` flag  updates on Velero CR are not getting reflected in the environment

kaovilai:
Fix `enableRestic` flag  updates on Velero CR are not getting reflected in the environment  #218

Edit: Interesting behavior below have since been fixed with this PR. The issue was [predicate](https://github.com/openshift/oadp-operator/blob/0148113846537ca268898cb4043f219da48e8ca0/controllers/predicate.go#L10) function returning false instead of true when object is desired to be updated.
Once this is solved, another problem. that came up is [isObjectOurs](https://github.com/openshift/oadp-operator/blob/0148113846537ca268898cb4043f219da48e8ca0/controllers/predicate.go#L28) function was using `object.GetObjectKind().GroupVersionKind()` function which do not work on unintialized object.
https://github.com/openshift/oadp-operator/blob/0148113846537ca268898cb4043f219da48e8ca0/controllers/predicate.go#L29
Fix inspired by https://github.com/kubernetes-sigs/controller-runtime/blob/0a9a7778cdb6d4cca3a3b9f268d90a85aa6f5fdf/pkg/webhook/conversion/conversion.go#L277-L289
related k8s issues:
- https://github.com/kubernetes/kubernetes/issues/80609

> [Fixed] Interesting behavior:
> changing `enableRestic: ` would sometimes only trigger reconcile `oc apply` followed by `oc replace` or vice versa.
> Repeated `oc replace` does not re-trigger reconcile.
> Repeated `oc apply` does not re-trigger reconcile.

CR used to `apply`
```yaml
kind: Velero
metadata:
  name: example-velero
  namespace: oadp-operator-system
spec:
  backupStorageLocations:
    - name: default
      default: true
      backupSyncPeriod: 2m0s
      provider: aws
      objectStorage:
        bucket: <bucket-name>
      credential:
        name: cloud-credentials
        key: cloud
      config:
        profile: default
        s3ForcePathStyle: "true"
        region: us-east
        s3Url: https://s3.private.us-east.cloud-object-storage.appdomain.cloud
  enableRestic: false
  defaultVeleroPlugins:
    - openshift
    - aws

```

**System info**
`oc version`
Client Version: 4.7.19
Server Version: 4.7.19
Kubernetes Version: v1.20.0+87cc9a4

image: quay.io/tkaovila/oadp-operator:restic_enable_false

Closes #218


kaovilai:
Maybe use `ctrl.Results{requeue: true}` in [_velero_controller_ ](https://github.com/openshift/oadp-operator/blob/master/controllers/velero_controller.go#L122)Reconcile function.

kaovilai:
comments from meet: check that objects we own have labels https://github.com/openshift/oadp-operator/blob/master/controllers/predicate.go#L34

controllers/restic.go
```diff
@@ -53,16 +56,23 @@ func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, erro
...
 			// no errors means there already is an existing DaeMonset.
 			// TODO: Check if restic is in use, a backup is running, so don't blindly delete restic.
 			// If velero Spec enableRestic exists and is false, attempt to delete.
-			if err := r.Delete(r.Context, ds); err != nil {
+			deleteOptionPropagationForeground := metav1.DeletePropagationForeground
+			if err := r.Delete(deleteContext, ds, &client.DeleteOptions{PropagationPolicy: &deleteOptionPropagationForeground}); err != nil {
+				r.EventRecorder.Event(ds, v1.EventTypeNormal, "DeleteDaemonSetFailed", "Got DaemonSet to delete but could not delete err:"+err.Error())
 				return false, err
 			}
+			r.EventRecorder.Event(ds, v1.EventTypeNormal, "DeletedDaemonSet", "DaemonSet deleted")
```
shawn-hurley, line 73: I think that we should have more consistent names for this with higher level buckets

kaovilai:
Per discussion controllers/predicate.go needs fixing

controllers/restic.go
```diff
@@ -53,16 +56,23 @@ func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, erro
...
 	}
-
 	if velero.Spec.EnableRestic != nil && !*velero.Spec.EnableRestic {
-
-		if err := r.Get(r.Context, r.NamespacedName, ds); err == nil {
+		deleteContext := context.Background()
+		if err := r.Get(deleteContext, types.NamespacedName{
+			Name:      ds.Name,
+			Namespace: r.NamespacedName.Namespace,
+		}, ds); err == nil {
```
dymurray, line 64: Change this so `if err != nil` record event, and return.

Then run deletion logic outside of conditional.

controllers/restic.go
```diff
@@ -53,17 +57,27 @@ func (r *VeleroReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, erro
...
+				return true, nil
 			}
+			return false, err
 		}
+		// no errors means there already is an existing DaeMonset.
+		// TODO: Check if restic is in use, a backup is running, so don't blindly delete restic.
+		// If velero Spec enableRestic exists and is false, attempt to delete.
+		deleteOptionPropagationForeground := metav1.DeletePropagationForeground
+		if err := r.Delete(deleteContext, ds, &client.DeleteOptions{PropagationPolicy: &deleteOptionPropagationForeground}); err != nil {
+			r.EventRecorder.Event(ds, v1.EventTypeNormal, "DeleteDaemonSetFailed", "Got DaemonSet to delete but could not delete err:"+err.Error())
```
dymurray, line 76: TODO: Come back and fix event recording to be consistent

dymurray:
Overriding so I can fix E2E tests

Issue #185, opened 2021-08-11
Add unsupportedOverrides configuration parameter to backdoor custom images


dymurray:
We need the ability to specify custom images for our dependencies, such as the openshift plugin. To do this in Ansible Operator we allowed the user to specify custom spec values such as `openshift_velero_plugin_fqin` to specify a fully qualified image name to override the value supplied to the operator via environment variables.

For golang operator, since we don't have the ability to pass in any key/value pair we want at the spec, we will want to adopt a pattern seen in other operators with a new spec value called `unsupportedOverrides` which would enable the user to specify these same values in a manner that makes it clear to the user that they are unsupported.

This new spec value should allow for configuration like this:

```
spec:
  openshiftVeleroPluginImage: quay.io/konveyor/openshift-velero-plugin:latest
  csiVeleroPluginImage: quay.io/dymurray/csi-plugin:foobar
... etc
```

PR #243, opened 2021-08-26
enable unsupportedOverrides for Velero image and plugin images

eemcmullan:
#185


pkg/credentials/credentials.go
```diff
@@ -129,7 +195,7 @@ func AppendPluginSpecificSpecs(velero *oadpv1alpha1.Velero, veleroDeployment *ap
 			veleroDeployment.Spec.Template.Spec.InitContainers = append(
 				veleroDeployment.Spec.Template.Spec.InitContainers,
 				corev1.Container{
-					Image:                    pluginSpecificMap.PluginImage,
+					Image:                    getPluginImage(velero),
```
dymurray, line 198: I think this needs to be reworked. You are calling `getPluginImage()` on every loop and `getPluginImage()` is looping through each plugin, finding if the plugin exists, and returning the image name if the plugin is found.

If I have a Velero CR with `aws` and `gcp` plugins installed, `getPluginImage()` will always return the `aws` image because its the first plugin found.

pkg/credentials/credentials.go
```diff
@@ -41,30 +41,96 @@ var (
...
+		return velero.Spec.UnsupportedOverrides[oadpv1alpha1.CSIImageKey]
+	}
+	if velero.Spec.UnsupportedOverrides[oadpv1alpha1.GCPImageKey] != "" {
+		return velero.Spec.UnsupportedOverrides[oadpv1alpha1.GCPImageKey]
+	}
+	// TODO: handle better
+	return ""
+}
+
+func getPluginImage(velero *oadpv1alpha1.Velero) string {
```
dymurray, line 84: I would recommend this function instead takes in the name of the plugin. Then you don't need a loop, and you can have this function return the image for a given plugin name. Something like this:

```
func getPluginImage(pluginName string, velero *oadpv1alpha1.Velero) string {
    switch (pluginName) {
        case oadpv1alpha1.DefaultPluginAWS:
            return getAWSPluginImage(velero)
    }
...
}

func getAWSPluginImage(velero *oadpv1alpha1.Velero) string {
 	if velero.Spec.UnsupportedOverrides[oadpv1alpha1.VeleroAWSPluginImageKey] != "" {
		return velero.Spec.UnsupportedOverrides[oadpv1alpha1.VeleroAWSPluginImageKey]
	}
	if os.Getenv("VELERO_AWS_PLUGIN_REPO") == "" {
		return VeleroAWSPluginImage
	}
	return fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_AWS_PLUGIN_REPO"), os.Getenv("VELERO_AWS_PLUGIN_TAG"))
}
```

And then do this for each of the plugins, adding each condition to the switch statement

pkg/credentials/credentials.go
dymurray: With this comment: https://github.com/openshift/oadp-operator/pull/243/files#r698782601

Now you could do:
```
getPluginImage(pluginSpecificMap.PluginName, velero)
```

pkg/credentials/credentials.go
```diff
@@ -32,7 +33,6 @@ var (
 			MountPath:          "/credentials",
 			EnvCredentialsFile: common.AWSSharedCredentialsFileEnvKey,
 			PluginName:         common.VeleroPluginForAWS,
-			PluginImage:        fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_AWS_PLUGIN_REPO"), os.Getenv("VELERO_AWS_PLUGIN_TAG")),
```
dymurray, line 35: Alternatively this just calls `getAWSPluginImage()`

pkg/credentials/credentials.go
```diff
@@ -41,30 +41,96 @@ var (
...
+func getPluginImage(velero *oadpv1alpha1.Velero) string {
+	// TODO: this assumes no velero image override..handle better
+	if velero.Spec.UnsupportedOverrides != nil {
+		return getPluginOverrideImage(velero)
+	}
+	// AWS
+	for _, plugin := range velero.Spec.DefaultVeleroPlugins {
+		if plugin == oadpv1alpha1.DefaultPluginAWS {
+			if os.Getenv("VELERO_AWS_PLUGIN_REPO") == "" {
+				return "quay.io/konveyor/velero-plugin-for-aws:konveyor-oadp"
```
dymurray, line 93: These are constants in the `common` package as well, I would use those when possible

controllers/velero_test.go
```diff
@@ -133,7 +134,7 @@ func TestVeleroReconciler_buildVeleroDeployment(t *testing.T) {
 							Containers: []corev1.Container{
 								{
 									Name:            common.Velero,
-									Image:           getVeleroImage(),
+									Image:           fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_REPO"), os.Getenv("VELERO_TAG")),
```
dymurray, line 137: We want this to use `getVeleroImage()` still. For this test it can pass an empty CR and it should just return the default

pkg/credentials/credentials.go
```diff
@@ -41,30 +41,97 @@ var (
...
+	}
+	if os.Getenv("VELERO_OPENSHIFT_PLUGIN_REPO") == "" {
+		return common.AzurePluginImage
+	}
+	return fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_AZURE_PLUGIN_REPO"), os.Getenv("VELERO_AZURE_PLUGIN_TAG"))
+}
+
+func getPluginImage(pluginName string, velero *oadpv1alpha1.Velero) string {
+	switch pluginName {
+
```
dymurray, line 116: NIT: remove empty line

pkg/credentials/credentials.go
```diff
@@ -129,7 +196,7 @@ func AppendPluginSpecificSpecs(velero *oadpv1alpha1.Velero, veleroDeployment *ap
 			veleroDeployment.Spec.Template.Spec.InitContainers = append(
 				veleroDeployment.Spec.Template.Spec.InitContainers,
 				corev1.Container{
-					Image:                    pluginSpecificMap.PluginImage,
+					Image:                    getPluginImage(pluginSpecificMap.PluginName, velero),
```
dymurray, line 199: +1 much better

Issue #252, opened 2021-08-30
BSL Validation to ensure BSL:Provider mapping is 1:1


shubham-pampattiwar:
Currently OADP assumes as well as only supports one BSL per provider, thus in order to ensure this we need a `BackupstorageLocation` specific validation.

PR #254, opened 2021-08-30
Add bsl-provider mapping validation

shubham-pampattiwar:
This validation ensures that there is 1:1 mapping relationship between BSL and provider.
Fixes https://github.com/openshift/oadp-operator/issues/252


controllers/bsl.go
```diff
@@ -215,3 +221,23 @@ func (r *VeleroReconciler) validateProviderPluginAndSecret(bslSpec velerov1.Back
...
 	return nil
 }
+
+func (r *VeleroReconciler) ensureBSLProviderMapping(velero *oadpv1alpha1.Velero) error {
+
+	providerBSLMap := map[string]int{}
+	for _, bsl := range velero.Spec.BackupStorageLocations {
+		provider := bsl.Provider
+
+		if providerBSLMap[provider] == 0 {
```
dymurray, line 231: Why do you need this? Can't you just always run `providerBSLMap[provider]++`?

shubham-pampattiwar: Done

dymurray (approved):

PR #255, opened 2021-08-30
ci-Dockerfile for test image 

hhpatel14:


alaypatel07 (changes requested):
we need to update the make file with a `test-ci` command

alaypatel07:
also, the tests on this PRs kind of useless as of now

build/ci-Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+RUN go mod download
+
+# Copy the go source
+COPY main.go main.go
+COPY api/ api/
+COPY pkg/ pkg/
+COPY controllers/ controllers/
+COPY tests/ tests/
+
+RUN make envtest controller-gen kustomize
```
alaypatel07, line 21: "manifests generate fmt vet envtest" you have to run those here as well right?

Makefile
```diff
@@ -103,6 +103,10 @@ test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) -p path)" go test -mod=mod ./controllers/... -coverprofile cover.out
```
alaypatel07, line 105: You will need the ENVTEST variable defined before using it here like this https://github.com/openshift/oadp-operator/blob/ef445b09b303262d27f59c84a78e4c0a9e01bcb6/Makefile#L144

alaypatel07: if you don't, the test-ci command might fail eventually

alaypatel07 (approved):
/lgtm

PR #268, opened 2021-09-02
Add restore resource priorities configuration option

### LINKED ISSUE #176: Add restore resource priorities configuration option
Investigate if this is still configurable in Velero 1.6. If so, we want to add this as an installation parameter on our Velero CR.

kaovilai:
- add feature flag to velero pod `--features` if restoreResourcesVersionPriority configmap is defined (does not change user applied velero CR)
- implemented removeDuplicateValues function

Closes #176


controllers/restore_resource_version_priority.go
```diff
@@ -0,0 +1,52 @@
...
+			Name:      enableApiGroupVersionsConfigMapName,
+			Namespace: velero.Namespace,
+		},
+	}
+	// Create ConfigMap
+	op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, &configMap, func() error {
+		if err := controllerutil.SetControllerReference(velero, &configMap, r.Scheme); err != nil {
+			return err
+		}
+		configMap.Data = make(map[string]string, 1)
```
dymurray, line 36: Won't this always overwrite the configmap.Data field? Would it better to check if it's `nil` first?

controllers/velero.go
```diff
@@ -355,19 +355,13 @@ func (r *VeleroReconciler) buildVeleroDeployment(veleroDeployment *appsv1.Deploy
...
-			for _, featureFlag := range velero.Spec.VeleroFeatureFlags {
-				if featureFlag == enableCSIFeatureFlag {
-					foundCSIFeatureFlag = true
-					break
-				}
-			}
-			if !foundCSIFeatureFlag { // Not Found so append to feature flag
-				velero.Spec.VeleroFeatureFlags = append(velero.Spec.VeleroFeatureFlags, enableCSIFeatureFlag)
-			}
+			velero.Spec.VeleroFeatureFlags = append(velero.Spec.VeleroFeatureFlags, enableCSIFeatureFlag)
```
dymurray, line 358: Sorry if I'm missing something, if I set `enableCSIFeatureFlag` on the spec, then won't this add it twice if the plugin is found no matter what?

dymurray: Oh ignore, I see `removeDuplicateValues` now

dymurray (approved):
overall LGTM, couple of comments

controllers/restore_resource_version_priority.go
kaovilai: How would you suggest I handle the case if it's not nill? Don't we need to follow whatever CR says?

kaovilai: Analyze current configmap Data and append to it maybe.
If there is a conflicting key which already exists, the corresponding values in the CR take precedence

dymurray: Alright, good points. fine by me

PR #275, opened 2021-09-07
Ignore IDE dotfiles

kaovilai:
Allows dev to leave dotfiles in their local repo


shubham-pampattiwar (changes requested):
can you please add `.idea/` as well ?

shubham-pampattiwar (approved):

Issue #181, opened 2021-08-11
Add VolumeSnapshotLocation validation logic to reconciler


dymurray:
The ValidateVolumeSnapshotLocation reconciler needs to validate that the declared VSLs in the spec are valid for their given cloud provider and that their associated provider plugin is enabled.

If a user has supplied known bad value for the VSL config, then we will want to present an error to the user on the status subresource. If a user has added a VSL for `aws` provider and no `aws` default plugin is specified, we will want to warn the user that there may be a misconfiguration issue.

Here are all the documentation on the valid VSL configuration values:
https://github.com/vmware-tanzu/velero-plugin-for-aws/blob/main/volumesnapshotlocation.md
https://github.com/vmware-tanzu/velero-plugin-for-gcp/blob/main/volumesnapshotlocation.md
https://github.com/vmware-tanzu/velero-plugin-for-microsoft-azure/blob/main/volumesnapshotlocation.md

jaidevmane:
assign

PR #278, opened 2021-09-07
Fix VSL validation & add VSL validation unit tests

eemcmullan:
#181


controllers/vsl.go
```diff
@@ -11,6 +12,16 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+// provider specific object storage
+const (
+	Profile          = "profile"
```
dymurray, line 17: For these, also include for which cloud provider the constant is for.

something like:
```
AwsProfile
AzureSubscriptionId
```
```

File: controllers/vsl.go [hunk 2]
```diff
@@ -19,61 +30,91 @@ func (r *VeleroReconciler) ValidateVolumeSnapshotLocations(log logr.Logger) (boo
...
+			},
+			Spec: vslSpec,
+		}
 
-			//AWS
-			if vslSpec.Provider == "aws" {
+		// check for valid provider
+		if vslSpec.Provider != AWSProvider && vslSpec.Provider != GCPProvider &&
+			vslSpec.Provider != Azure {
+			return false, errors.New("invalid provider for VSL")
```
dymurray, line 46: I'm wondering if a user configures a custom VSL plugin, we may be preventing them from adding it with this logic.

I think it would be better if we warned the user here instead. Maybe we trigger an event? Or warn the user on the status field somehow.

dymurray: Actually looking below we fire an event for something similar here: https://github.com/openshift/oadp-operator/pull/278/files#diff-9d604b6817079fa43c57dd7d9eaa35f87b4a609d7dae607aa4b64ca00e3d2cf9R66

Lets do that

controllers/vsl.go
```diff
@@ -19,61 +30,91 @@ func (r *VeleroReconciler) ValidateVolumeSnapshotLocations(log logr.Logger) (boo
...
 
-				//checking the azure plugin, if not present, throw warning message
-				if !contains(velero.Spec.DefaultVeleroPlugins, "azure") {
-					r.Log.Info("VSL for Azure specified, but Azure plugin not present, might be a misconfiguration")
+		//Azure
+		if vslSpec.Provider == Azure {
+			// no other required fields for gcp
+			// check for invalid config key
+			for key := range vslSpec.Config {
+				if key != ApiTimeout && key != ResourceGroup &&
```
dymurray, line 101: Since this is only 4 keys the chained if conditions aren't terrible, but consider (in case more keys are added later)

```
validAzureKeys := []string{ApiTimeout, ResourceGroup, ...}

or

validAzureKeys := map[string]bool{
    ResourceGroup: true,
...
}
```
Then you can either use the `contains` func or just do `valid := validAzureKeys[key]` to see if its a valid key

dymurray (approved):

PR #287, opened 2021-09-10
Add Backup/Restore test suite

dymurray:
This needs to be reviewed. Just the current state of the test suite that was worked on together. I've ran this 20+ times without flaking after much trial and error so feel pretty good about that, I just have a couple things I want to update.


tests/e2e/install_suite_test.go
```diff
@@ -21,60 +26,199 @@ var _ = BeforeSuite(func() {
 
 	vel = &veleroCustomResource{
```
dymurray, line 27: This is a poorly named struct var to setup the testing context. I would like to rework this slightly.

dymurray:
Moving off WIP so that testing @hhpatel14 's PR actually runs e2e tests we are confident will pass

Issue #315, opened 2021-09-17
CI: Run only necessary tests


kaovilai:
- presubmits
  - on `*.go` `go.mod` `go.sum` `Makefile` `Dockerfile` `*.yaml` `*.yml` `build/*` changes
    - operator-e2e
    - images
  - on `*.go` `go.mod` `go.sum` `Makefile` changes
    - operator-unit-test

kaovilai:
ci-index is still running on .md changes.
ci-index is only dependent on by e2e tests therefore should run when the same file changes
- `*.go` `go.mod` `go.sum` `Makefile` `Dockerfile` `*.yaml` `*.yml` `build/*` changes

PR #320, opened 2021-09-21
What would CI do .md

kaovilai:
verifying https://github.com/openshift/oadp-operator/issues/315


kaovilai:
1 expected check YAS!!

PR #359, opened 2021-10-04
Adding minor change to local dev environment docs

savitharaghunathan:
Adding a line to make sure that the oadp-operator repo in quay.io is set to public.

### DIFF
--- a/docs/developer/local_dev.md
+++ b/docs/developer/local_dev.md
@@ -17,7 +17,9 @@
     ```
 
     *Note:* <IMAGE_ID> can be found out by running `podman images` after the image has been built.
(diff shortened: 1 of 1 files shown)


docs/developer/local_dev.md
```diff
@@ -17,7 +17,9 @@
     ```
 
     *Note:* <IMAGE_ID> can be found out by running `podman images` after the image has been built.
-    <IMAGE_TAG> can be any tag that you would would like to assign to the image.
+    <IMAGE_TAG> can be any tag that you would would like to assign to the image. If the `oadp-operator`
```
dymurray, line 20: I would update this to say:
>Ensure that the registry which is hosting the container image is publicly available such that the image can be pulled from your OpenShift cluster

Issue #331, opened 2021-09-22
Configuring cloud plugins without credentials causes invalid volumeMounts


dymurray:
**Describe the bug**
If I set `gcp`, `azure`, and `aws` as default velero plugins, but don't configure their respective secrets, the Velero CR shows no errors but the velero deployment, restic daemonset, and registry deployment fail to startup.

```
[dymurray@galt oadp-operator]$ oc get pods
NAME                                                 READY     STATUS              RESTARTS   AGE
oadp-operator-7dc5cb884c-f5tvb                       1/1       Running             0          10m
oadp-velero-sample-1-aws-registry-68b6579874-j6mws   0/1       CrashLoopBackOff    6          5m57s
restic-4sg7h                                         0/1       ContainerCreating   0          7m26s
restic-cg8bq                                         0/1       ContainerCreating   0          7m26s
restic-skt2d                                         0/1       ContainerCreating   0          7m26s
velero-74f64fcd5c-mq2lk                              0/1       Init:0/5            0          7m26s

```

The errors:
```
  Warning  FailedMount  1m (x2 over 3m)   kubelet, ip-10-0-155-20.ec2.internal  Unable to attach or mount volumes: unmounted volumes=[cloud-credentials-gcp cloud-credentials-azure], unattached volumes=[scratch certs cloud-credentials-gcp cloud-credentials-azure cloud-credentials velero-token-wf2sn host-pods]: timed out waiting for the condition

```
**To Reproduce**
Set default Velero plugins:
```
- csi
- gcp
- azure
```
But only configure AWS credentials.

**Expected behavior**
I would expect the Velero CR to display an error that I've misconfigured the CR.

**Version of operator being used**
 - 0.3.0

dymurray:
To resolve this... we want to add to the `ValidateBackupStorageLocation` reconciler logic that will check if a specific cloud provider plugin is specified, that the associated expected secret exists before trying to mount it in the container.

i.e. if I specify:
```
defaultVeleroPlugins:
- aws
- gcp
```

Then we must verify `cloud-credentials` and `cloud-credentials-gcp` secrets exist.

PR #369, opened 2021-10-05
WIP: Fixing volume mount issue

savitharaghunathan:
Fixes issue #331 

Work in progress. Do not merge 

/hold


controllers/velero.go
```diff
@@ -531,3 +530,38 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
...
+	velero := oadpv1alpha1.Velero{}
+	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
+		return false, err
+	}
+
+	secretName := "nosecret"
+	var defaultPlugin oadpv1alpha1.DefaultPlugin
+	for _, defaultPlugin = range velero.Spec.DefaultVeleroPlugins {
+		//Switching only for supported cloud providers
+		switch defaultPlugin {
```
shubham-pampattiwar, line 547: As we want to run this validation only for the default cloud provider plugins lets use the following pattern that we have used for cloud provider specific things:
```
for _, plugin = range velero.Spec.DefaultVeleroPlugins {
		if pluginSpecificMap, ok := PluginSpecificFields[plugin]; ok && pluginSpecificMap.IsCloudProvider {
			secretName := pluginSpecificMap.SecretName
		}
	}
```

shubham-pampattiwar (changes requested):

controllers/velero.go
```diff
@@ -531,3 +530,31 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
 
 	return ResourcesReqs
 }
+
+// For later: Move this code into validator.go when more need for validation arises
+// TODO: if multiple default plugins exist, ensure we validate all of them.
+// Right now its sequential validation
+func (r *VeleroReconciler) ValidateVeleroDeployment(log logr.Logger) (bool, error) {
```
dymurray, line 537: Consider renaming this to `ValidateVeleroPiugins` since that's what it's doing today. We can then add a generic `ValidateVeleroConfig` or something down the line thats calls this when we have more validations to add.

controllers/velero.go
```diff
@@ -531,3 +530,31 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
...
+	velero := oadpv1alpha1.Velero{}
+	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
+		return false, err
+	}
+
+	secretName := "nosecret"
```
dymurray, line 543: I would initialize this as an empty string `secretName := ""`

controllers/velero.go
```diff
@@ -531,3 +530,31 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
...
+	var defaultPlugin oadpv1alpha1.DefaultPlugin
+	for _, plugin := range velero.Spec.DefaultVeleroPlugins {
+		if pluginSpecificMap, ok := credentials.PluginSpecificFields[plugin]; ok && pluginSpecificMap.IsCloudProvider {
+			secretName = pluginSpecificMap.SecretName
+		}
+	}
+
+	if secretName != "nosecret" {
```
dymurray, line 551: `if secretName != ""`

controllers/velero.go
savitharaghunathan: This was my initial PR. I had an empty string and for the condition, I equated the length of the string to zero. It failed on Openshift clusters < 4.7 but passed in my local cluster (which is running 4.8.x). Will try what you have suggested. Thanks @dymurray

controllers/velero.go
```diff
@@ -531,3 +530,31 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
...
+	velero := oadpv1alpha1.Velero{}
+	if err := r.Get(r.Context, r.NamespacedName, &velero); err != nil {
+		return false, err
+	}
+
+	secretName := ""
+	var defaultPlugin oadpv1alpha1.DefaultPlugin
+	for _, plugin := range velero.Spec.DefaultVeleroPlugins {
+		if pluginSpecificMap, ok := credentials.PluginSpecificFields[plugin]; ok && pluginSpecificMap.IsCloudProvider {
+			secretName = pluginSpecificMap.SecretName
```
shubham-pampattiwar, line 547: @savitharaghunathan Do you mind using the function specified here for getting the secret name, this function assumes the default name and key but also checks if the user has specified a different name and key for the particular provider https://github.com/openshift/oadp-operator/blob/7439793284b7112f22427478c5a275881fc04e2a/controllers/registry.go#L476

savitharaghunathan: @shubham-pampattiwar To get the `credential *corev1.SecretKeySelector` value, do I have to loop through bslspec? If so, then how it is different from the validations we are performing in the `bsl.go` ?

shubham-pampattiwar: @savitharaghunathan So there are 2 types of validations here based on the relationships:
1. _Cloud Provider Plugin - Secret Validation:_ This validation should make sure that there exists a credential secret if a cloud plugin is specified in the Velero CR.
2. _BSL - Secret Validation:_ This Validation should make sure that there exists a credential secret for an instance of BSL.

We already have the Second validation in place, but we do not have the first one. This PR should be implementing the first one, right ?

savitharaghunathan: Updated the code to reflect the changes we discussed yesterday.

controllers/velero.go
```diff
@@ -531,3 +530,31 @@ func (r *VeleroReconciler) getVeleroResourceReqs(velero *oadpv1alpha1.Velero) co
...
+		}
```
dymurray, line 548: You can add a `break` here since we found the plugin

PR #388, opened 2021-10-11
add golangci linters to github actions ci

weshayutin:
Information re: golangci
https://golangci-lint.run/

The configuration is set to only lint
new code atm.  Let's not fail jobs on unrelated
old code until this is well socialized. See
inline comments in config files.

Test execution in github:
https://github.com/weshayutin/oadp-operator/pull/4/checks?check_run_id=3860994813

To view lint issues in code base
remove param, "new-from-rev: HEAD~1"
from .golangci.yml and execute locally

### DIFF
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -1,0 +1,34 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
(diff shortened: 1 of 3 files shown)


dymurray:
I'm going to close this since we don't have github actions. With travis enabled I think it would be cool to revisit this as a travis job.

PR #394, opened 2021-10-13
adding docs for custom plugin images

jaidevmane:
Adding documentation outlining custom plugin images support in OADP Operator


kaovilai:
Is it possible to list where to find custom images?

jaidevmane:
Do you think a good approach would be to list the repos of the plugins so developers can work on the source code, make tweaks and use them in the above?

kaovilai:
> Do you think a good approach would be to list the repos of the plugins so developers can work on the source code, make tweaks and use them in the above?

Sure. I would link quay repo + GitHub repo of konveyor vetted plugins. It's more for people who are not on the same schedule as us to use say the newest possible velero image.

PR #398, opened 2021-10-14
e2e Prevent namespace terminating errors when looping on 1 test

kaovilai:
when looping on one test, it is possible to get namespace terminating errors when installing application.


kaovilai:
if an app failed to backup.. it would not have gotten to uninstall phase.. and will need to implement namespace removal before each run or after as well.

PR #407, opened 2021-10-19
Fix B/R e2e tests

eemcmullan:


tests/e2e/backup_restore_suite_test.go
```diff
@@ -91,11 +91,13 @@ var _ = Describe("AWS backup restore tests", func() {
 
 			// wait for backup to not be running
 			Eventually(isBackupDone(vel.Client, namespace, backupName), timeoutMultiplier*time.Minute*4, time.Second*10).Should(BeTrue())
-			Expect(getVeleroContainerFailureLogs(vel.Namespace)).To(Equal([]string{}))
 
 			// check if backup succeeded
 			succeeded, err := isBackupCompletedSuccessfully(vel.Client, namespace, backupName)
```
kaovilai, line 96: This would never return `false, nil`.
It would always print an error here but the error will not contain pod logs.

pod logs should be collected and printed out either in these functions when returning false, or before running these functions.

tests/e2e/backup_restore_suite_test.go
```diff
@@ -112,11 +114,13 @@ var _ = Describe("AWS backup restore tests", func() {
 			err = createRestoreFromBackup(vel.Client, namespace, backupName, restoreName)
 			Expect(err).ToNot(HaveOccurred())
 			Eventually(isRestoreDone(vel.Client, namespace, restoreName), timeoutMultiplier*time.Minute*4, time.Second*10).Should(BeTrue())
-			Expect(getVeleroContainerFailureLogs(vel.Namespace)).To(Equal([]string{}))
 
 			// Check if restore succeeded
 			succeeded, err = isRestoreCompletedSuccessfully(vel.Client, namespace, restoreName)
```
kaovilai, line 119: This would never return `false, nil`.
It would always print an error here but the error will not contain pod logs.

pod logs should be collected and printed out either in these functions when returning false, or before running these functions.

kaovilai (changes requested):
Right now

PR #410, opened 2021-10-21
Enable ingestion of the exposed metrics by openshift-montoring

shubham-pampattiwar:
This PR enable the ingestion of the velero metrics by openshift monitoring. We are deploying the service and serviceMonitor required for the metrics to be scraped by prometheus. We propose the following metrics for the monitoring and telemetry purposes:
```
"velero_backup_total"
"velero_restore_total"
```


config/rbac/role.yaml
```diff
@@ -143,3 +146,28 @@ rules:
...
+  verbs:
+  - get
+  - create
+  - list
+  - delete
+  - update
+  - patch
+  - watch
+- apiGroups:
+  - rbac.authorization.k8s.io
```
shawn-hurley, line 162: I worry about adding these permissions to the operator, is there any other way of achieving this?

smarterclayton:
Is this cardinality 13 or higher for series count planned to send to telemetry?

shawn-hurley:
@smarterclayton is the cardinality of 13 or lower for each metric, or in total for all metrics?

shawn-hurley:
@smarterclayton never mind you said series, not metrics sorry

IIUC this means that that we can only send 13 series to monitoring. The key distinction here is going to be that when a label is added to the data, you increase the cardinality by 1.

config/rbac/role.yaml
shubham-pampattiwar: @shawn-hurley we need these permissions so that the operator creates the appropriate role/rolebindings needed for the prometheus operator to scrape the metrics from our ns.

shawn-hurley: Hmm, isn't that already done by OLM when installing if you have the correct annotation?

shubham-pampattiwar:
> Is this cardinality 13 or higher for series count planned to send to telemetry?

@smarterclayton After re-evaluating our telemetry needs, we have decided to use only two metrics to be ingested by openshift-monitoring: `velero_backup_total` and `velero_restore_total` . Both these metrics do not have any kind of label, so I believe the cardinality would be 2.

config/rbac/role.yaml
shubham-pampattiwar: @shawn-hurley PTAL I have removed the prometheus role/rolebinding creation by OADP operator.

dymurray (approved):

PR #428, opened 2021-11-01
Install_suite e2e Refractor

savitharaghunathan:
* Overall refactoring of Install_suite e2e
* Added more checks to Velero deployment


eemcmullan (approved):

tests/e2e/velero_deployment_suite_test.go
```diff
@@ -0,0 +1,345 @@
...
+	. "github.com/onsi/gomega"
+	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
+	velero "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+)
+
+var _ = Describe("Configuration testing for Velero Custom Resource", func() {
+
+	var _ = BeforeEach(func() {
```
dymurray, line 18: I'm wondering if this would be better in `BeforeSuite`. This way we create the secret before the test suite runs, and then we don't need to keep creating it before each test. What do you think?

tests/e2e/velero_helpers.go
```diff
@@ -247,3 +248,58 @@ func (v *veleroCustomResource) IsDeleted() wait.ConditionFunc {
 		return false, err
 	}
 }
+
+//check if bsl matches the spec
+func doesBSLExist(namespace string, bsl velero.BackupStorageLocationSpec, spec *oadpv1alpha1.VeleroSpec) wait.ConditionFunc {
+	return func() (bool, error) {
+		if len(spec.BackupStorageLocations) > 0 {
```
dymurray, line 255: NIT: 
```
if len(spec.BackupStorageLocations) == 0 {
  return false, errors.New()
}
for _, b := range ....
```

tests/e2e/velero_helpers.go
```diff
@@ -247,3 +248,58 @@ func (v *veleroCustomResource) IsDeleted() wait.ConditionFunc {
...
+		return false, errors.New("no backup storage location configured. Expected BSL to be configured")
+
+	}
+}
+
+//check if vsl matches the spec
+func doesVSLExist(namespace string, vslspec velero.VolumeSnapshotLocationSpec, spec *oadpv1alpha1.VeleroSpec) wait.ConditionFunc {
+	return func() (bool, error) {
+
+		if len(spec.VolumeSnapshotLocations) > 0 {
```
dymurray, line 275: NIT: if the length of VSLs = 0, return an error. Then you don't have to indent everything else

dymurray (commented):
Overall this looks GREAT. It's really cool you were able to get so many working entries with this refactor. I had one question about the BeforeEach func, and a couple nitpicks. Other than that LGTM!

tests/e2e/velero_deployment_suite_test.go
savitharaghunathan: Makes sense. It should be a part of initial setup.

PR #433, opened 2021-11-03
Refactor OADP API to support future use cases

dymurray:


api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+	ResourceAllocations corev1.ResourceRequirements `json:"resourceAllocations,omitempty"`
+	// Timeout defines the Restic timeout, default value is 1h
 	// +optional
-	ResticResourceAllocations corev1.ResourceRequirements `json:"resticResourceAllocations,omitempty"`
-	// ResticTimeout defines the Restic timeout, default value is 1h
+	Timeout string `json:"timeout,omitempty"`
+}
+
+type ServerConfig struct {
+	*VeleroConfig
```
shawn-hurley, line 100: I think this might have issues,

```ServerConfig {
   tolerations: .......
  }
```

Will be undefined for either the velero config or restic config. I think that we should have NamedFields for these.

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+	*ResticConfig
+}
+
+type BackupStorage struct {
+	*velero.BackupStorageLocationSpec
```
shawn-hurley, line 105: I think this should be named to start with, otherwise adding some after the fact will be hard I think.

shawn-hurley (commented):
Except for the naming of the fields instead of inlining them I think this looks good to me!

api/v1alpha1/velero_types.go
dymurray: Great point, thanks Shawn

api/v1alpha1/velero_types.go
dymurray: Great point

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+type VeleroConfig struct {
+	// FeatureFlags defines the list of features to enable for Velero instance
 	// +optional
-	VeleroFeatureFlags []string `json:"veleroFeatureFlags,omitempty"`
-	// VeleroTolerations defines the list of tolerations to be applied to Velero deployment
+	FeatureFlags []string `json:"featureFlags,omitempty"`
+	// Tolerations defines the list of tolerations to be applied to Velero deployment
 	// +optional
-	VeleroTolerations []corev1.Toleration `json:"veleroTolerations,omitempty"`
+	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
```
alaypatel07, line 63: feels to me like Tolerationsp[], ResourceAllocations[], and NodeSelector could be common for both velero deployment as well as restic demonsets. I wonder if it is a good idea to do the following?
```
type PodConfig {
    Tolerations
    ResourceAllocations
    NodeSelector
}

type VeleroConfig struct {
    *PodConfig
    ...
}

type ResticConfig struct {
    *PodConfig
}
```

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
-	// ResticResourceAllocations defines the CPU and Memory resource allocations for the restic Pod
+	ResourceAllocations corev1.ResourceRequirements `json:"resourceAllocations,omitempty"`
+	// Timeout defines the Restic timeout, default value is 1h
 	// +optional
-	ResticResourceAllocations corev1.ResourceRequirements `json:"resticResourceAllocations,omitempty"`
-	// ResticTimeout defines the Restic timeout, default value is 1h
+	Timeout string `json:"timeout,omitempty"`
+}
+
+type ServerConfig struct {
```
alaypatel07, line 99: NIT: Wonder if `ComponentConfig` if more expressive?

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+
+type VolumeSnapshot struct {
+	*velero.VolumeSnapshotLocationSpec
+}
+
+// VeleroSpec defines the desired state of Velero
+type BackupApplicationSpec struct {
+	// BackupStorageLocations defines the list of desired configuration to use for BackupStorageLocations
+	// +optional
+	BackupStorageLocations []BackupStorage `json:"backupStorageLocations"`
```
alaypatel07, line 116: NIT: since we are creating a new type `BackupStorage`, should we consider moving away from naming the field `backupStorageLocations`? `spec.backupStorageLocations` feels very close to a velero storage location API I wonder when we have our own configurations in the future if we would want it to be named like this.

api/v1alpha1/velero_types.go
```diff
@@ -109,44 +132,40 @@ type VeleroSpec struct {
...
 // VeleroStatus defines the observed state of Velero
-type VeleroStatus struct {
+type BackupApplicationStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
 
 // Velero is the Schema for the veleroes API
```
alaypatel07, line 151: nit: doc change required

api/v1alpha1/velero_types.go
```diff
@@ -109,44 +132,40 @@ type VeleroSpec struct {
...
-type Velero struct {
+type BackupApplication struct {
```
alaypatel07, line 152: NIT: consider a generic name not tied to backup or restore? `DataProtectionApplication` we have named the project oadp so having the main CR like that would be in sync, but I'm not too tied to naming.

api/v1alpha1/velero_types.go
dymurray, line 105: Include a nested metadata type to support setting name, labels, annotations, etc...

api/v1alpha1/velero_types.go
```diff
@@ -109,44 +132,40 @@ type VeleroSpec struct {
...
 }
 
 // VeleroStatus defines the observed state of Velero
-type VeleroStatus struct {
+type BackupApplicationStatus struct {
 	Conditions []metav1.Condition `json:"conditions,omitempty"`
 }
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
```
alaypatel07, line 150: I wonder if we should add `shortName=...`, `printerColumn` to the list of kube builder annotations before GA. Having some crude validation a the API level will only be great https://book.kubebuilder.io/reference/generating-crd.html#validation

alaypatel07 (commented):
I have not spent a lot of time with this so I'm not super tied to any suggestion, but added some nits and questions on renaming/struct composition.

api/v1alpha1/velero_types.go
shubham-pampattiwar: +1

api/v1alpha1/velero_types.go
shubham-pampattiwar, line 152: Just a couple of naming suggestions (existing ones are cool too):
`BackupApplication` --> `BackupRestoreApplication` or `DataProtectionApp` (as @alaypatel07 suggested)
`BackupStorage` --> `BackupDataStorage` or just `DataStorage`
`VolumeSnapshot` --> `VolumeSnapshotStorage`or jsut `VolumeStorage`

dymurray: `DataProtectionApplication`

api/v1alpha1/velero_types.go
dymurray, line 116: Change this to:
```
BackupLocations []BackupLocation
```

```
type BackupLocation struct {
    VeleroBackupStorageLocation *velero.BackupStorageLocation
}
```

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+	*VeleroConfig
+	*ResticConfig
+}
+
+type BackupStorage struct {
+	*velero.BackupStorageLocationSpec
+}
+
+type VolumeSnapshot struct {
+	*velero.VolumeSnapshotLocationSpec
```
dymurray, line 109:
```
type VolumeSnapshot struct {
    Velero *velero.VolumeSnapshotLocationSpec
    DataMover *oadp.DataMoverConfig
    ...
}
```

api/v1alpha1/velero_types.go
dymurray: I decided to go with ApplicationConfig in spirit of `DataProtectionApplication`

api/v1alpha1/velero_types.go
```diff
@@ -54,47 +54,70 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
-	// VolumeSnapshotLocations defines the list of desired configuration to use for VolumeSnapshotLocations
-	// +optional
-	VolumeSnapshotLocations []velero.VolumeSnapshotLocationSpec `json:"volumeSnapshotLocations"`
-	// VeleroFeatureFlags defines the list of features to enable for Velero instance
+type VeleroConfig struct {
+	// FeatureFlags defines the list of features to enable for Velero instance
 	// +optional
-	VeleroFeatureFlags []string `json:"veleroFeatureFlags,omitempty"`
-	// VeleroTolerations defines the list of tolerations to be applied to Velero deployment
+	FeatureFlags []string `json:"featureFlags,omitempty"`
```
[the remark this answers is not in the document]
dymurray: I think that should belong outside of the `VeleroConfig` object. These flags are used explicitly within velero, and eventually we may support other OADP specific features we can put in OADP specific config.

api/v1alpha1/groupversion_info.go
```diff
@@ -28,7 +28,7 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "oadp.openshift.io", Version: "v1alpha1"}
 
-	Kind = "Velero"
```
kaovilai, line 31: Should this be v1alpha2? A new group version.

api/v1alpha1/oadp_types.go
```diff
@@ -54,47 +54,77 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+}
+
+// BackupLocation defines the configuration for the DPA backup storage
+type BackupLocation struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Velero            *velero.BackupStorageLocationSpec `json:"velero"`
+}
+
+// VolumeSnapshot defines the configuration for the DPA snapshot store
+type VolumeSnapshot struct {
```
kaovilai, line 114: Going with XLocation pattern. Should this be SnapshotLocation? DiskSnapshotLocation, StorageSnapshotLocation

api/v1alpha1/groupversion_info.go
dymurray: technically its a new API... so unsure if it matters

PR #438, opened 2021-11-07
Rewrite Travis Build

jmontleon:


jmontleon:
This has run several times here without issue:
https://app.travis-ci.com/github/konveyor/travis-multiarch-test/builds

The failures seem to come down to two issues causing the failures:
- The alternate architectures run using LXD for the most part. The output for the LXD containers seems to get buffered. If enough output is not generated within a ten minute period that the console output needs to be written out the job will fail on a 10m00s timeout Travis watches for to detect hung tasks. There is a travis_wait script that will write a line every minute to prevent jobs from failing from timeouts, but as this is run in the LXD container it's not nearly enough to force the buffer to be written out. But using -v for dependency download and build produces plenty of output to circumvent the issue.
- The second issue is that there appear to be pretty frequent network connection errors downloading dependencies within the container, so I run `go mod vendor`, add the folder to the container, and adjust the build accordingly. I haven't seen any network issues since doing this.

PR #442, opened 2021-11-08
Adding profile support for registry 

### LINKED ISSUE #435: Add support for multiple AWS credential profiles
**Is your feature request related to a problem? Please describe.**

Currently, OADP Operator does not support multiple AWS credential profiles for the secret to be used for BSL/VSL/Registry.

**Describe the solution you'd like**

We would like to add support for multiple AWS profiles. This might require AWS secret parsing changes so that the secret data is appropriately consumed by the Registry environment vars.

**Additional context**

Reference: https://github.com/openshift/oadp-operator/blob/0a6008d2a5e3603827727d7b57904b4040540325/controllers/registry.go#L505

deepakraj1997:
Resolves #435

### DIFF
--- a/controllers/registry.go
+++ b/controllers/registry.go
@@ -547,7 +547,6 @@ func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string
 			}
 			cleanedLine := strings.ReplaceAll(line, " ", "")
 			parsedProfile := awsProfileRegex.ReplaceAllString(cleanedLine, "")
-			fmt.Println(parsedProfile, awsProfile, parsedProfile == awsProfile)
 			if parsedProfile == awsProfile {
 				// check for end of arr
 				if index+1 >= len(splitString) {
@@ -557,7 +556,6 @@ func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string
 					if profLine == "" {
 						continue
 					}
-					fmt.Println(profLine)
 					matchedAccessKey := awsAccessKeyRegex.MatchString(profLine)
 
 					if err != nil {
--- a/controllers/registry.go
+++ b/controllers/registry.go
@@ -37,7 +37,6 @@ const (
 	RegistryStorageS3AccesskeyEnvVarKey      = "REGISTRY_STORAGE_S3_ACCESSKEY"
 	RegistryStorageS3BucketEnvVarKey         = "REGISTRY_STORAGE_S3_BUCKET"
 	RegistryStorageS3RegionEnvVarKey         = "REGISTRY_STORAGE_S3_REGION"
-	RegistryStorageS3ProfileEnvVarKey        = "REGISTRY_STORAGE_S3_PROFILE"
 	RegistryStorageS3SecretkeyEnvVarKey      = "REGISTRY_STORAGE_S3_SECRETKEY"
 	RegistryStorageS3RegionendpointEnvVarKey = "REGISTRY_STORAGE_S3_REGIONENDPOINT"
 	RegistryStorageS3RootdirectoryEnvVarKey  = "REGISTRY_STORAGE_S3_ROOTDIRECTORY"
@@ -85,10 +84,6 @@ var cloudProviderEnvVarMap = map[string][]corev1.EnvVar{
 		},
 		{
 			Name:  RegistryStorageS3RegionEnvVarKey,
-			Value: "",
-		},
-		{
-			Name:  RegistryStorageS3ProfileEnvVarKey,
 			Value: "",
 		},
 		{
@@ -411,14 +406,6 @@ func (r *VeleroReconciler) getAWSRegistryEnvVars(bsl *velerov1.BackupStorageLoca
 
 		if awsEnvVars[i].Name == RegistryStorageS3RegionEnvVarKey {
 			awsEnvVars[i].Value = bsl.Spec.Config[Region]
-		}
-
-		if awsEnvVars[i].Name == RegistryStorageS3ProfileEnvVarKey {
-			if value, exists := bsl.Spec.Config[Profile]; exists {
-				awsEnvVars[i].Value = value
-			} else {
-				awsEnvVars[i].Value = "default"
-			}
 		}
 
 		if awsEnvVars[i].Name == RegistryStorageS3SecretkeyEnvVarKey {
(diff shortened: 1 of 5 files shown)


controllers/registry.go
```diff
@@ -37,6 +37,7 @@ const (
 	RegistryStorageS3AccesskeyEnvVarKey      = "REGISTRY_STORAGE_S3_ACCESSKEY"
 	RegistryStorageS3BucketEnvVarKey         = "REGISTRY_STORAGE_S3_BUCKET"
 	RegistryStorageS3RegionEnvVarKey         = "REGISTRY_STORAGE_S3_REGION"
+	RegistryStorageS3ProfileEnvVarKey        = "REGISTRY_STORAGE_S3_PROFILE"
```
deepakraj1997, line 40: Added this for consistency. Not really required.

shubham-pampattiwar: @deepakraj1997 As this env var is not actually required or consumed by the registry deployment can we get rid of it ? IMO It's presence might give a false sense that it is being used somewhere for configuration purposes.

controllers/registry.go
```diff
@@ -383,9 +389,12 @@ func (r *VeleroReconciler) getAWSRegistryEnvVars(bsl *velerov1.BackupStorageLoca
 		r.Log.Info(fmt.Sprintf("Error fetching provider secret %s for backupstoragelocation %s/%s", secretName, bsl.Namespace, bsl.Name))
 		return nil, err
 	}
-
+	awsProfile := "default"
+	if value, exists := bsl.Spec.Config[Profile]; exists {
```
shubham-pampattiwar, line 393: +1, nice use of the Config spec map.

controllers/registry.go
```diff
@@ -502,7 +519,7 @@ func (r *VeleroReconciler) getSecretNameAndKey(credential *corev1.SecretKeySelec
 	return secretName, secretKey
 }
 
-func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string) (string, string, error) {
+func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string, awsProfile string) (string, string, error) {
 
 	AWSAccessKey, AWSSecretKey := "", ""
 	// this logic only supports single profile presence in the aws credentials file
```
shubham-pampattiwar, line 525: Please remove this comment, as this PR adds support for multiple profiles.

controllers/registry.go
```diff
@@ -519,44 +536,59 @@ func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string
...
-				return AWSAccessKey, AWSSecretKey, errors.New("secret parsing error")
+			awsProfileRegex, err := regexp.Compile(`\[|\]`)
+			if err != nil {
+				return AWSAccessKey, AWSSecretKey, errors.New("parseAWSSecret faulty regex: keyNameRegex")
 			}
-			AWSAccessKey = splitLine[1]
-			continue
-		}
+			cleanedLine := strings.ReplaceAll(line, " ", "")
+			parsedProfile := awsProfileRegex.ReplaceAllString(cleanedLine, "")
```
shubham-pampattiwar, line 549: @deepakraj1997 Can you please add a log for the case when the supplied aws profile is not found in the parsing of the credential secret ? like the config had the name `[test-profile]` but this profile was not to be found while parsing then this log might come in handy. I was thinking something like 
`if (index +1) == len(splitString) { 
errors.New("Supplied profile not found in parsing...") some thing similar 
}
`

shubham-pampattiwar (changes requested):
@deepakraj1997 Thank you for the PR. Overall changes look great ! Just added some suggestions.

controllers/registry.go
deepakraj1997: Thanks for reviewing. Removed.

controllers/registry.go
deepakraj1997: Removed the env vars.

controllers/registry.go
deepakraj1997: Hey shubham, I think if we do that we wouldnt know where the exact error is (Like if its access key/secret missing) and the lines from 585-592 will become dead code.  Maybe I am overthinking about the edge cases...I have added an alternative check to pinpoint where the error would be.

controllers/registry.go
deepakraj1997: Removed.

deepakraj1997:
@shubham-pampattiwar thanks for the review. I have made the requested changes. Do review when u have time. Thanks!

controllers/registry.go
shubham-pampattiwar: @deepakraj1997 yeah makes sense, the alternative check works. Thanks !

controllers/registry.go
```diff
@@ -502,10 +506,9 @@ func (r *VeleroReconciler) getSecretNameAndKey(credential *corev1.SecretKeySelec
...
 }
 
-func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string) (string, string, error) {
+func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string, matchProfile string) (string, string, error) {
 
-	AWSAccessKey, AWSSecretKey := "", ""
-	// this logic only supports single profile presence in the aws credentials file
+	AWSAccessKey, AWSSecretKey, AWSProfile := "", "", ""
 	splitString := strings.Split(string(secret.Data[secretKey]), "\n")
 	keyNameRegex, err := regexp.Compile(`\[.*\]`) //ignore lines such as [default]
```
shubham-pampattiwar, line 513: @deepakraj1997  NIT: The existing comment for this regex utility is misleading. Can you please remove/update this comment ?

controllers/registry.go
```diff
@@ -519,46 +522,66 @@ func (r *VeleroReconciler) parseAWSSecret(secret corev1.Secret, secretKey string
...
+						AWSSecretKey = splitLine[1]
+						continue
+					}
+				}
 			}
-			AWSSecretKey = splitLine[1]
-			continue
 		}
 	}
+	if AWSProfile == "" {
```
shubham-pampattiwar, line 581: NIT: AWS Profile -> just profile

shubham-pampattiwar (approved):
Just added a couple of nits. Apart from that LGTM ! Thank you !

controllers/registry.go
deepakraj1997: Yep, my bad, didnt notice this comment.

controllers/registry.go
deepakraj1997: changed.

dymurray (approved):

PR #458, opened 2021-11-17
Added yaml templating for creating Velero CR (e2e tests)

nitishSr:
- This PR provides a function to parse a template file (yaml) and provision tests to create Velero CR through the yaml template
- This will provide more flexibility to be able to use yaml files directly as sub standard templates, instead of defining veleroCustomResource struct for every new kind of variations
- The idea is to have yaml templates, and let test provide varying data, that can enable verification of different providers (aws, azure etc) and storage providers (minio, noobaa)


tests/e2e/velero_yaml_suite_test.go
```diff
@@ -0,0 +1,40 @@
...
+})
+
+var _ = Describe("Test Velero CR creation with", func() {
+	It("One Backup Storage Location templated from yaml", func() {
+		// Create Velero CR and verify it is successful
+		veleroTemplate, _ := filepath.Abs("templates/velero_bsl_template.tmpl")
+		params := veleroTemplateParams{
+			"example-vel",
+			"us-west-1",
+			"http://192.156.2.4:9000",
```
dymurray, line 31: @nitishSr I don't think we want this value here. Will fail in CI env

dymurray: But I really like the way we are reading from yaml and I would like to update the test suite to use this.

nitishSr: Thank you for review @dymurray , I have updated the test with changes.

dymurray (approved):

tests/e2e/templates/velero_bsl_template.tmpl
```diff
@@ -0,0 +1,26 @@
+apiVersion: oadp.openshift.io/v1alpha1
+kind: Velero
```
kaovilai, line 2: The problem is we have phased out this `kind` from master branch so this PR may need updating.

kaovilai: specifically... https://github.com/openshift/oadp-operator/pull/459

kaovilai (changes requested):

tests/e2e/templates/velero_bsl_template.tmpl
nitishSr: @kaovilai I have updated the yaml as per new API changes. However, after the test run, I can see, there is failure in some other test, which is not affected by this PR - velero_deployment_suite_test.go:103

nitishSr:
> /test operator-e2e

/retest

go.sum
```diff
@@ -1071,7 +1085,12 @@ golang.org/x/tools v0.0.0-20210106214847-113979e3529a/go.mod h1:emZCQorbCU4vsT4f
 golang.org/x/tools v0.1.0/go.mod h1:xkSsbof2nBLbhDlRMhhhyNLN/zl3eTqcnHD5viDpcZ0=
 golang.org/x/tools v0.1.2 h1:kRBLX7v7Af8W7Gdbbc908OJcdgtK8bOz9Uaj8/F1ACA=
 golang.org/x/tools v0.1.2/go.mod h1:o0xws9oXOQQZyjljx8fwUC0k7L1pTE6eaCbjGeHmOkk=
+<<<<<<< HEAD
```
dymurray, line 1088: @nitishSr There are still rebase bits here that need to be removed. For `go.sum` I recommend just checking it out and letting it get regenerated.

so:
```
git checkout go.sum
go mod tidy
```

nitishSr:
@dymurray , @kaovilai could you please help me with this PR, one or other test not affected by this PR keeps failing in CI check.

dymurray:
@nitishSr sure thing, I'll take a look

tests/e2e/velero_helpers.go
```diff
@@ -364,3 +365,37 @@ func verifyVeleroResourceLimits(namespace string, limits corev1.ResourceList) wa
 		return true, nil
 	}
 }
+
+// Create Velero Custom Resource Definition from yaml
+func (v *dpaCustomResource) CreateVeleroFromYaml(yamlTemplatePath string, data interface{}) error {
```
mperetzred, line 370: Maybe better to use CreateDpaFromYaml or BuildFromYaml as Velero CR changed to DPA CR

tests/e2e/templates/velero_bsl_template.tmpl
```diff
@@ -0,0 +1,27 @@
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataProtectionApplication
+metadata:
+ namespace: openshift-adp
+ name: {{ .VeleroName }}
```
mperetzred, line 5: Better to use DpaName or something to make it more generic

tests/e2e/velero_helpers.go
mperetzred, line 370: Maybe better to use CreateDpaFromYaml or BuildFromYaml as Velero CR changed to DPA CR

tests/e2e/templates/velero_bsl_template.tmpl
mperetzred, line 5: Better to use DpaName or something to make it more generic

tests/e2e/velero_yaml_suite_test.go
```diff
@@ -0,0 +1,42 @@
...
+	// Delete Velero CR once the test has finished
+	err := vel.Delete()
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = Describe("Test Velero CR creation with", func() {
+	It("One Backup Storage Location templated from yaml", func() {
+		// Create Velero CR and verify it is successful
+		veleroTemplate, _ := filepath.Abs("templates/velero_bsl_template.tmpl")
+		params := veleroTemplateParams{
```
mperetzred, line 29: Think better to use key:value syntax for initializing in struct

tests/e2e/velero_yaml_suite_test.go
```diff
@@ -0,0 +1,42 @@
+package e2e
+
+import (
+	"path/filepath"
+
+	. "github.com/onsi/ginkgo"
+	. "github.com/onsi/gomega"
+)
+
+type veleroTemplateParams struct {
```
mperetzred, line 10: Better to use dpaTemplateParams or something to make it more generic

nitishSr: Done

tests/e2e/velero_yaml_suite_test.go
nitishSr: Done

tests/e2e/templates/velero_bsl_template.tmpl
nitishSr: Done

tests/e2e/velero_helpers.go
nitishSr: Done

mperetzred (approved):

tests/e2e/velero_yaml_suite_test.go
```diff
@@ -0,0 +1,42 @@
...
+var _ = AfterEach(func() {
+	// Delete Velero CR once the test has finished
+	err := vel.Delete()
+	Expect(err).ToNot(HaveOccurred())
+})
+
+var _ = Describe("Test Velero CR creation with", func() {
+	It("One Backup Storage Location templated from yaml", func() {
+		// Create Velero CR and verify it is successful
+		veleroTemplate, _ := filepath.Abs("templates/velero_bsl_template.tmpl")
```
dymurray, line 28: NIT: template isn't a BSL, name should be:
```
templates/dpa_template.tmpl
```

nitishSr: > NIT: template isn't a BSL, name should be:
> 
> ```
> templates/dpa_template.tmpl
> ```

Sorry, missed that while addressing last review comments. Will update

tests/e2e/velero_helpers.go
```diff
@@ -364,3 +365,37 @@ func verifyVeleroResourceLimits(namespace string, limits corev1.ResourceList) wa
...
+	}
+
+	// Create custom resource
+	err = v.Client.Create(context.Background(), &dpaSpec)
+	if apierrors.IsAlreadyExists(err) {
+		return errors.New("found unexpected existing DPA")
+	} else if err != nil {
+		return err
+	}
+	log.Printf("DPA created \n")
```
dymurray, line 399: Please remove this

tests/e2e/velero_yaml_suite_test.go
```diff
@@ -0,0 +1,44 @@
...
+	ClusterProfile  string
+	Provider        string
+	CredentialsName string
+	BucketName      string
+	Prefix          string
+}
+
+var _ = AfterEach(func() {
+	// Delete DPA once the test has finished
+	err := vel.Delete()
```
dymurray, line 22: So this is actually not cleaning up the DPA CR that you created. By moving to this template creation workflow the existing functions won't be useful here. `vel.Delete()` depends on the DPA CR struct existing in memory on the `vel` resource.

You could write a new function:
```
func DeleteDPAFromYaml() error {
    // Take in the DPA template and issue a client.Delete() based on the template name
}
```

dymurray (changes requested):
The problem is that when the `velero_yaml_suite_test.go` file gets executed, it's creating a DPA CR and failing to clean it up after the test runs.

When the e2e tests for backup/restore run after this test suite, there will exist 2 DPA CRs at once which breaks OADP

dymurray (commented):
@nitishSr PR looks much better now. I'm curious, how does this relate to the changes Maya posted where we are creating these CRs from JSON templates?

Is there a benefit of having both or should we pick one and run with it?

nitishSr:
> @nitishSr PR looks much better now. I'm curious, how does this relate to the changes Maya posted where we are creating these CRs from JSON templates?
> 
> Is there a benefit of having both or should we pick one and run with it?

@dymurray my idea was to keep this as placeholder and create an alternate way of writing test scripts, where we can populate the yaml template with different table values. The current changes from Maya, put a json with static values, which is indeed one way, (can be made dynamic). But I just wanted to have more options to actually create any resource (DPA, Backup, Restore etc..) dynamically, without having to worry about initializing the dpa struct

nitishSr:
Will raise it with directory structure changes

PR #459, opened 2021-11-17
API v1 rewrite for GA

dymurray:


config/manager/kustomization.yaml
```diff
@@ -12,5 +12,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: quay.io/konveyor/oadp-operator
-  newTag: latest
+  newName: ttl.sh/oadp-operator-967a78f
```
kaovilai, line 15: Oppsies.. you probably `git add .`

These were most likely coming from `make deploy-olm`.
It can be reverted by running `make bundle` again.
Alternatively we could add `make bundle` to the end of `make deploy-olm`

kaovilai (dismissed):

config/manager/kustomization.yaml
dymurray: ahh thanks. This file always changes even with `IMG=<> make deploy` I just usually ignore it

api/v1alpha1/oadp_types.go
```diff
@@ -54,47 +54,81 @@ const CSIPluginImageKey UnsupportedImageKey = "csiPluginImageFqin"
...
+
+// BackupLocation defines the configuration for the DPA backup storage
+type BackupLocation struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+	Velero            *velero.BackupStorageLocationSpec `json:"velero"`
+}
+
+// SnapshotLocation defines the configuration for the DPA snapshot store
+type SnapshotLocation struct {
+	metav1.ObjectMeta `json:"metadata,omitempty"`
```
alaypatel07, line 119: is the entire surface of `metav1.ObjectMeta` required here? From what I recall we needed stuff like name, labelselector etc. I'm a little concerned about introducing the object meta here that we will not be able to support in the future.

kaovilai (approved):

api/v1alpha1/oadp_types.go
dymurray: That's a fair point. My thought on this was to support labels/annotations. But you're right, we are opening the surface area pretty far.

config/samples/oadp_v1alpha1_dpa.yaml
```diff
@@ -0,0 +1,31 @@
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataProtectionApplication
+metadata:
+  name: velero-sample
```
apinnick, line 4: You might want to change the name to something like `dpa-sample`.

docs/upgrade.md
```diff
@@ -72,35 +72,42 @@ spec:
   enable_restic: true
 ```
 
-And a sample Velero CR for version 0.3 or later:
+And a sample Velero CR (DPA) for version 0.3 or later:
```
kaovilai, line 75: NIT: we probably want a separate 0.5 or later section.

PR #460, opened 2021-11-17
Exit loop if none of the secret file pattern matches

deepakraj1997:


shubham-pampattiwar (changes requested):
@deepakraj1997 Do you mind adding unit test for this scenario as well as other credential file parsing scenarios covering boundary cases ?

deepakraj1997:
@shubham-pampattiwar  Thank you for reviewing, I did add a test case last time for testing profile but missing this corner case. I have added a new test case for this.

dymurray (approved):

shubham-pampattiwar (approved):

PR #461, opened 2021-11-17
`make bundle` fix v2. change project layout to kubebuilder/v3, add `make deploy-olm`

kaovilai:
- `make bundle` should now work. When submitting new versions to community operators, simply run this command and rename to x.y.z and upload the bundle directory to marketplace.
- deprecates `deploy/` path for bundling operator.
- Adds `make deploy-olm` for easier olm deployment of current code.


kaovilai:
Ideally role.yaml defining Service Account permissions for the operator which eventually end up in CSV would have more scoped permissions.
https://github.com/openshift/oadp-operator/blob/463897a8ecc558d9a2a5c1a964f94d401556a5df/config/rbac/role.yaml#L9-L14
However it is currently generated via `manifests` `make` target which uses `controller-gen cli` to do so. It does not appear that the behaviors are configurable.

Ideally we can match permissions from
https://github.com/redhat-openshift-ecosystem/community-operators-prod/blob/main/operators/oadp-operator/0.4.2/manifests/oadp-operator.v0.4.2.clusterserviceversion.yaml#L320-L419

kaovilai:
Actually.. could just remove manager-role from controller-gen then it won't generate it.

dymurray (approved):

PR #482, opened 2021-11-30
add velero deployment to CSV so velero SA/roles are added to CSV then removing it

kaovilai:
This fixes `make bundle` not adding velero roles/SA to CSV.

Velero placeholder deployment is added to config so `operator-sdk generate bundle` add cluster permission rules for velero SA in CSV.
`yq` is used to remove placeholder deployment from final bundle CSV.


config/velero/velero.yaml
```diff
@@ -0,0 +1,120 @@
+# This is a placeholder deployment config so operator-sdk add cluster-roles for velero to ClusterServiceVersion
+# It will be removed by yq during make bundle
+kind: Deployment
```
djzager, line 3: My only recommendation for this file is to strip it to make it less something that needs to be maintained.

As I understand it, the only thing that matters is that the deployment's serviceAccountName.

djzager (approved):
I personally prefer this to manually maintaining the csv. Have an ack.

config/velero/velero.yaml
kaovilai: @djzager just pulled this from a running development deployment.. nothing much to maintain. Just need a container to be a valid deployment

dymurray:
This is definitely a weird way to solve things, but it gets the job done for now. We are going to want to revisit this at some point

PR #491, opened 2021-12-02
remove default secret volumes when using cloud storage provider

### LINKED ISSUE #490: Bug: Unable to handle bucket only OADP velero applications
### Contact Details

_No response_

### Describe bug

When using only cloud storage location as a backup storage location we still attempt to load a default secret into the image even if it is not required.

### What happened?

Velero unable to start up

### OADP Version

0.4.x (Beta)

### OpenShift Version

_No response_

### Velero pod logs

_No response_

### Restic pod logs

_No response_

### Operator pod logs

_No response_

### New issue

- [X] This issue is new

shawn-hurley:
Adding back removed tests
refactoring so the logic is shared between validation and customization.

fixes #490


controllers/velero.go
```diff
@@ -611,3 +617,39 @@ func (r *DPAReconciler) getResticResourceReqs(dpa *oadpv1alpha1.DataProtectionAp
 
 	return ResourcesReqs
 }
+
+func (r DPAReconciler) noDefaultCredentials(dpa oadpv1alpha1.DataProtectionApplication) (map[string]bool, bool, error) {
```
kaovilai, line 621: A function comment on return vals would be nice so you don't have to inspect the function for what is returned.

dymurray (approved):

PR #497, opened 2021-12-06
Timeout  for must-gather failed backups

savitharaghunathan:
Failed backups can cause must-gather to be slow. This PR implements a timeout flag which can be passed as an argument to the gather command.

To test:
Build and push the must-gather image to your quay repo and then run,
```oc adm must-gather --image=<your repo path> -- /usr/bin/gather_with_timeout <timeout_value_in_seconds>```
eg:
``` oc adm must-gather --image=quay.io/sraghuna/must-gather1:43 -- /usr/bin/gather_with_timeout 3s```

/assign @dymurray


kaovilai:
Do you think there should be a default timeout and only if that's not enough.. customize?

savitharaghunathan:
I am not sure if we need a default value. Most of my test runs I found that the `velero describe` was faster without specifying the timeout value. Only reason we are doing this is to avoid must-gather run for hours when there is a failedvalidation error due to BSL location.

kaovilai:
It would be cool if you could make it spit out info about this function when must-gather has been running for x minutes. (15/30/60?)

savitharaghunathan:
> It would be cool if you could make it spit out info about this function when must-gather has been running for x minutes. (15/30/60?)

It will be definitely useful when the tool is actually hanging due to failed backups. But if the cluster has a lot of valid resources and it takes a long time to process, I am afraid spitting this command would prompt the user to rerun the gather again. Rerunning it with timeout will not improve the time in this case, as all the resources are valid and the process is not hanging.
I am actually torn between having an arbitrary timeout irrespective of the cluster resources state vs giving multiple options for the user. If we decide on having an arbitrary timeout then what that value would be. All the test runs I did it went anywhere from 2s to 7s.

savitharaghunathan:
Another implementation could actually check for failed resources and add a timeout to those, and let the other resources run normally. This will avoid prompting the user and giving them multiple options.

kaovilai:
> I am afraid spitting this command would prompt the user to rerun the gather again.

How about spitting it out at the beginning? then it is not an indication of hanging. They can decide at 0 minutes which route they want to take.

dymurray (approved):

PR #498, opened 2021-12-06
Add Cluster Profile argument for CSI test case

hhpatel14:
Cluster Profile defines which cloud provider the cluster is running on. This determines the storage drivers defined in volumesnapshotclass object which is unique to the cluster's cloud provider.

Provider flag is Object Storage provider and is not always the cloud provider that the e2e cluster is running on.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 OADP_TEST_NAMESPACE ?= openshift-adp
 REGION ?= us-east-1
 PROVIDER ?= aws
+CLUSTER_PROFILE ?= aws
 CREDS_SECRET_REF ?= cloud-credentials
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,6 +1,7 @@
 OADP_TEST_NAMESPACE ?= openshift-adp
 REGION ?= us-east-1
 PROVIDER ?= aws
```
kaovilai, line 3: @deepakraj1997 @hhpatel14 @dymurray may want to revisit renaming provider to something more BSL/BL/ObjectStorage specific

kaovilai (approved):

dymurray (approved):

Issue #503, opened 2021-12-09
Indentation needs to be checked


apinnick:
https://github.com/openshift/oadp-operator/blob/5a0e688429c5127bb4f5f931e1c934c2db2349b0/config/samples/oadp_v1alpha1_dpa.yaml#L27

I think the "- velero" block might be missing an indent.

PR #505, opened 2021-12-10
Fix Indentation in Sample CR

eemcmullan:
#503


dymurray (approved):

dymurray:
I think this technically shouldn't matter, but it does look better :)

PR #507, opened 2021-12-10
Adding support for service principal credential for Azure plugin

savitharaghunathan:
Adding support for service principal credential for Azure plugin

### DIFF
--- a/controllers/registry.go
+++ b/controllers/registry.go
@@ -134,6 +134,16 @@ var cloudProviderEnvVarMap = map[string][]corev1.EnvVar{
 	},
 }
 
+type azureCredentials struct {
+	subscriptionID     string
+	tenantID           string
+	clientID           string
+	clientSecret       string
+	resourceGroup      string
+	cloudName          string
+	strorageAccountKey string
+}
+
 func (r *DPAReconciler) ReconcileRegistries(log logr.Logger) (bool, error) {
 	dpa := oadpv1alpha1.DataProtectionApplication{}
 	if err := r.Get(r.Context, r.NamespacedName, &dpa); err != nil {
@@ -439,7 +449,7 @@ func (r *DPAReconciler) getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocat
 	}
 
 	// parse the secret and get azure storage account key
-	AzureStorageKey, err := r.parseAzureSecret(secret, secretKey)
+	azcreds, err := r.parseAzureSecret(secret, secretKey)
 	if err != nil {
 		r.Log.Info(fmt.Sprintf("Error parsing provider secret %s for backupstoragelocation %s/%s", secretName, bsl.Namespace, bsl.Name))
 		return nil, err
@@ -455,7 +465,7 @@ func (r *DPAReconciler) getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocat
 		}
 
 		if azureEnvVars[i].Name == RegistryStorageAzureAccountkeyEnvVarKey {
-			azureEnvVars[i].Value = AzureStorageKey
+			azureEnvVars[i].Value = azcreds.strorageAccountKey
 		}
 
 	}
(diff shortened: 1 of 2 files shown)


controllers/registry.go
```diff
@@ -439,11 +469,22 @@ func (r *DPAReconciler) getAzureRegistryEnvVars(bsl *velerov1.BackupStorageLocat
...
 		return nil, err
 	}
+	if len(bsl.Spec.Config["storageAccountKeyEnvVar"]) != 0 {
+		if azcreds.strorageAccountKey == "" {
+			r.Log.Info("Expecting storageAccountKeyEnvVar value set present in the credentials")
+			return nil, errors.New("no strorageAccountKey value present in credentials file")
+		}
+	} else {
+		r.Log.Info("Checking for service principal credentials")
+		if len(azcreds.subscriptionID) == 0 && len(azcreds.tenantID) == 0 && len(azcreds.clientID) == 0 && len(azcreds.clientSecret) == 0 && len(azcreds.resourceGroup) == 0 {
```
shubham-pampattiwar, line 484: Do we not need `strorageAccountKey` value as well for Service principal ?

controllers/registry.go
```diff
@@ -617,30 +687,72 @@ func (r *DPAReconciler) parseAzureSecret(secret corev1.Secret, secretKey string)
...
-			r.Log.Info("Error finding storage key for the supplied Azure credential")
-			return AzureStorageKey, err
-		}
+		matchedSubscriptionId := azureSubscriptionIdRegex.MatchString(line)
+		matchedTenantId := azureTenantIdRegex.MatchString(line)
+		matchedCliendId := azureClientIdRegex.MatchString(line)
+		matchedClientsecret := azureClientSecretRegex.MatchString(line)
+		matchedResourceGroup := azureResourceGroupRegex.MatchString(line)
 
 		if matchedStorageKey {
```
shubham-pampattiwar, line 696: Would it make sense to have switch-case here ?

shubham-pampattiwar (commented):
Thank you for the PR @savitharaghunathan, changes look sane to me.  Added some questions and Nits,

controllers/registry.go
savitharaghunathan: we need `strorageAccountKey`. Hence checking for the one in bslspec. If the config parameter for `storageAccountKeyEnvVar` is not present, then service principal creds needs to be validated

savitharaghunathan: Irresepctive of whatever method we need the `strorageAccountKey` in credential file.

controllers/registry.go
savitharaghunathan: Dylan and I, spoke about the same. It might involve assigning bits to every field and then using that in switch case. I need to look more into it. Happy to change this block to use switch

controllers/registry.go
dymurray: just a NIT for this... when we have this many conditionals on an if statement looks better if you do:
```
if len(azcreds.subscriptionID) == 0 &&
    len(azcreds.tenantID) == 0 && 
    len(foo)
   ...
```
```

File: controllers/registry.go (hunk 2 above)
Comment (savitharaghunathan) [reply controllers/registry.go]: Changed to `switch` case.

File: controllers/registry.go [hunk 3]
```diff
@@ -617,30 +690,63 @@ func (r *DPAReconciler) parseAzureSecret(secret corev1.Secret, secretKey string)
...
+		case matchedStorageKey:
+			storageKeyValue, err := r.getMatchedKeyValue("AZURE_STORAGE_ACCOUNT_ACCESS_KEY=", line)
+			if err != nil {
+				return azcreds, err
 			}
-			AzureStorageKey = storageKeyValue
-			r.Log.Info(fmt.Sprintf("Azure storage key value after parsing: %s", AzureStorageKey))
-			continue
+			azcreds.strorageAccountKey = storageKeyValue
+		case matchedSubscriptionId:
```
dymurray, line 706: Thiis looks great

dymurray (approved):

controllers/registry.go
```diff
@@ -43,9 +43,13 @@ const (
 	RegistryStorageS3RootdirectoryEnvVarKey  = "REGISTRY_STORAGE_S3_ROOTDIRECTORY"
 	RegistryStorageS3SkipverifyEnvVarKey     = "REGISTRY_STORAGE_S3_SKIPVERIFY"
 	// Azure registry env vars
-	RegistryStorageAzureContainerEnvVarKey   = "REGISTRY_STORAGE_AZURE_CONTAINER"
-	RegistryStorageAzureAccountnameEnvVarKey = "REGISTRY_STORAGE_AZURE_ACCOUNTNAME"
-	RegistryStorageAzureAccountkeyEnvVarKey  = "REGISTRY_STORAGE_AZURE_ACCOUNTKEY"
+	RegistryStorageAzureContainerEnvVarKey       = "REGISTRY_STORAGE_AZURE_CONTAINER"
+	RegistryStorageAzureAccountnameEnvVarKey     = "REGISTRY_STORAGE_AZURE_ACCOUNTNAME"
+	RegistryStorageAzureAccountkeyEnvVarKey      = "REGISTRY_STORAGE_AZURE_ACCOUNTKEY"
+	RegistryStorageAzureSPNClientIDEnvVarKey     = "REGISTRY_STORAGE_AZURE_SPN_CLIENT_ID"
```
kaovilai, line 49: We are [the only ones](https://github.com/search?q=REGISTRY_STORAGE_AZURE_SPN_CLIENT_ID&type=code) with `REGISTRY_STORAGE_AZURE_SPN_CLIENT_ID` in our code. I'm curious how we ended up with `_SPN_` here.

PR #516, opened 2021-12-17
[WIP] adding negative error testing to e2e test suite

jaidevmane:
Adding negative error testing to the testing suite, sort of like a first pass at implementing this feature. will definitely need more updates in the future


weshayutin:
Unrelated to this change... please hold off on merging so we can land multicloud tests.
https://github.com/openshift/oadp-operator/pull/568
Sorry for the inconvience folks.  @dymurray

PR #518, opened 2021-12-20
 Registry should not be deployed when Azure SP is used

savitharaghunathan:
Registry should not be deployed when Azure SP is used

### DIFF
--- a/api/v1alpha1/oadp_types.go
+++ b/api/v1alpha1/oadp_types.go
@@ -29,7 +29,7 @@ const ReconciledReasonError = "Error"
 const ReconcileCompleteMessage = "Reconcile complete"
 
 const OadpOperatorLabel = "openshift.io/oadp"
-const RegistryDeploymentLabel = "openshift.io/registryDeployed"
+const RegistryDeploymentLabel = "openshift.io/oadp-registry"
 
 type DefaultPlugin string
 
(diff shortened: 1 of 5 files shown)


api/v1alpha1/oadp_types.go
```diff
@@ -29,6 +29,7 @@ const ReconciledReasonError = "Error"
 const ReconcileCompleteMessage = "Reconcile complete"
 
 const OadpOperatorLabel = "openshift.io/oadp"
+const RegistryDeploymentLabel = "openshift.io/registryDeployed"
```
dymurray, line 32: NIT: `I would do openshift.io/oadp-registry`

controllers/bsl.go
```diff
@@ -151,14 +151,20 @@ func (r *DPAReconciler) updateBSLFromSpec(bsl *velerov1.BackupStorageLocation, d
 	if err != nil {
 		return err
 	}
-
+	registryDeployment := "True"
+	if bslSpec.Provider == "azure" {
+		if len(bslSpec.Config["storageAccountKeyEnvVar"]) == 0 {
```
dymurray, line 156: This looks great, please just add a comment here explaining why we are doing this

api/v1alpha1/oadp_types.go
savitharaghunathan: resolved. Thanks, Dylan :)

controllers/bsl.go
savitharaghunathan: Added a blurb. Let me know if that makes sense or not

dymurray (approved):

PR #528, opened 2022-01-10
Adding secrets watch

savitharaghunathan:
Adding secrets watch to help with reconciliation of secrets outside of the objects owned by OADP operator


controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+
+		secretName, _ := r.getSecretNameAndKey(bslSpec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Velero.Provider))
+		secret, _ := r.getProviderSecret(secretName)
+		sec := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: r.NamespacedName.Namespace,
+			},
+		}
+		secret.SetLabels(map[string]string{oadpv1alpha1.OadpOperatorLabel: "True", "dpaName": dpa.Name, "namespace": dpa.Namespace})
```
dymurray, line 109: Just curious do we need to be setting the namespace label? All these secrets should be in the same namespace as the operator so feels redundant to me.

NIT: I would also change `dpaName` to `dataprotectionapplication: <name>`.

controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+		op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, &sec, func() error {
+			sec = secret
+			return nil
+
+		})
+		if err != nil {
+			return false, err
+		}
+		//if err then return
```
dymurray, line 118: Move this to 114 please

controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+		if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated {
+			// Trigger event to indicate BSL was created or updated
+			r.EventRecorder.Event(&secret,
+				corev1.EventTypeNormal,
+				"SecretReconciled",
+				fmt.Sprintf("performed %s on Secret %s", op, secretName),
+			)
+		}
+		r.Log.Info(fmt.Sprintf("labels for %s : %v", secretName, secret.GetLabels()))
+		//}
```
dymurray, line 128: Remove this please

controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+		//if err then return
+		if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated {
+			// Trigger event to indicate BSL was created or updated
+			r.EventRecorder.Event(&secret,
+				corev1.EventTypeNormal,
+				"SecretReconciled",
+				fmt.Sprintf("performed %s on Secret %s", op, secretName),
+			)
+		}
+		r.Log.Info(fmt.Sprintf("labels for %s : %v", secretName, secret.GetLabels()))
```
dymurray, line 127: Lets remove the logging as well

controllers/dpa_controller.go
```diff
@@ -136,14 +139,75 @@ func (r *DPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
 		WithEventFilter(veleroPredicate(r.Scheme)).
 		Complete(r)
 }
 
+type labelHandler struct {
+}
+
+func (l *labelHandler) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	// check for the label & add it to the queue
+
```
dymurray, line 155: NIT:
```
	if evt.Object.GetLabels()[oadpv1alpha1.OadpOperatorLabel] == "" {
	    return
	}
	
	dpaname := evt.Object.GetLabels()["dpaName"]
	...

```

dymurray: removes need for else

controllers/dpa_controller.go
```diff
@@ -136,14 +139,75 @@ func (r *DPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	if evt.Object.GetLabels()[oadpv1alpha1.OadpOperatorLabel] != "" {
+		dpaname := evt.Object.GetLabels()["dpaName"]
+		namespace := evt.Object.GetLabels()["namespace"]
+		q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+			Name:      dpaname,
+			Namespace: namespace,
+		}})
+
+	} else {
```
dymurray, line 179: same as above

controllers/dpa_controller.go
dymurray: You likely don't need to log anything as well in the case that no label is found

controllers/bsl.go
savitharaghunathan: When I populated the namespace from the secret object it did not work. The event never kicked off 🤷‍♀️  So I ended up adding a label to work around. @dymurray do you think its hacky? I think it is but I dont know any another way to get it working

savitharaghunathan: I am going to test it again. I changed a whole bunch of things and this might work again.

savitharaghunathan: I must have done something wrong yesterday. It works now. Thanks @dymurray for catching it :)

controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
 				Namespace: r.NamespacedName.Namespace,
 			},
 		}
+		// Add the following labels to the bsl secret,
+		//	 1. oadpApi.OadpOperatorLabel: "True"
+		// 	 2. dataprotectionapplication: <name>
```
shawn-hurley, line 100: I think that we should namespace this label.

controllers/bsl.go
```diff
@@ -95,8 +95,40 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+		secretName, _ := r.getSecretNameAndKey(bslSpec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Velero.Provider))
+		secret, _ := r.getProviderSecret(secretName)
+		sec := corev1.Secret{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      secretName,
+				Namespace: r.NamespacedName.Namespace,
+			},
+		}
+		secret.SetLabels(map[string]string{oadpv1alpha1.OadpOperatorLabel: "True", "dataprotectionapplication": dpa.Name})
+		op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, &sec, func() error {
```
shawn-hurley, line 112: I don't know if we should be creating here probably just want to update.

controllers/dpa_controller.go
```diff
@@ -136,14 +139,69 @@ func (r *DPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+type labelHandler struct {
+}
+
+func (l *labelHandler) Create(evt event.CreateEvent, q workqueue.RateLimitingInterface) {
+	// check for the label & add it to the queue
+
+	if evt.Object.GetLabels()[oadpv1alpha1.OadpOperatorLabel] == "" {
+		return
+	}
+	dpaname := evt.Object.GetLabels()["dataprotectionapplication"]
```
shawn-hurley, line 159: handle the case where dpaname is not set.

controllers/predicate.go
```diff
@@ -1,7 +1,7 @@
 package controllers
 
 import (
-	oadpApi "github.com/openshift/oadp-operator/api/v1alpha1"
+	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
```
shawn-hurley, line 4: I think  this change may be outside the scope of the PR.

shawn-hurley (changes requested):
I think we just have to clean up what to do when the label is not set, the label name, and if we should be creating the provider secret ever.

shawn-hurley:
one more thing, what when we create VSL?

savitharaghunathan:
> I think we just have to clean up what to do when the label is not set, the label name, and if we should be creating the provider secret ever
@shawn-hurley  Thanks for the review. Quick question: Do we want to handle the case of us creating provider secret in this PR?

controllers/dpa_controller.go
savitharaghunathan: done

controllers/predicate.go
savitharaghunathan: Removed it from the PR!

controllers/bsl.go
```diff
@@ -145,6 +156,23 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
 	return true, nil
 }
 
+func (r *DPAReconciler) UpdateLabels(secretName string, namespace string, dpaName string) (bool, error) {
```
dymurray, line 159: NIT: can we rename to `UpdateCredentialsSecretLabels` just to be explicit?

controllers/bsl.go
```diff
@@ -145,6 +156,23 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+	var secret corev1.Secret
+	secret, err := r.getProviderSecret(secretName)
+	if err != nil {
+		return false, err
+	}
+	if secret.Name != "" {
```
dymurray, line 165: Perhaps this should error out if `secret.Name == ""`? That would imply we didn't find the secret right? Or maybe not even error but just return with no err?

dymurray (approved):
LGTM aside from a couple comments

controllers/bsl.go
shawn-hurley: it would mean that get provider secret failed in a very unexpected way right? I think we can assume that this case is covered in that function ?

savitharaghunathan: good catch..on it right now

controllers/dpa_controller.go
```diff
@@ -136,14 +140,69 @@ func (r *DPAReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+	}
+	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
+		Name:      dpaname,
+		Namespace: namespace,
+	}})
+
+}
+func (l *labelHandler) Update(evt event.UpdateEvent, q workqueue.RateLimitingInterface) {
+
+	if evt.ObjectNew.GetLabels()[oadpv1alpha1.OadpOperatorLabel] != "" {
```
shawn-hurley, line 183: add ```if evt.Object.GetLabels()[oadpv1alpha1.OadpOperatorLabel] == "" || dpaname == "" {
 		return
 	}
 	```

shawn-hurley (approved):
LGTM expect for NITs

controllers/bsl.go
savitharaghunathan: done

controllers/bsl.go
savitharaghunathan: returning an error is secret is not found

controllers/dpa_controller.go
savitharaghunathan: Updated it. I had to use `ObjectNew` instead of `Object`. Update function supports only `ObjectNew` & `ObjectOld`. Let me know if I have to add a case for `ObjectOld` as well :)

shawn-hurley: Nope I think ObjectNew is the only thing we care about

PR #530, opened 2022-01-11
OADP-193: Fix restic enablement with custom BSL secret

dymurray:


pkg/credentials/credentials.go
```diff
@@ -156,6 +156,16 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds
 				continue
 			}
 
+			pluginNeedsCheck, foundInBSLorVSL := providerNeedsDefaultCreds[string(plugin)]
```
shawn-hurley, line 159: Do we have to make sure to strip out velero.io here?

shawn-hurley (approved):
LGTM just a little question

pkg/credentials/credentials.go
dymurray: We do not, this gets stripped in `noDefaultCredentials` func

dymurray:
Now also included additional fix for OADP 150

shawn-hurley (approved):

PR #531, opened 2022-01-12
New structural changes to test repo

nitishSr:
This PR tries to introduce new structure for organising test repo as per below structure - 
```
.
└── oadp-operator/
    └── tests/
        └── e2e/
            ├── templates/
            │   ├── velero_bsl_template.tmpl
            │   └── credentials
            ├── lib/
            │   ├── kube_helpers.go
            │   └── velero_helpers.go
            ├── utils/
            │   └── common.go
              velero_deployment_suite_test.go
              bsl_suite_test.go
              vsl_suite_test.go
              e2e_suite_test.go
```


dymurray:
@nitishSr I like the structural changes, LGTM. Let me know when this is ready for review

tests/e2e/backup_restore_suite_test.go
```diff
@@ -1,4 +1,4 @@
-package e2e
+package e2e_test
```
shubham-pampattiwar, line 1: qq: Is the name change intentional or accidental ? The PR description does not mention about `e2e_test` package https://github.com/openshift/oadp-operator/pull/531#issue-1100332242

shubham-pampattiwar (approved):
@nitishSr +1 on the refactoring, just added one question, else LGTM.

kaovilai:
+1
I think exporting functions would allow folks to reuse our e2e functions elsewhere.

tests/e2e/utils/common_utils.go
```diff
@@ -0,0 +1,16 @@
+package utils
+
+import (
+	"io/ioutil"
+)
+
+func ReadFile(path string) ([]byte, error) {
```
kaovilai, line 7: I think we can remove this? This function doesn't do anything. Can remove later after merge. Thoughts?

kaovilai (approved):

tests/e2e/utils/common_utils.go
kaovilai: I see elsewhere we are using os.ReadFile as well which seems to do the same thing.

kaovilai: Looks like os.ReadFile is more direct over io/ioutil now... at least from reading https://pkg.go.dev/io/ioutil@go1.17.6#ReadFile

tests/e2e/backup_restore_suite_test.go
nitishSr: @shubham-pampattiwar yes that is intentional change. Just wanted to be more specific here, to reflect that its a test package (taken as reference from Ginkgo docs)

tests/e2e/utils/common_utils.go
nitishSr: @kaovilai I do think the same. I just kept it as it is, considering these changes were made by some earlier PR. I can definitely try to replace it altogether with os.ReadFile() if possible.

PR #534, opened 2022-01-13
Credential Usage Docs

eemcmullan:


docs/credentials.md
```diff
@@ -0,0 +1,131 @@
...
+          key: cloud
+```
+
+    
+### Credentials for VolumeSnapshotLocation:
+
+- `VolumeSnapshotLocation` will **always** expect the default secret name to exist, 
+which are described above.
+
+- To use seperate credentials for `BackupStorageLocation` and `VolumeSnapshotLocation`,
```
dymurray, line 44: @eemcmullan I think I said this to you yesterday and I did forget one thing here... a user can actually take advantage of `profile` config key to use one secret with separate credentials. This only works for the AWS plugin but good to call out.

So one secret:
```
[backupStorage]
aws_access_key_id=...
aws_secret_access_key=...
[volumeSnapshot]
aws_access_key_id=...
aws_secret_access_key=...
```

Then the BSL:
```
     ...
     profile: "backupStorage"
```

and under snapshotlocations set it to the other profile

eemcmullan: @dymurray updated

PR #540, opened 2022-01-16
made tests' structure more generic

mperetzred:
Changed the tests so they would use DPA settings from json file (using the json tags of DPA structs/nested structs), like for example:
https://github.com/openshift/oadp-operator/pull/540/files#diff-740eb61cce497a1813e959f31a1b99b40e39e4c0cbf13d001da2360203c139cc
So it would be more generic and flexible and could be used for other cloud providers.

On the makefile the params will be add to this script:
https://github.com/openshift/oadp-operator/pull/540/files#diff-c4c5b54f7d75374a7c6729ede838cd2c5d323a3851db04a30e7b6d2ad503558d
and create a file which will be used as the settings for the DPA.

Updated the doc accordingly:
https://github.com/openshift/oadp-operator/blob/ad53ecfea14bc6caee58655300a35785d1cc5875/docs/developer/TESTING.md#debugging-e2e-tests-with-visual-studio

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -289,11 +289,11 @@ catalog-build: opm ## Build a catalog image.
 catalog-push: ## Push a catalog image.
 	$(MAKE) docker-push IMG=$(CATALOG_IMG)
 
-S3_BUCKET := $(shell cat $(OADP_S3_BUCKET) | awk '/velero-bucket-name/  {gsub(/"/, "", $2); print $2}')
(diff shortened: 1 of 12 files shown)


mperetzred:
*with awk
a type...

kaovilai (approved):

dymurray (approved):

PR #554, opened 2022-01-25
Added CSI Example to Docs

eemcmullan:


docs/examples/csi_example.md
```diff
@@ -0,0 +1,252 @@
+<h1 align="center">Stateful Application Backup/Restore - MSSQL</h1>
+<h2 align="center">CSI Volume Snapshotting with AWS EBS</h2>
+
+### Prerequisites
+* OADP operator is [installed](/docs/install_olm.md).
+* Create a credentials secret for AWS:
+
+   `oc create secret generic cloud-credentials --namespace openshift-adp --from-file cloud=<CREDENTIALS_FILE_PATH>`
+
+* Make sure your DataProtectionApplication (DPA) CR is similar to below. Note
```
shubham-pampattiwar, line 10: NIT: Just add save this config as `config/samples/oadp_v1alpha1_dpa.yaml` as you are directly asking the user to create DPA isntance using this file OR just mention that create DPA instance using the above config in `Install Velero` step.

docs/examples/csi_example.md
```diff
@@ -0,0 +1,252 @@
...
+apiVersion: storage.k8s.io/v1 
+kind: StorageClass 
+metadata: 
+  name: gp2-csi
+  annotations:
+    storageclass.kubernetes.io/is-default-class: 'true'
+provisioner: ebs.csi.aws.com
+parameters: 
+  type: gp2
+reclaimPolicy: Delete
```
shubham-pampattiwar, line 106: Shouldn't the `reclaimPolicy` be set to `Retain` ?

shubham-pampattiwar (approved):
Thanks for the CSI example @eemcmullan 
+1 on the steps and ymls, added a nit/question, rest looks good to me.

docs/examples/csi_example.md
kaovilai: +1 Retain

kaovilai: If StorageClass is set to `Retain` I believe upon restore you would end up with two PVs, the original retained one, and a newly restored one with content from snapshot.
The original PV would be dangling and not connected to anything.. either that or new pv would fail to create due to some kind of conflict which I think velero would be ok with and "skip, already exists"

eemcmullan: Yes what @kaovilai mentioned was my understanding was well. I can verify though.

kaovilai: "skip, already exists" when restoring PV is undesired because the data could've diverged from initial snapshot and we wanted stuff from snapshot in the PV.

dymurray (approved):

PR #563, opened 2022-02-02
KubeVirt Default Image

kaovilai:
Closes [OADP-139](https://issues.redhat.com/browse/OADP-139)


pkg/credentials/credentials.go
```diff
@@ -114,6 +114,16 @@ func getAzurePluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
 	return fmt.Sprintf("%v/%v/%v:%v", os.Getenv("REGISTRY"), os.Getenv("PROJECT"), os.Getenv("VELERO_AZURE_PLUGIN_REPO"), os.Getenv("VELERO_AZURE_PLUGIN_TAG"))
 }
 
+func getKubeVirtPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
```
kaovilai, line 117: Current quirk. If `VELERO_KUBEVIRT_PLUGIN_REPO`  set, it would break. due to using `PROJECT=konveyor` default.
If they want to set it to something else, should be using UnsupportedOverrides `kubevirtPluginImageFqin`.
May want to revisit this or leave as is.

kaovilai: quirk resolved after konveyor now contains kubevirt

dymurray (approved):

PR #565, opened 2022-02-06
Changed backup suite to be more generic & changed mssql app to mysql

mperetzred:
- Changed backup suite to be more generic so it could be used later with other CSI drivers, different backup spec, etc.
- Changed mssql app to mysql because mssql fails with restic as it uses deploymentconfig (restore gets stuck on InProgress; known bug).


tests/e2e/sample-applications/mysql-persistent/mysql-persistent-template.yaml
```diff
@@ -0,0 +1,151 @@
...
+    kind: PersistentVolumeClaim
+    metadata:
+      name: mysql
+      namespace: mysql-persistent
+      labels:
+        app: mysql
+    spec:
+      accessModes:
+      - ReadWriteOnce
+      storageClassName: gp2
```
Mathieu-Ferraton, line 87: The gp2 storageClass is specific to AWS, and the deployment fails due to pending pvc for clusters deployed outside aws. Maybe this could be left blank, so that we use the clsuter's default storage class?

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi-template.yaml
```diff
@@ -0,0 +1,151 @@
...
+      labels:
+        app: mysql
+    spec:
+      ports:
+      - name: mysql
+        port: 3306
+      selector:
+        name: mysql
+  - apiVersion: v1
+    kind: PersistentVolumeClaim
```
Mathieu-Ferraton, line 78: Same comment, the pvc is declared twice

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-template.yaml
mperetzred: yes, agree. The MSSQL had the storageclasses hardcoded, so made it in a similar way, cause I didn't want to change the logic. I plan later to refactor the apps, but not as part of this commit cause it's already too big.

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi-template.yaml
mperetzred: fixed

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-template.yaml
```diff
@@ -0,0 +1,151 @@
...
+      namespace: mysql-persistent
+      labels:
+        app: mysql
+    spec:
+      ports:
+      - name: mysql
+        port: 3306
+      selector:
+        name: mysql
+  - apiVersion: v1
```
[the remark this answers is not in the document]
mperetzred: fixed

weshayutin:
@dymurray @kaovilai if we change the e2e mssql to mysql should we also change the example app to mysql in the same review or a follow up review? https://github.com/openshift/oadp-operator/blob/master/docs/examples/stateful.md

weshayutin:
@tiger my vote would be to change the example app as the app does not consistently come up properly most times.. probably works 50% of the time at best outside of e2e.    Let's get it down in a follow up patch would be my vote.  @mperetzred

weshayutin:
@mperetzred just in case you didn't see.. https://github.com/openshift/oadp-operator/pull/565 merged.  So you are unblocked :)  Thank you for your patience.

kaovilai:
> @mperetzred just in case you didn't see.. #565 merged. So you are unblocked :) Thank you for your patience.

/unhold

kaovilai (approved):

PR #568, opened 2022-02-10
Multi cloud e2e master rebase

deepakraj1997:
https://hackmd.io/Gzmvnu90RC20f-6OYnwskA

PR #573, opened 2022-02-17
E2E Tests: Fix azure templating name. 

deepakraj1997:
I am in process of removing this completely and making it platform agnostic as per review made here: https://github.com/openshift/oadp-operator/pull/568#discussion_r805298142

This is just a temporary patch


kaovilai:
What about here? Is the bool not working?
https://github.com/openshift/oadp-operator/pull/573/files#diff-f757ae4a7b575d13d9943ac5e3fc8fa0d225335f88d0db06f0e5eabf823f270aR88

deepakraj1997:
@kaovilai The bool is working perfectly fine. Its just that we need be able to test VSL and BSL in both CI and non-CI env.

kaovilai (approved):

PR #574, opened 2022-02-17
CloudStorage: Only delete and remove finalizer if there exists an annotation.

kaovilai:
New:
- Annotation `oadp.openshift.io/cloudstorage-delete: "true"` is now required to trigger bucket deletion

Fix:
- If bucket is deleted successfully at the cloud provider, a finalizer on the CR is removed so the CR goes deletion continues
- Add checks such that finalizer is only added if deletionTimestamp is nil.

Example annotation
```yaml
metadata:
  annotations:
    oadp.openshift.io/cloudstorage-delete: "true"
```

If user wants to remove the CloudStorage custom resource from the cluster without deleting bucket from cloud provider, they can do so by
- Don't add `oadp.openshift.io/cloudstorage-delete: "true"` annotation
- `oc delete cloudstorage/<name>`
- remove finalizer 
```
kubectl patch -n openshift-adp cloudstorage/<name> \
    --type json \
    --patch='[ { "op": "remove", "path": "/metadata/finalizers" } ]'
```


kaovilai:
Annotation ("oadp.openshift.io/cloudstorage-delete": "true") to activate bucket deletion in cloud providers, finalizer will be removed for you.

To remove CR without deleting bucket in cloud storage, user just have to remove finalizer.

controllers/bucket_controller.go
```diff
@@ -75,10 +76,32 @@ func (b BucketReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	if err != nil {
 		return result, err
 	}
+	if bucket.Annotations[oadpCloudStorageDeleteAnnotation] == "true" && bucket.DeletionTimestamp != nil {
```
shawn-hurley, line 79: What happens if Deletion Timestamp is not nil AND no annotation?

what happens if the annotation is True, or t or something like that? do we think that we should parse it for a boolean?

kaovilai: Deletion Timestamp is not nil AND no annotation
- Delete() is not called, object still contains finalizer so it will live in the cluster.
- User have two choices, remove the finalizer to remove Custom Resource without deleting cloud provider bucket  OR add annotation so that we processes the bucket deletion request with cloud provider, if successful, finalizer will be removed for user.

kaovilai: @shawn-hurley do you mean
```suggestion
	if bucket.Annotations[oadpCloudStorageDeleteAnnotation] == true && bucket.DeletionTimestamp != nil {
```
or
```suggestion
	if strings.ToLower(bucket.Annotations[oadpCloudStorageDeleteAnnotation]) == "true" && bucket.DeletionTimestamp != nil {
```

kaovilai: > I think I was thinking of using strconv.ParseBool helper
Sounds like a great idea.

> Wouldn't the rest of the reconcile be called? do we want that behavior if the deletion timestamp is not nil but we are not going to clean it up?
I see what you mean. How does this sound?
```suggestion
	if bucket.DeletionTimestamp != nil {
	     if !strconv.ParseBool(bucket.Annotations[oadpCloudStorageDeleteAnnotation]) {
	         return ctrl.Result{Requeue: false}, nil
	     }
	
```

shawn-hurley (approved):
LGTM after updating the event names

PR #581, opened 2022-03-01
Makefile refactor `deploy-tmp` to `velero-role-tmp` expose `apply-velerosa-role`

kaovilai:
Helps create velero SA roles when running operator controller-manager locally in IDE so velero and restic pods have permissions.
Normally OLM would handle this task when installing operator from OperatorHub.

Breaks out SA role creation code away from `make deploy` so it can be used separately.


shubham-pampattiwar (approved):
+1 @kaovilai

PR #590, opened 2022-03-08
pr template

kaovilai:


shawn-hurley:
I personally don't like this kind of template, but if folks on the team find them useful then go for it. 

Can you just remove me from the reviewers?

.github/pull_request_template.md
```diff
@@ -0,0 +1,8 @@
+# Summary of changes
+<!-- Fixes #(issue), doc updates, etc -->
+
+# PR Change Checklist:
```
mateusoliveira43, line 4: I do not like checklists as well. This should be automated checked by our CI and documented in a contributing doc

But having sections that can be apply to every PR, I think is really good

@shawn-hurley since you were against this one, would like your opinion about having something like
```md
## Description

<!-- what is the goal of this PR, if it is related to any issue, etc  -->

## What was done

<!-- how was the thought process of the changes made in this PR  -->

## How to test

<!-- how to test the changes this PR introcues  -->
```

PR #597, opened 2022-03-16
Datamover CRD design

savitharaghunathan:
Adding Data Mover CRD design


docs/design/datamover.md
```diff
@@ -0,0 +1,239 @@
...
+As an application developer, I would like to save the CSI snaphots in a S3 bucket. 
+
+Story 2:
+As a cluster admin, I would like to be able to restore CSI snapshots if disaster happens.
+
+## Design & Implementation details
+
+This design supports adding the data mover feature to the OADP operator and facilitates integrating various vendor implemented data movers. 
+
+![DataMover CRD](../images/datamovercrd.png)
```
shubham-pampattiwar, line 64: Please update this image path.

savitharaghunathan: Should I create a PR to just merge the image first? I think the path is correct. WDYT?

docs/design/datamover.md
```diff
@@ -0,0 +1,239 @@
...
+## Release Signoff Checklist
+
+- [ ] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open questions
+
+* PVC mover/ volumesnapshot mover
```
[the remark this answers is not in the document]
savitharaghunathan: Yes. From my understanding there is also another question here - are we going to support both? If yes, how?

docs/design/datamover.md
```diff
@@ -0,0 +1,239 @@
...
+    resticSecret:
+      name: <secret_name>
+```
+
+We will provide a sample codebase which the vendors will be able to extend and implement their own data movers. 
+
+
+### Default OADP Data Mover controller
+
+VolSync will be used as the default Data Mover for OADP and `restic` will be the supported method for backup & restore of PVCs. Restic repository details are configured in a `secret` object which gets used by the VolSync's resources. This design takes advantage of VolSync's two resources - `ReplicationSource` & `ReplicationDestination`. `ReplicationSource` object helps with taking a backup of the PVCs and using restic to move it to the storage specified in the restic secret. `ReplicationDestination` object takes care of restoring the backup from the restic repository. There will be a 1:1 relationship between the replication src/dest CRs and PVCs.
```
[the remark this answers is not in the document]
savitharaghunathan: Good point. I will add a line or two about VolSync installation.

docs/design/datamover.md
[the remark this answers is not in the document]
savitharaghunathan: @dymurray Expanded the question. PTAL when you get a moment. Thanks :)

docs/design/datamover.md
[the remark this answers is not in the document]
savitharaghunathan: Added couple of lines on the installation.

docs/design/datamover.md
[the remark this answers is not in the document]
dymurray: Looks better now, thanks!

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+
+Note: We will be supporting VolSync as the default data mover. 
+
+The DataMoverBackup Controller will watch for DataMoverBackup CR. Likewise, DataMoverRestore Controller will watch for DataMoverRestore CR. Both of these CRs will have a reference to a DataMoverType. 
+
+`DataMoverType` is a cluster scoped Custom Resource that will have details about the data mover provisioner. The provisioner will be responsbile for creating a DataMover plugin that will create a `DataMoverBackup` CR & `DataMoverRestore` CR. They will also be responsible for implementing the logic for DataMoverBackup & DataMoverRestore controller that corresponds to their data mover. The spec will also include a field to identify the PVCs that would be moved with the given provisioner. 
+
+```
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataMoverType
```
alaypatel07, line 72: nit: I wonder if DataMoverClass is a better-suited name?

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+metadata:
+  annotations:
+    oadp.openshift.io/default: "true"
+  name: <name>
+spec:
+  provisioner: <VolSync>
```
alaypatel07, line 78: nit: change this to field key `mover`?

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+  pvcIdentifier: <tagname>
```
alaypatel07, line 79: I am curious if we can have something more generic like a `selector` that is devoid of implementation type, i.e. pvc or snap

I could imagine having a separate field to reflect if the provisioner is going to work on pvcs or snapshots.  But that decision can be differed in the future (I see you have already tracked this in open questions) a key field as suggested will be more accomadative to those changes.

savitharaghunathan: Thanks for the review, Alay. tbh I was struggling to come up with this field name. It makes sense to have a generic field name.

savitharaghunathan: Resolved the feedback changes. Let me know if this field name is okay or not. Thanks again :)

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+
+This design supports adding the data mover feature to the OADP operator and facilitates integrating various vendor implemented data movers. 
+
+![DataMover CRD](../images/datamovercrd.png)
+
+Note: We will be supporting VolSync as the default data mover. 
+
+The DataMoverBackup Controller will watch for DataMoverBackup CR. Likewise, DataMoverRestore Controller will watch for DataMoverRestore CR. Both of these CRs will have a reference to a DataMoverClass. 
+
+`DataMoverClass` is a cluster scoped Custom Resource that will have details about the data mover provisioner. The provisioner (identified by the field `mover`) will be responsbile for creating a DataMover plugin that will create a `DataMoverBackup` CR & `DataMoverRestore` CR. They will also be responsible for implementing the logic for DataMoverBackup & DataMoverRestore controller that corresponds to their data mover. The spec will also include a field (`dmselector`) to identify the PVCs that would be moved with the given provisioner.
```
alaypatel07, line 68: extreme nit picking:

I wonder if we should use the term "provisioner"?

I would think using "mover" term in this context makes more sense, using pharses like "mover provisioner" IMO does not buy anything buy confusion, you could repharse this as follows:

```
`DataMoverClass` is a cluster scoped Custom Resource that will have details about the data mover. The specified mover will be registered in the system by creating the datamoverclass CR, addig a velero plugin that will create the appropriate resources for datamovement of a single datamoverclass and a controller that will reconcile the objects created by the plugin. The datamoverclass spec will also include a field (`dmselector`) to identify the PVCs that would be moved with the given provisioner.
```

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+```
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataMoverClass
+metadata:
+  annotations:
+    oadp.openshift.io/default: "true"
+  name: <name>
+spec:
+  mover: <VolSync>
+  dmselector: <tagname>
```
alaypatel07, line 79: should this be pvcSelector or snapshotSelector?

since we are not sure, I am also open to keeping it just `selector` and differ the decision pvc vs snapshot

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+
+```
+
+The above `DataMoverClass` name will be referenced in `DataMoverBackup` & `DataMoverRestore` CRs. This will help in selecting the data mover implementation during runtime. If the `DataMoverClass` name is not defined, then the default `DataMoverClass` will be used, which in this case will be `VolSync`
+
+### Data Mover Backup
+
+Assuming that the `DataMover Enable` flag is set to true in the DPA config, when a velero backup is created, it triggers DataMover plugin to create the `DataMoverBackup` CR in the app namespace. The plugin looks up for the PVCs in the user namespace mentioned in the velero backup and creates a `DataMoverBackup` CR for every PVC in that namespace.
```
alaypatel07, line 87: > The plugin looks up for the PVCs in the user namespace mentioned in the velero backup and creates a `DataMoverBackup` CR for every PVC in that namespace.

Should this be as follows?

> The plugin looks up for the PVCs in the user namespace mentioned in the velero backup and creates a `DataMoverBackup` CR for every PVC in that namespace that is filtered by the datamoverclass.spec.selector

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+
+```
+apiVersion: oadp.openshift.io/v1alpha1
+kind: DataMoverBackup
+metadata:
+  name: <name>
+spec:
+  DataMoverClass: <DataMoverClass name> 
+  - type: <VolumeSnapshot|PVC>
+    sourceClaimRef:
```
alaypatel07, line 100: [sourceRef:](https://github.com/kubernetes/kubernetes/blob/b456e70fe939271ebda96c03cdf4094ab0bdc4e9/pkg/apis/core/types.go#L477) can we model the ref based on that go type and remove the type field in line 99?

docs/design/datamover.md
kaovilai: Looks fine to me @ https://github.com/savitharaghunathan/oadp-operator/blob/data_mover_enhancement/docs/design/datamover.md

GitHub PR .md previews don't show (new??) images.

docs/design/datamover.md
savitharaghunathan: Edited :) Thanks, Alay!

docs/design/datamover.md
savitharaghunathan: updated!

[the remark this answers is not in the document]
savitharaghunathan: Thanks for the review, Scott. Clarified that the plugin is a  BackupItemAction plugin. Should I capture the race condition information anywhere?

docs/design/datamover.md
```diff
@@ -0,0 +1,237 @@
...
+      name: <snapshot_content_name>|<pvc_name>
+      namespace: <namespace>  //optional 
+    config:     //optional based on the datamover impl
+
+```
+### Data Mover Restore
+When a velero restore is triggered, the DataMover plugin looks for `DataMoverBackup` in the backup resources. If it encounters a `DataMoverBackup` resource, then the plugin will create a `DataMoverRestore` CR in the app namespace. It will populate the CR with the details obtained from the `DataMoverBackup` resource.
```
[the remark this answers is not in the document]
savitharaghunathan: Clarified that the plugin is a `RestoreItemAction` plugin

docs/design/datamover.md
savitharaghunathan: @kaovilai I think so.

PR #594, opened 2022-03-11
Running must gather after each tests

deepakraj1997:
- Was trying to run using oc client, but after some research, it may very cumbersome to run `oc adm ...` commands using oc go client.
- Running must gather using shell.

PR #599, opened 2022-03-17
Adding gather_logs_without_zip script for the use case of (#594)

deepakraj1997:
- Added gather_logs_without_zip script for must-gather to enable dumping of pod logs and DPA CR instance for each tests.
- Adding this in separate PR to facilitate build and push of oadp-must-gather image for e2e testing of [594](https://github.com/openshift/oadp-operator/pull/594).
- Avoiding dumping of other OADP CRs such as backups, restores etc., as it collects all of them (Including from previous test cases) for each test case.
- The purpose of #599 and #594 is to enable the collection of test logs so that they are easily searchable through our customized [ci-search](https://github.com/weshayutin/ci-search)

### DIFF
--- a/must-gather/collection-scripts/gather_logs_without_zip
+++ b/must-gather/collection-scripts/gather_logs_without_zip
@@ -1,0 +1,50 @@
+#!/bin/bash
+
+# Default to gathering 72h of logs unless specified as var
+logs_since="${logs_since:-72h}"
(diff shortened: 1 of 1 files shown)


savitharaghunathan (commented):
/lgtm
Thanks, Deepak

weshayutin:
@deepakraj1997 nice!

PR #600, opened 2022-03-17
remove support for `make deploy`

kaovilai:


sseago:
Nit: We should update the PR name/commit message to say that we're *removing* `make deploy` rather than merely deprecating it, since normally deprecation will keep the functionality around for a while.

sseago (approved):

PR #618, opened 2022-03-29
ImageStreamsBackup Custom Resource design.

kaovilai:
Related to [OADP-363](https://issues.redhat.com/browse/OADP-363)

### DIFF
--- a/docs/design/imagestream.md
+++ b/docs/design/imagestream.md
@@ -105,6 +105,7 @@ spec:
 The spec fields in the ImageStreamsRestore are identical to the ImageStreamsBackup except:
 - startBackup
 - veleroBackupReference
+- ImageStreamsBackupReference
 
 Source registry path in dockerImageReference will be ignored if inaccessble (ie. from another cluster's internal registry) and only the sha256 digest will be used when looking up image from object storage to restore to internal registry.
 


kaovilai:
> Why a new controller rather than a plugin that is able to hand back the data to be backed up by coping the manifefsts and blobs?

New controller approach allow us to rely on existing vetted solutions of tarring up images without complicated deduplication logic instead of within the confines of velero plugin that requires you to
- double check what's already uploaded in object storage to avoid duplication which may either require that we redownload what's already uploaded to check digest or trust the manifests is correct.

dymurray:
@shawn-hurley @kaovilai I think it's important we decide the pros and cons of going down each path. Our biggest concern with the plugin approach was that we wouldn't be able to dedup the layers in a single backup. The hope with an API/Controller approach was that we could bunch images into one action and get the dedup functionality.

It's worth noting that both approaches involve a change to the velero plugins to work properly... but if we went down this approach the gaps missing would be identical to data mover, which is a plugin creating a custom resource and having to wait for it complete for the backup to be complete.

The plugin backup method requires us to introduce a new concept of a "pre backup complete plugin" which gets wonky and would also require us to be writing data to a transient location in the pod. I'm unsure which approach is better... but I like the idea of exploring this being separate from the velero backup process if our goal is to simply enable backup/restore of the internal registry. What do you all think?

kaovilai:
> What do you all think?
I'm on the fence about both approach :) So no comment.

PR #641, opened 2022-04-14
Add DPA to `oc get all`

kaovilai:
Thoughts?

### DIFF
--- a/api/v1alpha1/oadp_types.go
+++ b/api/v1alpha1/oadp_types.go
@@ -189,7 +189,7 @@ type DataProtectionApplicationStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
-//+kubebuilder:resource:path=dataprotectionapplications,shortName=dpa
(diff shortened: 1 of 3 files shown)


shawn-hurley:
I would rather have backups and/or schedules in all but IDK especially for most ns's this is just going to be empty.

kaovilai:
Right. But if you are debugging and your context is in the namespace where DPA is created, it will show.

kaovilai:
> for most ns's this is just going to be empty.
Doesn't affect people day to day usage of their own namespace which is a good thing right? allows you to query all DPA when in the context of OADP namespace.

shawn-hurley:
Pretty sure you adding a list call, for something that is always empty. I just don't know if from a wider perspective this would make us good citizens because we know those namespaced list calls will mostly be empty. Just my two cents

kaovilai:
> adding a list call

I'm going to have to read more what this means.

Sounds like this would add burden to oc get outside the namespace context as well.

Issue #621, opened 2022-04-05
Bug: Bad reading of the AWS credentials file


lebrisg:
### Contact Details

lebrisg@hotmail.com

### Describe bug

When creating an AWS credentials file on Windows 10, the file is created with cr/nl characters at the end of every line.
Then, the oadp-*-aws-registry pod can't read the aws_access_key_id and displays an error message: CreateContainerConfigError (see operator pod error message below).
The same file created on Linux with only nl characters at the end of every line works perfectly fine.

### What happened?

Whatever platform [Linux/Microsoft Windows/MacOs] you use to create the credentials, the AWS keys should be read fine.

### OADP Version

1.0.1 (Stable)

### OpenShift Version

4.10

### Operator pod logs

```shell
2022-04-05T14:21:53.588Z        INFO    controller.dataprotectionapplication    Error finding AWS Profile for the supplied AWS credential       {"reconciler group": "oadp.openshift.io", "reconciler kind": "DataProtectionApplication", "name": "dpa-sample", "namespace": "openshift-adp"}
2022-04-05T14:21:53.588Z        INFO    controller.dataprotectionapplication    Error parsing provider secret cloud-credentials for backupstoragelocation openshift-adp/dpa-sample-1    {"reconciler group": "oadp.openshift.io", "reconciler kind": "DataProtectionApplication", "name": "dpa-sample", "namespace": "openshift-adp"}
```

### New issue

- [X] This issue is new

[OADP-463](https://issues.redhat.com//browse/OADP-463)

kaovilai:
Thanks for the report. Tracking on OADP JIRA at https://issues.redhat.com/browse/OADP-463

PR #652, opened 2022-04-29
Replace carriage return when getting provider secrets for registry + E2E changes

kaovilai:
[OADP-463](https://issues.redhat.com//browse/OADP-463)
Close https://github.com/openshift/oadp-operator/issues/621

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -4,6 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strings"
 
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
 	"github.com/openshift/oadp-operator/pkg/common"
@@ -15,7 +16,6 @@ type DefaultPluginFields struct {
 	IsCloudProvider    bool
 	SecretName         string
 	MountPath          string
-	BslSecretName      string
 	BslMountPath       string
 	EnvCredentialsFile string
 	PluginImage        string
@@ -33,7 +33,6 @@ var (
 			IsCloudProvider:    true,
 			SecretName:         "cloud-credentials",
 			MountPath:          "/credentials",
-			BslSecretName:      "bsl-cloud-credentials-aws",
 			BslMountPath:       "/bsl-cloud-credentials-aws",
 			EnvCredentialsFile: common.AWSSharedCredentialsFileEnvKey,
 			PluginName:         common.VeleroPluginForAWS,
@@ -43,7 +42,6 @@ var (
 			IsCloudProvider:    true,
 			SecretName:         "cloud-credentials-gcp",
 			MountPath:          "/credentials-gcp",
-			BslSecretName:      "bsl-cloud-credentials-gcp",
 			BslMountPath:       "/bsl-cloud-credentials-gcp",
 			EnvCredentialsFile: common.GCPCredentialsEnvKey,
 			PluginName:         common.VeleroPluginForGCP,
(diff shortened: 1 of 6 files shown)


tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -499,13 +533,23 @@ var _ = Describe("Configuration testing for DPA Custom Resource", func() {
 					if installCase.DpaSpec.BackupLocations[0].Velero.Config != nil {
 						installCase.DpaSpec.BackupLocations[0].Velero.Config["credentialsFile"] = "bsl-cloud-credentials-aws/cloud"
 					}
+					if installCase.TestCarriageReturn {
+						installCase.DpaSpec.BackupLocations[0].Velero.Config["credentialsFile"] = "bsl-cloud-credentials-aws-with-carriage-return/cloud"
```
kaovilai, line 537: Using carriage return secret

tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -62,6 +63,39 @@ var _ = Describe("Configuration testing for DPA Custom Resource", func() {
 			},
 			WantError: false,
 		}, nil),
+		Entry("Default velero CR, test carriage return", InstallCase{
+			Name:         "default-cr",
+			BRestoreType: RESTIC,
+			TestCarriageReturn: true,
```
kaovilai, line 69: Added a test for carriage return case

tests/e2e/utils/common_utils.go
```diff
@@ -37,3 +38,9 @@ func WriteFile(credFile string, data []byte) error {
 	err := ioutil.WriteFile(credFile, data, 0644)
 	return err
 }
+
+func ReplaceSecretDataNewLineWithCarriageReturn(data []byte) []byte {
```
kaovilai, line 42: New function to replace newline with carriage return in secret data.

tests/e2e/e2e_suite_test.go
```diff
@@ -103,6 +107,8 @@ var _ = BeforeSuite(func() {
 			cloudCreds := GetAzureCreds(cloudCredData)
 			err = CreateCredentialsSecret(cloudCreds, namespace, "bsl-cloud-credentials-azure")
 			Expect(err).NotTo(HaveOccurred())
+			err = CreateCredentialsSecret(utils.ReplaceSecretDataNewLineWithCarriageReturn(cloudCreds), namespace, "bsl-cloud-credentials-azure-with-carriage-return")
```
kaovilai, line 110: Create another secret to use to test carriage return case.

dymurray (approved):
LGTM, thanks for the tests

pkg/credentials/credentials.go
```diff
@@ -33,7 +33,6 @@ var (
 			IsCloudProvider:    true,
 			SecretName:         "cloud-credentials",
 			MountPath:          "/credentials",
-			BslSecretName:      "bsl-cloud-credentials-aws",
```
deepakraj1997, line 36: Kinda followed the logic we had for the previous ones as I included this for the case of CI, now that we are pushing this for new feature, we expect the user to have created the secret name with key given in the `credentialsFile` right ?

kaovilai: I think user should be able to pick which secret is used. My change didn't break functionality so I think it's okay.

kaovilai: Prior to this PR, user was expected to have created a secret anyway but the name was hardcoded. This meant I couldn't use a different secret for the carriage return case unless I overwrite the current secret with the fixed name.

This only affected the credentialFile config case which most customers won't be using. With this PR the secret name for registry is now flexible if they specified bsl secret with credentialsFile config.

kaovilai: >  we expect the user to have created the secret name with key given in the credentialsFile right?

Yes.

Issue #373, opened 2021-10-07
Bug: Velero pod not running after sending Velero CR to OADP operator 


Frank51:
### Contact Details

yunqian.chen@dell.com

### Describe bug

A OADP-Operator 0.3.0 was deployed on the openshift cluster. And then a Velero CR was sent to OADP-Operator via 'oc apply -f <Velero-CR-json-file>', but velero pod is not running on the namespace. 

I also tried sending this json file via Openshift UI to oadp-operator, but it shows that the 'spec.backupStorageLocations' is 'Required value'.

Here is the Velero CR json file I applied.

{
	"kind": "Velero",
	"apiVersion": "konveyor.openshift.io/v1alpha1",
	"metadata": {
		"name": "oadp-velero",
		"namespace": "velero-ppdm",
		"creationTimestamp": null
	},
	"spec": {
		"custom_velero_plugins": [{
			"image": "velero-dd-pluin-uri",
			"name": "powerprotect-velero-dd"
		}],
		"default_velero_plugins": ["openshift"],
		"enable_restic": false,
		"olm_managed": false,
		"velero_feature_flags": "EnableLocalMode"
	}
}

### What happened?

The velero pod should running on the namespace. But the no velero pod showing on the namespace.

### OADP Version

0.3.x (Beta)

### OpenShift Version

4.6

### Velero pod logs

```shell
Velero pod is not running.
```

### Operator pod logs

```shell
I1006 21:38:33.529251       1 request.go:665] Waited for 1.029013433s due to client-side throttling, not priority and fairness, request: GET:https://172.30.0.1:443/apis/autoscaling.openshift.io/v1beta1?timeout=32s
2021-10-06T21:38:35.343Z	INFO	controller-runtime.metrics	metrics server is starting to listen	{"addr": ":8080"}
2021-10-06T21:38:35.345Z	INFO	setup	starting manager
I1006 21:38:35.346433       1 leaderelection.go:248] attempting to acquire leader lease velero-ppdm/8b4defce.openshift.io...
2021-10-06T21:38:35.347Z	INFO	controller-runtime.manager	starting metrics server	{"path": "/metrics"}
I1006 21:38:51.546526       1 leaderelection.go:258] successfully acquired lease velero-ppdm/8b4defce.openshift.io
[...]

Issue #349, opened 2021-09-29
AWS BSL should allow empty region


kaovilai:
**Describe the bug**

AWS BSL cannot have empty region. This is not how it should be according to [BSL docs for aws plugin](https://github.com/vmware-tanzu/velero-plugin-for-aws/blob/main/backupstoragelocation.md)
Region can be queried from aws s3 api or not needed through means such as `s3Url`

**To Reproduce**
Create Velero CR with region empty

**Expected behavior**
Velero reconcile

**Provide the following logs**
Got error reconcile condition on velero object.
`region for AWS backupstoragelocation config cannot be empty`

**Version of operator being used**
 - 0.3.0

**Additional context**
Fix by changing the following
https://github.com/openshift/oadp-operator/blob/5a5d4439b0ccbbc200a606cb133b5ad56e8e7e07/controllers/bsl.go#L144-L146

dymurray:
This validation logic should be changed to check if a region is not specified, the user must supply `s3Url`. If `s3Url` and `region` are missing, then it should fail and display the error.

shubham-pampattiwar:
We do not need `s3Url` as well. Both are optional. The AWS BSL can be created with just specifying the `bucket` and nothing in the `config` spec. We need the `region` for registry deployment though. So more like the `region` validation for AWS BSL should be modified to be called only for B/R of images `velero.Spec.BackupImages == nil || *velero.Spec.BackupImages)` . This will be same as the `prefix` validation I believe https://github.com/openshift/oadp-operator/blob/v0.3.0/controllers/bsl.go#L148

Issue #12, opened 2020-06-16
Add full support for Velero customized install


dymurray:
https://velero.io/docs/v1.4/customize-installation/

- [x] Enable custom plugins
- [x] Install in any namespace
- [x] #175
- [x] Enable Restic integration
- [x] Enable Features (CSI)
- [x] Customize resource requests + limits
- [x] Configure more than one storage location for backup/volumes (minimum 1 per provider ideally N per provider)
- [x] Do not configure BSL during install (optional)
- [x] Use storage provider with self-signed certificate
- [x] #176

0.3.0 Regression
- [x] #379

yipng:
Can the restore-resource-priorities velero server option be included as part of this customization as well?

kaovilai:
0.3.0 release regression from 0.2
Regressed feature
- [ ] Do not configure BSL during install (optional)

PR #662, opened 2022-05-02
[OADP-460] Don't mount cloud provider secret volumes with NoDefaultBackupLocation

kaovilai:
- [ ] Add/fix e2e test

Fixes [OADP-460](https://issues.redhat.com//browse/OADP-460)

This is a follow up to #420, #607, #349, #380, #373, and #12

### DIFF
--- a/controllers/validator.go
+++ b/controllers/validator.go
@@ -20,16 +20,16 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error) 
 
 	if dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
 		if len(dpa.Spec.BackupLocations) != 0 {
-			return false, errors.New("DPA CR Velero configuration cannot have backup locations if NoDefaultBackupLocation is set")
+			return false, errors.New("DPA CR Velero configuration cannot have backup locations if noDefaultBackupLocation is set")
 		}
 	} else {
 		if len(dpa.Spec.BackupLocations) == 0 {
-			return false, errors.New("no backupstoragelocations configured, ensure a backupstoragelocation has been configured or use the noDefaultLocationBackupLocation flag")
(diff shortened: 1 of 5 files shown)


controllers/validator.go
```diff
@@ -18,12 +18,18 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 		return false, errors.New("DPA CR Velero configuration cannot be nil")
 	}
 
-	if len(dpa.Spec.BackupLocations) == 0 && !dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
-		return false, errors.New("no backupstoragelocations configured, ensure a backupstoragelocation has been configured or use the noDefaultLocationBackupLocation flag")
+	if dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
+		if len(dpa.Spec.BackupLocations) != 0 {
```
dymurray, line 22: Not sure this is true, we really are just saying if we have any BSLs defined in the DPA spec that none of them have `default: true` right?

dymurray: after digging, `velero install` takes the same approach https://github.com/vmware-tanzu/velero/blob/77be862b9c8d4d0b9187a8239059e6ed9b54cd1b/pkg/cmd/cli/install/install.go#L307-L309

so its fine

dymurray (dismissed):
VISIACK, changes look sane

controllers/validator.go
dymurray: We only use install lib for the deployment, so we really can choose our own behavior here. Awaiting merge for more input

kaovilai: @shubham-pampattiwar this flag is only used in [func AllResources](https://github.com/vmware-tanzu/velero/blob/cb11bf1924273428890b4f75e012c6e18cf8fed7/pkg/install/resources.go#L251) which we skipped entirely from using.

dymurray (approved):
VISIACK

PR #664, opened 2022-05-04
Makefile CLUSTER_TYPE auto discovery

kaovilai:


Makefile
```diff
@@ -1,5 +1,5 @@
 OADP_TEST_NAMESPACE ?= openshift-adp
-CLUSTER_TYPE ?= aws
+CLUSTER_TYPE ?= $(shell oc get infrastructures cluster -o jsonpath='{.status.platform}' | awk '{print tolower($0)}')
```
deepakraj1997, line 2: Just a heads up here, OC client, if you look [this](https://github.com/openshift/oadp-operator/pull/664/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R35-R39), you can see that OC CLI has been exposed at this DIR, its working probably cause this directory is in PATH, which wasnt the case when I added the CLI in the release repository. In future if this happens to fail due to client not being found, we can use this directly from the path.

kaovilai: @deepakraj1997 I don't know which line you were referencing. The link didn't contain line number

kaovilai: Nvm.. the link contained line number.. but it just wasn't part of this PR's diff and the link didn't work.

deepakraj1997: Oop, sorry, looks like you got it resolved. Thanks!

PR #673, opened 2022-05-06
OADP 486 - Update velero service account permissions

dymurray:


kaovilai:
If we are removing velero-privileged here maybe remove the SCC from reconciler in go too?

dymurray:
I didn't even realize we were still creating that SCC. I don't want to introduce too much churn into 1.0.3.... I will keep that for now
open a separate issue to remove the SCC bits

kaovilai (approved):

PR #684, opened 2022-05-10
OADP-145: post-restore script for restores with DCs and defaultVolumesToRestic

sseago:
When restoring a backup with defaultVolumesToRestic=true and
DeploymentConfigs in the backup, this script must be run after the
completion of the restore to clean up the restic restore actions.

The openshift plugins will disconnect DC pods from the ReplicationController
when defaultVolumesToRestic is true and set DC replicas to 0.
This script will remove the restore pods and reset DC replicas so that
new pods will be created that are properly reconnected to the DC.

### DIFF
--- a/docs/scripts/dc-restic-post-restore.sh
+++ b/docs/scripts/dc-restic-post-restore.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+set -e
+
+OADP_NAMESPACE=${OADP_NAMESPACE:=openshift-adp}
+
+if [[ $# -ne 1 ]]; then
+  echo "usage: ${BASH_SOURCE} restore-name"
+  exit 1
+fi
+
+echo using OADP Namespace $OADP_NAMESPACE
+echo restore $1
+
+echo Deleting disconnected restore pods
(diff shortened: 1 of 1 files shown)


docs/scripts/dc-restic-post-restore.sh
```diff
@@ -0,0 +1,28 @@
...
+if [[ $# -ne 1 ]]; then
+  echo "usage: ${BASH_SOURCE} restore-name"
+  exit 1
+fi
+
+echo using OADP Namespace $OADP_NAMESPACE
+echo restore $1
+
+echo Deleting disconnected restore pods
+oc delete pods -l openshift.io/disconnected-from-dc=$1
```
kaovilai, line 15: Per https://github.com/openshift/openshift-velero-plugin/pull/141/files#r891349632 please update all new labels to `oadp.openshift.io`

kaovilai (changes requested):

docs/scripts/dc-restic-post-restore.sh
sseago: Updated.

kaovilai (approved):

PR #694, opened 2022-05-12
make bundle related images fix

kaovilai:
Looks like `operator-sdk generate bundle` gets Related images from [FindRelatedImages](https://github.dev/operator-framework/operator-sdk/blob/fb15b62e9c0defee208086ccab79112c68843454/internal/cmd/operator-sdk/generate/internal/relatedimages.go#L29) which iterate over manifests of type deployment. It then check that the container has env RELATED_IMAGE_

Looks like a regression with operator-sdk that appeared around [v1.19.1 release](https://github.com/operator-framework/operator-sdk/releases/tag/v1.19.1)

Edit: submitted a fix to operator-sdk https://github.com/operator-framework/operator-sdk/pull/5765 is approved, this will close.

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -698,6 +698,24 @@ spec:
                         value: latest
                       - name: VELERO_KUBEVIRT_PLUGIN_TAG
                         value: v0.2.0
+                      - name: RELATED_IMAGE_velero
+                        value: quay.io/konveyor/velero:latest
+                      - name: RELATED_IMAGE_velero_restic_restore_helper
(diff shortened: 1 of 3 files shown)


bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -807,19 +825,19 @@ spec:
     - image: quay.io/konveyor/velero:latest
       name: velero
     - image: quay.io/konveyor/velero-restic-restore-helper:latest
-      name: velero_restic_restore_helper
+      name: velero-restic-restore-helper
```
kaovilai, line 828: side-effect, `_` are now `-`

kaovilai (commented):
sideeffects of specifying `RELATED_IMAGE_` in deployment.

PR #704, opened 2022-05-19
Fix OADP 526

shubham-pampattiwar:
Fixes https://issues.redhat.com/browse/OADP-526

### DIFF
--- a/controllers/restic.go
+++ b/controllers/restic.go
@@ -172,6 +172,7 @@ func (r *DPAReconciler) buildResticDaemonset(dpa *oadpv1alpha1.DataProtectionApp
 	}
 
 	// get resource requirements for restic ds
+	// ignoring err here as it is checked in validator.go
 	resticResourceReqs, _ := r.getResticResourceReqs(dpa)
 
 	installDs := install.DaemonSet(ds.Namespace,
--- a/controllers/restic.go
+++ b/controllers/restic.go
@@ -171,8 +171,11 @@ func (r *DPAReconciler) buildResticDaemonset(dpa *oadpv1alpha1.DataProtectionApp
 		return nil, fmt.Errorf("ds cannot be nil")
 	}
 
+	// get resource requirements for restic ds
+	resticResourceReqs, _ := r.getResticResourceReqs(dpa)
+
 	installDs := install.DaemonSet(ds.Namespace,
-		install.WithResources(r.getResticResourceReqs(dpa)),
+		install.WithResources(resticResourceReqs),
 		install.WithImage(getVeleroImage(dpa)),
 		install.WithAnnotations(dpa.Spec.PodAnnotations),
 		install.WithSecret(false))
--- a/controllers/velero.go
+++ b/controllers/velero.go
@@ -392,11 +392,14 @@ func (r *DPAReconciler) buildVeleroDeployment(veleroDeployment *appsv1.Deploymen
 	}
 	r.ReconcileRestoreResourcesVersionPriority(dpa)
 
+	// get resource requirements for velero deployment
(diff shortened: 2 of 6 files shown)


controllers/restic.go
```diff
@@ -171,8 +171,11 @@ func (r *DPAReconciler) buildResticDaemonset(dpa *oadpv1alpha1.DataProtectionApp
 		return nil, fmt.Errorf("ds cannot be nil")
 	}
 
+	// get resource requirements for restic ds
+	resticResourceReqs, _ := r.getResticResourceReqs(dpa)
```
pranavgaikwad, line 175: Ideally, this error should be surfaced into the DPA CR so that users are aware of the invalid values they put

controllers/velero.go
```diff
@@ -690,20 +715,42 @@ func (r *DPAReconciler) getResticResourceReqs(dpa *oadpv1alpha1.DataProtectionAp
 	}
 
 	if dpa != nil && dpa.Spec.Configuration != nil && dpa.Spec.Configuration.Restic != nil && dpa.Spec.Configuration.Restic.PodConfig != nil {
-		// Set custom limits and requests values if defined on VELERO Spec
-		if dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests != nil {
-			ResourcesReqs.Requests[corev1.ResourceCPU] = resource.MustParse(dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests.Cpu().String())
-			ResourcesReqs.Requests[corev1.ResourceMemory] = resource.MustParse(dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests.Memory().String())
+		// Set custom limits and requests values if defined on Restic Spec
+		if dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests.Cpu() != nil && dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests.Cpu().Value() != 0 {
+			parsedQuantity, err := resource.ParseQuantity(dpa.Spec.Configuration.Restic.PodConfig.ResourceAllocations.Requests.Cpu().String())
```
pranavgaikwad, line 720: +1 to using `ParseQuantity()`

pranavgaikwad (changes requested):
It'd be nice to have a test for invalid resource inputs. Otherwise looks great. Thanks for quick turnaround!

controllers/velero.go
```diff
@@ -392,11 +392,14 @@ func (r *DPAReconciler) buildVeleroDeployment(veleroDeployment *appsv1.Deploymen
 	}
 	r.ReconcileRestoreResourcesVersionPriority(dpa)
 
+	// get resource requirements for velero deployment
+	veleroResourceReqs, _ := r.getVeleroResourceReqs(dpa)
```
kaovilai, line 396: I would add a comment here that err is checked in validator.go otherwise just check err anways.

controllers/restic.go
kaovilai: I would add a comment about err check that is done in validator.go

pranavgaikwad:
I just realized that since resource is a typed field. Validation on the CR itself will kick in and prevent users from putting stupid values. @shubham-pampattiwar approving

pranavgaikwad (approved):

controllers/restic.go
shubham-pampattiwar: Done

controllers/velero.go
shubham-pampattiwar: Done, updated PR.

kaovilai (approved):

kaovilai:
From OADP-526 Fix version says 1.0.3
/cherrypick oadp-1.0

PR #710, opened 2022-05-25
Update velero/restic/registry to check owner types

kaovilai:


tests/e2e/lib/dpa_helpers.go
```diff
@@ -269,19 +298,22 @@ func GetVeleroPods(namespace string) (*corev1.PodList, error) {
 	return podList, nil
 }
 
-func AreVeleroPodsRunning(namespace string) wait.ConditionFunc {
+func AreVeleroDeploymentReplicasReady(namespace string) wait.ConditionFunc {
 	return func() (bool, error) {
-		podList, err := GetVeleroPods(namespace)
+		deploymentList, err := GetVeleroDeploymentList(namespace)
```
kaovilai, line 303: switching to Get

kaovilai:
Feedback:
- use Get instead of List when possible
- check owner observedGeneration also when checking for updated pods. applies to Deployment and Daemonset.

tests/e2e/lib/restic_helpers.go
```diff
@@ -47,64 +47,57 @@ func HasCorrectNumResticPods(namespace string) wait.ConditionFunc {
...
-		// loop until pod status is 'Running' or timeout
-		for _, podInfo := range (*podList).Items {
-			if podInfo.Status.Phase != "Running" {
-				return false, err
-			}
+		if lastObservedReadyResticDaemeonSet != nil &&
+			lastObservedReadyResticDaemeonSet.CreationTimestamp.Equal(&rds.CreationTimestamp) && // same object
+			!reflect.DeepEqual(lastObservedReadyResticDaemeonSet.Spec, rds.Spec) &&
+			// if spec is not deep equal, something updated. Consider ready only if observedGeneration has increased.
+			rds.Status.ObservedGeneration == lastObservedReadyResticDaemeonSet.Status.ObservedGeneration {
```
kaovilai, line 68: DeepEqual check seems redundant here. it would make more sense to deepequal an updated spec passing in as an optional parameter.

tests/e2e/lib/dpa_helpers.go
```diff
@@ -82,23 +85,41 @@ func WithBackupImages(backupImage bool) DpaCROption {
...
+	}
+}
+
+var veleroPrefix = "velero-e2e-" + string(uuid.NewUUID())
+
+func GetVeleroPrefix() string {
+	return veleroPrefix
+}
+
+var dpa *oadpv1alpha1.DataProtectionApplication
```
shawn-hurley, line 108: I don't know if we should do it this way. 

Global Singletons are scary, and it is just as scary that you are pass around the reference and I am not seeing any deep copies. 

I wonder if we can re-state the goals of what we want this to do (sounds like some sort of fixtures) and maybe try and come up with a different approach?

kaovilai: Would it help if I remove the pointer? I didn't mean to pass reference around. Was just unexporting it. it was `Dpa` before with the same pointer type..
```suggestion
var dpa oadpv1alpha1.DataProtectionApplication
```

kaovilai: This var is supposed to be loaded once and for accessing an original copy loaded from JSON per comment in line below it

kaovilai: resolved by only allowing access to dpa via getters which returns a copy.

Redesigns will be followed up in another PR.

kaovilai:
pod spec checking could take a week or longer to implement. WIP here https://github.com/kaovilai/oadp-operator/pull/48. Probably be separate PR.

kaovilai:
TODO: GCP, When not using credentialsFile, it is expected that [sreviceAccount is added to config.](https://github.com/vmware-tanzu/velero-plugin-for-gcp/blob/c48ee5e4c11c2e0f4fb88f784dc5a50aa12790d3/velero-plugin-for-gcp/object_store.go#L162)

kaovilai:
eh flake
```
velero container error line#757: time="2022-05-28T00:39:52Z" level=error msg="Error backing up item" backup=openshift-adp/mongo-e2e-501f0cb1-de1e-11ec-aff0-0a580a82be25 error="error getting volume info: rpc error: code = Unknown desc = googleapi: Error 404: The resource 'projects/openshift-gce-devel/zones/us-central1-a/disks/ci-op-4dy5npht-4818d-h-pvc-d3eea69a-45e7-4556-8cc2-4ab39f475fd6' was not found, notFound" logSource="pkg/backup/backup.go:417" name=mongo-64fdc89c69-szslb
```

kaovilai:
other than one spec flake. witnessed successful b/r nodefaultbackuplocation on gcp

kaovilai:
TODO: registry flake.
/test 4.9-operator-e2e-azure

kaovilai:
Don't think we need to use credentialsFile anymore looking at the structure. Initially introduced in multi-cloud e2e pr. Seems like bsl.Spec.Credentials works in all cases.
May attempt removing when we refactor e2e.

kaovilai:
breakthrough greenery other than registry flakes for commit https://github.com/openshift/oadp-operator/pull/710/commits/7556d8f060fb29e39f31f4229f8ee03e62ac9295
https://prow.ci.openshift.org/pr-history/?org=openshift&repo=oadp-operator&pr=710

kaovilai:
restic flake
/test 4.7-operator-e2e-azure

tests/e2e/backup_restore_suite_test.go
```diff
@@ -11,7 +11,7 @@ import (
 	. "github.com/onsi/gomega"
 	"github.com/openshift/oadp-operator/api/v1alpha1"
 	. "github.com/openshift/oadp-operator/tests/e2e/lib"
-	utils "github.com/openshift/oadp-operator/tests/e2e/utils"
+	i "github.com/openshift/oadp-operator/tests/e2e/lib/init"
```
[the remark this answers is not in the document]
kaovilai: init actually wouldn't work. reserve word lol.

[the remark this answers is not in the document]
kaovilai: or `in`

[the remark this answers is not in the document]
alaypatel07: what about `libinit`?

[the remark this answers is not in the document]
kaovilai: yes sounds good.

.github/workflows/go-tests.yml
```diff
@@ -23,7 +23,7 @@ jobs:
 
     - uses: actions/setup-go@v2
       with:
-        go-version: '^1.16.6' # The Go version to download (if necessary) and use.
+        go-version: '^1.18' # The Go version to download (if necessary) and use.
```
shawn-hurley, line 26: Make sure that the downstream is ready for this change.

tests/e2e/lib/init/init.go
```diff
@@ -0,0 +1,104 @@
+package init
+
+import (
+	"flag"
+	"os"
+	"time"
+)
+
+// Common vars obtained from flags passed in ginkgo.
+var credFile, namespace, credSecretRef, instanceName, provider, ci_cred_file, settings, artifact_dir, oc_cli string
```
shawn-hurley, line 10: There has to be a better way to do this stuff than global variables IMO. I also am very concerned with the package name.

kaovilai:
Plan of action from 1:1
- gonna split this into smaller PRs by reverting to a good known state b64e96e

kaovilai:
will reapply incrementally fixes for the broken test case.

tests/e2e/lib/init/init.go
kaovilai: re: package variable. only accessed via getters. these are string and has no pointers.

PR #711, opened 2022-05-25
OADP-535 allow for nullable resource allocations

dymurray:
This change allows for null resource allocations fields.

**NOTE** this includes a manual update to the DPA CRD to allow for `null` requests and allocations fields. This means that `make manifests` will no longer give us exactly what we need and will require folks to constantly add it back in. We need to figure out a better path forward for this.
[OADP-535](https://issues.redhat.com//browse/OADP-535)

### DIFF
--- a/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
+++ b/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
@@ -213,6 +213,7 @@ spec:
                                   anyOf:
                                   - type: integer
                                   - type: string
+                                  nullable: true
                                   pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                                   x-kubernetes-int-or-string: true
                                 description: 'Limits describes the maximum amount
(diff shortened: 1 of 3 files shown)


bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -216,6 +216,7 @@ spec:
                                   x-kubernetes-int-or-string: true
```
shubham-pampattiwar, line 216: Don't  we need `nullable: true` on line number 209 as well as on line number 353.

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -208,6 +208,7 @@ spec:
                           resourceAllocations:
                             description: ResourceAllocations defines the CPU and Memory
                               resource allocations for the restic Pod
+                            nullable: true
                             properties:
                               limits:
                                 additionalProperties:
```
shubham-pampattiwar, line 214: Don't  we need `nullable: true` specifically for limits as well as requests ?

bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -216,6 +217,7 @@ spec:
...
                                 description: 'Limits describes the maximum amount
                                   of compute resources allowed. More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/'
+                                nullable: true
                                 type: object
                               requests:
                                 additionalProperties:
```
pranavgaikwad, line 223: needs `nullable` for the additionalProperties.

pranavgaikwad (changes requested):

pranavgaikwad (approved):
ACK, thank you for the fix!

PR #727, opened 2022-06-03
Add datamover controller deployment

shubham-pampattiwar:
This PR adds a new reconcile function `ReconcileDataMoverController` which enables the deployment of volume-snapshot-mover
Related to the first task of this issue https://github.com/konveyor/volume-snapshot-mover/issues/69


controllers/dpa_controller.go
```diff
@@ -102,6 +102,7 @@ func (r *DPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		r.ReconcileResticDaemonset,
 		r.ReconcileVeleroServiceMonitor,
 		r.ReconcileVeleroMetricsSVC,
+		r.ReconcileDataMoverController,
```
savitharaghunathan, line 105: Are we adding datamover by default?

shubham-pampattiwar: Just for this PR, did not want to post a long PR, I have a [TODO ](https://github.com/openshift/oadp-operator/pull/727/files#diff-a97f576da4c6bee59ebf0b5c4462c39c6eb7d1906bd517dbcef41acdec0263a0R30)added and will be addressed in the plugin install PR which will be based on `datamover` as a default plugin listed on the DPA CR

controllers/datamover.go
```diff
@@ -0,0 +1,120 @@
...
+		if dataMoverDeployment.ObjectMeta.CreationTimestamp.IsZero() {
+			dataMoverDeployment.Spec.Selector = &metav1.LabelSelector{
+				MatchLabels: map[string]string{
+					"component": "datamover-controller",
+				},
+			}
+		}
+
+		// update the Deployment template
+		err := r.builddataMoverDeployment(dataMoverDeployment, &dpa)
```
dymurray, line 43: `buildDataMoverDeployment`

dymurray (commented):
This should be configurable from the DPA spec, a user needs to explicitly enable this feature for 1.1 since we are tech-preview.

controllers/dpa_controller.go
kaovilai: shots fired.

kaovilai:
`MySQL application NoDefaultBackupStorageLocation` was removed from master. Rebase to test without this entry.

controllers/datamover.go
shubham-pampattiwar: resolved

shubham-pampattiwar:
> This should be configurable from the DPA spec, a user needs to explicitly enable this feature for 1.1 since we are tech-preview.

@dymurray Added a new spec `enableDataMover` which needs to be specifically set to `true` by the user in order to enable the data-mover functionality

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -433,6 +433,24 @@ spec:
                 - patch
                 - update
                 - watch
+            - apiGroups:
```
kaovilai, line 436: These changes should be going to https://github.dev/shubham-pampattiwar/oadp-operator/blob/add-datamover-controller/config/rbac/role.yaml then run `make bundle`

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -441,6 +441,11 @@ spec:
                         type: string
                     type: object
                 type: object
+              enableDataMover:
```
kaovilai, line 444: run `make bundle` to propagate CRD changes to bundle/

kaovilai (changes requested):
Please apply all bundle changes in config/ then use `make bundle` to generate updated bundle. Reach out if you have any questions.

api/v1alpha1/oadp_types.go
```diff
@@ -184,7 +186,9 @@ type DataProtectionApplicationSpec struct {
 	// BackupImages is used to specify whether you want to deploy a registry for enabling backup and restore of images
 	// +optional
 	BackupImages *bool `json:"backupImages,omitempty"`
-
+	// EnableDataMover is used to specify whether you want to deploy the volume snapshot mover controller and a modified csi datamover plugin
+	// +optional
+	EnableDataMover bool `json:"enableDataMover,omitempty"`
```
dymurray, line 191: I was wondering, is it worth adding a new type `TechPreviewConfiguration` that hides this behind a pointer? This would allow us to change this API as we go to GA if upstream velero makes this totally native.

Then if we have other tech preview features in the future we can always use this field.

controllers/datamover.go
```diff
@@ -0,0 +1,162 @@
...
+		Template: corev1.PodTemplateSpec{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels: map[string]string{
+					"component": common.DataMoverController,
+				},
+			},
+			Spec: corev1.PodSpec{
+				RestartPolicy:      corev1.RestartPolicyAlways,
+				Containers:         datamoverContainer,
+				ServiceAccountName: "openshift-adp-controller-manager",
```
dymurray, line 141: do we have this SA declared as a const anywhere?

dymurray (commented):
Overall changes look sane to me. Couple questions

api/v1alpha1/oadp_types.go
galt-tr: yeah but this will be a supported field, just maybe something we don't keep around in future versions. So I don't think `unsupportedOverrides` is the correct place for it

bundle/manifests/oadp-operator.clusterserviceversion.yaml
shubham-pampattiwar: Will it fine if I just update the `role.yaml` and not run `make bundle` because it breaks our csv

api/v1alpha1/oadp_types.go
shubham-pampattiwar: @dymurray I am not sure if `TechPreviewConfigutarion` is the right naming here, I am unsure. Can we not rely on the docs to let the users know that this is a tech-preview feature ? Or is it the right way to do from the API side as you suggested ?

dymurray: yeah but this will be a supported field, just maybe something we don't keep around in future versions. So I don't think unsupportedOverrides is the correct place for it

shubham-pampattiwar: Cool, I like `features` more and its combination with the docs. Will be similar to what `Velero` does

bundle/manifests/oadp-operator.clusterserviceversion.yaml
shubham-pampattiwar: Done updated the PR

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
shubham-pampattiwar: Done updated the PR

controllers/datamover.go
shubham-pampattiwar: Done updated the PR

api/v1alpha1/oadp_types.go
shubham-pampattiwar: @dymurray updated the PR, PTAL, Thanks !

pkg/credentials/credentials.go
```diff
@@ -101,6 +101,9 @@ func getAWSPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
 }
 
 func getCSIPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
+	if dpa.Spec.Features != nil && dpa.Spec.Features.EnableDataMover {
+		return common.CSIDataMoverPluginImage
```
kaovilai, line 105: Please respect unsupportedOverrides of CSIPluginImageKey

kaovilai: Is this piece unused? I see the following function not referring to this at all.
```
func (r *DPAReconciler) getDataMoverImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
```

pkg/common/common.go
```diff
@@ -3,23 +3,29 @@ package common
...
-	RegistryImage        = "quay.io/konveyor/registry:latest"
-	KubeVirtPluginImage  = "quay.io/konveyor/kubevirt-velero-plugin:v0.2.0"
+	VeleroImage             = "quay.io/konveyor/velero:latest"
+	OpenshiftPluginImage    = "quay.io/konveyor/openshift-velero-plugin:latest"
+	AWSPluginImage          = "quay.io/konveyor/velero-plugin-for-aws:latest"
+	AzurePluginImage        = "quay.io/konveyor/velero-plugin-for-microsoft-azure:latest"
+	GCPPluginImage          = "quay.io/konveyor/velero-plugin-for-gcp:latest"
+	CSIPluginImage          = "quay.io/konveyor/velero-plugin-for-csi:latest"
+	CSIDataMoverPluginImage = "quay.io/konveyor/velero-plugin-for-csi:data-mover"
+	DataMoverImage          = "quay.io/konveyor/volume-snapshot-mover:latest"
```
kaovilai, line 26: Can you comment the differences between these two?

kaovilai (changes requested):
please add a few comments regarding what each images does.

pkg/credentials/credentials.go
shubham-pampattiwar: Done

pkg/common/common.go
shubham-pampattiwar: Added comments

api/v1alpha1/oadp_types.go
```diff
@@ -30,6 +30,7 @@ const ReconcileCompleteMessage = "Reconcile complete"
 
 const OadpOperatorLabel = "openshift.io/oadp"
 const RegistryDeploymentLabel = "openshift.io/oadp-registry"
+const DataMoverDeploymentLabel = "openshift.io/oadp-data-mover"
```
kaovilai, line 33: Might want to change this after we finish discussion at https://github.com/openshift/openshift-velero-plugin/pull/141/files#r891349632 about oadp labels.

kaovilai: can change later. testing now. visiack.

dymurray (dismissed):
Tested PR and worked for me, agree with tiger on the unsupported override field for CSI plugin.

Please also rename the data mover deployment to be `volume-snapshot-mover`

dymurray:
Please add missing CRDs for volume snapshot mover as well

api/v1alpha1/oadp_types.go
shubham-pampattiwar: Thank you !

kaovilai:
Could be another PR.
Add example annotation to 
- config/samples/volumesnapshotbackup.yaml
- config/samples/volumesnapshotrestore.yaml

Add above to 
- config/samples/kustomization.yaml

Would resolve operator-sdk warnings as well as provide default template in openshift web console yaml view when creating these resources.
```
WARN[0000] Warning: Value datamover.oadp.openshift.io/v1alpha1, Kind=VolumeSnapshotBackup: provided API should have an example annotation 
WARN[0000] Warning: Value datamover.oadp.openshift.io/v1alpha1, Kind=VolumeSnapshotRestore: provided API should have an example annotation 
```

kaovilai (approved):

savitharaghunathan (approved):

eemcmullan (approved):
Tested these changes and was able to complete a backup and restore successfully using data mover.

savitharaghunathan:
backup/restore workflow worked with these changes! 
/pony party

PR #737, opened 2022-06-14
OADP-363 plugin-registry design, registry deployment removal proposal

kaovilai:
[OADP-363](https://issues.redhat.com//browse/OADP-363)

Implementation
https://github.com/openshift/openshift-velero-plugin/pull/145
https://github.com/openshift/oadp-operator/pull/743

linked lucidchart
[plugin-registry](https://lucid.app/documents/view/14dd9d7d-38a5-40f7-a378-c89a99a1ace6)

Rich markdown with working in-line image can be viewed at https://github.com/kaovilai/oadp-operator/blob/velero-plugin-registry-designdoc/docs/design/plugin-registry.md

Simplified diagram of changes.
[Plugin Registry Imagestream Backup Workflow - Page 1 (2)](https://lucid.app/documents/view/14dd9d7d-38a5-40f7-a378-c89a99a1ace6)


docs/design/plugin-registry.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Security, Risks, and Mitigations
+
+As more backups and/or new Backup Storage Location are added and thus caching more initialized registries configurations, we may want to periodically garbage collect unused or outdated backup storage location registry configurations to optimize memory usage.
+
+## Design Details
+
+OADP-Operator controller manager will
+ * set environment variables on the velero container
+   * tell openshift-velero-plugin to initialize [udistribution transport](https://github.com/kaovilai/udistribution/blob/main/pkg/image/udistribution/docker_transport.go#L36) to be used to talk to storage drivers such as s3
```
kaovilai, line 67: will be adding lucidchart to https://github.com/kaovilai/udistribution/ and add a summary here of what udistribution repo contains.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,105 @@
...
+
+Do we want registry logs to be included in velero backup and/or restore logs?
+
+## Summary
+
+removes a listening HTTP server and connect [openshift-velero-plugin](https://github.com/openshift/openshift-velero-plugin/) directly with the [distribution/distribution](https://github.com/distribution/distribution) code to serve requests from [container/image/copy](https://github.com/containers/image/tree/main/copy) method as needed. A "library" to help achieve this is [kaovilai/udistribution](https://github.com/kaovilai/udistribution) which may be moved to konveyor org.
+
+## Motivation
+
+Current image backup workflow depends on registry deployment with exposed insecure routes allowing anyone knowing the URL of the exposed routes to push or pull images to the registry.
```
shubham-pampattiwar, line 41: Maybe you can add more to the `Motivation` section like:
- How existing registry is of brittle nature and prone to transient errors
- Also, that this alternative approach will help avoid maintaining a dependency on registry deployment, the operator will not have to worry about this deployment resource

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,105 @@
...
+
+### Goals
+
+Exposed routes are eliminated.
```
shubham-pampattiwar, line 45: Can be added: "Remove dependency on registry deployment"

docs/design/plugin-registry.md
shubham-pampattiwar: @kaovilai thank you, yes this will be very helpful to understand the workflow.

kaovilai: This is the published lucidchart. It will be added to PR soon.
https://lucid.app/documents/view/14dd9d7d-38a5-40f7-a378-c89a99a1ace6

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
```
dymurray, line 45: FYI in markdown use a list
```
- <>
- <>
```

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+Remove dependency on registry deployment.
+Be backward compatible with existing image backup workflow.
```
dymurray, line 47: Additional goal is keep deduplication across backups

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Non-Goals
+
+Creating another custom resource and a controller to replace usage of [distribution/distribution](https://github.com/distribution/distribution).
+
+## Proposal
+
+We propose that insecure routes are eliminated by removing registry deployment, and making network requests to [distribution/distribution](https://github.com/distribution/distribution) storage drivers such as [s3-aws](https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go) from within velero pod via openshift-velero-plugin.
```
dymurray, line 55: Believe this should say "making network requests via storage drivers". I.e. we are making network requests to the backing store using native golang drivers. And this is abstracted away from a new udistribution library.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+
+### Implementation Details/Notes/Constraints [optional]
+
+The logs relating to registry will be mixed into velero pod logs
```
dymurray, line 59: How noisy are these logs? Is this something we can control with the debug flag?

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+## Design Details
+
+[![plugin-registry](plugin-registry.svg)](https://lucid.app/documents/view/14dd9d7d-38a5-40f7-a378-c89a99a1ace6)
+
+[udistribution](https://github.com/kaovilai/udistribution) library has been created to help with the following:
+- Initializing Registry [App](https://github.com/distribution/distribution/blob/b5e2f3f33dbc80d2c40b5d550541467477d5d36e/registry/handlers/app.go#L58) from configuration resolved from environment variables passed via [function parameters](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/client/client.go#L58).
+- Add a custom container/images docker transport with modified [makeRequestToResolvedURLOnce()](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/image/udistribution/docker_client.go#L559) to use [Registry App ServeHTTP method](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/image/udistribution/docker_client.go#L613) instead of sending HTTP requests to a listening HTTP server.
+
+OADP-Operator controller manager will
+ * set environment variables on the velero container
```
dymurray, line 74: What if a user has multiple s3 BSLs? How does the operator specify which environment variables to use? Wouldn't this require setting the env vars per backup run?

What if instead the plugin reads these env vars from a configmap?

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+   * tell openshift-velero-plugin to initialize [udistribution transport](https://github.com/kaovilai/udistribution/blob/main/pkg/image/udistribution/docker_transport.go#L36) to be used to talk to storage drivers such as s3
+
+openshift-velero-plugin will
+ * resolve registry secret for a given Backup Storage Location by looking up the secret in the namespace of the Backup Storage Location.
```
dymurray, line 78: Sounds like perhaps we *aren't* setting these env vars on the velero container then right?

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+ * initialize [udistribution transport](https://github.com/kaovilai/udistribution/blob/main/pkg/image/udistribution/docker_transport.go#L36) to be used to talk to storage drivers such as s3
+ * substitute transport name given to `image.Copy()` from `docker://` to [generated udistribution transport name](https://github.com/openshift/openshift-velero-plugin/blob/1600327cb3f6f9f60ade880aef8fe16d34e6fb04/velero-plugins/imagecopy/imagestream.go#L102) for bsl when plugin registry is requested.
+
+### Test Plan
+- Check for no user experience changes
+- Image backup continues to work and upload to s3/azure/gcp
+
```
dymurray, line 85: Add e2e test that backs up application with internal image and verify the image exists in the object store

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+  - @shawn-hurley
+  - @shubham-pampattiwar
+creation-date: 2022-06-14
+last-updated: 2022-06-14
+status: implementable
+---
+
+# Plugin Registry
+
+Plugin registry enhancement replaces current usage of [registry](https://github.com/konveyor/distribution/) [deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/) insecure [routes](https://access.redhat.com/documentation/en-us/openshift_container_platform/4.10/html/networking/configuring-routes)
```
dymurray, line 22: Replaces with what? :)

docs/design/plugin-registry.md
kaovilai: Operator only set one env (not storage location name) to use plugin registry. Secrets, bucket names, can be discovered by velero plugin.

docs/design/plugin-registry.md
kaovilai: Yes. No need to pass multiple envs like before.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+- Initializing Registry [App](https://github.com/distribution/distribution/blob/b5e2f3f33dbc80d2c40b5d550541467477d5d36e/registry/handlers/app.go#L58) from configuration resolved from environment variables passed via [function parameters](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/client/client.go#L58).
+- Add a custom container/images docker transport with modified [makeRequestToResolvedURLOnce()](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/image/udistribution/docker_client.go#L559) to use [Registry App ServeHTTP method](https://github.com/kaovilai/udistribution/blob/d7f491d7c354caa1df6893d20c735b9c08c20108/pkg/image/udistribution/docker_client.go#L613) instead of sending HTTP requests to a listening HTTP server.
+
+OADP-Operator controller manager will
+ * set environment variables on the velero container
+   * tell openshift-velero-plugin to initialize [udistribution transport](https://github.com/kaovilai/udistribution/blob/main/pkg/image/udistribution/docker_transport.go#L36) to be used to talk to storage drivers such as s3
```
savitharaghunathan, line 75: Are we only supporting S3 for now? If so, documenting it somewhere would help.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,114 @@
...
+- Check for no user experience changes
+- Image backup continues to work and upload to s3/azure/gcp
+
+#### Test Changes
+- Remove check for ready registry deployments
+- Deploy a dummy bsl registry deployment and check for removal
+
+### Upgrade / Downgrade Strategy
+
+Upon upgrading, operator will attempt to for detected BSLs remove any remaining registry deployments.
```
savitharaghunathan, line 92: Possible reword maybe? its a bit confusing

docs/design/plugin-registry.md
sseago: You can get logs per container:
```
oc logs <pod-name> -c <conatiner_name>
```

docs/design/plugin-registry.md
kaovilai: +1

docs/design/plugin-registry.md
kaovilai: The registry sidecar container idea was thrown out.

kaovilai:
In-line image will work when viewed at https://github.com/kaovilai/oadp-operator/blob/velero-plugin-registry-designdoc/docs/design/plugin-registry.md

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,131 @@
...
+
+- Exposed routes are eliminated.
+- Remove dependency on registry deployment.
+- Be backward compatible with existing image backup workflow.
+  - images backed up will continue to be in the same directory in object storage as they were prior to this enhancement.
+- Keep deduplication across backups.
+
+### Non-Goals
+
+Creating another custom resource and a controller to replace usage of [distribution/distribution](https://github.com/distribution/distribution).
```
eemcmullan, line 55: What benefit would this provide?

kaovilai: This fits under non goals and what is avoided by this design. It was originally one of the proposed solutions to resolve insecure registry route.

Creating custom resource and a controller would mean we develop our own controller to watch for imagebackup custom resource created from plugin which OADP then process to upload image data into object storage.

https://github.com/openshift/oadp-operator/blob/befe8eac1b7cd6a503d193c870b7c14e14f66d96/docs/design/plugin-registry.md?plain=1#L124-L127

kaovilai: Similar to how data mover create VolumeSnapshotBackup, VolumeSnapshotRestore

shubham-pampattiwar (commented):
+1 Great approach Thank you @kaovilai. Added some comments. Also, it would be really great if you could help me understand more and address the following:
- A high-level component diagram, +1 on the current lucid chart, but could just add a component diagram without the implementation details, or just explain the involved components/functions like Whats is registry App, Transport, serveHTTP method etc. I believe this will be really helpful in understanding the approach flow. (A brief description about each would also do without a diagram)
- Are the requests made via udistribution secure ? 
- How does the image get transported ? I understand there is a working PoC, but lets say we have an app image `foo` that we want to transport to s3 , how would the new approach go about  it?

docs/design/plugin-registry.md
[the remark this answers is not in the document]
kaovilai: Thanks.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,131 @@
...
+## Release Signoff Checklist
+
+- [x] Enhancement is `implementable`
+- [ ] Design details are appropriately documented from clear requirements
+- [ ] Test plan is defined
+- [ ] User-facing documentation is created
+
+## Open Questions [optional]
+
+Do we want registry logs to be included in velero backup and/or restore logs?
```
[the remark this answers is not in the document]
dymurray: No I think there are 2 things we are talking about here. 
1. the plugin logs that records the process of backing up and restoring images.
2. The registry library doing the actual push/pull of the images.

I think 1. should be in the backup/restore for sure. 2. feels overly verbose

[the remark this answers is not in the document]
kaovilai: Yes. Plugin logs. However the logs generated from registry will not be in backup logs in object storage since the velero backup logger isn't being passed to the ServeHTTP method.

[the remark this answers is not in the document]
kaovilai: We do want them but I believe that per current implementation it is not possible.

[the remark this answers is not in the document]
kaovilai: On the positive side, this means that we can leave log level as debug for the registry component without additional configuration flags.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,131 @@
...
+- Be backward compatible with existing image backup workflow.
+  - images backed up will continue to be in the same directory in object storage as they were prior to this enhancement.
+- Keep deduplication across backups.
+
+### Non-Goals
+
+Creating another custom resource and a controller to replace usage of [distribution/distribution](https://github.com/distribution/distribution).
+
+## Proposal
+
```
dymurray, line 58: More from a high level this proposal is to

>Redesign the image backup/restore workflow to instead consume a library directly from within the plugin rather than standing up a transient registry deployment.

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,131 @@
...
+We propose that insecure routes are eliminated by removing registry deployment, and to instead have openshift-velero-plugin make network requests to registry APIs internally via [golang ServeHTTP function provided by distribution/distribution](https://github.com/distribution/distribution/blob/3e4f8a0ab1476a9516d02d76fbf869480541657f/registry/handlers/app.go#L624) which will then through storage drivers such as [`s3-aws`](https://github.com/distribution/distribution/blob/main/registry/storage/driver/s3-aws/s3.go),`gcs` , and `azure` write to desired object storage.
+
+
+### Implementation Details/Notes/Constraints [optional]
+
+First, a udistribution client are to be initialized using [`udistribution.NewTransportFromNewConfig()`](https://github.com/kaovilai/udistribution/blob/2b5e16ac1f8efa0bbcdc513ae7103d4f56f3befa/pkg/image/udistribution/docker_transport.go#L50) function.
```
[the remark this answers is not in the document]
dymurray: I see its more explained in line 88, but in this section stay away from in depth code references just give a high level description of the library

docs/design/plugin-registry.md
```diff
@@ -0,0 +1,131 @@
...
+
+Current image backup workflow depends on registry deployment with exposed insecure routes allowing anyone knowing the URL of the exposed routes to push or pull images to the registry.
+There are some transient errors such as pod health probe timeouts that are resolved by this design.
+This approach will eliminate registry deployment dependency which the operator will no longer have to reconcile.
+
+### Goals
+
+- Exposed routes are eliminated.
+- Remove dependency on registry deployment.
+- Be backward compatible with existing image backup workflow.
```
savitharaghunathan, line 49: Is it possible for both workflows to coexist? For example, backup A chooses to use the workflow proposed in this design and backup B decides to current workflow. That brings to my next question - can this selection be configurable?

savitharaghunathan: Thanks for the clarification, Tiger

savitharaghunathan: Test case example makes a lot of sense. Can we add this to test plan?

docs/design/plugin-registry.md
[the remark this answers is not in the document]
kaovilai: Had to reread section description in https://github.com/konveyor/enhancements/blob/master/guidelines/enhancement_template.md
Will move this to design detail and summary here.

docs/design/plugin-registry.md
kaovilai: yup

shubham-pampattiwar (approved):

PR #747, opened 2022-06-29
OADP-599: Auto create s3 restic secrets for volsync from BSLs

savitharaghunathan:


controllers/datamover.go
```diff
@@ -160,3 +180,149 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+	}
+
+	return nil, nil
+}
+
+func (r *DPAReconciler) createResticSecret(dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
+
+	// obtain restic secret name from the config
+	// if no name is mentioned in the config, ten assume default restic secret name
+	resticsecretname := "restic-secret"
```
eemcmullan, line 273: should this be a const value?

controllers/datamover.go
```diff
@@ -160,3 +180,149 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+	resticSecret := corev1.Secret{}
+	if err := r.Get(r.Context, types.NamespacedName{Namespace: r.NamespacedName.Namespace, Name: resticsecretname}, &resticSecret); err != nil {
+		r.Log.Error(err, "unable to fetch Restic Secret")
+		return false, err
+	}
+
+	// validate restic secret
+	val := r.validateDataMoverCredential(&resticSecret)
+	if !val {
+		return false, fmt.Errorf("invalid restic credential")
```
eemcmullan, line 289: Since this only checks for a password, should the error reflect that? Something like "no password supplied in restic secret." wdyt?

controllers/datamover.go
```diff
@@ -160,3 +180,149 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+		// TODO: Azure & GCP
+	}
+
+	return nil, nil
+}
+
+func (r *DPAReconciler) createResticSecret(dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
+
+	// obtain restic secret name from the config
+	// if no name is mentioned in the config, ten assume default restic secret name
```
eemcmullan, line 272: NIT: then

controllers/datamover.go
```diff
@@ -160,3 +180,149 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+// Check if there is a valid user supplied restic secret
+func (r *DPAReconciler) validateDataMoverCredential(resticsecret *corev1.Secret) bool {
+	if resticsecret == nil {
+		return false
+	}
+	for key, val := range resticsecret.Data {
+		if key == ResticPassword {
+			if len(val) != 0 {
+				return false
+			}
```
eemcmullan, line 193: should this be the other way around? returning true if there is a supplied password

api/v1alpha1/oadp_types.go
```diff
@@ -165,6 +165,9 @@ type Features struct {
 	// EnableDataMover is used to specify whether you want to deploy the volume snapshot mover controller and a modified csi datamover plugin
 	// +optional
 	EnableDataMover bool `json:"enableDataMover,omitempty"`
+	// User suplied Restic Secret name
```
eemcmullan, line 168: NIT: supplied*

controllers/datamover.go
savitharaghunathan: ha ha 100% :D what was I thinking. Thanks Emily

controllers/datamover.go
```diff
@@ -160,3 +180,149 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+				)
+			}
+		}
+		// TODO: Azure & GCP
+	}
+
+	return nil, nil
+}
+
+func (r *DPAReconciler) createResticSecret(dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
```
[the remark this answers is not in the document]
savitharaghunathan: no. I dont see any use case where it needs to be deleted or updated. These stay as long as we have oadp operator running.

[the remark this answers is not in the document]
savitharaghunathan: Do you have any use case in mind?

controllers/datamover.go
savitharaghunathan: changed condition

savitharaghunathan:
/hold
todo: Add plugin support for passing restic secret name in VSB & VSR CR
Add support for restic secret names in snap mover

controllers/datamover.go
```diff
@@ -160,3 +180,153 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+				awsProfile = value
+			}
+
+			key, secret, err := r.parseAWSSecret(bslSecret, secretKey, awsProfile)
+			if err != nil {
+				r.Log.Info(fmt.Sprintf("Error parsing provider secret %s for backupstoragelocation", secretName))
+				return nil, err
+			}
+
+			repo := "s3:s3.amazonaws.com/" + bsl.Spec.ObjectStorage.Bucket
```
shubham-pampattiwar, line 222: I think this is not always the case for s3 repo path. The URL format of a bucket is either of two options: http://s3.amazonaws.com/[bucket_name]/ http://[bucket_name].s3.amazonaws.com/, not entirely sure about this.

controllers/datamover.go
```diff
@@ -160,3 +180,153 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+
+			rsecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%s-resticsecret", bsl.Name),
```
shubham-pampattiwar, line 226: `resticsecret` <-> `restic-secret`

controllers/datamover.go
```diff
@@ -160,3 +180,153 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+					Namespace: bsl.Namespace,
+					Labels: map[string]string{
+						"openshift.io/oadp-bsl-name": bsl.Name,
```
shubham-pampattiwar, line 229: Please add the labels as a string constant in `oadp_types.go`

controllers/datamover.go
savitharaghunathan: @shubham-pampattiwar I am using the format that we use in restic repo secret. Which format would you prefer?

controllers/datamover.go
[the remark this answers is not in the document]
eemcmullan: Gotcha. Just wanted to double check

controllers/datamover.go
```diff
@@ -160,3 +180,153 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+					},
+				},
+			}
+
+			op, err := controllerutil.CreateOrUpdate(r.Context, r.Client, rsecret, func() error {
+
+				if len(pass) == 0 {
```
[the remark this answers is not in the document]
savitharaghunathan: looks like I am checking that already in L#191. I will remove this redundant check

controllers/datamover.go
kaovilai: https://restic.readthedocs.io/en/stable/030_preparing_a_new_repo.html?highlight=minio#minio-server

shubham-pampattiwar: The repo URL is a bit dynamic here to be created by us, IMO lets keep this on the user to supply via a secret.

savitharaghunathan: Made changes to obtain repo from the user

eemcmullan (approved):

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -445,6 +445,24 @@ spec:
                 description: Features defines the configuration for the DPA to enable
                   the OADP tech preview features
                 properties:
+                  dataMoverCredential:
```
dymurray, line 448: I'm wondering should we move these configs to be:
```
features:
  datamover:
    enable: true
    credentials: <secret>
```

This way if we have more configs we can stick it under a top level `datamover` key?

shubham-pampattiwar (changes requested):
Added a few suggestions.

kaovilai:
Don't forget `make bundle` I expect to see changes in `config/` replicated in `bundle/`

controllers/datamover.go
```diff
@@ -20,6 +22,19 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 )
 
+const (
+	ResticPassword   = "RESTIC_PASSWORD"
+	ResticRepository = "RESTIC_REPOSITORY"
+	ResticsecretName = "dm-restic-secret"
```
eemcmullan, line 28: this will be changed to 'dm-credential' right?

savitharaghunathan: yes. Submitting a commit right away

savitharaghunathan: done

kaovilai:
@savitharaghunathan please run `make bundle` to duplicate changes to bundle/

And revert changes related to allowing null in CRD

savitharaghunathan:
@kaovilai I did run make bundle. I had to manually put the nullable true back.

kaovilai (approved):

savitharaghunathan:
@kaovilai Is there any file changes you are expecting to see other than the two that are in the commits?

kaovilai:
Nope. Just misread the changes

kaovilai:
Last
/retest
was 23 hours ago
Resolution to label length related failures were last resolved 15 hours ago
https://github.com/openshift/release/tree/master/ci-operator/config/openshift/oadp-operator

kaovilai:
@hhpatel14 any reason https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-master__4.10.yaml is different to 
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-master__4.9.yaml ?

Specifically
```diff
+  openshift_release_golang-1.16:
+    name: builder
+    namespace: ocp
+    tag: rhel-8-golang-1.16-openshift-4.9
-  openshift_release_golang-1.16:
-    name: release
-    namespace: openshift
-    tag: golang-1.16
```

controllers/datamover.go
```diff
@@ -129,6 +144,11 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
 			fmt.Sprintf("performed %s on datamover deployment %s/%s", op, dataMoverDeployment.Namespace, dataMoverDeployment.Name),
 		)
 	}
+
+	_, err = r.createResticSecret(&dpa)
```
shubham-pampattiwar, line 148: Maybe rename the function to `CreateDataMoverSecret` ? And Also, not add this here, instead add this in the main OADP Oeprator Batch Reconcile function ?

controllers/datamover.go
```diff
@@ -186,3 +206,158 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+		if key == ResticPassword {
+			if len(val) == 0 {
+				return false
+			}
+		}
+	}
+	return true
+}
+
+func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtectionApplication, bsl velerov1.BackupStorageLocation, dmresticsecretname string, pass []byte) (*corev1.Secret, error) {
```
shubham-pampattiwar, line 226: Once you move this function to the main batch reconcile, I believe you will need to add the check whether dm is enabled else skip this function's execution altogether.

controllers/datamover.go
```diff
@@ -186,3 +206,158 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+				repobase = bsl.Spec.Config["s3Url"]
+			}
+
+			repo := repobase + bsl.Spec.ObjectStorage.Bucket
+			rsecret := &corev1.Secret{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      fmt.Sprintf("%s-volsync-restic", bsl.Name),
+					Namespace: bsl.Namespace,
+					Labels: map[string]string{
+						oadpv1alpha1.OadpBSLnameLabel: bsl.Name,
```
shubham-pampattiwar, line 259: As the secret is create by OADP Operator, maybe add the operatort label as well, so that we do not miss on the reconciles via events on these secrets

controllers/datamover.go
```diff
@@ -186,3 +206,158 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+		}
+		// TODO: Azure & GCP
+	}
+
+	return nil, nil
+}
+
+func (r *DPAReconciler) createResticSecret(dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
+
+	// obtain restic secret name from the config
```
shubham-pampattiwar, line 303: Once we move this to Reconcile batch, you will also have to fetch fresh DPA instance.

shubham-pampattiwar (commented):
Changes look sane to me, Added some suggestions though. Thank you @savitharaghunathan

controllers/datamover.go
```diff
@@ -58,7 +73,7 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
 		},
 	}
 
-	if (dpa.Spec.Features == nil) || (dpa.Spec.Features != nil && !dpa.Spec.Features.EnableDataMover) {
+	if (dpa.Spec.Features == nil) || (dpa.Spec.Features != nil && dpa.Spec.Features.DataMover != nil && !dpa.Spec.Features.DataMover.Enable) {
```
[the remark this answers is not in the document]
savitharaghunathan: if `dpa.Spec.Feature` is nil, wouldnt `dpa.Spec.Feature.Datamover` be nil as well?

savitharaghunathan:
@shubham-pampattiwar Thanks for the elaborate review. I updated the func according to the last feedback. When you get a moment can you take a look at it please?

controllers/datamover.go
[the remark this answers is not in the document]
shubham-pampattiwar: So currently this might work, but lets say we add more features and Datamover is not the only tech preview feature, then it might right?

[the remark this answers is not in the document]
savitharaghunathan: Makes sense. will add that to the conditional.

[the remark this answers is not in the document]
shubham-pampattiwar: Thank you !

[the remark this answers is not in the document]
savitharaghunathan: updated

controllers/datamover.go
```diff
@@ -186,3 +203,176 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+func (r *DPAReconciler) ReconcileDataMoverResticSecret(log logr.Logger) (bool, error) {
+
+	// fetch latest DPA instance
+	dpa := oadpv1alpha1.DataProtectionApplication{}
+	if err := r.Get(r.Context, r.NamespacedName, &dpa); err != nil {
+		return false, err
+	}
+
+	// create secret only if data mover is enabled
+	if dpa.Spec.Features.DataMover != nil && dpa.Spec.Features.DataMover.Enable {
```
shubham-pampattiwar, line 315: I think you need `dpa.Spec.Features != nil` as well, else you might hit a nil ptr @savitharaghunathan

shubham-pampattiwar, line 315: Suggestion: Create a function which has all these checks on dpa and returns to you whether datamover is enabled or not, might make the code more readable and we can also escape from bugs if a dev forgets to adda nil check. WDTY ? @savitharaghunathan

savitharaghunathan: on it

controllers/datamover.go
```diff
@@ -186,3 +203,176 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
...
+		backupStorageLocationList := velerov1.BackupStorageLocationList{}
+
+		// Fetch the configured backupstoragelocations
+		if err := r.List(r.Context, &backupStorageLocationList, bslListOptions); err != nil {
+			return false, err
+		}
+
+		for _, bsl := range backupStorageLocationList.Items {
+			if strings.Contains(bsl.Name, dpa.Name) {
+				_, err := r.createResticSecretsPerBSL(&dpa, bsl, dmresticsecretname, res_pass)
```
shubham-pampattiwar, line 367: @savitharaghunathan Are we assuming that the given datamover secret will work for the BSL configured ? According to this current code, if I am not wrong, we are creating a dm secret per BSL and password provided in this secret will work for all the BSLs configured ?
For e.g: 
1. OADP has 2 BSLs - AWS and GCP. User supplied one dm secret,this PR will be creating restic secret using the dm secret for both the BSLs, right ? I think we should only create the restic secret for the dm one.
2. OADP has 2 BSLs - Both AWS. Same issue.

**Suggested Solution:**
We add a BSLName and Provider field to Deature -> DataMover/ And we create the restic secret only for the BSL specified here. WDYT ?

shubham-pampattiwar: Please correct me if I am wrong here.

bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -1,3 +1,5 @@
+
+---
```
kaovilai, line 2: I just tried running make bundle. These two lines don't exist here. Yes they exists in config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml but it doesn't belong in bundle/manifests.

Just let `make bundle` do its thing. Hopefully that fixes it?
```suggestion
```

kaovilai (changes requested):

bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
savitharaghunathan: Let me try that. Thanks @kaovilai

savitharaghunathan: updated

shubham-pampattiwar (approved):
LGTM ! Thank you so much @savitharaghunathan !

controllers/datamover.go
```diff
@@ -186,3 +203,187 @@ func (r *DPAReconciler) getDataMoverLabels() map[string]string {
 	labels[oadpv1alpha1.DataMoverDeploymentLabel] = "True"
 	return labels
 }
+
+// Check if there is a valid user supplied restic secret
+func (r *DPAReconciler) validateDataMoverCredential(resticsecret *corev1.Secret) bool {
+	if resticsecret == nil {
+		return false
+	}
+	for key, val := range resticsecret.Data {
```
dymurray, line 212: We want to update this to check if the key exists at all, right now I get no error if the key isn't set

config/crd/bases/datamover.oadp.openshift.io_volumesnapshotbackups.yaml
```diff
@@ -40,6 +40,14 @@ spec:
               protectedNamespace:
                 description: Namespace where the Velero deployment is present
                 type: string
+              resticSecretRef:
+                description: Restic Secret reference for given BSL
+                properties:
+                  name:
+                    description: 'Name of the referent. More info: https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#names
+                      TODO: Add other useful fields. apiVersion, kind, uid?'
```
dymurray, line 48: I would remove this from the description

PR #762, opened 2022-07-18
[OADP-592]: must-gather adding insecure tls support 

savitharaghunathan:


must-gather/Dockerfile
```diff
@@ -1,8 +1,8 @@
-FROM registry.access.redhat.com/ubi8/go-toolset:1.14.7 as gobuilder
+FROM registry.access.redhat.com/ubi8/go-toolset:1.17.10 as gobuilder
 
-RUN go get github.com/google/pprof
+RUN export GO111MODULE=on && go install -v github.com/google/pprof@latest
```
kaovilai, line 3: Why do we need this export?

must-gather/collection-scripts/gather_with_tls
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+echo "insecure-skip-tls-verify: $1"
```
kaovilai, line 2: `gather_without_tls` name was used earlier. Is this the correct file name?

kaovilai (changes requested):

must-gather/Dockerfile
kaovilai: https://github.com/openshift/oadp-operator/blob/e4f87247e56bc78530b9c78554b5175806d00a95/must-gather/README.md#support-for-insecure-tls-connections

must-gather/collection-scripts/gather_with_tls
kaovilai: https://github.com/openshift/oadp-operator/blob/e4f87247e56bc78530b9c78554b5175806d00a95/must-gather/README.md#support-for-insecure-tls-connections

must-gather/Dockerfile
savitharaghunathan: Otherwise cannot install `github.com/google/pprof`.I tried a bunch of things and this was the first one that worked.

must-gather/collection-scripts/gather_with_tls
savitharaghunathan: This is not the file name. It is just a variable that we use to receive the cli args

savitharaghunathan: when the script gets invoked, this value gets printed on the console. It is more of output and has no specific use

must-gather/Dockerfile
kaovilai: "since the default behavior in Go 1.16 is now GO111MODULE=on, this trick isn’t necessary anymore"
Maybe the first thing that work is the same to not specifying anything

savitharaghunathan: Updated. Thanks, Tiger

kaovilai:
It would be cool if we could have travis build images for must-gather PRs that we can try out..

kaovilai (approved):

shubham-pampattiwar (approved):
VISIACK

PR #763, opened 2022-07-19
OADP-552: Add privileged pod security labels to operator namespace

shubham-pampattiwar:
This PR adds the privileged pod security labels to the OADP operator ns, these are need for OADP Operator to work properly on OCP 4.11

Ref PRs: 
https://github.com/konveyor/mig-operator/pull/838
https://github.com/openshift/cluster-kube-apiserver-operator/pull/1234

### DIFF
--- a/main.go
+++ b/main.go
@@ -184,7 +184,12 @@ func getWatchNamespace() (string, error) {
 
 // setting privileged pod security labels to OADP operator namespace
 func addPodSecurityPriviledgedLabels(watchNamespaceName string) error {
-	setupLog.Info("patching namespace with PSA labels")
+	setupLog.Info("patching operator namespace with PSA labels")
+
+	if len(watchNamespaceName) == 0 {
+		return fmt.Errorf("cannot add privileged pod security labels, watchNamespaceName is empty")
+	}
+
 	kubeconf := ctrl.GetConfigOrDie()
 	clientset, err := kubernetes.NewForConfig(kubeconf)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


JaydipGabani (approved):

main.go
```diff
@@ -78,6 +81,13 @@ func main() {
 			"the manager will watch and manage resources in all namespaces")
 	}
 
+	// setting privileged pod security labels to operator ns
+	err = addPodSecurityPriviledgedLabels(watchNamespace)
```
kaovilai, line 85: Handle case where watchNamespace is `""` as returned from
https://github.com/openshift/oadp-operator/blob/85a0a7f04652ad79668ff70772a39d2a2db91a7d/main.go#L180

kaovilai (changes requested):

main.go
kaovilai: Maybe exit the function early with an error instead of trying to process `""`

shubham-pampattiwar: @kaovilai PTAL, update the PR https://github.com/openshift/oadp-operator/pull/763/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R189

kaovilai (approved):

main.go
```diff
@@ -171,3 +181,40 @@ func getWatchNamespace() (string, error) {
 	}
 	return ns, nil
 }
+
+// setting privileged pod security labels to OADP operator namespace
+func addPodSecurityPriviledgedLabels(watchNamespaceName string) error {
```
dymurray, line 186: typo

dymurray (approved):

PR #773, opened 2022-07-27
OADP-648 Don't set default pod resource limits

kaovilai:
[OADP-648](https://issues.redhat.com//browse/OADP-648)

We want to leave empty the resource limits to prevent a user from failing backups/restores due to these limits and they can always adjust as necessary.

### DIFF
--- a/controllers/restic_test.go
+++ b/controllers/restic_test.go
@@ -3,10 +3,11 @@ package controllers
 import (
 	"context"
 	"fmt"
-	"k8s.io/apimachinery/pkg/api/resource"
 	"os"
 	"reflect"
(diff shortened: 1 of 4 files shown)


dymurray (approved):
VISIACK

shubham-pampattiwar (commented):
@kaovilai Do we want to document this at least in OADP upstream that we are only setting requests and not limits, and that this differs from the upstream Velero configuration.

kaovilai:
Sure thing. The resource limits doc?

shubham-pampattiwar (approved):
Thanks for the update @kaovilai VISIACK

PR #777, opened 2022-07-29
OADP-652: Adding noobaa s3 repo

savitharaghunathan:


kaovilai (approved):

eemcmullan (approved):
I tested these changes using noobaa, and data mover backup completed. Lgtm

controllers/datamover.go
```diff
@@ -250,13 +250,14 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
...
-			if bsl.Spec.Config["s3Url"] != "" {
+			repobase := "s3.amazonaws.com"
 
+			if bsl.Spec.Config["s3Url"] != "" {
 				repobase = bsl.Spec.Config["s3Url"]
+				repobase = strings.TrimPrefix(repobase, "s3:")
 			}
 			repobase = strings.TrimSuffix(repobase, "/")
-			repo := repobase + "/" + bsl.Spec.ObjectStorage.Bucket
+			repo := "s3:" + repobase + "/" + bsl.Spec.ObjectStorage.Bucket
```
shubham-pampattiwar, line 260: @savitharaghunathan Do we have to add the `s3` prefix ? Why are we not directly using the `s3Url` specified under BSL config ? I believe this might work for noobaa case but might not work for other s3 compatible storage services like minio.

savitharaghunathan: `s3Url` from bsl config caused VolSync to error. Hence adding `s3:` prefix.

shubham-pampattiwar: @eemcmullan Thanks for the info https://restic.readthedocs.io/en/latest/030_preparing_a_new_repo.html#minio-server
Seems this is the expected format ! Approving the PR !

shubham-pampattiwar (approved):

controllers/datamover.go
savitharaghunathan: Thank you @shubham-pampattiwar @eemcmullan

PR #780, opened 2022-08-02
OADP-660: Add datamover support for azure blob storage

shubham-pampattiwar:
Related volume-snapshot controller PR: https://github.com/konveyor/volume-snapshot-mover/pull/128

### DIFF
--- a/controllers/datamover.go
+++ b/controllers/datamover.go
@@ -2,6 +2,7 @@ package controllers
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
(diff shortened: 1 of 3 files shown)


controllers/datamover.go
```diff
@@ -290,14 +294,82 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
...
+			if op == controllerutil.OperationResultCreated || op == controllerutil.OperationResultUpdated {
+				r.EventRecorder.Event(rsecret,
+					corev1.EventTypeNormal,
+					"ResticSecretReconciled",
+					fmt.Sprintf("%s restic secret %s", op, rsecret.Name),
+				)
+			}
+
+		}
 	}
```
kaovilai, line 366: - Add a case for empty provider
- Add a default case for unsupported providers

kaovilai (changes requested):

controllers/datamover.go
shubham-pampattiwar: @kaovilai we do not support empty provider or any other provider other than aws/azure/gcp, its already validated in bsl.go https://github.com/openshift/oadp-operator/blob/master/controllers/bsl.go#L61

savitharaghunathan (commented):
visual ack
/lgtm

kaovilai (approved):

savitharaghunathan (approved):
Tested back/restore workflow and it worked.

PR #784, opened 2022-08-09
OADP-708, OADP-1526 validate secret names in BSL and VSL, remove required default VSL secret if no VSL.

kaovilai:
[OADP-708](https://issues.redhat.com//browse/OADP-708)

additionally defined `veleroIOPrefix` const which fixes some typos in TrimPrefixes.

TODO: remove required default secrets validation for VSL, they are now optional.
If secret is defined, check secret exists in cluster during validation

### DIFF
--- a/controllers/validator.go
+++ b/controllers/validator.go
@@ -7,6 +7,7 @@ import (
 
 	"time"
 
+	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/go-logr/logr"
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
 	"github.com/openshift/oadp-operator/pkg/credentials"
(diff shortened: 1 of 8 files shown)


controllers/vsl.go
```diff
@@ -54,7 +54,7 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
 	}
 
 	for _, vsl := range dpa.Spec.SnapshotLocations {
-		provider := strings.TrimPrefix(vsl.Velero.Provider, "velero.io")
```
kaovilai, line 57: a few of the trimprefixes were missing `/`

controllers/velero.go
```diff
@@ -824,7 +825,7 @@ func (r DPAReconciler) noDefaultCredentials(dpa oadpv1alpha1.DataProtectionAppli
 		if vsl.Velero != nil {
 			// To handle the case where we want to manually hand the credentials for a cloud storage created
 			// Bucket credentials via configuration. Only AWS is supported
-			provider := strings.TrimPrefix(vsl.Velero.Provider, "velero.io")
```
kaovilai, line 827: a few of the trimprefixes were missing `/`

kaovilai (commented):
a few of the trimprefixes were missing `/`

kaovilai:
/retest
There have been [api timeout failure flake](https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/784/pull-ci-openshift-oadp-operator-master-4.10-operator-e2e-azure/1559245067464478720#1:build-log.txt%3A889).

sseago (approved):

kaovilai:
switched to using map of empty struct (zero size) instead of appending to array for list of secrets to check.
Added missing required secret for GCP

controllers/validator_test.go
```diff
@@ -514,9 +515,16 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
...
-			wantErr: true,
-			want:    false,
+			objects: []client.Object{
+				&corev1.Secret{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      "cloud-credentials-gcp",
+						Namespace: "test-ns",
+					},
+				},
+			},
```
kaovilai, line 525: Looks like we want to allow gcp without secrets for when they are using
`Option 2: Using Workload Identity` which is applicable when they are on GKE cluster.

kaovilai (commented):
need to make some changes.

controllers/validator.go
```diff
@@ -78,6 +81,9 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 	return true, nil
 }
 
+// empty struct to use as map value
+type empty struct{}
```
sseago, line 85: Instead of using a map to emulate a set, it might be better to just use a Set implementation like https://github.com/deckarep/golang-set -- we used that in MTC

kaovilai: +1

kaovilai: but empty struct has size of 0 bytes.. The set linked will be bigger than 0 bytes.

kaovilai: resolved via https://github.com/openshift/oadp-operator/pull/784/commits/06a3c10bad8a107dca1dce76fedf138e2edc7554

kaovilai:
obsolete by https://github.com/openshift/oadp-operator/pull/989

PR #791, opened 2022-08-16
fixing post-restore script to work for long restore names.

sseago:


sseago:
Note that the truncation/sha256 label generation is already implemented in the plugin when generating the labels. This just adds the same logic to the cleanup script.

kaovilai (approved):
This has worked when doing local e2e testing and updates.

PR #799, opened 2022-08-18
Update DC troubleshooting to reflect 1.1 changes

sseago:


docs/TROUBLESHOOTING.md
```diff
@@ -203,6 +203,12 @@ If the issue still persists, [create a new issue](https://github.com/openshift/o
     velero restore create --from-backup=<BACKUP_NAME> -n openshift-adp --include-namespaces <TARGET_NAMESPACE> --include-resources replicationcontroller,deploymentconfig,templateinstances.template.openshift.io --restore-volumes=true
     ```
 
+-  (OADP 1.1+) **Error:** `DeploymentConfigs restore with spec.Replicas==0 or DC pods fail to restart if they crash if using Restic or restore hooks`
+
+    **Solution:**
+    
+    This is expected behavior on restore if you are restoring DeploymentConfigs and are either using Restic for volume restore or you have post-restore hooks. The pod and DC plugins make these modifications to ensure that Restic and hooks work properly, and [dc-restic-post-restore.sh](../docs/scripts/dc-restic-post-restore.sh) should have been run immediately after a successful restore. Usage for this script is `dc-restic-post-restore.sh <restore-name>`
+
```
kaovilai, line 211: Do you think 1.1 section should be higher and 1.0 lower?

sseago: Oh, probably. I'll swap them.

sseago: updated

kaovilai (approved):

PR #796, opened 2022-08-18
Use yq to add nullables to bundle CRD

kaovilai:
remove placeholder deployment previously used to get operator-sdk to honor the service accounts roles.
Now using operator-sdk `--extra-service-accounts` flag feature instead.

Nullables that we override from `make manifests` are now added to make bundle by default.

The format changes are a side effect of yq processing.

manifests generated from kustomize has 2 spaces for indents except for lists where it has 0 indents.

yq forces 2 spaces on lists and turn multi-line description into a single line.

file now affected by yq formatting
- dpa crd

file no longer affected by yq formatting
- csv

PR #809, opened 2022-08-26
pulling in velero upstream changes as of 26 Aug 2022

sseago:
This includes the first round of kopia-related changes, which
may temporarily break things.

The most visible change is the renaming of the "resticrepository"
CRD to "backuprepository".

This also points oadp-operator back to latest velero/plugin images (from 1.9 branches)
and updates builder image to latest go version (currently 1.18.4).


weshayutin:
Looks good to me.. 
@rayfordj you will want to take a look at this one :)

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -215,27 +215,23 @@ spec:
...
-                                nullable: true
                                 type: object
                               requests:
                                 additionalProperties:
                                   anyOf:
                                   - type: integer
                                   - type: string
-                                  nullable: true
                                   pattern: ^(\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))(([KMGTPE]i)|[numkMGTPE]|([eE](\+|-)?(([0-9]+(\.[0-9]*)?)|(\.[0-9]+))))?$
                                   x-kubernetes-int-or-string: true
```
kaovilai, line 229: Becareful removing nullables here. It's removed when `make generate` runs I think.

https://github.com/openshift/oadp-operator/pull/796 might mitigate this issue.

kaovilai (changes requested):

config/crd/bases/oadp.openshift.io_dataprotectionapplications.yaml
sseago: @kaovilai Thanks for catching that. I hadn't intended on including this change in the PR. I'll remove it (it's one of the files I have to generally ignore from `make bundle` since it makes changes even when I didn't touch it.

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM quay.io/konveyor/builder:v1.17.2 as builder
+FROM quay.io/konveyor/builder as builder
```
kaovilai, line 2: Should we be using a tagged version such that it is clear what go version is used?

kaovilai (approved):

kaovilai:
From scrum the intention is that openshift:master does not have to be for 1.1.z. We have oadp-1.1 for that.
So openshift:master may point to a higher velero version.

shubham-pampattiwar (approved):

PR #811, opened 2022-08-29
Enable e2e tests to execute against downstream build

weshayutin:
Reusing the e2e tests across upstream and downstream builds
will benefit the engineering teams by increasing the coverage
across teams.  Additionally this should enable QE to execute
the dev teams e2e tests against a build.

* New Variable = "stream" ( ok name? )
  * allow cli -stream or env var OADP_STREAM to set value [up,down]
* Only change required was the subscription label

Tested and passed on internal build

### DIFF
--- a/tests/e2e/subscription_suite_test.go
+++ b/tests/e2e/subscription_suite_test.go
@@ -109,7 +109,6 @@ var _ = Describe("Subscription Config Suite Test", func() {
 					},
 				},
 			},
-			stream: stream,
 		}),
 		Entry("NO_PROXY set", SubscriptionConfigTestCase{
 			SubscriptionConfig: operators.SubscriptionConfig{
@@ -120,7 +119,6 @@ var _ = Describe("Subscription Config Suite Test", func() {
 					},
 				},
 			},
-			stream: stream,
 		}),
 		Entry("HTTPS_PROXY set", SubscriptionConfigTestCase{
 			SubscriptionConfig: operators.SubscriptionConfig{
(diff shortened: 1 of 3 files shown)


tests/e2e/subscription_suite_test.go
```diff
@@ -36,11 +36,12 @@ var _ = Describe("Subscription Config Suite Test", func() {
 	type SubscriptionConfigTestCase struct {
 		operators.SubscriptionConfig
 		failureExpected *bool
+		stream string
```
kaovilai, line 39: So the intention is to have in a single suite run test both streams?

If only one stream then it doesn't need to be added to the test case config.

weshayutin: OH.. so each Entry in the table can access the vars from the cli?   I wasn't able to get that to work but happy to change it once I figure the HOW.. :)  Thanks for the review

kaovilai: It is currently exposed but a refactoring is due. Wanted to move it out to another package like here https://github.com/openshift/oadp-operator/pull/710/files#diff-7520c0381b468591b4362f559e3043268715f1862cc6f9ac8fed25367be166e9

weshayutin:
@kaovilai ok.. now that I know I didn't break anything I'll go after the adjustments you asked for.
Thanks for the review!

kaovilai:
we love a good
/retest
/lgtm

kaovilai (approved):

PR #818, opened 2022-09-09
add a simple golang version check to warn developers

weshayutin:
For new developers it may not be clear why some of the make commands are not working properly.  For example [1] changed the minimum required golang to successfully use `make deploy-olm`. It was also not 100% clear what golang version would work across platforms.

It may be helpful to specify the required version and expose it via the makefile.

[1]https://github.com/openshift/oadp-operator/pull/809

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -66,6 +66,7 @@ endif
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
 ENVTEST_K8S_VERSION = 1.21
 
+
 .PHONY:ginkgo
 ginkgo: # Make sure ginkgo is in $GOPATH/bin
 	go get -d github.com/onsi/ginkgo/ginkgo
@@ -148,6 +149,9 @@ help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 ##@ Development
+# golang version check
+golang_check: golang_check ## Ensure the golang version meets the requirements
+	./docs/developer/golang_version_check
 
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	# Commenting out default which overwrites scoped config/rbac/role.yaml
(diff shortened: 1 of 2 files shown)


docs/developer/golang_version_check
```diff
@@ -0,0 +1,31 @@
+#!/usr/bin/sh
```
kaovilai, line 1: If it's gonna be a script.. it better be .sh

Makefile
```diff
@@ -148,6 +149,9 @@ help: ## Display this help.
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_0-9-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
 ##@ Development
+# golang version check
+golang_check: golang_check ## Ensure the golang version meets the requirements
```
kaovilai, line 153: Recursion?
```suggestion
golang_check: ## Ensure the golang version meets the requirements
```

docs/developer/golang_version_check
```diff
@@ -0,0 +1,31 @@
...
+MINIMUM_GO_MAJOR_VERSION=1
+MINIMUM_GO_MINOR_VERSION=18
+MINIMUM_GO_RELEASE_VERSION=6
```
kaovilai, line 4: 1.18.3 worked for me. Your issue was probably environment specific. Or specific to 1.18.5

kaovilai (changes requested):

docs/developer/golang_version_check
weshayutin: ya.. fine w/ that.  Teams go either way.. if it's executable and #! <something> it's not needed.  
happy to add

Makefile
weshayutin: sorry.. but you are incorrect sir..   
the first one is for the make command, the second is for the help parser

docs/developer/golang_version_check
weshayutin: This is why a version check script may help.
We may want to say:
1.18.3 works fine
1.18.4 works fine
1.18.5 DOES NOT WORK
1.18.6 works
1.19.x DOES NOT WORK 
¯\_(ツ)_/¯

kaovilai: Can you outline which version you used rpm?

Are they the ones that don't work?

kaovilai:
Current build and ci docker images used for tests is using 1.17 as requested by go.mod so not sure if the version here is valid. But if this check isn't enforced on every build/test then I am ok merging.

PR #819, opened 2022-09-09
Datamover blog: Update diagrams and add links

shubham-pampattiwar:


kaovilai:
Can you add another circle for volume snapshot content? If it's reused
across the two namespaces.

shubham-pampattiwar:
@kaovilai we want to keep the flow diagram a bit more high level, hence skipping vsc block representation

kaovilai (approved):

PR #830, opened 2022-09-22
update 1.0 e2e to master with 1.0 accommodations

kaovilai:
1.0 accommodations
- volumesnapshot used is v1 in master. revert to v1beta1 in 1.0.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
 OADP_TEST_NAMESPACE ?= openshift-adp
-CLUSTER_TYPE ?= aws
 
 # CONFIGS FOR CLOUD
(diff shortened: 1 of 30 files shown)


kaovilai:
This is failing because in 1.0 we do not have https://github.com/openshift/oadp-operator/blob/master/docs/scripts/dc-restic-post-restore.sh

kaovilai:
https://github.com/openshift/openshift-velero-plugin/pull/141 does not exist in oadp-1.0 so update this to go back to two stage restic restore.

kaovilai:
Closing for now. Might revisit later.

PR #832, opened 2022-09-23
updating plugin/velero images to use 1.0 branches.

sseago:


kaovilai:
Replacing env with RELATED_IMAGE requires [velero.go changes](https://github.com/openshift/oadp-operator/commit/4ea48910d9f17a56bfa0b8d1283aa2604cff39b9#diff-992e40b76fc7cf12af63ec6c2cdcc70321669228469781c0d2bc527dc4c28627)
as well as [credentials.go changes](https://github.com/openshift/oadp-operator/commit/4ea48910d9f17a56bfa0b8d1283aa2604cff39b9#diff-1c259cca045064227710d0f341606e8ba73adda92dc126782acf62e2667af5ba) to update the ENV used.

kaovilai:
Missing registry env https://github.com/sseago/oadp-operator/pull/5

kaovilai (approved):

controllers/velero_test.go
```diff
@@ -3651,12 +3651,12 @@ func TestDPAReconciler_getVeleroImage(t *testing.T) {
...
 			},
 			pluginName: common.Velero,
-			wantImage:  "quay.io/konveyor/velero:latest",
+			wantImage:  "quay.io/konveyor/velero:oadp-1.0",
 			setEnvVars: map[string]string{
 				"REGISTRY":    "quay.io",
 				"PROJECT":     "konveyor",
 				"VELERO_REPO": common.Velero,
-				"VELERO_TAG":  "latest",
+				"VELERO_TAG":  "oadp-1.0",
```
rayfordj, line 3659: Shouldn't these blocks (here & in [pkg/credentials/credentials_test.go](https://github.com/openshift/oadp-operator/pull/832/files#diff-0a12f5c252d8d2a8e406ccd3c83025bd31c3b523c18c11e65c73a84d6b4ae26d)) be changed to use `RELATED_IMAGE_` in order to continue to work as expected?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -639,22 +617,22 @@ spec:
...
+      name: velero-plugin-for-microsoft-azure
+    - image: quay.io/konveyor/velero-plugin-for-gcp:oadp-1.0
+      name: velero-plugin-for-gcp
+    - image: quay.io/konveyor/velero-plugin-for-csi:oadp-1.0
+      name: velero-plugin-for-csi
     - image: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
-      name: kubevirt_plugin
-  version: 1.0.2
+      name: kubevirt-velero-plugin
+  version: 1.0.4
```
rayfordj, line 638:
```suggestion
  version: 1.0.5
```

rayfordj (changes requested):
Off to a strong start

* We're already building 1.0.5 images for the next OADP 1.0.z release
* We're no longer adding 1.0.z to `stable`; 1.0.3 was the last.
* The *_test.go files are still using `REGISTRY, PROJECT, _REPO, _TAG` instead of `RELATED_IMAGE_`

I realize some of my comments throughout the files are generated using the **Makefile**, but I wanted to highlight them as I came across them for posterity.

Thanks!

controllers/velero_test.go
kaovilai: I guess so.. the function checks if expected env is defined, else default to values in the code.. which I guess is the same as this test hence it didn't see any differences.

kaovilai:
Can't push to remote but can edit one file at a time :D.
Curious if you have this checked 
image

kaovilai:
need to update https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-oadp-1.0__4.7.yaml#L79 to stable-1.0 to unblock e2e.

hhpatel14:
> need to update https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-oadp-1.0__4.7.yaml#L79 to stable-1.0 to unblock e2e.

updating!

hhpatel14:
https://github.com/openshift/release/pull/32611

controllers/velero_test.go
```diff
@@ -3651,12 +3651,9 @@ func TestDPAReconciler_getVeleroImage(t *testing.T) {
...
 			},
 			pluginName: common.Velero,
-			wantImage:  "quay.io/konveyor/velero:latest",
+			wantImage:  "notquay.io/someorg/velero:sometag",
 			setEnvVars: map[string]string{
-				"REGISTRY":    "quay.io",
-				"PROJECT":     "konveyor",
-				"VELERO_REPO": common.Velero,
-				"VELERO_TAG":  "latest",
+				"RELATED_IMAGE_VELERO": "notquay.io/someorg/velero:sometag",
```
rayfordj, line 3656: @kaovilai , deliberately set invalid image?

kaovilai: This is a unit test. It only checks if image returned by the function is one that is not default.

If we check the same image as hardcoded image, we wouldn't know if ENV actually work or if it came from a hardcoded backup.

rayfordj: Good enough for me :sunglasses:

rayfordj (approved):
/lgtm

controllers/velero_test.go
kaovilai: UnsupportedOverrides comes first then ENV then hardcoded backup image.

You could make a release that only require a tag change and only change this env in CSV instead of having to recompile everything.

kaovilai: this "invalid" image actually won't get pulled :)

rayfordj: Yes, end result is the same.  I realized it was in test (and probably intentionally left as that value), but I wanted to double-check.  Thanks!

kaovilai:
This PR unblocks oadp-1.0 e2e :) .. let the retests begin :D

PR #848, opened 2022-10-06
OADP-823: Address oadp-1.0.z with ocp 4.11 csi compatibility issue

shubham-pampattiwar:


controllers/validator.go
```diff
@@ -100,6 +103,30 @@ func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {
...
+		// check csi compatibility with cluster version
+		// velero version <1.9 expects API group snapshot.storage.k8s.io/v1beta1, while OCP 4.11 (k8s 1.24) has only snapshot.storage.k8s.io/v1
+		for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
+			if plugin == oadpv1alpha1.DefaultPluginCSI {
+				clusterVersion, err := getClusterVersion()
+				if err != nil {
+					return false, err
+				}
+
+				if clusterVersion.Major == "1" && clusterVersion.Minor >= "24" {
```
weshayutin, line 116: looks good! thank you

weshayutin:
looks good to me..  I suspect we should add a unit test for this issue.  Should be easy enough and can be in a follow up. 
Thanks

controllers/validator.go
kaovilai: Why are we doing nested for loops for no reason? The outerloop is already for each plugin. I would put this logic in line 90.
https://github.com/openshift/oadp-operator/blob/65159dd313799cc3ca52a0a4b31623098ee62900/controllers/validator.go#L89-L120

sseago: This works until k8s 2.0, but I can only assume that oadp 1.0 won't be supported at all on whatever version of OpenShift that would be.

kaovilai (changes requested):
Please don't do 
```
for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
  for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
  }
}
```

sseago (approved):
lgtm

controllers/validator.go
shubham-pampattiwar: @kaovilai @sseago earlier I had a function for this and then relocated the logic, copy/paste oversight, will rectify, Thank you !

controllers/validator.go
```diff
@@ -100,6 +103,30 @@ func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {
...
+					return false, errors.New("when using csi, velero version <1.9 expects API group snapshot.storage.k8s.io/v1beta1, while OCP 4.11+ (k8s 1.24+) has only snapshot.storage.k8s.io/v1, please disable CSI support, or if CSI is needed, upgrade to OADP 1.1+")
```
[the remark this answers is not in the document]
kaovilai:
```suggestion
					return false, errors.New("csi plugin on OADP 1.0 (velero version <1.9) requires API snapshot.storage.k8s.io/v1beta1. On OCP 4.11+ (k8s 1.24+), If CSI is needed, upgrade to OADP 1.1+")
```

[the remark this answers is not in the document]
shubham-pampattiwar: @eemcmullan the suggested error message makes it seem like csi is not supported with prior velero 1.9 and priror OCP versions, which is not the case right ? We need to tell the actual reason imo. but yeah agreed on making concise, any more suggestions with the reason embedded in error message ?

controllers/validator.go
shubham-pampattiwar: Done, updated the PR !

kaovilai (approved):
Agree with Emily on condensing error string a bit. otherwise
/lgtm

sseago (approved):

sseago (approved):

kaovilai (approved):

PR #846, opened 2022-10-05
master: Makefile Automate operator-sdk install, preserve nullables.

kaovilai:
Might fix `make deploy-olm` image override workarounds by forcing use of a GitHub release binary which hopefully uses tagged version of opm instead of latest by default.

PR #859, opened 2022-10-21
oadp-1.1: yq format when for `make test`. pin gomega to v1.22

kaovilai:
Cherrypick #846 to oadp-1.1


sseago (approved):

build/custom-ci-Dockerfile
```diff
@@ -1,8 +1,9 @@
-FROM openshift/origin-release:golang-1.16
+FROM openshift/origin-release:golang-1.17
```
kaovilai, line 1: note that this image tag `golang-1.17` doesn't actually exists at least in docker hub.

Will check and maybe remove this file if it's not used anywhere.

kaovilai: Will delete in a follow up later.

kaovilai: looks to be a leftover from ansible days

weshayutin:
Do we want this change on OADP-1.1.1 ?

kaovilai:
Let's punt for 1.1.2

dymurray (approved):

Issue #576, opened 2022-02-23
Bug: The `velero-privileged` SCC is causing the CIS benchmark to fail


alexisph:
### Describe bug

As per the title, the `velero-privileged` SCC is causing the CIS benchmark to fail.

The Red Hat OADP operator is installed in our disconnected v4.8 cluster. The Compliance operator is checking compliance to the CIS Red Hat OpenShift Container Platform 4 Benchmark V1.1 and this rule is failing:

> Limit Container Capabilities
> Rule ID: xccdf_org.ssgproject.content_rule_scc_limit_container_allowed_capabilities
> Result: fail
> Severity: medium
> Description: Containers should not enable more capabilites than needed as this opens the door for malicious use. To enable only the required capabilities, the appropriate Security Context Constraints (SCCs) should set capabilities as a list in allowedCapabilities.
> Rationale: By default, containers run with a default set of capabilities as assigned by the Container Runtime which can include dangerous or highly privileged capabilities. Capabilities should be dropped unless absolutely critical for the container to run software as added capabilities that are not required allow for malicious containers or attackers.

Verification:

```sh
oc get scc -o json | jq '[.items[] | select(.metadata.name != "privileged") | .metadata.name, .allowedCapabilities]'

[
  "anyuid",
  null,
  "hostaccess",
  null,
  "hostmount-anyuid",
  null,
  "hostnetwork",
  null,
  "log-collector-scc",
  null,
  "machine-api-termination-handler",
  null,
  "node-exporter",
  null,
  "nonroot",
  null,
  "restricted",
  null,
  "velero-privileged",
  [
    "*"
  ]
]
```

The `velero-privileged` SCC was created by the OADP operator. Please consider limiting its allowed capabilities.

Thanks in advance.

### What happened?

The CIS benchmark failed because the created SCC is too open.
Expected that the created SCC would be limited to the absolutely necessary set of capabilities.

### OADP Version

0.5.x (Stable)

### OpenShift Version

4.8

### Operator pod logs

```shell
2022-02-23T10:33:02.565Z        DEBUG   events  Normal  {"object": {"kind":"SecurityContextConstraints","name":"velero-privileged","uid":"e8b64dd3-1870-449f-b395-df69d4b4bac7","apiVersion":"security.openshift.io/v1","resourceVersion":"290547648"}, "reason": "VeleroSecurityContextConstraintsReconciled", "message": "performed created on velero scc velero-privileged"}
[...]

PR #860, opened 2022-10-26
Remove creation/usage of velero-privileged SCC (OADP-290)

hhpatel14:
Fixes https://github.com/openshift/oadp-operator/issues/576


kaovilai (approved):
lgtm!

PR #872, opened 2022-11-29
velero fork update, 11/29/2022

sseago:


kaovilai (approved):

kaovilai:
Can you run `make bundle` ?

kaovilai:
Could do as a follow up

sseago:
Oh, wait. I thought I did. Is something missing? let me check.

sseago:
@kaovilai done. I usually do that with these. I guess I forgot. Should we have a prow check that runs `make bundle` and errors out if there are changes?

sseago:
But once CI passes, we need to merge this since the 1.2 builds are otherwise out-of-sync (although the CRD changes are trivial here, in comments only, so it's unlikely to actually break things).

PR #886, opened 2023-01-25
oadp-1.1: OADP-1150 Add Env to dpa.Spec.PodConfig

kaovilai:
Makes Velero and Restic pods Environment variables configurable.

[OADP-1150](https://issues.redhat.com//browse/OADP-1150)

PR #889, opened 2023-01-25
master: OADP-1150 Add Env to PodConfig

kaovilai:
Makes Velero and Restic pods Environment variables configurable.

OADP-1150

cp from #886


kaovilai:
some panics due to nil response in e2e.
https://github.com/openshift/oadp-operator/issues/893

/retest

eemcmullan (approved):

PR #895, opened 2023-01-30
OADP-1067: Set env vars on datamover deployment for batching

eemcmullan:


api/v1alpha1/oadp_types.go
```diff
@@ -176,10 +176,10 @@ type DataMover struct {
 	Timeout string `json:"timeout,omitempty"`
 	// the number of batched volumeSnapshotBackups that can be inProgress at once, default value is 10
 	// +optional
-	MaxConcurrentBackupVolumes *int64 `json:"maxConcurrentBackupVolumes,omitempty"`
+	MaxConcurrentBackupVolumes string `json:"maxConcurrentBackupVolumes,omitempty"`
```
kaovilai, line 179: Why string?

eemcmullan: Was going for consistency: https://github.com/openshift/oadp-operator/blob/master/api/v1alpha1/oadp_types.go#L176

kaovilai: Timeout was string because it also have text descriptors.. like 10m vs 10h. Works for me

kaovilai:
although not required, if there are any test failures, consider rebasing because this PR is 2 commits behind master branch

api/v1alpha1/oadp_types.go
kaovilai: A more convenient reason imo would be to that there is no conversion required in corev1.EnvVar
```
		dataMoverContainer.Env = append(dataMoverContainer.Env, corev1.EnvVar{
			Name:  DataMoverConcurrentBackup,
			Value: dpa.Spec.Features.DataMover.MaxConcurrentBackupVolumes,
		})
```

I would add validation to prevent negative inputs, 0, trailing/prefix whitespace etc.. either via kubebuilder CRD and/or reconciler.

kaovilai (approved):

shubham-pampattiwar (approved):

PR #897, opened 2023-02-01
OADP-1231: Add config values for data mover

eemcmullan:


api/v1alpha1/oadp_types.go
```diff
@@ -180,6 +180,19 @@ type DataMover struct {
 	// the number of batched volumeSnapshotRestores that can be inProgress at once, default value is 10
 	// +optional
 	MaxConcurrentRestoreVolumes string `json:"maxConcurrentRestoreVolumes,omitempty"`
+	// defines configurations for VolSync options
+	// +optional
+	MoverConfig *MoverConfig `json:"moverConfig,omitempty"`
```
shubham-pampattiwar, line 185: IMO `MoverConfig` may sound like configuration parameters for datamover, I believe our aim is to set options for the Volume, so maybe `VolumeOptions` or something on similar lines. wdyt ?

api/v1alpha1/oadp_types.go
```diff
@@ -180,6 +180,19 @@ type DataMover struct {
 	// the number of batched volumeSnapshotRestores that can be inProgress at once, default value is 10
 	// +optional
 	MaxConcurrentRestoreVolumes string `json:"maxConcurrentRestoreVolumes,omitempty"`
+	// defines configurations for VolSync options
```
shubham-pampattiwar, line 183: maybe refrain from specifying `volsync` here ?

weshayutin:
not quite ready for review

controllers/datamover.go
```diff
@@ -164,6 +166,61 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
 	return true, nil
 }
 
+func (r *DPAReconciler) ReconcileDataMoverConfigs(log logr.Logger) (bool, error) {
```
shubham-pampattiwar, line 169: NIT: rename to `ReconcileDatamoverVolumeOptions`

controllers/datamover.go
```diff
@@ -612,6 +669,54 @@ func (r *DPAReconciler) checkIfDataMoverIsEnabled(dpa *oadpv1alpha1.DataProtecti
 	return false
 }
 
+func (r *DPAReconciler) checkDataMoverConfigs(dpa *oadpv1alpha1.DataProtectionApplication) bool {
```
shubham-pampattiwar, line 672: NIT: rename to `checkDatamoverVolumeOptions`

controllers/datamover.go
```diff
@@ -164,6 +166,61 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
...
+	}
+
+	// check configMap already exists
+	confMap, confMapExists, err := r.checkConfigMapExists()
+	if err != nil {
+		return false, err
+	}
+
+	// check for existing configMap but no data mover configs set
+	if !r.checkDataMoverConfigs(&dpa) && confMapExists {
```
shubham-pampattiwar, line 184: I believe we should delete the config map only if datamover is disabled by the user, right ?

controllers/datamover.go
```diff
@@ -164,6 +166,61 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
...
+		// create configmap to pass values to data mover CRs
+		cm := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      DataMoverConfigMapName,
+				Namespace: dpa.Namespace,
+			},
+		}
+
+		op, err := controllerutil.CreateOrPatch(r.Context, r.Client, &cm, func() error {
+			cm.Data = r.makeConfigMapData(&dpa)
```
shubham-pampattiwar, line 204: Maybe follow the `build<ResourceName>` pattern here as well as we have done through the OADP Operator tasks where we pass on the resource pointer and build its specs

shubham-pampattiwar (commented):
Added a few comments

controllers/datamover.go
eemcmullan: This deletes the configmap if the user deletes the config options on DPA, as then we would not want the VSB to use these options anymore. wdyt?

api/v1alpha1/oadp_types.go
```diff
@@ -183,6 +183,29 @@ type DataMover struct {
...
 	// +optional
 	PruneInterval string `json:"pruneInterval,omitempty"`
+	// defines configurations for data mover volume options
+	// +optional
+	VolumeOptions *VolumeOptions `json:"volumeOptions,omitempty"`
+}
+
+// MoverConfig defines configurations for VolSync options
+type VolumeOptions struct {
+	// storageClassName can be used to override the StorageClass of the source PVC
```
kaovilai, line 193: What about destination PVC?

api/v1alpha1/oadp_types.go
```diff
@@ -190,10 +190,12 @@ type DataMover struct {
 
 // MoverConfig defines configurations for VolSync options
 type VolumeOptions struct {
-	// storageClassName can be used to override the StorageClass of the source PVC
+	// storageClassName can be used to override the StorageClass of the source
```
kaovilai, line 193: Can we have different storage class used for source and destination? They might both be using the same driver but different deletionPolicy for example.

eemcmullan: @shubham-pampattiwar wdyt of this? ^ I think then we'd need to do the same for each of the new types for backup and restore, which would bring us to a lot of config values.

shubham-pampattiwar: @eemcmullan @kaovilai yeah IMO it would make sense. Lets create a struct something like this:
```
type DatamoverVolumeOptions struct {
    SrcVolumeOptions VolumeOptions
    DestinationVolumeOptions VolumeOptions
}
```

controllers/datamover.go
```diff
@@ -165,6 +167,65 @@ func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, err
...
+	}
+
+	// create configmap only if data mover is enabled and has config values
+	if r.checkIfDataMoverIsEnabled(&dpa) && r.checkDataMoverVolumeOptions(&dpa) {
+
+		// create configmap to pass values to data mover CRs
+		cm := corev1.ConfigMap{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      DataMoverConfigMapName,
+				Namespace: dpa.Namespace,
```
shubham-pampattiwar, line 200: I believe you need to add OADP Operator labels to the configMap or else the operator predicate would filter out the reconcile events and OADP operator will not reconcile on the changes related to this configMap

api/v1alpha1/oadp_types.go
```diff
@@ -183,6 +183,36 @@ type DataMover struct {
 	// defines how often (in days) to prune the datamover snapshots from the repository
 	// +optional
 	PruneInterval string `json:"pruneInterval,omitempty"`
+	// defines configurations for data mover volume options
+	// +optional
+	DataMoverVolumeOptions *DataMoverVolumeOptions `json:"volumeOptions,omitempty"`
+}
+
+type DataMoverVolumeOptions struct {
+	SourceVolumeOptions      *VolumeOptions `json:"sourceVolumeOptions,omitempty"`
```
shubham-pampattiwar, line 192: +1

api/v1alpha1/oadp_types.go
```diff
@@ -183,6 +183,36 @@ type DataMover struct {
...
+	DestinationVolumeOptions *VolumeOptions `json:"destinationVolumeOptions,omitempty"`
+}
+
+// MoverConfig defines configurations for VolSync options
```
shubham-pampattiwar, line 196: NIT: MoverConfig <-> VolumeOptions in comment

shubham-pampattiwar (changes requested):

controllers/datamover.go
```diff
@@ -628,6 +692,98 @@ func (r *DPAReconciler) checkIfDataMoverIsEnabled(dpa *oadpv1alpha1.DataProtecti
...
+	cm.Data = cmMap
+
+	return nil
+}
+
+func (r *DPAReconciler) checkConfigMapExists() (*corev1.ConfigMap, bool, error) {
+
+	// check configMap already exists
+	confmap := corev1.ConfigMap{}
+	err := r.Get(context.Background(), types.NamespacedName{Name: DataMoverConfigMapName, Namespace: r.NamespacedName.Namespace}, &confmap)
```
kaovilai, line 775: I would make this function generic accepting any name/namespace.

As is, it can be a bit hard to know that DPAReconciler.checkConfigMapExists is specific to data mover without really digging.

kaovilai (changes requested):

controllers/datamover.go
kaovilai: Just the name as parameter would be fine, we can reasonably assume anything coming out of `DPAReconciler` expect namespace to be `r.NamespacedName.Namespace`.

kaovilai: Or rename the function to be more data mover specific

kaovilai:
This configmap is read by the plugin correct?

eemcmullan:
@kaovilai No it is read by the VSM controller

kaovilai (approved):

shubham-pampattiwar:
related controller PR: https://github.com/migtools/volume-snapshot-mover/pull/191

shubham-pampattiwar (approved):

PR #898, opened 2023-02-01
init use same resources as the main container

jordiprats:
When deploying a velero instance, the initContainer doesnt have any resources defined, instead, use the same resources that have been defined for the main one.

I'd be nice to be able to set resources for each container, meanwhile we can use whatever is defined for the main one

### DIFF
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -255,15 +255,7 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds 
 // add plugin specific specs to velero deployment
 func AppendPluginSpecificSpecs(dpa *oadpv1alpha1.DataProtectionApplication, veleroDeployment *appsv1.Deployment, veleroContainer *corev1.Container, providerNeedsDefaultCreds map[string]bool, hasCloudStorage bool) error {
 
-	init_container_resources := corev1.ResourceRequirements{}
-	if len(veleroDeployment.Spec.Template.Spec.Containers) > 0 {
-		for _, container := range veleroDeployment.Spec.Template.Spec.Containers {
-			if container.Name == "velero" {
-				init_container_resources = container.Resources
-				break
-			}
-		}
-	}
+	init_container_resources := veleroContainer.Resources
 
 	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
 		if pluginSpecificMap, ok := PluginSpecificFields[plugin]; ok {
--- a/pkg/credentials/credentials.go
+++ b/pkg/credentials/credentials.go
@@ -254,6 +254,17 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds 
 
 // add plugin specific specs to velero deployment
 func AppendPluginSpecificSpecs(dpa *oadpv1alpha1.DataProtectionApplication, veleroDeployment *appsv1.Deployment, veleroContainer *corev1.Container, providerNeedsDefaultCreds map[string]bool, hasCloudStorage bool) error {
+
+	init_container_resources := corev1.ResourceRequirements{}
+	if len(veleroDeployment.Spec.Template.Spec.Containers) > 0 {
+		for _, container := range veleroDeployment.Spec.Template.Spec.Containers {
+			if container.Name == "velero" {
+				init_container_resources = container.Resources
+				break
+			}
+		}
+	}
+
 	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
 		if pluginSpecificMap, ok := PluginSpecificFields[plugin]; ok {
 			veleroDeployment.Spec.Template.Spec.InitContainers = append(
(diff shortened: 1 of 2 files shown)


kaovilai:
@jordiprats Thanks for your contribution!

Just fyi the init containers for plugins in velero do not live very long and its only job is to "copies the binary into a shared emptyDir volume for the Velero server to access"
https://velero.io/docs/v1.9/custom-plugins/#docs

Do you still think the change here is necessary? I doubt initContainers need much resources at all for just running essentially `cp`

``` Dockerfile
ENTRYPOINT ["cp", "/plugins/velero-plugin-for-aws", "/target/."]
```
https://github.com/vmware-tanzu/velero-plugin-for-aws/blob/main/Dockerfile#L38

jordiprats:
> Do you still think the change here is necessary? I doubt initContainers need much resources at all for just running essentially `cp`

I do, if you have ResourceQuotas not having the resources defined Kubernetes won't allow the Pod to run the initContainers. There are workarounds, of course, but just as it is allowed to configure resources for the main container, it should be allowed to set them for the initContainers (or at least use the same that we are going to use for the main container)

kaovilai:
1. it shouldn't need as much as the main container. I would push for a fixed amount.. failing that, an amount that's a percentage from main container above a minimum.
2. line 339 also need the change.
https://github.com/openshift/oadp-operator/blob/d8b4ba7564687d60231a2bde585e175d67b545e7/pkg/credentials/credentials.go#L339

jordiprats:
@kaovilai To use a fixed amount instead of copying the resources that are set for the main container would be trivial to make, but I'm wondering if it's worth having hard-coded variables (and what they should be)

From the scheduler perspective, it's going to take into consideration the highest amount to be able to schedule it (since once the initContainer is done it will have to run the main one) so assigning a different value won't make a difference. Furthermore, since we can consider it it is going to be a short-lived container setting a fixed amount might cause issues in the future.

For these reasons, I would avoid setting any hard-coded values, but it's something I could do easily if you want me to.

kaovilai (approved):
Lgtm.

@shubham-pampattiwar please review.

eemcmullan (approved):

pkg/credentials/credentials.go
```diff
@@ -254,6 +254,12 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds
 
 // add plugin specific specs to velero deployment
 func AppendPluginSpecificSpecs(dpa *oadpv1alpha1.DataProtectionApplication, veleroDeployment *appsv1.Deployment, veleroContainer *corev1.Container, providerNeedsDefaultCreds map[string]bool, hasCloudStorage bool) error {
+
+	init_container_resources := corev1.ResourceRequirements{}
+	if len(veleroDeployment.Spec.Template.Spec.Containers) > 0 {
```
shubham-pampattiwar, line 259: You are trying to grab the container resources assigned to velero container, right ? This code needs a container name check, currently this snippet just grabs the first container that it encounters in velero deployment. The check will make it future proof and free of inaccuracies.

shubham-pampattiwar (dismissed):
Thanks for the PR @jordiprats . Added my review.

jordiprats:
@shubham-pampattiwar I've updated the PR with the requested changes. Could you please take a look?

pkg/credentials/credentials.go
```diff
@@ -254,6 +254,17 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds
...
+		for _, container := range veleroDeployment.Spec.Template.Spec.Containers {
+			if container.Name == "velero" {
+				init_container_resources = veleroDeployment.Spec.Template.Spec.Containers[0].Resources
+				break
+			}
+		}
+	}
```
kaovilai, line 266: length check are not necessary for _for_ loops.
```suggestion
	for _, container := range veleroDeployment.Spec.Template.Spec.Containers {
		if container.Name == "velero" {
			init_container_resources = container.Resources
			break
		}
	}
```

kaovilai (changes requested):

pkg/credentials/credentials.go
jordiprats: OMG, I forgot to change this as well. @kaovilai PR updated!

kaovilai (approved):

pkg/credentials/credentials.go
```diff
@@ -254,6 +254,17 @@ func AppendCloudProviderVolumes(dpa *oadpv1alpha1.DataProtectionApplication, ds
...
+
+	init_container_resources := corev1.ResourceRequirements{}
+	if len(veleroDeployment.Spec.Template.Spec.Containers) > 0 {
+		for _, container := range veleroDeployment.Spec.Template.Spec.Containers {
+			if container.Name == "velero" {
+				init_container_resources = container.Resources
+				break
+			}
+		}
+	}
```
kaovilai, line 266: Here's another heart breaking thought..
```suggestion
	init_container_resources := veleroContainer.Resources
```

kaovilai (changes requested):

pkg/credentials/credentials.go
kaovilai: sorry noticed just now.

kaovilai: We kind of do this logic already here https://github.com/openshift/oadp-operator/blob/090604ba0304cfe37ffbec53c9dc9f8f14bf4424/controllers/velero.go#L442-L448

kaovilai:
unit test failing @eemcmullan 
```
--- FAIL: TestVolumeSnapshotRestoreReconciler_buildReplicationDestination (0.00s)
    --- FAIL: TestVolumeSnapshotRestoreReconciler_buildReplicationDestination/Given_vsr_and_repdest_and_secret,_should_pass (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0x130c00b]
```

kaovilai:
oops wrong PR comment..

kaovilai (approved):
/lgtm

PR #899, opened 2023-02-07
OADP-1269: Integrate velero-plugin-for-vsm with OADP Operator

shubham-pampattiwar:


eemcmullan (approved):

dymurray (approved):
LGTM

controllers/validator.go
```diff
@@ -57,6 +57,18 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
+		// Also check if the VSM plugin is specified or not
+		VSMPluginPresent := false
+		for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
+			if plugin == oadpv1alpha1.DefaultPluginVSM {
+				VSMPluginPresent = true
+			}
+		}
+
+		if !VSMPluginPresent {
+			return false, errors.New("datamover feature is enabled, need to specify vsm as a default plugin in DataProtectionApplication CR")
```
kaovilai, line 70: Can we not do this automatically behind the scenes? or even update DPA to reflect this.

dymurray: yeah good point, we do this for the CSI plugin as well. Poor UX to make them do this manually.

shubham-pampattiwar: @kaovilai @dymurray AFAIK We don't update/patch DPA for CSI plugin, rather what we do is if the `csi` plugin is specified as a `defaultPlugin` under DPA,  we append the `EnableCSI` features flag to Velero container command. IMO The DPA spec is something the user specifies and OADP Operator acts on completing the specified tasks. Rather I think We should update this validation to also check for csi plugin along with vsm plugin if datamover feature is enabled because volume-snapshot-mover/datamover would be useless if do not have CSI volumesnapshots. And yes we can document all this in our docs. WDYGT ?

kaovilai: Should we consider update/patch DPA for CSI plugin as well then?

If we insist on what you put in is what you get, then we could just add those plugins behind without updating DPA as if you enable it you want it to work the first time. We can reduce the steps you need to do to get minimal working DPA. This would be akin to appending EnableCSI to container command but for plugin.

Automation/Autocompletion can reduce support volumes.

Otherwise this new plugin change is gonna break existing data mover DPAs.. which some tech preview users may not be alerted to when upgrading which is within the realm of acceptability of a tech preview feature.

kaovilai: If we have a GUI, this would be a simple prompt.

kaovilai: I think it is acceptable that setting certain dpa spec adds plugin additional from what you specified.

kaovilai: [Operators in Kubernetes](https://kubernetes.io/docs/concepts/extend-kubernetes/operator/#operators-in-kubernetes)
> Kubernetes is designed for automation. Out of the box, you get lots of built-in automation from the core of Kubernetes. You can use Kubernetes to automate deploying and running workloads, and you can automate how Kubernetes does that.

I look at DPA like a _Deployment_ to a _Pod_.

Deployment automate certain fields/metadata of a Pod. DPA automates deployment of Velero on OpenShift..

shubham-pampattiwar: This analogy is incorrect IMO, your suggested change is updating DPA spec based on some other spec specified on the same resource spec (DPA).

kaovilai: In slack we discussed to just document the required plugins to set with data mover flag in addition to the DPA error here.
WDYT @dymurray

kaovilai (approved):

controllers/validator.go
dymurray: It feels like an obvious thing for us to simply add the plugin when data mover is set. Doesn't need to be on the DPA spec for us to include the plugin. Right?

dymurray: Talked in slack, I'm fine with the conclusion

PR #918, opened 2023-03-13
OADP-639: Pass custom CA configuration to volsync

mrnold:
For OADP-639: add custom CA configuration to a secret that can be looked up by volsync, so that restic can handle self-signed certificates.


controllers/datamover.go
```diff
@@ -558,6 +559,59 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
 	return nil, nil
 }
 
+func (r *DPAReconciler) createResticCustomCASecret(dpa *oadpv1alpha1.DataProtectionApplication, bsl velerov1.BackupStorageLocation) (*corev1.Secret, error) {
+	insecureSkipTLSVerify, skipPresent := bsl.Spec.Config["insecureSkipTLSVerify"]
+	if !skipPresent || insecureSkipTLSVerify != "false" {
+		return nil, nil
+	}
+	if bsl.Spec.ObjectStorage.CACert == nil {
```
eemcmullan, line 567: This check would maybe be better here: https://github.com/openshift/oadp-operator/blob/master/controllers/bsl.go#L17

controllers/datamover.go
```diff
@@ -558,6 +559,59 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
...
+			Namespace: bsl.Namespace,
+			Labels: map[string]string{
+				oadpv1alpha1.OadpBSLnameLabel:     bsl.Name,
+				oadpv1alpha1.OadpBSLProviderLabel: bsl.Spec.Provider,
+				oadpv1alpha1.OadpOperatorLabel:    "True",
+			},
+		},
+	}
+
+	mutateSecret := func() error {
```
eemcmullan, line 583: I'm not sure if this will create an issue with the volsync controller. Just something to keep in mind.

controllers/datamover.go
```diff
@@ -673,6 +727,12 @@ func (r *DPAReconciler) ReconcileDataMoverResticSecret(log logr.Logger) (bool, e
 				if err != nil {
 					return false, err
 				}
+
+				_, err = r.createResticCustomCASecret(&dpa, bsl)
```
eemcmullan, line 731: We may want a separate reconcile func for this secret so we do not have multiple objects being reconciled on in one

controllers/datamover.go
```diff
@@ -558,6 +559,59 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
...
+			Labels: map[string]string{
+				oadpv1alpha1.OadpBSLnameLabel:     bsl.Name,
+				oadpv1alpha1.OadpBSLProviderLabel: bsl.Spec.Provider,
+				oadpv1alpha1.OadpOperatorLabel:    "True",
+			},
+		},
+	}
+
+	mutateSecret := func() error {
+		err := controllerutil.SetControllerReference(dpa, secret, r.Scheme)
```
eemcmullan, line 584: I'm not sure if this will create an issue with the volsync controller. Just something to keep in mind.

controllers/datamover.go
mrnold: I moved this to its own function: https://github.com/openshift/oadp-operator/pull/918/commits/1958bbebcde1443a4fbba428a201ea07086fadf2

controllers/datamover.go
mrnold: That's a better place to fail if there's a configuration problem, but I still need to create the secret based on the value of insecureSkipTLSVerify. I pulled this out into its own function so it can be called from both places, if that makes sense: https://github.com/openshift/oadp-operator/pull/918/commits/3c55c383d78ec04e0585eef6e1ece18a73b16a41

controllers/datamover.go
mrnold: I copied this from the other secret creation code without fully understanding what it does. What kind of problems are you anticipating?

eemcmullan: My concern was that the volsync controller may have also been setting a controller reference on this secret: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference
Although I just looked through their controller and do not see them doing so, which means this should be fine.

eemcmullan:
@mrnold LGTM. Will review once this is out of draft and the other relevant PRs are up. Thank you!

controllers/bsl.go
```diff
@@ -60,6 +60,11 @@ func (r *DPAReconciler) ValidateBackupStorageLocations(log logr.Logger) (bool, e
 			default:
 				return false, fmt.Errorf("invalid provider")
 			}
+
+			_, err := r.resticCustomCAEnabled(*bslSpec.Velero, &dpa)
```
shubham-pampattiwar, line 64: This function is supposed to be a validate function, right ? The boolean returned by this function should not be ignored, or else how would the reconcile loop break if validation fails ?

controllers/bsl.go
```diff
@@ -347,3 +352,15 @@ func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(dpa *oadpv1a
 	}
 	return nil
 }
+
+func (r *DPAReconciler) resticCustomCAEnabled(bslSpec velerov1.BackupStorageLocationSpec, dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
+	insecureSkipTLSVerify, skipPresent := bslSpec.Config["insecureSkipTLSVerify"]
+	if !skipPresent || insecureSkipTLSVerify != "false" {
```
shubham-pampattiwar, line 358: Why are we returning `false` here ?

controllers/bsl.go
```diff
@@ -347,3 +352,15 @@ func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(dpa *oadpv1a
...
+		return false, nil
+	}
+	if bslSpec.ObjectStorage.CACert == nil {
```
shubham-pampattiwar, line 361: Maybe to simplify things, we need to add another condition like if `insecureSkipTLSVerify` == `false` && `bslSpec.ObjectStorage.CACert == nil`

controllers/datamover.go
```diff
@@ -681,6 +671,27 @@ func (r *DPAReconciler) ReconcileDataMoverResticSecret(log logr.Logger) (bool, e
 	return true, nil
 }
 
+func (r *DPAReconciler) listBackupStorageLocations() (*velerov1.BackupStorageLocationList, error) {
```
shubham-pampattiwar, line 674: +1

controllers/datamover.go
```diff
@@ -792,3 +803,78 @@ func (r *DPAReconciler) parseGCPSecret(secret corev1.Secret, secretKey string) (
...
+	}
+
+	backupStorageLocationList, err := r.listBackupStorageLocations()
+	if err != nil {
+		return false, err
+	}
+
+	if r.checkIfDataMoverIsEnabled(&dpa) {
+		for _, bsl := range backupStorageLocationList.Items {
+			if strings.Contains(bsl.Name, dpa.Name) {
```
shubham-pampattiwar, line 820: Please add a TODO here to change the logic when we allow users to specify BSL name

shubham-pampattiwar (commented):
Added a few comments

shubham-pampattiwar:
@mrnold @eemcmullan Why are not using the existing restic secret for this purpose ?

controllers/bsl.go
kaovilai: if insecureSkipTlsVerify is true or does not exists, then customCA is not used for communication.

controllers/datamover.go
kaovilai: also it might be more efficient to check the other way around.. dpa.spec.backupLocations instead of listing all BSLs that may contain those created outside of DPA.

controllers/bsl.go
mrnold: The function validates the custom CA field, where the only invalid condition is specifying "insecureSkipTLSVerify=false" without including a caCert. The returned boolean indicates whether or not "insecureSkipTLSVerify" is set, which is used later to decide if the custom CA field should be passed to restic. In this context, it does not matter if it is set to true or false, just that the configuration is valid (indicated by a nil err).

I guess it could be split up into separate validation and isEnabled functions?

mrnold:
> @mrnold @eemcmullan Why are not using the existing restic secret for this purpose ?

The existing restic secret maps directly to the Repository field in volsync's restic replication source/destination specs. There is a CustomCA field directly alongside Repository, so from that perspective it makes sense to have two secrets to match the types. The format of the Repository secret is specified by volsync, so I didn't think I should go adding new keys to it in case volsync decides to add a conflicting key name. It is certainly cleaner to use one secret from the OADP operator's point of view, but it feels more like the contents belong to volsync.

controllers/bsl.go
kaovilai: Reading it again, I agree with @mrnold . Validate does not need to know the bool returned by this function, only if it errors.

mrnold:
I have changed things around to use the existing restic secret, just to see what it looks like. It might be all around simpler to a degree that makes it worth changing my related pull requests to match (https://github.com/migtools/volume-snapshot-mover/pull/198 and https://github.com/migtools/velero-plugin-for-vsm/pull/8).

mrnold:
Okay, I have seen a good backup and restore (as far as it is working for the moment), so I think this is ready for review now.

controllers/datamover.go
```diff
@@ -393,6 +394,7 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
 				pruneInterval = strings.ReplaceAll(pruneInterval, `"`, "")
 				pruneInterval = strings.ReplaceAll(pruneInterval, `'`, "")
 			}
+			resticCustomCA := bsl.Spec.ObjectStorage.CACert
```
eemcmullan, line 397: Should there be a check first as to whether or not this exists? 
`if len(bsl.Spec.ObjectStorage.CACert) > 0`

controllers/datamover.go
```diff
@@ -519,6 +522,7 @@ func (r *DPAReconciler) createResticSecretsPerBSL(dpa *oadpv1alpha1.DataProtecti
 			if len(dpa.Spec.Features.DataMover.PruneInterval) > 0 {
 				pruneInterval = dpa.Spec.Features.DataMover.PruneInterval
 			}
+			resticCustomCA := bsl.Spec.ObjectStorage.CACert
```
eemcmullan, line 525: here as well

controllers/datamover.go
mrnold: I tried this and it didn't make sense in this particular spot because if the length is empty it just ends up passing an empty string to buildDataMoverResticSecretForAWS, so the check didn't really cause any difference in behavior.

But it does make sense to check inside of buildDataMoverResticSecretForAWS, because that way an empty CACert field doesn't get passed along to the next secret. So I added changes there instead, which I think should better cover the concern you have brought up.

eemcmullan (approved):

kaovilai (approved):

sseago (approved):

PR #940, opened 2023-03-27
velero upstream main rebase as of Mar 27, 2023

sseago:


kaovilai:
Do you think it would make sense to bump to v1.11 here? just to signal to human readers what version we intend to be compatible with.
 https://github.com/openshift/oadp-operator/blob/84636765553916f76f8f056464ed180c8902ce56/go.mod#L18

Also would it make sense to up go.mod to 1.19 by now?

sseago:
Hmm. Not sure about 1.11 -- I'm wary of bumping that up when there is no actual release for it yet. With the replace directive, it doesn't really matter, but I think I'd prefer to wait for 1.11 release to bump up the primary dep. I'm open to persuasion there if anyone else has strong opinions. 

As for golang version, that's probably worth doing, but maybe as a separate PR?

kaovilai (approved):

PR #716, opened 2022-05-28
go 1.18 `make kustomize` fix, GH Actions: go 1.18, submit-coverage fix 

kaovilai:
kustomize v4 can be installed via go install
Github Actions: go-test go version bump to 1.18

kustomize v3->v4 breaking changes related to supported url types don't affect our repo

PR #962, opened 2023-04-12
oadp-1.0: Update ci-Dockerfile, Makefile for go 1.19

kaovilai:


build/ci-Dockerfile
```diff
@@ -1,6 +1,6 @@
 # Build the manager binary
 # TODO! Find a real ubi8 image for golang 1.16
```
kaovilai, line 2:
```suggestion
```

kaovilai:
missing bin/controller-gen in ci-Dockerfile

sseago (approved):

kaovilai:
So instead of using ci-Dockerfile, reverting to regularly [scheduled replacing base images used in CI for now](https://github.com/openshift/release/compare/f4675fc84e53ad9f084ce4b62d2cacc7df261109..ca6f60e9b132a863b71c9ce7d2f6b46256217de9#diff-e11094be0aac72be384cc4521caeaf6643a480db0a2e4ce15993cbb2e97753c9R15) which ignores the FROM: in this Dockerfile.

Can come back to using the same FROM as this ci-Dockerfile when we fix bin/controller-gen install.

kaovilai:
so their base images don't have controller-gen working either.. debugging :D

kaovilai:
cherrypicking from https://github.com/openshift/oadp-operator/pull/716

kaovilai:
re: go 1.18+ have different behavior for go-get so now we have to use go install

weshayutin (approved):

kaovilai:
waiting for e2e to fail, override, then we should be good to use ci-Dockerfile in CI including `FROM` line.. since the prior issue was fixed via Makefile cherrypick

PR #984, opened 2023-04-18
oadp-1.1: go get openshift/velero:oadp-1.1

kaovilai:


kaovilai:
After https://github.com/openshift/velero/pull/262 will need to up openshift/velero again.

PR #1001, opened 2023-05-03
add 8 and 3 vol b/r manifest and b/r crs

weshayutin:
add b/r cr's for 8vol csi manifest
add 3vol manifest and b/r cr's


kaovilai (approved):
/skip

PR #1012, opened 2023-05-16
OADP-1844: Validation to ensure vsm plugin is removed from DPA if DM is false

shubham-pampattiwar:


controllers/validator.go
```diff
@@ -52,7 +52,15 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
 	}
-	//
+
+	// check if the VSM plugin is specified or not
+	VSMPluginPresent := false
+	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
+		if plugin == oadpv1alpha1.DefaultPluginVSM {
+			VSMPluginPresent = true
+		}
+	}
```
kaovilai, line 62: Why don't we do this closer to line 78 where var is used?

controllers/validator.go
```diff
@@ -62,19 +70,15 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
-			}
-		}
-
 		if !VSMPluginPresent {
 			return false, errors.New("datamover feature is enabled, need to specify vsm as a default plugin in DataProtectionApplication CR")
 		}
 	}
 
+	if !r.checkIfDataMoverIsEnabled(&dpa) && VSMPluginPresent {
+		return false, errors.New("datamover feature is disabled, need to remove vsm as a default plugin in DataProtectionApplication CR")
```
eemcmullan, line 79: could possibly shorten this to something like "datamover is disabled-remove vsm as a default plugin"

controllers/validator.go
kaovilai: icic

weshayutin (approved):

PR #1038, opened 2023-05-24
Doc with Datamover Advanced VolumeOptions Example

eemcmullan:


docs/examples/datamover_advanced_voloptions.md
```diff
@@ -0,0 +1,349 @@
...
+
+`oc get vsb <vsb-name> -n <app-ns> -ojsonpath="{.status.phase}` 
+
+- There should now be a snapshot(s) in the object store that was given in the restic secret.
+- You can check for this snapshot in your targeted `backupStorageLocation` with a
+prefix of `/<OADP-namespace>`
+
+<h4> For Restore <a id="restore"></a></h4>
+
+- Make sure the application namespace is deleted, as well as any volumeSnapshotContents
```
kaovilai, line 254: Is this just a precaution and there shouldn't be anything that cause faulty restores if VSCs are not cleared?

weshayutin (approved):
Thank you!

docs/examples/datamover_advanced_voloptions.md
```diff
@@ -0,0 +1,374 @@
...
+        ocs-storagecluster-cephfs:
+          sourceVolumeOptions:
+            accessMode: ReadOnlyMany
+            cacheAccessMode: ReadWriteMany
+            cacheStorageClassName: ocs-storagecluster-cephfs
+            moverSecurityContext: true
+            storageClassName: ocs-storagecluster-cephfs-shallow
+          destinationVolumeOptions:
+            moverSecurityContext: true
+        ocs-storagecluster-ceph-rbd:
```
shubham-pampattiwar, line 363: @eemcmullan @weshayutin Why do we need ceph rbd in volumeoptions mapping ? Are we trying to override something for rbd ? if not then why specify ?

eemcmullan: It'll work either way. I added it to show that field was there, but can remove

shubham-pampattiwar: yeah I agree it will work but lets this whole feature is to override options if there is a diff in the needed options vs the default options. So lets remove it so that the users do not get confused IMO.

PR #1067, opened 2023-06-09
Fix 3 CSI Volumes sample app deployment issue

shubham-pampattiwar:


hhpatel14 (approved):

eemcmullan (approved):

weshayutin (changes requested):
why provide the command and not have the manifest yaml to update the change as part of the template? No one is going to read that.
-1

kaovilai (approved):

kaovilai:
> why provide the command and not have the manifest yaml to update the change

I guess it isn't always needed??

shubham-pampattiwar:
@kaovilai @weshayutin yes this command is not always needed, we only the issue for OCP 4.10 AFAIK.

weshayutin:
@shubham-pampattiwar does applying the template in 4.11+ do anything negative?

PR #1068, opened 2023-06-09
Add a way to override velero/datamover replicas for debug purposes.

mrnold:
For local debugging, I need a way to avoid starting the velero and volume-snapshot-mover pods. This pull request adds two environment variables to control the respective deployment replicas, which seems like the simplest way to get this to work. There should be no change in behavior when they are not set.

### DIFF
--- a/controllers/datamover.go
+++ b/controllers/datamover.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/go-logr/logr"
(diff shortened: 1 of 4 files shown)


kaovilai:
You can edit CSV replicas to 0 then you can edit deployment replicas of these without operator undoing them

mrnold:
> You can edit CSV replicas to 0 then you can edit deployment replicas of these without operator undoing them

If I understand this correctly, that's spec.install.spec.deployments(openshift-adp-controller-manager).spec.replicas in the OADP ClusterServiceVersion instance? That's what I have been using to shut down oadp-operator. But I also want to launch oadp-operator as a debug target, and when I do it immediately sets the velero/datamover replicas to 1.

kaovilai:
Ok cool make sense.

/approve
/lgtm

shubham-pampattiwar (approved):

mrnold:
Thanks, I added some unit tests.

PR #1078, opened 2023-06-30
OADP-2175: Make use of restic secret to pass datamover retain policy instead of datamover config map

shubham-pampattiwar:
In OADP 1.2, we used datamover config map in order to pass the values of retain policy from DPA CR to replicationSource CR, but this meant that we had to assign volume options as well due to the nature of changes that went in the volume options feature PR. (Shared config map of volume options and retain policy), This PR uses restic secret to pass the retain policy from DPA to ReplicationSource insterad of the Volume Options CM. 
Related VSM controller changes: https://github.com/migtools/volume-snapshot-mover/pull/245

### DIFF
--- a/controllers/datamover.go
+++ b/controllers/datamover.go
@@ -66,6 +66,15 @@ type gcpCredentials struct {
 	googleApplicationCredentials string
 }
 
+type retainPolicy struct {
+	daily   string
+	weekly  string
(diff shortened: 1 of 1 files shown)


kaovilai (approved):
/hold ? if you don't need additional review feel free to unhold.

Issue #769, opened 2022-07-25
Bug: velero metrics stripped


felicianmv:
### Contact Details

felician.moldovan@flex.com

### Describe bug

Hi guys,

I'm trying to setup alerting and reporting for our OADP instance ( 1.0.3 ) and Prometheus/Grafana is the logical solution, but from what I see data related to Velero are only 2: velero_backup_total and velero_restore_total.

Checking upstream Velero docs and /metrics URL shows that Velero actually is exposing a lot more data ( vanila_metrics.txt attached ), but Prometheus doesn't pick them up.
[vanilla_metrics.txt](https://github.com/openshift/oadp-operator/files/9180866/vanilla_metrics.txt)

On the OADP namespace I found a ServiceMonitor object ( openshift-adp-velero-metrics-sm ) which seems to be filtering data recorded by Prometheus, specifically this section:
`
   metricRelabelings:
    - action: keep
      regex: velero_backup|velero_restore
      sourceLabels:
      - __name__
`

I changed the above section into 
`
   metricRelabelings:
    - action: keep
      regex: velero_backup.*|velero_restore.*
      sourceLabels:
      - __name__
`
and now I have all the data to setup monitoring.

### What happened?

Is there a specific reason for that filter in the ServiceMonitor?
Is it fine to use my solution above for production?
Any plans to remove the filter in a future version of OADP?

### OADP Version

1.0.3 (Stable)

### OpenShift Version

4.8

### New issue

- [X] This issue is new

shubham-pampattiwar:
@felicianmv Thank you for reaching out and raising the concerns. We added a service monitor to filter out the metrics exposed  via Velero because most of the metrics exposed did not have bounded [cardiniality](https://www.robustperception.io/cardinality-is-key/) and we were concerned that this might affect openshift's in-cluster monitoring stack adversely.

QuingKhaos:
@shubham-pampattiwar Can we please have a `exposeAllMetrice: true` setting on the operator, so we can override this as long as we are aware of the cardinality consequences? We need to create alerts on production backup schedules.

kaovilai:
> Do not check teh Enable Monitoring checkbox during installation of OADP Operator

Then you can have your own prometheus instance record everything.

PR #1081, opened 2023-07-05
[oadp-1.2] OADP-1887: Disable cluster monitoring and remove ServiceMonitor

mpryc:
Disable Velero cluster monitoring and remove ServiceMonitor from the Velero deployment.

First change to remove cluster monitoring of OADP that will be replaced by the user workload monitoring (UWM).

Related Issues:
  https://github.com/openshift/oadp-operator/issues/769
  https://issues.redhat.com/browse/OADP-1887
  https://issues.redhat.com/browse/OADP-661

The openshift-adp-velero-metrics-svc is left to easy process of enabling UWM.

Once UWM is enabled, it will require setting up ServiceMonitor and configuring alerts or dashboards that are crucial for a particular use-case.

Enablement of user workload monitoring with additional documentation is not part of this PR to easilly re-add the cluster monitoring in the future by reverting this change.


mateusoliveira43:
From the issues, it seems this fix should be to `1.2` and `1.1` versions. How is the process of updating released versions of OADP? Or is just change in master?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -151,11 +151,11 @@ metadata:
     categories: OpenShift Optional
     certified: "false"
     containerImage: quay.io/konveyor/oadp-operator:latest
-    createdAt: "2020-09-08T12:21:00Z"
+    createdAt: "2023-07-05T10:14:04Z"
```
mateusoliveira43, line 154: what command updates the operator files?

mateusoliveira43:
There is any way of testing these changes, @mpryc ?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mpryc: I have used `make bundle`

mpryc:
> There is any way of testing these changes, @mpryc ?

@mateusoliveira43 Sure. Build and deploy new operator. Log in to cluster (also had to modify Makefile to use podman instead of docker, see below):

```shell
make build-deploy
make deploy-olm
```

Modified Makefile

```diff
diff --git a/Makefile b/Makefile
index 94cfa11..bf250a4 100644
--- a/Makefile
+++ b/Makefile
@@ -212,10 +212,10 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # related enhancements that may remove the need to manually install qemu-user-static https://bugzilla.redhat.com/show_bug.cgi?id=2061584
 DOCKER_BUILD_ARGS ?= --platform=linux/amd64
 docker-build: ## Build docker image with the manager.
-	docker build -t $(IMG) . $(DOCKER_BUILD_ARGS)
+	podman build -t $(IMG) . $(DOCKER_BUILD_ARGS)
 
 docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+	podman push ${IMG}
 
 ##@ Deployment
 
@@ -361,7 +361,7 @@ nullable-crds-config:
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) . $(DOCKER_BUILD_ARGS)
+	podman build -f bundle.Dockerfile -t $(BUNDLE_IMG) . $(DOCKER_BUILD_ARGS)
 
 .PHONY: bundle-push
 bundle-push: ## Push the bundle image.
```

After OADP is deployed create DPA and check if the oadp service is missing from the:

```shell
$ oc get servicemonitor -n openshift-adp
```

However the SVC should be there:
```shell
$ oc get svc -n openshift-adp openshift-adp-velero-metrics-svc
NAME                               TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
openshift-adp-velero-metrics-svc   ClusterIP   172.30.38.244   <none>        8085/TCP   3h29m
```

Also the `openshift-adp` namespace should be without the ` openshift.io/cluster-monitoring` label:

```shell
$ oc get label openshift-adp
error: the server doesn't have a resource type "label"
(.venv) [migi@fedora mpryc-oadp-operator-prometheus]$ oc describe ns openshift-adp
Name:         openshift-adp
Labels:       kubernetes.io/metadata.name=openshift-adp
              olm.operatorgroup.uid/3419e1b4-c27a-4b63-808c-69046ce67152=
              pod-security.kubernetes.io/audit=privileged
              pod-security.kubernetes.io/enforce=privileged
              pod-security.kubernetes.io/warn=privileged
Annotations:  openshift.io/sa.scc.mcs: s0:c26,c0
              openshift.io/sa.scc.supplemental-groups: 1000650000/10000
              openshift.io/sa.scc.uid-range: 1000650000/10000
              volsync.backube/privileged-movers: true
Status:       Active

No resource quota.

No LimitRange resource.
```

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mateusoliveira43: for the team: when is this field updated?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -155,7 +155,7 @@ metadata:
     description: OADP (OpenShift API for Data Protection) operator sets up and installs
       Data Protection Applications on the OpenShift platform.
     olm.skipRange: '>=0.0.0 <99.0.0'
-    operatorframework.io/cluster-monitoring: "true"
+    operatorframework.io/cluster-monitoring: "false"
```
shubham-pampattiwar, line 158: Shouldn't we remove this annotation altogether, instead of marking it as false ?

config/manager/manager.yaml
```diff
@@ -3,7 +3,7 @@ kind: Namespace
 metadata:
   labels:
     control-plane: controller-manager
-    openshift.io/cluster-monitoring: "true"
+    openshift.io/cluster-monitoring: "false"
```
shubham-pampattiwar, line 6: Same question as earlier.

shawn-hurley:
Do we currently send back metrics when enabled, and if we do, are we able to send these back to the telemetry through User Workload Monitoring,

mpryc:
> Do we currently send back metrics when enabled, and if we do, are we able to send these back to the telemetry through User Workload Monitoring,

There is relevant change in thecluster-monitoring-operator, however we will be pulling it out as well:  https://github.com/openshift/cluster-monitoring-operator/pull/1536

mateusoliveira43:
with this PR will thsi message disappear when installing OADP?
Screenshot from 2023-07-06 16-27-50

mpryc:
> with this PR will thsi message disappear when installing OADP? Screenshot from 2023-07-06 16-27-50

correct

bundle/manifests/oadp-operator.clusterserviceversion.yaml
kaovilai: it was populated when repo first initialized. Only recent operator-sdk version update this field after every `make bundle`. We have yet to upgrade this repo to the newer version.

kaovilai:
makefile edits for podman is unnecessary as if you prefer podman, installing https://packages.fedoraproject.org/pkgs/podman/podman-docker/ would be a more universal solution for your environment.

deploy/disconnected-prep.sh
mateusoliveira43, line 1: for the team: how is this script used? did not see any mentions to it with `grep -Iinr 'disconnected-prep.sh' .`

shubham-pampattiwar: @mateusoliveira43 This scrip is used in releasing a OADP Operator for disconnected environment (air gaped openshift clusters)

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mpryc: @shubham-pampattiwar 

I am not sure about that. I always prefer to explicitly tell what is the intended behavior as:
 - current the defaults may change over time
 - I don't know how the update/upgrade scenario for existing namespaces will behave and here we explicitly say what is the intended label for that setting, which can be reflected better in the documentation to check for the false or true on that label, otherwise we also need to check if that label exists
 - user have knowledge by looking at the label that the cluster-monitoring is disabled, so a query on the cluster based on labels can quickly identify the ones where cluster-monitoring is enabled (true) but also negative filter is nice to have, or at least doesn't hurt

let me know what do you think as I am happy to remove that in case you disagree

shawn-hurley: I would probably remove the label as well to be safe and future-proof. but just my two cents it's a nit IMO.

mpryc: Ok I will propose change to this PR and remove the labels.

mpryc: New revision to that PR removed all *cluster-monitoring* labels and also part that was adding this for the disconnected environment.

@shawn-hurley @shubham-pampattiwar

shubham-pampattiwar (approved):

sseago (approved):

PR #1091, opened 2023-07-17
point to 1.2 branch for velero and csi plugin on master

sseago:


shubham-pampattiwar (approved):

pkg/common/common.go
```diff
@@ -23,12 +23,12 @@ const (
 
 // Images
 const (
-	VeleroImage          = "quay.io/konveyor/velero:latest"
+	VeleroImage          = "quay.io/konveyor/velero:oadp-1.2"
```
kaovilai, line 26: don't actually need these. IIRC we just read from ENV first.. only if ENV is empty/malformed do these hardcoded backups take effect.

kaovilai: def non issue.. just a note.

sseago: @kaovilai I was just following what we did for oadp-1.2 branch. In any case, if it's harmless, lets just leave it for now.

kaovilai:
keeping builder:v1.19 for now to minimize changes. added GOBIN to path as a temp workaround

shubham-pampattiwar (approved):
/lgtm

kaovilai (approved):

PR #1094, opened 2023-07-19
use latest builder golang image

sseago:


kaovilai:
once e2e is running I think we can merge.. that validates that nothing else is breaking.. since e2e is using ci-Dockerfile already.

sseago:
Merging now and then re-testing with openshift/release PR

PR #1098, opened 2023-07-20
fix: defaultPlugins validations

mateusoliveira43:
Add defaultPlugins yaml validation, that already exist in Go code, to OADP CRD.

## What has been done to achieve the goal

Edited files in `api/v1alpha1/` folder and run `make bundle`.

## How to test

Run `make deploy-olm` and try creating a DPA with a invalid defaultPlugin. After testing, run `make undeploy-olm` for clean up.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -152,7 +152,8 @@ help: ## Display this help.
 
 ##@ Development
 
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+# run make bundle instead
+manifests: controller-gen
 	# Commenting out default which overwrites scoped config/rbac/role.yaml
 	# GOFLAGS="-mod=mod" $(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	GOFLAGS="-mod=mod" $(CONTROLLER_GEN) $(CRD_OPTIONS) webhook paths="./..." output:crd:artifacts:config=config/crd/bases
--- a/Makefile
+++ b/Makefile
@@ -152,8 +152,7 @@ help: ## Display this help.
 
 ##@ Development
 
-# run make bundle instead
-manifests: controller-gen
+manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	# Commenting out default which overwrites scoped config/rbac/role.yaml
 	# GOFLAGS="-mod=mod" $(CONTROLLER_GEN) $(CRD_OPTIONS) rbac:roleName=manager-role webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 	GOFLAGS="-mod=mod" $(CONTROLLER_GEN) $(CRD_OPTIONS) webhook paths="./..." output:crd:artifacts:config=config/crd/bases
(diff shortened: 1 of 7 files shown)


Makefile
```diff
@@ -152,7 +152,8 @@ help: ## Display this help.
 
 ##@ Development
 
-manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
+# run make bundle instead
+manifests: controller-gen
```
mateusoliveira43, line 156: remove help to avoid my mistake of running manifests instead of bundle. Or there is scenarios that is good to run only manifests?

api/v1alpha1/cloud_storage_types.go
```diff
@@ -36,8 +36,8 @@ type CloudStorageSpec struct {
 	Tags map[string]string `json:"tags,omitempty"`
 	// region for the bucket to be in, will be us-east-1 if not set.
 	Region string `json:"region,omitempty"`
-	// +kubebuilder:validation:Enum=aws
 	// provider is the provider of the cloud storage
+	// +kubebuilder:validation:Enum=aws;azure;gcp
```
mateusoliveira43, line 40: this one was missing or was it on purpose?

mateusoliveira43: going to revert then

Makefile
kaovilai: If we did not have the nullables target we added, perhaps manifests would be more useful to run occasionally.. but we still wanna keep it around, please add back the comments around what it does.

api/v1alpha1/oadp_types.go
```diff
@@ -304,6 +305,7 @@ type DataProtectionApplicationSpec struct {
 	//   - resticRestoreImageFqin
 	//   - kubevirtPluginImageFqin
 	// +optional
+	// +kubebuilder:validation:MinProperties: 1
```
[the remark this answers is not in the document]
mateusoliveira43: that would be the idea, but it was not added to the yaml and still did not think how to add the keys as the properties of the object

api/v1alpha1/oadp_types.go
```diff
@@ -42,6 +42,7 @@ const OadpBSLProviderLabel = "openshift.io/oadp-bsl-provider"
 const DataMoverLabel = "openshift.io/volume-snapshot-mover"
 const StorageClassLabel = "openshift.io/vsm-storageclass"
 
+// +kubebuilder:validation:Enum=aws;gcp;azure;csi;vsm;openshift;kubevirt
```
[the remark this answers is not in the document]
sseago: "user need to add openshift plugin" -- I don't know that we'd want to enforce this via validation. It may be that someone will want to remove it for testing purposes -- i.e. "the bug seems to be in the openshift plugin, because when I remove it, the issue goes away". Documentation should certainly tell users that under normal circumstances they should include it, but I don't think there's any reason to require it via validation.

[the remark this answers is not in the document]
mateusoliveira43: right, good thinking

weshayutin:
@mateusoliveira43 seems like our test reg-ex may be botched.. IMHO the unit and e2e should trigger on this.

mateusoliveira43:
I think there is no need to rush @weshayutin , I can try to look into it (I only saw one user use this field wrong in the 2 months I am here)

mateusoliveira43:
@weshayutin I think was false alarm, regex working fine, is just that I am still not part of OpenShift org, once some gives ok to test, all test will be triggered, because I modified a Go file.

Issue #1046, opened 2023-05-31
Design: Enable Kopia file copy


weshayutin:
Technical requirements to add kopia as the default file copy option.
 
- Upstream Design: https://github.com/vmware-tanzu/velero/blob/v1.10.0/design/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md

 ## To-do

- [x ] ResticRepository CR is renamed as BackupRepository CR ( as velero 1.10 )
- [ ] There is a new installation flag --pod-volume-backup-uploader, we need to add support for this via DPA
- [ x ] Restic daemonset is renamed as node-agent, we might have already tackled this as part of our routine rebase
- [ ] Documentation: There is a Velero CLI change, Velero restic repo get is not Velero repo get
- [ ] Documentation: Restic name presense is removed from APIs as well like defaultVolumesToRestic is now defaultVolumesToFSBackup
   - [ ] Eng:  if defaultVolumesToFSBackup is a server param this needs investigation
- [ ] Make Kopia the default
- [ ] Document user requirements
  - [ ] need to rebackup after installing 1.3.0 or flip option to use use restic.

Decent issue for newer folks

weshayutin:
related to #1057

shubham-pampattiwar:
Changes that we need to be aware and support via OADP involve the following (but not limited to):
- `ResticRepository` CR is renamed as `BackupRepository` CR
- There is a new installation flag `--pod-volume-backup-uploader`, we need to add support for this via DPA
- Restic daemonset is renamed as `node-agent`, we might have already tackled this as part of our routine rebase 
- There is a Velero CLI change, `Velero restic repo get` is not `Velero repo get` 
- Restic name presense is removed from APIs as well like `defaultVolumesToRestic` is now `defaultVolumesToFSBackup`

sseago:
Investigation required for proper documentation. Main concern is what happens when you need to a restore a backup with a different uploader than the currently configured one. For example, you configure the new OADP 1.3 install to use kopia. What happens when you attempt to restore a restic backup (from 1.2 or 1.3). Will it succeed? Will it fail with a sane error message (i.e. "restic uploader is not configured...")? What is the proper procedure for restoring from restic when kopia is enabled (or vice versa)? "pause schedules; set DPA to use restic uploader; (don't run any backups); restore; upon restore completion, set DPA to use kopia uploader; unpause schedules", etc.

tsanders-rh:
[...]

PR #1101, opened 2023-07-25
Node Agent Configuration design and design template

mpryc:
New configuration for the file system backup & restore design that will be used by the OADP and allow to choose Restic or Kopia uploader type.

_template.md borrowed from the Upstream vmware-tanzu/velero/main/design/_template.md

WIP implementation: #1102 

Closes https://github.com/openshift/oadp-operator/issues/1046

### DIFF
--- a/docs/design/_template.md
+++ b/docs/design/_template.md
@@ -1,0 +1,53 @@
+# Design proposal template `<replace with your proposal's title>`
+
+_Note_: The preferred style for design documents is one sentence per line.
+*Do not wrap lines*.
+This aids in review of the document as changes to a line are not obscured by the reflowing those changes caused and has a side effect of avoiding debate about one or two space after a period.
+
+_Note_: The name of the file should follow the name pattern `<short meaningful words joined by '-'>_design.md`, e.g:
+`listener-design.md`.
+
+## Abstract
+One to two sentences that describes the goal of this proposal and the problem being solved by the proposed change.
+The reader should be able to tell by the title, and the opening paragraph, if this document is relevant to them.
+
+## Background
+One to two paragraphs of exposition to set the context for this proposal.
+
+## Goals
+- A short list of things which will be accomplished by implementing this proposal.
+- Two things is ok.
+- Three is pushing it.
+- More than three goals suggests that the proposal's scope is too large.
+
+## Non Goals
+- A short list of items which are:
+- a. out of scope
+- b. follow on items which are deliberately excluded from this proposal.
+
+
+## High-Level Design
+One to two paragraphs that describe the high level changes that will be made to implement this proposal.
+
+## Detailed Design
+A detailed design describing how the changes to the product should be made.
+
+The names of types, fields, interfaces, and methods should be agreed on here, not debated in code review.
+The same applies to changes in CRDs, YAML examples, and so on.
+
+Ideally the changes should be made in sequence so that the work required to implement this design can be done incrementally, possibly in parallel.
+
+## Alternatives Considered
+If there are alternative high level or detailed designs that were not pursued they should be called out here with a brief explanation of why they were not pursued.
+
+## Security Considerations
+If this proposal has an impact to the security of the product, its users, or data stored or transmitted via the product, they must be addressed here.
+
+## Compatibility
+A discussion of any compatibility issues that need to be considered
+
+## Implementation
+A description of the implementation, timelines, and any resources that have agreed to contribute.
+
+## Open Issues
+A discussion of issues relating to this proposal for which the author does not know the solution. This section may be omitted if there are none.
--- a/docs/design/node_agent_config-design.md
+++ b/docs/design/node_agent_config-design.md
@@ -1,0 +1,225 @@
+# NodeAgentConfig configuration with restic/kopia
+Date: 2023-07-25
+
+## Abstract
+
+New configuration for the file system backup & restore that will be used by the OADP and allow to choose Restic or Kopia uploader type.
+
+## Background
+
+For the file system backup and restore the Velero may use Restic or Kopia as the uploader mechanism. A number of tests were [performed](https://velero.io/docs/main/performance-guidance/) by the Velero community to compare those mechanisms. In many cases Kopia uploader is a much better performing mechanism and as such should be added to the OADP operator.
+
+Please refer to the upstream [kopia uploader integration design](https://github.com/vmware-tanzu/velero/blob/main/design/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md) for underlying Velero design and backup & restore workflow.
+
+Upstream [kopia](https://github.com/kopia/kopia) project which is used by Velero and configured by the OADP as this design proposes.
+
+## Goals
+
+- A new option `nodeAgentConfig` to allow configuration of restic or kopia uploader
+- Backwards compatibility with the current OADP configuration schema options, namely `restic`
+- Preparation for the future deprecation of the current `restic` configuration option (from OADP 1.4+)
+- Allow new schema(s) to be used by the datamover node agent
+- Enablment of datamover node agent
+- Deprecation of the `restic` configuration option
+- New environment option `FS_PV_HOSTPATH` that is used as a replacement for `RESTIC_PV_HOSTPATH`. See [Compatibility](#compatibility) section for more details.
+- Removal of the `restic-restore-action-config` ConfigMap with direct replacement by `fs-restore-action-config`. See [Compatibility](#compatibility) section for more details.
+
+## Non Goals
+- Removal of the `restic` configuration option
+- Removal of the `RESTIC_PV_HOSTPATH` environment option
+- Support for the downgreade of OADP operator with new configuration options
+- E2E tests for the `kopia` or `restic` uploader, however they should be added in the near future to cover dpa testing of the new fields and we need backup/restore e2e tests which test both kopia and restic (and datamover eventually) using this new struct.
+
+## High-Level Design
+
+Since new `nodeAgentConfig` configuration option is a sibling of the `restic` one, the new common structure `NodeAgentCommonFields` will be created which will be exactly the same data structure as current `ResticConfig` and will be used by both `Restic` and the new `NodeAgentConfig`. The only difference between `NodeAgentConfig` and `Restic` is an addition of one `UploaderType` option to the `NodeAgentConfig` that will be either `kopia` or `restic`.
+
+When `nodeAgentConfig` is used, the `UploaderType` option is a required one, so the user have to select either `kopia` or `restic`.
+
+## Detailed Design
+
+
+### New data structures
+
+A new structure will be added, that includes the ResticConfig inline and extends this with one new parameter `UploaderType`:
+
+```go
+type NodeAgentConfig struct {
+	// Embedding NodeAgentCommonFields
+	// +optional
+	NodeAgentCommonFields `json:",inline"`
+
+	// The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'
+	// +kubebuilder:validation:Enum=restic;kopia
+	// +kubebuilder:validation:Required
+	UploaderType string `json:"uploaderType"`
+}
+```
+
+The `NodeAgentCommonFields` structure is 1-1 as the current `Restic` structure.
+
+```go
+type NodeAgentCommonFields struct {
+	// enable defines a boolean pointer whether we want the daemonset to
+	// exist or not
+	// +optional
+	Enable *bool `json:"enable,omitempty"`
+	// supplementalGroups defines the linux groups to be applied to the NodeAgent Pod
+	// +optional
+	SupplementalGroups []int64 `json:"supplementalGroups,omitempty"`
+	// timeout defines the NodeAgent timeout, default value is 1h
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
+	// Pod specific configuration
+	PodConfig *PodConfig `json:"podConfig,omitempty"`
+}
+```
+
+Current `Restic` structure will embedd the `NodeAgentCommonFields` without any additional options or changes.
+
+The above `NodeAgentConfig` is a member of `ApplicationConfig`, which already includes `ResticConfig`, however we do not replace the `ResticConfig`, instead for backwards compatibility we add new `NodeAgentConfig` parameter:
+
+```go
+// ApplicationConfig defines the configuration for the Data Protection Application
+type ApplicationConfig struct {
+	Velero *VeleroConfig `json:"velero,omitempty"`
+	// (deprecation warning) ResticConfig is the configuration for restic server.
+	// Restic is for backwards compatibility and is replaced by the nodeAgentConfig
+	// Restic will be removed with the OADP 1.4
+	// +kubebuilder:deprecatedversion:warning=1.3
+	// +optional
+	Restic *ResticConfig `json:"restic,omitempty"`
+
+	// NodeAgentConfig is needed to allow selection between kopia or restic
+	// +kubebuilder:validation:Optional
+	NodeAgent *NodeAgentConfig `json:"nodeAgentConfig,omitempty"`
+}
+```
+
+### Configuration YAML options
+
+The `nodeAgentConfig` configuration options and `restic` configuration options will be exactly the same and will contain same options as the current `restic` schema with additional `uploaderType` field under `nodeAgentConfig`. The part of YAML that presents the additions:
+
+```yaml
+configuration:
+	description: configuration is used to configure the data protection application's server config
+	properties:
+
+	nodeAgentConfig:
+		description: NodeAgent is needed to allow selection between kopia or restic
+		properties:
+
+		[...] // <Same as all current restic configuration options>
+
+		uploaderType:
+			description: The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'
+			enum:
+			- restic
+			- kopia
+			type: string
+		type: object
+
+	restic:
+		description: (deprecation warning) Restic is for backwards compatibility and will be removed with the OADP 1.4+. Use nodeAgentConfig instead.
+		properties:
+
+		[...] // <Same as all current restic configuration options>
+
+```
+
+### Validations
+
+It is important to disallow user from using both options `restic` and `nodeAgentConfig` in OADP 1.3 together (error state).
+
+### Deprecation of the `restic` coiguration option in OADP 1.3 and it's future removal
+
+The `restic` configuration option will be deprecated in OADP 1.3. It will be removed in a future OADP release. However, restic backup functionality is still fully-supported, but restic users are encouraged to use the new `nodeAgentConfig` struct instead so that they won't be impacted on upgrade when the legacy struct is removed in a future release.
+
+There were few alternatices considered (see [Alternatives Considered](#alternatives-considered)). We will have three places where the deprecation information will be presented to the user:
+1. Description of the `resic` property will have the deprecation warning
+2. If the `restic` is used, the DPA event will contain a `warning` message to inform user that the `restic` is deprecated, that will appear in the DPA `Events`.
+3. If the `restic` is used, the application log will have `warning` message to inform user that the `restic` is deprecated
+4. Release notes for OADP 1.3 will contain information about new configuration option `nodeAgentConfig` that should be used instead of `restic`
+
+
+## Alternatives Considered
+
+- leave `restic` as the only option and do not allow to use `kopia`
+- remove the `restic` and use `kopia` as the default and only available option
+
+### Schema structure
+
+There were alternative schema structure for the `Restic` and new `NodeAgentConfig` considered, however because structures may be directly used by other `go` applications, outside of the API schema, the decision was to use structure as described in the [New data structures](#new-data-structures) section:
+- Keeping `Restic` as is and including inline into `NodeAgentConfig`
+- Duplicating all the fields from `Restic` within `NodeAgentConfig`
+- Moving all the fields from the `Restic` to the `NodeAgentConfig` and ignoring new field which will appear in the `Restic`
+
+### Deprecation warnings
+For informing user about deprecation of the `restic` we considered few additional options:
+- Embedding `warning` within the OpenShift console. This would be the nicest way as the push notifications will appear in the main OCP console, however it would require bigger implementation effort.
+
+- Creating custom information on the DPA object itself, so whenever user would describe created object that had the `restic` the deprecation warning would appear in the `events` section. That option also requires publishing custom `events` fromt he reconcile function and still requires user to pull that information rather then push method.
+
+- Using kubebuilder annotation
+	```
+	// +kubebuilder:deprecatedversion:warning=<string>
+	```
+	This option is not for one particular onfiguration field, but entire CRD, which is outside of our desired needs as we do not want to deprecate entire CRD and create a new version of it, just one field rename.
+
+- Adding additional Reconcile condition with warning message
+	Currently there are two `Reasons` that the main DPA reconcail status may have: 
+	```
+	const ReconciledReasonComplete = "Complete"
+	const ReconciledReasonError = "Error"
+	```
+
+	There is also message that is attached for each of this `Reason`. For an `Error` reason there is propagated error message and for the `Complete` reason there is one string, which is:
+	```
+	const ReconcileCompleteMessage = "Reconcile complete"
+	```
+
+	Additional Reconcile reason, which would have it's own message could be added, that will be similar to `Complete`, so the operator is functioning properly, however there are some messages that require user attention.
+	```
+	const ReconciledReasonWarning = "Warning"
+	```
+
+	This however would require refactoring of the `Reconcile` functions, and that would require separate design doc and separate implementation. 
+
+## Security Considerations
+
+The enablement of an extra upload mechanism could potentially introduce security implications due to the integration of a new backend, which might contain vulnerabilities.
+
+## Compatibility
+
+### Current `restic` configuration option
+
+This design does not change current configuration options, however it proposes addition of the deprecation warning to the current `Restic` schema. The `Restic` configuration option will be removed from OADP 1.4.
+
+### RESTIC_PV_HOSTPATH environment option
+
+This design does not remove `RESTIC_PV_HOSTPATH` used to redefine the default `/var/lib/kubelet/pods`, however it adds the environment variable `FS_PV_HOSTPATH` that may be used the same way as `RESTIC_PV_HOSTPATH`.
+
+The `RESTIC_PV_HOSTPATH` takes precedence over `FS_PV_HOSTPATH`
+
+### Replacing `restic-restore-action-config` with `fs-restore-action-config`
+
+The `restic-restore-action-config` was [removed](https://github.com/vmware-tanzu/helm-charts/commit/7484d8e8365ab91da698e5b5d6346153cde30af4) in the Velero v1.10.0 it should also be removed and replaced with the [fs-restore-action-config](https://github.com/vmware-tanzu/helm-charts/blob/c15d13a916c255c93ef48e8bf9c59a3ba198d5ca/charts/velero/templates/NOTES.txt#L62).
+
+## Implementation
+
+Implementation for the new data structure will follow the [New data structures](#new-data-structures) design.
+
+There are two uses of the additional `--uploader-type` that will specify restic or kopia uploader type.
+
+- One within call to the method from the `github.com/vmware-tanzu/velero` package:
+  ```go
+  func Deployment(namespace string, opts ...podTemplateOption) *appsv1.Deployment
+  ```
+
+- Second is within the OADP `pkg/velero/server/args.go` that is used for testing and development purposes.
+
+We will also rename the `restic.go` module to be named `nodeagent.go` and relevant functions within the `restic.go`, so they are more generic.
+
+## Open Issues
+
+- n/a


docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+	// Embedding ResticConfig
+	// +kubebuilder:validation:Optional
+	ResticConfig `json:",inline"`
+
+	// The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=restic;kopia
+	UploaderType string `json:"uploaderType,omitempty"`
+}
+```
```
mateusoliveira43, line 45: would be interesting to add the yaml as well? I think it is easier to see the final result (that will be generated from that code)

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Detailed Design
+
+A new structure will be added, that includes the ResticConfig inline and extends this with one new parameter `UploaderType`:
+
+```go
+type NodeAgentConfig struct {
+	// Embedding ResticConfig
+	// +kubebuilder:validation:Optional
+	ResticConfig `json:",inline"`
```
mateusoliveira43, line 38: leftover comma?

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+	Velero *VeleroConfig `json:"velero,omitempty"`
+	// (deprecation warning) ResticConfig is the configuration for restic server.
+	// Restic is for backwards compatibility and will be replaced by nodeagentconfig
+	// with the OADP 1.4+
+	// +optional
+	Restic *ResticConfig `json:"restic,omitempty"`
+
+	// NodeAgentConfig is needed to allow selection between kopia or restic
+	// +kubebuilder:validation:Optional
+	NodeAgent *NodeAgentConfig `json:"nodeagentconfig,omitempty"`
```
mateusoliveira43, line 61: nodeAgentConfig to follow CamelCase?

docs/design/node_agent_config-design.md
mpryc: It will be part of the implementation PR. This is design, so should be hand written not auto-generated. Usually auto-generated code happens *after* implementation, and implementation *after* design, so it's a chicken-egg wish here.

docs/design/node_agent_config-design.md
mpryc: No it's not:
 https://github.com/openshift/oadp-operator/blob/master/api/v1alpha1/oadp_types.go#L150-L153

mateusoliveira43: ?

weshayutin:
@mpryc in the background of your design can you please link to the following with some context.
https://github.com/vmware-tanzu/velero/blob/main/design/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md

weshayutin:
@mpryc it's not very clear to me what changes are required in our operator to allow for the choice between restic/kopia as the nodeagent.  Can you please speak to that a bit.

docs/design/node_agent_config-design.md
kaovilai: +1 nodeAgentConfig

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+```
+
+## Alternatives Considered
+None.
+
+## Security Considerations
+
+The enablement of an extra upload mechanism could potentially introduce security implications due to the integration of a new backend, which might contain vulnerabilities.
+
+## Compatibility
```
weshayutin, line 72: This would be a good spot to document or just clarify, or point to upstream design for restic backups from 1.2.x etc.

weshayutin (commented):
Thank you @mpryc ++

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+
+- A new option `nodeagentconfig` to allow configuration of restic or kopia uploader
+- Backwards compatibility with the current OADP configuration schema options, namely `restic`
+- Preparation for the future deprecation of the current `restic` configuration option (from OADP 1.4+)
+- Allow new schema to be used by the datamover node agent
+
+## Non Goals
+- Enablment of datamover node agent
+- Deprecation of the "restic" configuration option
+- E2E tests for the `kopia` or `restic` uplaoder
```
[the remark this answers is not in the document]
shawn-hurley: a non-nit I would like to make sure that we know if we need more infrastructure or something to do the e2e testing of the new fields

docs/design/node_agent_config-design.md
shawn-hurley: agree the json key should be nodeAgentConfig everything else looks right.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+
+```go
+// ApplicationConfig defines the configuration for the Data Protection Application
+type ApplicationConfig struct {
+	Velero *VeleroConfig `json:"velero,omitempty"`
+	// (deprecation warning) ResticConfig is the configuration for restic server.
+	// Restic is for backwards compatibility and will be replaced by nodeagentconfig
+	// with the OADP 1.4+
+	// +optional
+	Restic *ResticConfig `json:"restic,omitempty"`
```
shawn-hurley, line 57: Is there validation, such that a user can not set Restic and NodeAgent?

shawn-hurley (changes requested):
I am confused on two points:

1. Why readding restic config to the node agent if we are removing it
2. Is the only configuration you need for Kopia the uploader type?

docs/design/_template.md
```diff
@@ -0,0 +1,53 @@
+# Design proposal template `<replace with your proposal's title>`
```
shubham-pampattiwar, line 1: NIT: remove the template part

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
+# NodeAgentConfig configuration with restic/kopia
+Date: 2023-07-25
+
+## Abstract
+
+New configuration for the file system backup & restore that will be used by the OADP and allow to choose Restic or Kopia uploader type.
+
+## Background
+
```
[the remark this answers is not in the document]
shubham-pampattiwar: +1 (Related issues/designs)

shubham-pampattiwar:
fixes: https://github.com/openshift/oadp-operator/issues/1046

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
kaovilai: I can help push e2e tests into the implementation PR if needed.

docs/design/node_agent_config-design.md
kaovilai: I think the idea here is to minimize duplicate struct fields. When Restic is fully gone from oadp that is when fields from resticconfig will be copied over instead of inherited inline.

docs/design/node_agent_config-design.md
mpryc: Ah ok got it, it's the json part, will change.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+	Restic *ResticConfig `json:"restic,omitempty"`
+
+	// NodeAgentConfig is needed to allow selection between kopia or restic
+	// +kubebuilder:validation:Optional
+	NodeAgent *NodeAgentConfig `json:"nodeagentconfig,omitempty"`
+}
+```
+
+## Alternatives Considered
+None.
```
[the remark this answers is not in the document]
shubham-pampattiwar: +1

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: This have to be ResticConfig, because we are using fields from the `ResticConfig struct` inline (same as copy-paste of those fields in that particular line). I just want those to not duplicate same code in 2 places.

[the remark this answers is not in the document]
mpryc: This is slightly different because we use that inline it's in the format `json:",inline"` meaning that *all* fields from ResticConfig will be included here on the same level, so we end up with:

```yaml
                    nodeagentconfig:
                      description: NodeAgent is needed to allow selection between kopia or restic
                      properties:
                        enable:
                          description: enable defines a boolean pointer whether we want the daemonset to exist or not
                          type: boolean
                        podConfig:
                          description: Pod specific configuration
                          properties:
                            env:
                              description: env defines the list of environment variables to be supplied to podSpec
                              items:
                                description: EnvVar represents an environment variable present in a Container.
                                properties:
                                  name:
                                    description: Name of the environment variable. Must be a C_IDENTIFIER.
                                    type: string
                                  value:
[...]
```

sseago: @shawn-hurley @kaovilai That was my first concern as well. If we want, though, we could restructure this and create a new NodeAgentFields struct and then have *that* struct inlined by both ResticConfig and NodeAgentConfig. Should be backwards-compatible for 1.2 ResticConfig as long as the inlined fields are identical to the prior inherited fields. Would that work, or does it break any important design patterns here? The main goal is to not have to copy-and-paste every field from ResticConfig into the new struct.

sseago: In any case, once we completely remove ResticConfig in a future release, we can just put all of these fields directly in NodeAgentConfig.

docs/design/node_agent_config-design.md
sseago: @mpryc @shawn-hurley We should definitely allow only one struct to be included. Either Restic *or* NodeAgent must be empty. Validation should fail if both are included, and validation should give a deprecation warning if Restic is included.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
sseago: On the e2e front, we need dpa testing of the new fields (similar to existing dpa tests -- making sure dpa changes have the desired effect on the installed velero), and we need backup/restore e2e tests which test both kopia and restic (and datamover eventually) using this new struct.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
sseago: @mpryc @shubham-pampattiwar As an alternative, we could  create a new `NodeAgentFields` struct that is identical to current `ResticConfig` (i.e. everything but the uploader-type) and this can be inlined by both `ResticConfig` and `NodeAgentConfig`.

We wouldn't want to call it `UploaderConfig` in any case because:
1) It doesn't include the `uploaderType` field.
2) We need this for datamover too, not just uploader/fsBackup, since to enable datamover you must include a `NodeAgentConfig`.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+- One within call to the method from the `github.com/vmware-tanzu/velero` package:
+  ```go
+  func Deployment(namespace string, opts ...podTemplateOption) *appsv1.Deployment
+  ```
+
+- Second is within the OADP `pkg/velero/server/args.go` that is used for testing and development purposes.
+
+## Open Issues
+
+- Should we also rename the `restic.go` module to be named `node_agent.go` and relevant functions within the `restic.go`, so they are more generic.
```
sseago, line 91: If restic.go is used for datamover/kopia as well, then we probably want to rename it.

sseago:
@shawn-hurley 
```
Why readding restic config to the node agent if we are removing it
Is the only configuration you need for Kopia the uploader type?
```
The first is an attempt to avoid copy-and-paste of all the fields from ResticConfig, but I suggested a possible alternative above -- having both ResticConfig and NodeAgentConfig inline the same shared fieldset which includes everything currently in ResticConfig. We could, of course, just duplicate the individual fields one-by-one and not worry about the copy-and-paste, but the shared struct should make validation easier too.

And yes, the only new field is uploader type. OADP 1.2 and Velero 1.11 already have the same restic vs. kopia config with all other node-agent config being identical for restic vs. kopia -- we just ignore the uploaderType field in 1.2, so we're adding it here. The other difference for 1.12/OADP 1.3 is that datamover will *also* require node agent config in place, but for datamover the uploader type is ignored -- kopia is always used for datamover, even if fs backup uses restic.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
shubham-pampattiwar: Cool, @mpryc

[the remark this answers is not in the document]
shubham-pampattiwar: `NodeAgentConfig` makes more sense.

shawn-hurley: That is what I was driving towards as well. I think the only thing I would do, is add 1 more struct that does not have UploaderType because that is a specific field for the new NodeAgentConfig

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
shawn-hurley: +1 on what @sseago said, we need to at least write down what we should be adding to the e2e tests here IMO

docs/design/node_agent_config-design.md
shawn-hurley: +100 on deprecation warning pretty sure there is a new header you can set that will do that. 

I wonder if we are going to be able to, unless we have a webhook though.

docs/design/node_agent_config-design.md
shawn-hurley: yeah, it could work with validation, but I would just as soon not give it as an option to the user if there is only one right answer you know?

sseago: Also, if UploaderType is optional, are we going to just use velero default for NodeAgentConfig (i.e. kopia)? Since this is a new struct, defaulting to Kopia here won't suprise anyone on upgrade, since upgraded DPAs will use ResticConfig. If we default  UploaderType to Restic here, we will have an upgrade problem in 1.4 if we make the default Kopia -- *unless* part of the design is for the controller to modify the NodeAgentConfig to fill in UploaderType when it's left blank. That solves the future default problem because upgrades with existing DPAs won't have an empty field to default.

mpryc: So my take on this is that it adds some lines and when deprecating ResticConfig one would also want to move NodeAgentCommonFields into NodeAgentConfig and remove NodeAgentCommonFields all together, which makes just extra struct that in fact is the ResticConfig struct. I may be missing something, but looks like artificial addition to me. 

In other words with deprecation of `ResticConfig` we will in the future just move the relevant fields into `NodeAgentConfig` similarly to what we will do with the `NodeAgentCommonFields`.

sseago: We're deprecating ResticConfig right now (for 1.3). So we will want to add deprecation warning condition to the DPA on reconcile (but not an error, since it's still functional). Removing it entirely will be in a future release. 

We wouldn't have to get rid of the common fields struct when removing ResticConfig (although it would simplify things to do so). I guess my take is that there are more lines with the common struct, but the resulting structs are clearer in their purpose. Having an optional field in ResticConfig with only one valid value seems confusing from a user point of view. So the extra complication on the golang struct side of things results in a simpler yaml/CRD API.

shawn-hurley: I don't know if I agree that we would want to add the NodeAgentCommonFields into NodeAgentConfig. It may be confusing, but I do think that in the future, we *may* have additional NodeAgents (for dealing with third-party DataMovers). It will be beneficial to have this common struct IMO.

mpryc: Ok I will add that common struct then. Thanks for all great comments and guiding me on this.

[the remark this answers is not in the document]
mpryc: I am closing this comment as it's the same as other thread. Currently we will have `NodeAgentCommonFields`, which is exactly the same structure as `ResticConfig` that will be included inline by `NodeAgentConfig` and `ResticConfig`. From `ResticConfig` we will remove all fields.

docs/design/node_agent_config-design.md
mpryc: It's in the `controllers/restic.go` and `controllers/restic_test.go`:
  https://github.com/openshift/oadp-operator/blob/master/controllers/restic.go

I feel like it's better to rename it to:
`controllers/nodeagent.go` and  `controllers/nodeagent_test.go`

docs/design/node_agent_config-design.md
kaovilai: We could have "Reconciled with warnings" status?

docs/design/node_agent_config-design.md
kaovilai: CamelCase only for exported fields. snakeCase for private. File names is pretty flexible other than `*_test.go` I think

docs/design/node_agent_config-design.md
mpryc: The fields is there and I will use it:
   https://book.kubebuilder.io/reference/markers/crd.html

It was added in thtat PR:
https://github.com/kubernetes-sigs/controller-tools/blob/ecaffbc2dc09cd5e09c07b692d899dbbc8fdda0e/pkg/crd/markers/crd.go#L52

mpryc: My understanding is that the above `// +kubebuilder:deprecatedversion:warning=<string>` is not working with field validations and just CRD's and their deprecation. If that's correct we can't really use it here as we are not deprecating CRD, just one field within already existing one.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,91 @@
...
+## Alternatives Considered
+None.
+
+## Security Considerations
+
+The enablement of an extra upload mechanism could potentially introduce security implications due to the integration of a new backend, which might contain vulnerabilities.
+
+## Compatibility
+
+This design does not change current configuration options, however it proposes addition of the deprecation warning to the current `Restic` schema. The `Restic` configuration option will be removed from OADP 1.4+.
```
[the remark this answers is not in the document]
mpryc: Discussed that on the scrum yesterday. There is no support or desire to support downgrade of OADP. We are in fact introducing new configuration options which are not and never will be backwards compatible.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+
+For the file system backup and restore the Velero may use Restic or Kopia as the uploader mechanism. A number of tests were [performed](https://velero.io/docs/main/performance-guidance/) by the Velero community to compare those mechanisms. In many cases Kopia uploader is a much better performing mechanism and as such should be added to the OADP operator.
+
+Please refer to the upstream [kopia uploader integration design](https://github.com/vmware-tanzu/velero/blob/main/design/unified-repo-and-kopia-integration/unified-repo-and-kopia-integration.md) for underlying Velero design and backup & restore workflow.
+
+Upstream [kopia](https://github.com/kopia/kopia) project which is used by Velero and configured by the OADP as this design proposes.
+
+## Goals
+
+- A new option `nodeagentconfig` to allow configuration of restic or kopia uploader
```
mateusoliveira43, line 18: NIT nodeAgentConfig (`CTRL+H` in VSCode or `sed` to do it only once in the file)

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+	SupplementalGroups []int64 `json:"supplementalGroups,omitempty"`
+	// timeout defines the Restic timeout, default value is 1h
+	// +optional
+	Timeout string `json:"timeout,omitempty"`
+	// Pod specific configuration
+	PodConfig *PodConfig `json:"podConfig,omitempty"`
+}
+```
+
+Current `Rescrit` structure will embedd the `NodeAgentCommonFields` without any additional options.
```
mateusoliveira43, line 74: NIT restic?

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+	NodeAgent *NodeAgentConfig `json:"nodeAgentConfig,omitempty"`
+}
+```
+
+### Configuration YAML options
+
+The generated YAML manifest for the `nodeAgentConfig` and `restic` configuration options will be as follows:
+
+```yaml
+configuration:
```
mateusoliveira43, line 99: to avoid this big yaml, something like this would be valid?

> the updated DPA will look like this:
```yaml
metadata:
  ...
spec:
  ...
  configuration:
    ...
    restic:
      <Not changed, but added deprecated warning in description>
    nodeAgentConfig:
      uploaderType: <The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'>
      <same fields restic has>
```

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+
+
+## Alternatives Considered
+
+- leave `restic` as the only option and do not allow to use `kopia`
+- remove the `restic` and use `kopia` as the default and only available option
+
+### Schema structure
+
+There were alternative schema structure for the `Restic` and new `NodeAgentConfig` considered, however because structures may be directly used outside of the API schema, the decision was to use structure as described in the [Schema structure](#schema-structure) section:
```
mateusoliveira43, line 475: I am lost in this section. what is the goal of this section? This seems like Alternatives Considered and has a link to itself

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+## Goals
+
+- A new option `nodeagentconfig` to allow configuration of restic or kopia uploader
+- Backwards compatibility with the current OADP configuration schema options, namely `restic`
+- Preparation for the future deprecation of the current `restic` configuration option (from OADP 1.4+)
+- Allow new schema(s) to be used by the datamover node agent
+
+## Non Goals
+- Enablment of datamover node agent
+- Deprecation of the "restic" configuration option
```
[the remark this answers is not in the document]
kaovilai: I think it's meant to say, out of scope.. for now.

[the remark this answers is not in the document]
sseago: Sure -- but we are deprecating "restic" config now. As of 1.3, restic config is deprecated and users are encouraged to use nodeAgentConfig instead, so we will print deprecation warnings for use of restic config in 1.3. We want to deprecate it now so 1.3 users will get warnings that it goes away in the future, so that they have more time to move to the new config.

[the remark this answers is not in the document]
sseago: So deprecation of restic config is in scope for this proposal.

docs/design/node_agent_config-design.md
mpryc: Ok, but we are adding a new `nodeAgentConfig` which will contain all options that `restic` did +`uploaderType` so in fact entire `nodeAgentConfig` is what's changing. This way should we include entire part of the yaml for `nodeAgentConfig` ?

docs/design/node_agent_config-design.md
mpryc: Ok will rename one.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+
+## Goals
+
+- A new option `nodeagentconfig` to allow configuration of restic or kopia uploader
+- Backwards compatibility with the current OADP configuration schema options, namely `restic`
+- Preparation for the future deprecation of the current `restic` configuration option (from OADP 1.4+)
+- Allow new schema(s) to be used by the datamover node agent
+
+## Non Goals
+- Enablment of datamover node agent
```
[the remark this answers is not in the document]
mpryc: Ok will remove that. My understanding was different. Non goals is that even if they *will* allow enablment, the goal of this proposal is to focus on kopia and do not use it for datamover. If it's a side-effect or design to fit the future purpose it's ok to have it, but it's not part of kopia enablment design.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: Ok understood. We are deprecating, but we are not removing it. So the goal is to deprecate, non goal is to remove it.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+			type: array
+		timeout:
+			description: timeout defines the Restic timeout, default value is 1h
+			type: string
+		type: object
+```
+
+### Validations
+
+It is important to warn user that both options `restic` and `nodeagentconfig` are not allowed together.
```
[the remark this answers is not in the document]
mpryc: It's an error right.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+
+### Deprecation of the `restic` coiguration option from OADP 1.4+
+
+Since the `restic` configuration option will be deprecated with the OADP 1.4, it is important to inform user to prepare for such deprecation. There were few alternatices considered (see [Alternatives Considered](#alternatives-considered)). We will have three places where the deprecation information will be presented to the user:
+1. Description of the `resic` property will have the deprecation warning
```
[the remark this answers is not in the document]
mpryc: @sseago Could you point me to example how this could work? I checked and we can add warning of deprecation of entire CRD, but really one field. Please see also the considered alternatives for the deprecation methods, which I've discussed with @shawn-hurley, however I may be not aware of other options to inform user of the deprecation.

docs/design/node_agent_config-design.md
mateusoliveira43: I think there is no problem, just instead of going over every field of PodConfig (I thing is the biggest thing) just put 3 dots and say it is an object, for example:
```yaml
...
enable: <enable defines a boolean pointer whether we want the daemonset to exist or not>
podConfig:
  <same object with the same name in restic>
  ...
uploaderType: <The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'>
```

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
sseago: @mpryc Since datamover requires node agent configuration, I think we should explicitly consider datamover enablement in scope. That's part of why we renamed it "nodeAgentConfig" from "fsBackupConfig".

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
sseago: @mpryc I suspect that this needs to be handled as part of validation logic in the oadp controller rather than a controller-runtime flag. Basically when validating, if this field is present we need to add a warning condition. @shubham-pampattiwar  are we doing anything similar elsewhere? Adding DPA warning conditions during validation?

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
kaovilai: perhaps doesn't have to be done and tested in part-1 implementation?

docs/design/node_agent_config-design.md
mpryc: I am closing this comment. YAML was added. I feel like structure is much clearer, also as I understood structure may be used by other operators/applications directly without YAML file.

But I have added YAML, it's long so need to shorten it, but that's not within this comment.

docs/design/node_agent_config-design.md
mpryc: Closing as we have common structure and I feel like all other considered options are covered in the "Alternatives Considered"

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: added.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: Added

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
shawn-hurley: I think that we can do in order of what I would prefer:

1. Validation webhook that can add a warning IF it is set. 
2. Create a condition, about deprecation like Scott mentioned.
3. I also think an event on the object could work as well, if we don't want to do the above. 

I do love this discussion though, forcing the user to be aware of this is really good!

docs/design/node_agent_config-design.md
mpryc: Mistake, should be linked to "New data structures"

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: ok moved to the goals

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: moved to goals, non-goal added:
`- Removal of the "restic" configuration option`

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: changed

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+		type: object
+```
+
+### Validations
+
+It is important to warn user that both options `restic` and `nodeagentconfig` are not allowed together.
+
+### Deprecation of the `restic` coiguration option from OADP 1.4+
+
+Since the `restic` configuration option will be deprecated with the OADP 1.4, it is important to inform user to prepare for such deprecation. There were few alternatices considered (see [Alternatives Considered](#alternatives-considered)). We will have three places where the deprecation information will be presented to the user:
```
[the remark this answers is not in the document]
mpryc: Changed that part, so the users will know that we are deprecating restic in 1.3 and future removing.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: See below, but that's open to discuss:

```
1. Description of the `resic` property will have the deprecation warning
2. Application log will have `warning` message to inform user that the `restic` will be deprecated if the `restic` was used
3. Release notes for OADP 1.3 will contain information about new configuration option `nodeAgentConfig` that should be used instead of `restic`
```

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: Added

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: added

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: > I think that we can do in order of what I would prefer:
> 
> 1. Validation webhook that can add a warning IF it is set.
Wasn't there a general problem with webhooks that were problematic and in some events could cause more trouble? I am all for the best technical way to handle such deprecation warning, just not experience enough to make the call which way to go.

> 3. I also think an event on the object could work as well, if we don't want to do the above.
I am not sure if that's technically correct for one particular field within API version ? The doc is about deprecation of entire API and replacing it with the newer version of API. We are not going to bump the API version, right?

https://kubernetes.io/blog/2020/09/03/warnings/#deprecation-warnings

docs/design/node_agent_config-design.md
mateusoliveira43: Just to confirm: we are renaming `ResticConfig` to `NodeAgentCommonFields`? should we change here as well?

mpryc:
@shawn-hurley @kaovilai @sseago @shubham-pampattiwar 
Should we also deprecate as part of this `restic-restore-action-config` and allow to use: `fs-restore-action-config` ?

https://github.com/vmware-tanzu/helm-charts/blob/c15d13a916c255c93ef48e8bf9c59a3ba198d5ca/charts/velero/templates/NOTES.txt#L61C1-L63C11

kaovilai:
deprecate with warning that it's been moved.
To remove warning, rename field to fs-restore-action-config.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
shawn-hurley: I think we have some leeway here as we are in v1alpha1.

If we wanted to, an option here is to create a conversion webhook and bump to, say, v1alpha2 or v1beta1. 

I would just as soon just tell the user from one of the mechanisms that the field no longer has meaning in the future.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,517 @@
...
+- Enablment of datamover node agent
+- Deprecation of the "restic" configuration option
+- Support for the downgreade of OADP operator with new configuration options
+- E2E tests for the `kopia` or `restic` uplaoder, however they should be added in the near future to cover dpa testing of the new fields and we need backup/restore e2e tests which test both kopia and restic (and datamover eventually) using this new struct.
+
+## High-Level Design
+
+Since new `nodeagentconfig` configuration option is a sibling of the `restic` one, the new common structure `NodeAgentCommonFields` will be created which will be exactly the same data structure as current `ResticConfig` and will be used by both `Restic` and the new `NodeAgentConfig`. The only difference between `NodeAgentConfig` and `Restic` is an addition of one `UploaderType` option to the `NodeAgentConfig` that will be either `kopia` or `restic`.
+
+Without the `UploaderType` option a default `restic` will be used. That default option may change in the future.
```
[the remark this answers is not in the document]
mpryc: Changed to ensure `uploaderType` is required field.

docs/design/node_agent_config-design.md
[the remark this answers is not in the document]
mpryc: Changed to have an event within the DPA. Please note there is also extended **Deprecation warnings** with explanation of other options considered. It's also visible on the `oc describe dpa....` under `Events`

Screenshot from 2023-08-02 14-16-36

mpryc:
> deprecate with warning that it's been moved. To remove warning, rename field to fs-restore-action-config.

Actually I have removed that all together, please see design doc for details why.

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Replacing `restic-restore-action-config` with `fs-restore-action-config`
+
+The `restic-restore-action-config` was [removed](https://github.com/vmware-tanzu/helm-charts/commit/7484d8e8365ab91da698e5b5d6346153cde30af4) in the Velero v1.10.0 it should also be removed and replaced with the [fs-restore-action-config](https://github.com/vmware-tanzu/helm-charts/blob/c15d13a916c255c93ef48e8bf9c59a3ba198d5ca/charts/velero/templates/NOTES.txt#L62).
+
+## Implementation
+
+Implementation for the new data structure will follow the [New data structures](#new-data-structures) design.
+
+There are two uses of the additional `--uploader-type` that will specify restic or kopia uplaoder type.
```
sseago, line 212: nit: s/uplaoder/uploader/

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,225 @@
...
+- Enablment of datamover node agent
+- Deprecation of the `restic` configuration option
+- New environment option `FS_PV_HOSTPATH` that is used as a replacement for `RESTIC_PV_HOSTPATH`. See [Compatibility](#compatibility) section for more details.
+- Removal of the `restic-restore-action-config` ConfigMap with direct replacement by `fs-restore-action-config`. See [Compatibility](#compatibility) section for more details.
+
+## Non Goals
+- Removal of the `restic` configuration option
+- Removal of the `RESTIC_PV_HOSTPATH` environment option
+- Support for the downgreade of OADP operator with new configuration options
+- E2E tests for the `kopia` or `restic` uplaoder, however they should be added in the near future to cover dpa testing of the new fields and we need backup/restore e2e tests which test both kopia and restic (and datamover eventually) using this new struct.
```
sseago, line 31: nit: s/uplaoder/uploader/

sseago (commented):
couple minor typos noted, otherwise looks good.

docs/design/node_agent_config-design.md
mpryc: fixed

docs/design/node_agent_config-design.md
mpryc: fixed, thanks.

sseago (approved):

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,225 @@
...
+- Removal of the `restic` configuration option
+- Removal of the `RESTIC_PV_HOSTPATH` environment option
+- Support for the downgreade of OADP operator with new configuration options
+- E2E tests for the `kopia` or `restic` uploader, however they should be added in the near future to cover dpa testing of the new fields and we need backup/restore e2e tests which test both kopia and restic (and datamover eventually) using this new struct.
+
+## High-Level Design
+
+Since new `nodeAgentConfig` configuration option is a sibling of the `restic` one, the new common structure `NodeAgentCommonFields` will be created which will be exactly the same data structure as current `ResticConfig` and will be used by both `Restic` and the new `NodeAgentConfig`. The only difference between `NodeAgentConfig` and `Restic` is an addition of one `UploaderType` option to the `NodeAgentConfig` that will be either `kopia` or `restic`.
+
+When `nodeAgentConfig` is used, the `UploaderType` option is a required one, so the user have to select either `kopia` or `restic`.
```
shawn-hurley, line 37: I thought that we were defaulting this to restic if this is empty?

shawn-hurley (approved):
When we do decide to remove restic, we do have to determine how upgrades are going to work I think for now, the addition of the new things and deprecation warnings is sufficient for this design but wanted to call this out.

docs/design/node_agent_config-design.md
sseago: @shawn-hurley That was in an earlier iteration. We had a slack conversation where we had agreed to make it required. My  main concern, expressed in prior comments here, was if we default to restic, then we then have the "things may change behind users' backs on upgrade" issue to deal with in 1.4 when the default changes to kopia. If we just require a choice now, then there's no upgrade issue there. The alternative to this would be to default to restic but have the reconcile explicitly set the field to `restic` when it's left blank, which also solves the future default change problem, but with more complexity required.

shawn-hurley:
I agree with you completely Scott, and the default thing makes sense to me, thanks for reminding me. 

Approved from me and looks good

docs/design/node_agent_config-design.md
```diff
@@ -0,0 +1,225 @@
...
+- Creating custom information on the DPA object itself, so whenever user would describe created object that had the `restic` the deprecation warning would appear in the `events` section. That option also requires publishing custom `events` fromt he reconcile function and still requires user to pull that information rather then push method.
+
+- Using kubebuilder annotation
+	```
+	// +kubebuilder:deprecatedversion:warning=<string>
+	```
+	This option is not for one particular onfiguration field, but entire CRD, which is outside of our desired needs as we do not want to deprecate entire CRD and create a new version of it, just one field rename.
+
+- Adding additional Reconcile condition with warning message
+	Currently there are two `Reasons` that the main DPA reconcail status may have:
```
weshayutin, line 170: nit reconcail/reconcile
fix post merge

PR #1102, opened 2023-07-25
Adds kopia/restic uploader type

mpryc:
Implements #1101 part of the #1046 

Testing instructions:
1. Uninstall previous version of OADP
2. Ensure you are logged as an admin
  ```shell
  $ oc whoami
  system:admin
  ```
4. Remove left over CRD's:
  ```shell
  $ oc delete crd dataprotectionapplications.oadp.openshift.io volumesnapshotrestores.datamover.oadp.openshift.io volumesnapshotbackups.datamover.oadp.openshift.io cloudstorages.oadp.openshift.io
  ```

5a. You may want to use the following patch to not install or symlink podman to docker:

  ```shell
  $ patch -p1 < /below/patch/file/location.diff
  ```

```diff
index 94cfa11..bf250a4 100644
--- a/Makefile
+++ b/Makefile
@@ -212,10 +212,10 @@ run: manifests generate fmt vet ## Run a controller from your host.
 # related enhancements that may remove the need to manually install qemu-user-static https://bugzilla.redhat.com/show_bug.cgi?id=2061584
 DOCKER_BUILD_ARGS ?= --platform=linux/amd64
 docker-build: ## Build docker image with the manager.
-	docker build -t $(IMG) . $(DOCKER_BUILD_ARGS)
+	podman build -t $(IMG) . $(DOCKER_BUILD_ARGS)
 
 docker-push: ## Push docker image with the manager.
-	docker push ${IMG}
+	podman push ${IMG}
 
 ##@ Deployment
 
@@ -361,7 +361,7 @@ nullable-crds-config:
 
 .PHONY: bundle-build
 bundle-build: ## Build the bundle image.
-	docker build -f bundle.Dockerfile -t $(BUNDLE_IMG) . $(DOCKER_BUILD_ARGS)
+	podman build -f bundle.Dockerfile -t $(BUNDLE_IMG) . $(DOCKER_BUILD_ARGS)
 
 .PHONY: bundle-push
 bundle-push: ## Push the bundle image.
```

5b. Build and deploy OADP from this PR:
  ```shell
  $ make build-deploy
  $ make deploy-olm
  ```

6. Create DPA, ensure to choose `restic` or `kopia` and ensure to use `unsupportedOverrides` with provided `veleroImageFqin`

Make sure to specify the `backupLocations` accordingly to your cluster configuration and `s3` bucket

Part of the DPA for restic:
  ```yaml
  metadata:
    name: dpa-sample
  spec:
    configuration:
      nodeAgentConfig:
        enable: true
        uploaderType: restic
  [...]
    unsupportedOverrides:
      veleroImageFqin: 'docker.io/sseago/velero:kopia'
  ```
And for Kopia
  ```yaml
  metadata:
    name: dpa-sample
  spec:
    configuration:
      nodeAgentConfig:
        enable: true
        uploaderType: kopia
  [...]
    unsupportedOverrides:
      veleroImageFqin: 'docker.io/sseago/velero:kopia'
  ```

8. Create `Backup` object to trigger backup, ensure to include `defaultVolumesToFsBackup: true`
  ```yaml
  apiVersion: velero.io/v1
  kind: Backup
  metadata:
    name: mysql-persistent
    namespace: openshift-adp
  spec:
    includedNamespaces:
    - mysql-persistent
    storageLocation: dpa-sample
    defaultVolumesToFsBackup: true
    ttl: 720h0m0s
  ```


pkg/velero/server/config.go
```diff
@@ -81,8 +81,8 @@ type ServerConfig struct {
 	// Backup all volumes with pod volume file system backup by default.
 	// +optional
 	DefaultVolumesToFsBackup *bool `json:"default-volumes-to-fs-backup,omitempty"`
-	// uploaderType, oadp only support restic at the moment
-
+	// The type of uploader to transfer the data of pod volumes, the supported values are 'restic', 'kopia' (default "restic")
```
mateusoliveira43, line 84: here can't we add validation enum and remove code validation from `pkg/velero/server/args.go`?

api/v1alpha1/oadp_types.go
```diff
@@ -146,10 +146,29 @@ type ResticConfig struct {
 	PodConfig *PodConfig `json:"podConfig,omitempty"`
 }
 
+type NodeAgentConfig struct {
+	// Embedding ResticConfig
+	// +kubebuilder:validation:Optional
```
mateusoliveira43, line 151: this and `+optional` are the same? if yes, we could just one for standardization

api/v1alpha1/oadp_types.go
```diff
@@ -146,10 +146,29 @@ type ResticConfig struct {
...
+	// The type of uploader to transfer the data of pod volumes, the supported values are 'restic' or 'kopia'
+	// +kubebuilder:validation:Optional
+	// +kubebuilder:validation:Enum=restic;kopia
+	UploaderType string `json:"uploaderType,omitempty"`
+}
+
 // ApplicationConfig defines the configuration for the Data Protection Application
 type ApplicationConfig struct {
 	Velero *VeleroConfig `json:"velero,omitempty"`
+	// (deprecation warning) ResticConfig is the configuration for restic server.
```
mateusoliveira43, line 163: OpenAPI has a deprecated field, do not know if this works for this case https://github.com/kubernetes-sigs/kubebuilder/issues/2116#issuecomment-1050509977

api/v1alpha1/oadp_types.go
mpryc: It is, looking at the: https://book.kubebuilder.io/reference/markers/crd-validation.html

```
// +kubebuilder:validation:Optional
specifies that this field is optional, if fields are required by default.

// +optional
specifies that this field is optional, if fields are required by default.
```

pkg/velero/server/config.go
mpryc: I still prefer to have validation on the `pkg/velero/server/args.go` side.

The earlier the user is informed the fields are incorrect the better, which makes that is why we have for that field enumeration that only allows two entries:

```
	// +kubebuilder:validation:Enum=restic;kopia
```

However it's a code side responsibility to confirm that arguments used in the functions are correct. From the go language I don't have the warranty that the function ` StringArr` will *always* receive only those two fields, that's why I do prefer to have such validation in that function.

The `server/args.go` is mostly used outside of standard OADP usage (development/testing) and as such should show also errors when one would be using that function in non-standard way.

controllers/nodeagent.go
```diff
@@ -61,17 +69,32 @@ func getNodeAgentObjectMeta(r *DPAReconciler) metav1.ObjectMeta {
...
 	}
+
+	if dpa.Spec.Configuration.Restic != nil {
+		// V(-1) corresponds to the warn level
+		var deprecationMsg string = "(Deprecation Warning) Use nodeAgentConfig instead of restic, which is deprecated and will be removed with the OADP 1.4"
+		log.V(-1).Info(deprecationMsg)
+		r.EventRecorder.Event(&dpa, corev1.EventTypeWarning, "DeprecationResticConfig", deprecationMsg)
+	}
+
 	if dpa.Spec.Configuration.Restic == nil || dpa.Spec.Configuration.Restic != nil && (dpa.Spec.Configuration.Restic.Enable == nil || !*dpa.Spec.Configuration.Restic.Enable) {
```
mpryc, line 91: This logic is br8ken, fixing...

kaovilai: it is working for restic in 1.2

mpryc: That was comment to myself, but the issue with not creating DaemonSet was around my logic in if/else for the nil pointer check. I had:

```
if dpa.Spec.Configuration.Restic != nil && dpa.Spec.Configuration.Restic.blahblah != nil .... {
} else if dpa.Spec.Configuration.NodeAgent.blahblah != nil {
}
```

So that was wrong, because if `Restic` was used and we were missing just `Restic.blahblah`, then the check on the similar attribute for `NodeAgent.blahblah` was done, however this didn't mean the `NodeAgent` was used at all.

Fixed and now the `DaemonSet` is being created properly. Will upload new rev. shortly.

mpryc: It's fixed now.

api/v1alpha1/oadp_types.go
mpryc: No it does not. The deprecation woks on the entire CRD and not one field.

api/v1alpha1/oadp_types.go
mpryc: changed to be consistent

api/v1alpha1/oadp_types.go
```diff
@@ -130,26 +130,53 @@ type PodConfig struct {
...
+	// Restic is for backwards compatibility and is replaced by the nodeAgentConfig
+	// Restic will be removed with the OADP 1.4
+	// +kubebuilder:deprecatedversion:warning=1.3
+	// +optional
 	Restic *ResticConfig `json:"restic,omitempty"`
+
+	// NodeAgent is needed to allow selection between kopia or restic
+	// +optional
+	NodeAgent *NodeAgentConfig `json:"nodeAgentConfig,omitempty"`
```
mpryc, line 179: @weshayutin @sseago 

I know the design says 'nodeAgentConfig', but should we actually at this point name it 'nodeAgent' ?

Why? When I am playing with this, it seems for me more natural to write it without `Config` suffix as it's already under `configuration` section in our yaml/definition.:

```yaml
  configuration:
    nodeAgent:
      enable: true
      uploaderType: restic
```

Then:
```yaml
  configuration:
    nodeAgentConfig:
      enable: true
      uploaderType: restic
```

build/Dockerfile.bundle
```diff
@@ -7,7 +7,7 @@ LABEL operators.operatorframework.io.bundle.metadata.v1=metadata/
 LABEL operators.operatorframework.io.bundle.package.v1=oadp-operator
 LABEL operators.operatorframework.io.bundle.channels.v1=stable
 LABEL operators.operatorframework.io.bundle.channel.default.v1=stable
-LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.23.0
+LABEL operators.operatorframework.io.metrics.builder=operator-sdk-v1.28.1
```
mateusoliveira43, line 10: I would update operator-sdk in a separate PR

api/v1alpha1/oadp_types.go
sseago: Note that the yaml will still have `configuration.nodeAgent` since the *field name* is `NodeAgent`, while the field type is `NodeAgentConfig`.

That said, I don't oppose renaming the struct to be the same as the field name where it's used. I think we had `Config` following prior practice with `ResticConfig`, but having struct and field named `NodeAgent` is fine by me. Just update the design in this PR along with the impl change so we don't have inconsistent docs.

build/Dockerfile.bundle
sseago: This stuff gets updated as part of `make bundle` based on the operator-sdk used, so I think this is the right place, since it's an auto-generated file.

pkg/velero/server/args.go
```diff
@@ -39,6 +39,13 @@ func (a Args) StringArr(dpaFeatureFlags []string, logLevel string) ([]string, er
 	} else if boolptr.IsSetToFalse(a.DefaultVolumesToFsBackup) {
 		args = append(args, "--default-volumes-to-fs-backup=false")
 	}
+	if a.UploaderType != "" {
```
sseago, line 42: Shouldn't we also return an error if `UploaderType` is empty?

api/v1alpha1/oadp_types.go
sseago: I misread this. The field is indeed called `NodeAgent`, but the yaml says `nodeAgentConfig` -- so at a minimum, change yaml to `nodeAgent` -- I don't have a strong opinion on whether the struct itself changes name.

sseago:
Looks like timeout isn't being set properly with nodeAgentConfig. If you add timeout to restic, it's pulled into the velero deployment, but if you add to nodeAgentConfig it's not. Here's where you need to check for nodeAgentConfig and then restic, rather than just restic:
https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L429-L431
```
	if dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
		fsBackupTimeout = dpa.Spec.Configuration.Restic.Timeout
	}
```

pkg/velero/server/args.go
mpryc: We can return error or ignore and do not add that as a flag to the server args (empty arg is ignored in most CLI's). The ignore is current behavior but I am happy to change it to an error state, thought this isn't really error as it's empty=not passed. 

In here I was following also what is on line 49, which is logLevel if it's empty string it's ignored, so I think if we change it here we also change it to any other string being used as an arg passed to the server?

In our CSV we ensure it's required and it's either kopia or restic (enum type), this file `args.go` as I understand is used for testing/dev only and it's not really used by operator.

sseago: "ignore and do not add that as a flag to the server args" would be defaulting to whatever velero defaults to (kopia in 1.12).
Then again, since UploaderType is now required in the dpa config, I guess this will always be non-empy, since DPA creation itself fails if it's empty.

sseago:
Other than the timeout config bug, I was able to successfuly back up and restore the OADP mysql-persistent testing app with kopia fs backup using this PR.

build/Dockerfile.bundle
mateusoliveira43: oh I mean just the operator-sdk version (from `1.23.0` to `1.28.1`)

At least to update version, we need to also update Makefile https://github.com/openshift/oadp-operator/blob/master/Makefile#L303-L310

weshayutin:
@sseago nice! Thanks for testing mysql-persistent.  If mysql db was successful w/ just a regular kopia backup then we have at least a data point to investigate https://issues.redhat.com/browse/OADP-2360 w/ kopia + DM.  Thank you!

api/v1alpha1/oadp_types.go
mpryc: I have changed the config option to `nodeAgent` and kept the structure name `NodeAgentConfig` so it's similar to `restic` option and `ResticConfig` structure name.

pkg/velero/server/args.go
mpryc: I believe we are talking here about two different parts of OADP:

1. `uploaderType` in DPA is required, but *only* when `nodeAgent` is defined. We need to remember that `nodeAgent` is *optional*, which makes `uploaderType` also non-existing for the scenario where `nodeAgent` is missing.
2. The `if a.UploaderType != "" ` on line 42 relates to the server args, which are not really connected to DPA and are used to configure server for testing/development, which makes that `UploaderType` optional and if it's not there the string will be empty, so we should check here for an empty string case. See:
  https://github.com/openshift/oadp-operator/blob/65281decc7c532255b9e04b88eac7963f8cc57ba/pkg/velero/server/config.go#L85

sseago: ok. So we don't actually expect users to pass UploaderType via server args, though, do we? It's required with `nodeAgent`, but if they're not defining node agent, then 1 of 2 things is true:
1) they're using restic, so OADP is setting `UploaderType` to `restic` in the velero command
2) they're not using node agent at all, so it's irrelevant

So there's really no situation in which a user should be specifying UploaderTypevia server config, is there?

mpryc: you are right, I think the implementation needs to include that and pass `--use-node-agent` if `--uploader-type` is specified.

Should we have an option where `--use-node-agent` but *without* `--uploader-type` in upstream it is possible, then it defaults to the default `--uploader-type` but I am not convinced we should allow that.

My proposal is to inject  `--use-node-agent` whenever `uploaderType` is set in our server args?

api/v1alpha1/oadp_types.go
```diff
@@ -115,13 +115,13 @@ type PodConfig struct {
 	// labels to add to pods
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
-	// nodeSelector defines the nodeSelector to be supplied to podSpec
+	// nodeSelector defines the nodeSelector to be supplied to NodeAgent podSpec
```
shubham-pampattiwar, line 118: NIT: Remove `NodeAgent` from this kubebuilder comment/marker, `PodConfig` is not just specific for nodeAgent/Restic, it also used by Velero Pod.

api/v1alpha1/oadp_types.go
```diff
@@ -115,13 +115,13 @@ type PodConfig struct {
...
 	// +optional
 	NodeSelector map[string]string `json:"nodeSelector,omitempty"`
-	// tolerations defines the list of tolerations to be applied
+	// tolerations defines the list of tolerations to be applied to NodeAgent daemonset
```
shubham-pampattiwar, line 121: Same as above. NIT: Remove `NodeAgent` from this kubebuilder comment/marker, `PodConfig` is not just specific for nodeAgent/Restic, it also used by Velero Pod.

api/v1alpha1/oadp_types.go
```diff
@@ -115,13 +115,13 @@ type PodConfig struct {
...
 	// +optional
 	Tolerations []corev1.Toleration `json:"tolerations,omitempty"`
-	// resourceAllocations defines the CPU and Memory resource allocations for the Pod
+	// resourceAllocations defines the CPU and Memory resource allocations for the NodeAgent Pod
```
shubham-pampattiwar, line 124: NIT: same change as suggested for `tolerations` and `nodeSelector`

api/v1alpha1/oadp_types.go
```diff
@@ -269,7 +296,7 @@ type VolumeOptions struct {
 	// the data mover cache volume
 	//+optional
 	CacheStorageClassName string `json:"cacheStorageClassName,omitempty"`
-	// cacheCapacity determines the size of the restic metadata cache volume
+	// cacheCapacity determines the size of the nodeAgent metadata cache volume
```
shubham-pampattiwar, line 299: NIT: Please remove this change, the `VolumeOptions` are advanced options for Volsync datamover's restic, `nodeAgent` is not relevant here.

build/Dockerfile.bundle
shubham-pampattiwar: How was the bundle generated ? Not via `make bundle` ?

shubham-pampattiwar: I would suggest running (in the following sequence):
- make generate
- make manifests
- make bundle

controllers/nodeagent.go
```diff
@@ -61,17 +69,32 @@ func getNodeAgentObjectMeta(r *DPAReconciler) metav1.ObjectMeta {
...
 	ds := &appsv1.DaemonSet{
 		ObjectMeta: getNodeAgentObjectMeta(r),
 	}
-	if dpa.Spec.Configuration.Restic == nil || dpa.Spec.Configuration.Restic != nil && (dpa.Spec.Configuration.Restic.Enable == nil || !*dpa.Spec.Configuration.Restic.Enable) {
+
+	if dpa.Spec.Configuration.Restic != nil {
+		// V(-1) corresponds to the warn level
+		var deprecationMsg string = "(Deprecation Warning) Use nodeAgent instead of restic, which is deprecated and will be removed with the OADP 1.4"
+		log.V(-1).Info(deprecationMsg)
+		r.EventRecorder.Event(&dpa, corev1.EventTypeWarning, "DeprecationResticConfig", deprecationMsg)
```
shubham-pampattiwar, line 88: ++ @mpryc

controllers/nodeagent.go
```diff
@@ -133,49 +156,57 @@ func (r *DPAReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error)
...
 	if dpa == nil {
 		return nil, fmt.Errorf("dpa cannot be nil")
 	}
 	if ds == nil {
 		return nil, fmt.Errorf("ds cannot be nil")
 	}
 
+	var nodeAgentResourceReqs corev1.ResourceRequirements
+
 	// get resource requirements for restic ds
```
shubham-pampattiwar, line 198: NIT: // get resource requirements for nodeAgent ds

pkg/velero/server/args.go
sseago: I'm not sure we want it there at all. If they want to enable node agent, they should use the node agent config. Otherwise we end up having to duplicate all of this logic under server args too. But also, --use-node-agent isn't a server arg. Using node agent doesn't add an arg to the server, it creates a separate DaemonSet to create node agent pods. So I think we can leave UploaderType out of our server args package completely.

controllers/nodeagent.go
```diff
@@ -187,38 +218,57 @@ func (r *DPAReconciler) buildResticDaemonset(dpa *oadpv1alpha1.DataProtectionApp
...
 
-	return r.customizeResticDaemonset(dpa, ds)
+	return r.customizeNodeAgentDaemonset(dpa, ds)
 }
 
-func (r *DPAReconciler) customizeResticDaemonset(dpa *oadpv1alpha1.DataProtectionApplication, ds *appsv1.DaemonSet) (*appsv1.DaemonSet, error) {
-	if dpa.Spec.Configuration.Restic == nil {
-		// if restic is not configured, therefore not enabled, return early.
+func (r *DPAReconciler) customizeNodeAgentDaemonset(dpa *oadpv1alpha1.DataProtectionApplication, ds *appsv1.DaemonSet) (*appsv1.DaemonSet, error) {
+	if dpa.Spec.Configuration == nil || (dpa.Spec.Configuration.Restic == nil && dpa.Spec.Configuration.NodeAgent == nil) {
```
shubham-pampattiwar, line 231: Shouldn't this be `dpa.Spec.Configuration == nil || (dpa.Spec.Configuration.Restic == nil || dpa.Spec.Configuration.NodeAgent == nil)`

sseago: @shubham-pampattiwar Your test will always be true, since other validation prevents both Restic and NodeAgent from being configured. It's one or the other. So "no node agent" means that "either the whole configuration is nil or *both* restic and node agent are unconfigured"

sseago: So I think the above change is correct.

mpryc: It's correct. Applying De Morgan's law, we could have different:

```
A or B == not (A and B)
A and B == not A or not B
```

```go
if dpa.Spec.Configuration == nil || (dpa.Spec.Configuration.Restic == nil && dpa.Spec.Configuration.NodeAgent == nil) {}
```
equals to:

```go
if !(dpa.Spec.Configuration != nil && (dpa.Spec.Configuration.Restic != nil || dpa.Spec.Configuration.NodeAgent != nil)){}
```

In here we are checking if the following is not configured (bail out quick)
```
dpa.Spec.Configuration
```

Or if the `dpa.Spec.Configuration` is configured then we need to bail out when both `Restic` *and* `NodeAgent` are not configured, this means we don't apply logic.

Checking if both Restic and NodeAgent are present at the same time is done in other place with this PR:
https://github.com/openshift/oadp-operator/blob/7c07be18ba9ece757c00ff9c480a4c83e000a407/controllers/validator.go#L26-L28

controllers/validator.go
```diff
@@ -23,6 +23,10 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 		return false, errors.New("DPA CR Velero configuration cannot be nil")
 	}
 
+	if dpa.Spec.Configuration.Restic != nil && dpa.Spec.Configuration.NodeAgent != nil {
+		return false, errors.New("It's not allowed to have both 'restic' (deprecated in OADP 1.3) and 'nodeAgent' options simultaneously")
```
shubham-pampattiwar, line 27: ++ on the validation, NIT: maybe just re-frame the error message: " DPA CR cannot have `restic` as well as  `nodeAgent` enabled at the same time/simultaneously"

api/v1alpha1/oadp_types.go
mpryc: Fixed.

api/v1alpha1/oadp_types.go
mpryc: Fixed.

api/v1alpha1/oadp_types.go
```diff
@@ -130,26 +130,53 @@ type PodConfig struct {
...
+type ResticConfig struct {
+	// Embedding NodeAgentCommonFields
+	// +optional
+	NodeAgentCommonFields `json:",inline"`
+}
+
 // ApplicationConfig defines the configuration for the Data Protection Application
 type ApplicationConfig struct {
 	Velero *VeleroConfig `json:"velero,omitempty"`
+	// (deprecation warning) ResticConfig is the configuration for restic server.
```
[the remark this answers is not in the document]
mpryc: Changed.

api/v1alpha1/oadp_types.go
mpryc: Reverted that line change.

controllers/nodeagent.go
mpryc: Changed.

controllers/validator.go
mpryc: Changed to: 
"DPA CR cannot have restic (deprecated in OADP 1.3) as well as nodeAgent options at the same time"

I did not add "enable" because there is field `enabled` which may be misleading here, they can't be defined at all not only enabled: true.

controllers/nodeagent.go
kaovilai: > De Morgan's law

sending me back to college I see.

controllers/nodeagent.go
```diff
@@ -133,49 +156,57 @@ func (r *DPAReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error)
...
+
 	// get resource requirements for restic ds
 	// ignoring err here as it is checked in validator.go
-	resticResourceReqs, _ := getResticResourceReqs(dpa)
+	if dpa.Spec.Configuration.Restic != nil {
+		nodeAgentResourceReqs, _ = getResticResourceReqs(dpa)
+	} else if dpa.Spec.Configuration.NodeAgent != nil {
+		nodeAgentResourceReqs, _ = getNodeAgentResourceReqs(dpa)
+	} else {
+		return nil, fmt.Errorf("NodeAgent or Restic configuration cannot be nil")
```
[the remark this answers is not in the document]
mpryc: Not from the operator as it's enum of `kopia` or `restic` and is required field, however I was told that our structures and functions may be used (doesn't have to, just may be) by other operators or applications directly, without really interacting with CRDs and then I feel we have this non-harmful state here?

But if I am mistaken I am happy to remove it all together.

pkg/velero/server/config.go
mpryc: Commented out that part, so it's not relevant anymore.  See link on the commented out part.

build/Dockerfile.bundle
mpryc: @shubham-pampattiwar thanks for that, I've used above, however without first `make generate`.  I feel like `docs/developer/install_from_source.md` may need small tweaks.

pkg/velero/server/args.go
```diff
@@ -39,6 +39,17 @@ func (a Args) StringArr(dpaFeatureFlags []string, logLevel string) ([]string, er
...
+		// Commented out, please see:
+		//    https://github.com/openshift/oadp-operator/pull/1102#discussion_r1286054062
+		if a.UploaderType != "" {
+		uploaderType := strings.ToLower(a.UploaderType)
+		if uploaderType != "kopia" && uploaderType != "restic" {
+			return []string{}, fmt.Errorf("invalid uploader type, the supported values are 'restic' or 'kopia': %s", a.UploaderType)
+		}
+		args = append(args, fmt.Sprintf("--uploader-type=%s", uploaderType))
+	}
+	*/
```
sseago, line 52: Instead of commenting out, lets just remove these lines. I don't think we'll ever want to configure UploaderType here.

pkg/velero/server/config.go
```diff
@@ -81,8 +81,10 @@ type ServerConfig struct {
...
+	// Commented out, please see:
+	//    https://github.com/openshift/oadp-operator/pull/1102#discussion_r1286054062
+	// UploaderType string `json:"uploader-type,omitempty"`
```
sseago, line 87: Instead of commenting out, lets just remove these lines. I don't think we'll ever want to configure NodeAgent here.

sseago (changes requested):
Looks good overall. One additional thing -- for final merge, it's better to remove code we don't want than just commenting it out.

sseago (changes requested):

controllers/nodeagent.go
```diff
@@ -306,58 +366,58 @@ func (r *DPAReconciler) customizeResticDaemonset(dpa *oadpv1alpha1.DataProtectio
...
+	fsRestoreHelperCM := corev1.ConfigMap{
 		ObjectMeta: metav1.ObjectMeta{
-			Name:      ResticRestoreHelperCM,
+			Name:      FsRestoreHelperCM,
 			Namespace: r.NamespacedName.Namespace,
 		},
 	}
 
-	op, err := controllerutil.CreateOrPatch(r.Context, r.Client, &resticRestoreHelperCM, func() error {
+	op, err := controllerutil.CreateOrPatch(r.Context, r.Client, &fsRestoreHelperCM, func() error {
```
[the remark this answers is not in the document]
mpryc: Is it safe to remove the resticRestoreHelperCM ?

We have other options:
 1. Error (so user will know what's wrong)
 2. Re-label and give some warning, add some prefix to the labels so the user won't loose it
 3. Rename to fsRestoreHelper if fsRestoreHelper does not exists, otherwise option (1) or (2)
 4. Delete resticRestoreHelperCM if fsRestoreHelper is created or patched and have some warning

controllers/nodeagent.go
```diff
@@ -306,58 +366,58 @@ func (r *DPAReconciler) customizeResticDaemonset(dpa *oadpv1alpha1.DataProtectio
 	return ds, nil
 }
 
-func (r *DPAReconciler) ReconcileResticRestoreHelperConfig(log logr.Logger) (bool, error) {
+func (r *DPAReconciler) ReconcileFsRestoreHelperConfig(log logr.Logger) (bool, error) {
 	dpa := oadpv1alpha1.DataProtectionApplication{}
 	if err := r.Get(r.Context, r.NamespacedName, &dpa); err != nil {
 		return false, err
 	}
```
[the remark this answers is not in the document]
mpryc: @shubham-pampattiwar I don't understand this comment.

pkg/velero/server/config.go
mpryc: removed.

pkg/velero/server/args.go
mpryc: removed.

controllers/nodeagent.go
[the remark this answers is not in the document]
mpryc: I am going for an implementation that we talked over with @shawn-hurley. @sseago please let me know if that should be modified:

1. Check if the ConfigMap `restic-restore-action-config` exists.
2. If it exists check if it was created by OADP (by comparing the ConfigMap values to the one OADP had).
3a. If the ConfigMap was unmodified, we can delete it and create `fs-restore-action-config` (with of course event&warning on the DPA)
3b. If the ConfigMap `restic-restore-action-config` was modified, it means we should not delete it. Instead modify the `restic-restore-action-config` CM by adding some `-deprecated` suffix to the label, so velero won't find it and migrate this ConfigMap to the `fs-restore-action-config` with exactly same values (again event & warning on the DPA).

That will also modify a bit our design doc, so users know what to expect and also we deprecate that CM in the 1.4.

controllers/nodeagent.go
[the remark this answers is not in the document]
shubham-pampattiwar: @mpryc Velero uses labels to find that restore-helper-config-map, so it does not matter what the name of the config map is

controllers/nodeagent.go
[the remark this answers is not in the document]
sseago: @mpryc We shouldn't error, since the old configmap *will* exist on upgrade. It's not user-configurable, OADP created it, and OADP should remove it on upgrade. Currently this configmap is not user-configurable -- oadp sets it, so it should be sufficient to simply remove the old one and create the new one.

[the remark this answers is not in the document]
sseago: Currently if user modifies the configmap, the modifications are completely overwritten by the operator. We may need to make this configurable in a future enhancement, but we don't need to worry about custom changes for now, since they won't exist.

controllers/nodeagent.go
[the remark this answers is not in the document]
sseago: @shubham-pampattiwar While the name doesn't matter to velero, users upgrading from 1.2 will have two matching configmaps with the current impl, one with the old 1.2 image. So I think we need to make sure that the operator completely removes the old configmap when creating/updating the new one.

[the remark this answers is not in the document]
shubham-pampattiwar: yes makes sense, lets delete the old cm and create a new cm with a different name in OADP 1.3

controllers/nodeagent.go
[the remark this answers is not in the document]
mpryc: Ok removed in the last revision.

pkg/velero/server/args.go
mpryc: Removed that part as requested.

controllers/nodeagent.go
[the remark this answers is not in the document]
mpryc: added deletion logic

build/Dockerfile.bundle
mpryc: done. not updating this part in this PR.

controllers/velero.go
```diff
@@ -425,12 +437,8 @@ func (r *DPAReconciler) customizeVeleroContainer(dpa *oadpv1alpha1.DataProtectio
 	}
 
 	// Enable user to specify --fs-backup-timeout (defaults to 1h)
-	fsBackupTimeout := "1h"
-	if dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
-		fsBackupTimeout = dpa.Spec.Configuration.Restic.Timeout
-	}
-	// Append restic timeout option manually. Not configurable via install package, missing from podTemplateConfig struct. See: https://github.com/vmware-tanzu/velero/blob/8d57215ded1aa91cdea2cf091d60e072ce3f340f/pkg/install/deployment.go#L34-L45
-	veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--fs-backup-timeout=%s", fsBackupTimeout))
+	// Append FS timeout option manually. Not configurable via install package, missing from podTemplateConfig struct. See: https://github.com/vmware-tanzu/velero/blob/8d57215ded1aa91cdea2cf091d60e072ce3f340f/pkg/install/deployment.go#L34-L45
```
kaovilai, line 440: Filed issue: https://github.com/vmware-tanzu/velero/issues/6700

kaovilai (approved):

sseago (approved):

shubham-pampattiwar (approved):

PR #1103, opened 2023-07-25
Add .dockerignore

kaovilai:
Speed up Dockerfile builds by not allowing non code changes to invalidate build cache when `COPY ./ .`
Also reduces image size considerably.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,18 @@
+*.md
+docs
+blogs
+bin
+.vscode
+.github
+.devcontainer
+build
+cache
+.ci-operator.yaml
+.gitignore
+./*.yml
+./*.yaml
+OWNERS*
+PROJECT
+LICENSE
+Dockerfile
+!/docs/scripts/dc-restic-post-restore.sh # needed for `func RunResticPostRestoreScript` in tests/e2e/lib/velero_helpers.go


sseago:
@kaovilai actually, instead of cherry-picking to velero-datamover, we should just rebase the branch to master periodically, otherwise we'll get merge commits and such when we create the PR back to master from the topic branch.

kaovilai:
> otherwise we'll get merge commits and such

I think squash and merge should avoid that.

sseago:
But still, I think, from a workflow point of view, having to cp every change to master onto topic branches is undesirable. It clutters up the topic branch commit history. The usual practice with this sort of thing is to rebase regularly from master.

weshayutin (approved):
Thank you!

mpryc (commented):
Why is this required?

Dockerfile copies only subset of those:
```
COPY go.mod go.mod
COPY go.sum go.sum
# cache deps before building and copying source so that we don't need to re-download as much
# and so that source changes don't invalidate our downloaded layer
RUN go mod download

# Copy the go source
COPY main.go main.go
COPY api/ api/
COPY pkg/ pkg/
COPY controllers/ controllers/
```

So there is no `COPY . .`

Also why tests hack and other things are still there ?

shubham-pampattiwar (approved):

kaovilai:
re:
> So there is no COPY . .

https://github.com/openshift/oadp-operator/blob/05cd5496225de88cae71a460cb5b489eb5f9a455/build/ci-Dockerfile#L6

> why tests hack and other things are still there ?

`hack/` is tiny, no images; Can include here if needed.

Mostly trying to make CI faster (at least local development of the ci-Dockerfile)

mpryc:
> re:
> 
> > So there is no COPY . .
> 
> https://github.com/openshift/oadp-operator/blob/05cd5496225de88cae71a460cb5b489eb5f9a455/build/ci-Dockerfile#L6
> 
> > why tests hack and other things are still there ?
> 
> `hack/` is tiny, no images; Can include here if needed.
> 
> Mostly trying to make CI faster (at least local development of the ci-Dockerfile)

Ah ok I have not noticed that ci-Dockerfile. That makes sense. Thanks for clarification.

PR #1104, opened 2023-07-25
Refactor shell code to Go

mateusoliveira43:
## Description
Transform bash script into Go. The best things seems to be performance improvement, maintainability/readability and args and flags validation.

## How to test
Run the script with `go run docs/examples/getDataMoverResources.go [flags]` and compare with the bash one.


docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,217 @@
...
+	// TODO should we ignore errors?
+	if err != nil {
+		fmt.Print(string(stdout))
+		// TODO can't we get the return code from the command run?
+		os.Exit(1)
+	}
+
+	return string(stdout)
+}
+func captureShellOutputA(command []string) string {
```
mateusoliveira43, line 57: delete this func

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,217 @@
...
+	for _, line := range array {
+		if strings.Contains(line, match) {
+			matches = matches + line
+		}
+	}
+	return matches
+}
+
+func rootRun(cmd *cobra.Command, args []string) {
+	// this checks slow the script
```
mateusoliveira43, line 97: I wanted to add validations, but the way the script is intended to run I think it is not good. With `watch`, it would run this every iteration. Better to remove, right?

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,217 @@
...
+	// for i in $(oc -n openshift-adp get replicationdestination -A -o=custom-columns=:.metadata.name); do
+	//     oc -n openshift-adp get replicationdestination $i -o jsonpath='{"Name: "}{@.metadata.name}{" SyncDuration: "}{@.status.lastSyncDuration}{" sourcePVC: " }{@.spec.sourcePVC}' | column -t
+	// done
+}
+
+func init() {
+	rootCmd.Flags().BoolVarP(&backup, "backup", "b", false, "Check a backup")
+	rootCmd.Flags().BoolVarP(&restore, "restore", "r", false, "Check a restore")
+	rootCmd.MarkFlagsMutuallyExclusive("backup", "restore")
+	// https://github.com/spf13/cobra/pull/1952
```
mateusoliveira43, line 204: backup and restore should be required, but there is not this feature in cobra yet (but it is already merged)

docs/examples/getDataMoverResources.go
mateusoliveira43, line 1: I think my biggest question is how top run the script. I saw this discussion https://gist.github.com/posener/73ffd326d88483df6b1cb66e8ed1e0bd

what you suggest?

mateusoliveira43:
@weshayutin can you check if I am going in the right direction? `-b` should be already working, the others ones should be simple to implement, just want to check first

mpryc (commented):
@mateusoliveira43 my general question around this script is if you have considered using `go` libraries to access the k8s / OCP cluster instead of running shell that invokes `oc` command?

Here are few examples:
  https://github.com/kubernetes/client-go/tree/master/examples

mateusoliveira43:
@mpryc did not thought about that. I will take a look and test out

docs/examples/getDataMoverResources.go
kaovilai:
```
❯ go run docs/examples/getDataMoverResources.go --help
Check the OADP DataMover CR's in real time.

You need to log in to your cluster and deploy OADP prior to run this script

Usage:
  getDataMoverResources [flags]

Examples:
go run docs/examples/getDataMoverResources.go -b

Flags:
  -b, --backup    Check a backup
  -d, --details   Print a list of the relevant backup/restore CR's
  -h, --help      help for getDataMoverResources
  -r, --restore   Check a restore

```

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,315 @@
...
+	exitOnError(err)
+}
+
+func getKubernetesResourceList(resourceSpecification schema.GroupVersionResource, namespace string) *unstructured.UnstructuredList {
+	resource, err := kubernetesClient.Resource(resourceSpecification).Namespace(namespace).List(context.TODO(), metav1.ListOptions{})
+	exitOnError(err)
+	return resource
+}
+
+func backupSummaryKubeLib(details bool) {
```
mateusoliveira43, line 90: tested @mpryc suggestion. I liked :smile:  what you prefer so I can apply to the remaining: shell or Kubernetes lib?

kaovilai: go lib would be more reliable but adds go dependency that could come back to haunt us.

Don't go depend on something not meant for consumption like [k8s.io/kubectl/pkg/cmd/cp](https://gitlab.cee.redhat.com/app-mig/oadp-e2e-qe/-/commit/1281a66f866bf39b2a85b450302af5f7c7e1c374#a1f4f20d8f21da3825a49a53c8146f0230a76a67_39_36) tho :P

mateusoliveira43: yeah, that is a valid concern. Even without Kubernetes lib, we are using cobra, which is not direct as well

since Go does not have development dependencies, I think the best solution would be a separate repo and use [Makefile go-install-tool](https://github.com/openshift/oadp-operator/blob/master/Makefile#L284-L296) (which can become 2 repos changes to achieve a simple change in the future). Do you suggest another solution to overcome Go dependency problem?

kaovilai: I mean you could still consume kubectl via go like in https://gitlab.cee.redhat.com/app-mig/oadp-e2e-qe/-/blob/master/utils/pods/kube_exec.go#L37

Just consume the top main cobra (kubectl) command, rather than digging into subcommand for example.

I'm not as worried about client-go lib as we import that already.

mateusoliveira43: I think I still did not understand your suggestion @kaovilai 

do you suggest to have a function like this?
```go
func kubectl(args ...string) string {
  kubectl := cmd.NewDefaultKubectlCommand()
  kubectl.SetArgs(args...)
  outputBuffer := new(bytes.Buffer)
  kubectl.SetOut(outputBuffer)
  err := kubectl.Execute()
  exitOnError(err)
  return outputBuffer.String()
}

kubectl([]string{"get", "volumesnapshot", "-A"})
```

kaovilai: Correct.. if you are to do something with kubectl, that would be easier/faster to consume and reliable (shell independent).

kaovilai: if you can use client-go instead, that would be preferable but otherwise it's fine.
https://stackoverflow.com/questions/43314689/example-of-exec-in-k8ss-pod-by-using-go-client

kaovilai: oc get volumesnapshots -A should definitely use client-go, you can call length on the result easily.

mateusoliveira43: alright, thanks! going to study it and make the required changes

mateusoliveira43: used go client for everything :) resolving conversation, but feel free to open it again if you see any problems

docs/examples/datamover_resources.sh
mateusoliveira43, line 1: TODO run grep file_name and see if any place needs changes

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,318 @@
...
+	exitOnError(err)
+	return resource.Items
+}
+
+func getPodForDeployment(deployment *appsV1.Deployment) coreV1.Pod {
+	listOptions := metaV1.ListOptions{LabelSelector: labels.Set(deployment.Spec.Selector.MatchLabels).AsSelector().String()}
+	pods, err := KubernetesClient.CoreV1().Pods(namespace).List(context.TODO(), listOptions)
+	exitOnError(err)
+	if len(pods.Items) > 1 {
+		fmt.Printf("More then one Pod (%v) for deployment %v\n", pods.Items, deployment)
```
mateusoliveira43, line 141: typo then -> then

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,318 @@
...
+	batchingCompleted := 0
+	batchingProcessing := 0
+	batchingQueued := 0
+	increaseBatching := map[string]func(){
+		"Completed":  func() { batchingCompleted += 1 },
+		"Processing": func() { batchingProcessing += 1 },
+		"Queued":     func() { batchingQueued += 1 },
+	}
+	for _, volumeSnapshotBackup := range volumeSnapshotBackups {
+		// TODO there is an easier way to access the properties?
```
mateusoliveira43, line 204: is there a less verbose way of getting properties from object?

docs/examples/datamover_resources.sh
mateusoliveira43: found a problem: how to replace this https://github.com/openshift/oadp-operator/blob/master/docs/oadp_cheat_sheet.md?plain=1#L230 ?

is there something shorter than this?
```sh
git clone --depth=1 git@github.com:openshift/oadp-operator.git
go build -C oadp-operator docs/examples/getDataMoverResources.go
mv oadp-operator/getDataMoverResources ~/.local/bin/getDataMoverResources
rm -r oadp-operator
```

do I need to change anything in PROW? I thing I remember a regex for this file

docs/examples/getDataMoverResources.go
kaovilai: what do you mean less verbose?

mateusoliveira43: something like `volumeSnapshotBackup.Object["status"]["phase"]` instead of `volumeSnapshotBackup.Object["status"].(map[string]interface{})["phase"]`

docs/examples/datamover_resources.sh
mateusoliveira43: wait this PR https://github.com/openshift/oadp-operator/pull/1103 before merging this one?

weshayutin:
@mateusoliveira43 let's wrap this up vs. letting it sit if you have some time.  Ready for testing?

mateusoliveira43:
Yeah, already ready @weshayutin 
I will just fix conflict during the day

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,319 @@
...
+		Resource: "replicationsources",
+	}
+	ReplicationDestinationResource = schema.GroupVersionResource{
+		Group:    "volsync.backube",
+		Version:  "v1alpha1",
+		Resource: "replicationdestinations",
+	}
+)
+
+// TODO Does this work with built in DataMover?
```
mateusoliveira43, line 73: will test in my upcoming tasks

mateusoliveira43: tested

it works (it sees volumeSnapshotContents) but could be better

should I 
- merge like this (focused in OLD DataMover)
- add NEW DataMover stuff (DataUpload and DataDownload) and add new flag to select which to use
- remove OLD DataMover stuff (VSM and Volsync), and add NEW DataMover stuff
?

shubham-pampattiwar:
@mateusoliveira43 Currently this PR carters to OADP 1.2 datamover (Volsync + VSM based) , as we are moving away from this datamover to support Velero built-in datamover in OADP 1.3, wouldn't it make sense to make this work with the velero built-in datamover ?

weshayutin:
@mateusoliveira43 might as well create a flag for 1.2 vs. 1.3.  In 1.3. track DUCR's and DDCR's.

docs/examples/getDataMoverResources.go
```diff
@@ -0,0 +1,407 @@
...
+
+func init() {
+	cli.Flags().BoolVarP(&backup, "backup", "b", false, "Check a backup")
+	cli.Flags().BoolVarP(&restore, "restore", "r", false, "Check a restore")
+	cli.MarkFlagsMutuallyExclusive("backup", "restore")
+	// TODO https://github.com/spf13/cobra/pull/1952
+	// cli.MarkFlagsOneRequired("backup", "restore")
+	cli.Flags().BoolVarP(&details, "details", "d", false, "Print a list of the relevant backup/restore CR's")
+	cli.Flags().StringVarP(&namespace, "namespace", "n", "openshift-adp", "Namespace OADP was deployed")
+	cli.Flags().BoolVarP(&builtInDataMover, "velero", "v", true, "Check for Velero Built-in DataMover Resources, if true, otherwise, check for VSM/Volsync DataMover Resources")
```
mateusoliveira43, line 130: Added flag to select between Velero Built-in DataMover and VSM/Volsync DataMover :smile:

docs/examples/debug.md
```diff
@@ -1,11 +1,11 @@
 # Potentially helpful debug commands
 
-## watch datamover backup resources live
+## watch DataMover backup resources live
 ```
-watch -n 10 ./datamover_resources.sh -b -d 
+watch -n 10 go run getDataMoverResources.go -b -d
```
kaovilai, line 5: btw `go run` is just a quick and dirty way to run from go code directly.

For a published code, it would be good to `go build getDataMoverResources.go -o getDataMoverResources` then use compiled binary for faster speed.
```
./getDataMoverResources
```

mateusoliveira43: oh yeah, forget that, I'm used to Python ehhehe Going to add in the doc to user to compile first (and also tell about flag to switch between 1.2 and 1.3 datamover)

mrnold:
Who is the intended audience for this section of docs? This change is fine if it's meant for OADP developers, but are we also presenting these examples to users or product support personnel anywhere? I don't think the second group will want to have to compile a go program instead of copying a shell script.

mrnold (approved):

mateusoliveira43:
/hold

We will wait to have a easy way to users to run the script (without the need of having Go in the machines)

kaovilai:
> We will wait to have a easy way to users to run the script (without the need of having Go in the machines)

We can prebuilt the binaries for download.

mateusoliveira43:
Yeah, I think we also brainstormed container images, @kaovilai 

But will wait a future PR that Michal is working on, to see if there is an easier way

mateusoliveira43:
gave up on this

PR #1106, opened 2023-08-01
OADP-2112: Support the standardized STS configuration flow via OLM and CCO for OADP

shubham-pampattiwar:
This PR adds support for the standardized STS configuration flow via OLM and CCO.
In order to test the changes:
- Deploy a ROSA STS cluster (OCP 4.14+) by following the instructions given [here](https://mobb.ninja/docs/rosa/sts/)
- Create a role ARN by followingthe steps 1-3 in the "Getting Started" section of this [document](https://mobb.ninja/docs/misc/oadp/rosa-sts/)
- Now checkout this PR
- Build the operator image and push it to quay/docker
- Update the operator CSV in bundle to use this image
- Now follow the below instructions to bundle, index and catalogsource of the operator:
    - `export ORG=<YOUR_ORG>`
    - `export TAG=<YOUR_TAG>`
    - `podman build -f bundle.Dockerfile -t quay.io/$ORG/oadp-operator-bundle:$TAG`
    - `podman push quay.io/$ORG/oadp-operator-bundle:$TAG`
    - `opm index add --container-tool podman --bundles quay.io/$ORG/oadp-operator-bundle:$TAG --tag quay.io/$ORG/oadp-operator-index:$TAG`
    - `podman push quay.io/$ORG/oadp-operator-index:$TAG`
    - Now create a catalogsource using the operator index image:
```
apiVersion: operators.coreos.com/v1alpha1
kind: CatalogSource
metadata:
  name: oadp-operator-sts-catalog
  namespace: openshift-marketplace
spec:
  sourceType: grpc
  image: <YOUR-INDEX-IMAGE>
```
- Now from operatorhub install OADP operator created from our catalogsource
- During installation supply the role ARN created earlier
- Be sure to select upgrade strategy as `Manual` during installation
- Now create the cloudStorage CR and the DPA CR (follow the ROSA STS document: https://mobb.ninja/docs/misc/oadp/rosa-sts/)
- Finally you are all set to use OADP Operator for Backup/Restore operations :)

Related Enhancement link: [Tokenized Cloud Auth Enablement For Operators](https://github.com/openshift/enhancements/pull/1339)

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -484,10 +484,13 @@ spec:
     support snapshots, OADP backs up resources and PV data with Restic.
 
     * [Installing OADP for application backup and restore][install-doc]
+    
+    * [Installing OADP on a ROSA cluster and using STS, please follow the Getting Started Steps 1-3 in order to obtain the role ARN needed for using the standardized STS configuration flow via OLM][rosa-sts-doc]
 
     * [Frequently Asked Questions][faq-doc]
 
     [install-doc]: https://docs.openshift.com/container-platform/latest/backup_and_restore/application_backup_and_restore/installing/about-installing-oadp.html
+    [rosa-sts-doc]: https://mobb.ninja/docs/misc/oadp/rosa-sts/
     [faq-doc]: https://access.redhat.com/articles/5456281
   displayName: OADP Operator
   icon:
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -2,6 +2,7 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   annotations:
+    features.operators.openshift.io/token-auth-aws: "true"
     alm-examples: |-
       [
         {
@@ -496,6 +497,18 @@ spec:
     spec:
       clusterPermissions:
       - rules:
+        - apiGroups:
+          - apiextensions.k8s.io
+          resources:
+          - customresourcedefinitions
+          verbs:
+          - '*'
+        - apiGroups:
+          - cloudcredential.openshift.io
+          resources:
+          - '*'
+          verbs:
+          - '*'
         - apiGroups:
           - oadp.openshift.io
           resources:
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -501,17 +501,11 @@ spec:
       clusterPermissions:
       - rules:
         - apiGroups:
-          - apiextensions.k8s.io
-          resources:
-          - customresourcedefinitions
-          verbs:
-          - '*'
-        - apiGroups:
           - cloudcredential.openshift.io
           resources:
-          - '*'
+          - credentialsrequests
           verbs:
-          - '*'
+          - create
         - apiGroups:
           - oadp.openshift.io
           resources:
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -2,7 +2,6 @@ apiVersion: operators.coreos.com/v1alpha1
 kind: ClusterServiceVersion
 metadata:
   annotations:
-    features.operators.openshift.io/token-auth-aws: "true"
     alm-examples: |-
       [
         {
@@ -156,6 +155,7 @@ metadata:
     createdAt: "2020-09-08T12:21:00Z"
     description: OADP (OpenShift API for Data Protection) operator sets up and installs
       Data Protection Applications on the OpenShift platform.
+    features.operators.openshift.io/token-auth-aws: "true"
     olm.skipRange: '>=0.0.0 <99.0.0'
     operatorframework.io/suggested-namespace: openshift-adp
     operators.openshift.io/infrastructure-features: '["Disconnected"]'
@@ -484,7 +484,7 @@ spec:
     support snapshots, OADP backs up resources and PV data with Restic.
 
     * [Installing OADP for application backup and restore][install-doc]
-    
+
     * [Installing OADP on a ROSA cluster and using STS, please follow the Getting Started Steps 1-3 in order to obtain the role ARN needed for using the standardized STS configuration flow via OLM][rosa-sts-doc]
 
     * [Frequently Asked Questions][faq-doc]
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,6 +7,18 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
+  - apiextensions.k8s.io
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - '*'
+- apiGroups:
+  - cloudcredential.openshift.io
+  resources:
+  - '*'
+  verbs:
+  - '*'
+- apiGroups:
   - oadp.openshift.io
   resources:
   - '*'
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -7,17 +7,11 @@ metadata:
   name: manager-role
 rules:
 - apiGroups:
-  - apiextensions.k8s.io
-  resources:
-  - customresourcedefinitions
-  verbs:
-  - '*'
-- apiGroups:
   - cloudcredential.openshift.io
   resources:
-  - '*'
+  - credentialsrequests
   verbs:
-  - '*'
+  - create
 - apiGroups:
   - oadp.openshift.io
   resources:
--- a/controllers/bucket_controller.go
+++ b/controllers/bucket_controller.go
@@ -119,7 +119,7 @@ func (b BucketReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	if common.CCOWorkflow() {
 		// wait for the credential request to be processed and the secret to be created
 		installNS := os.Getenv("WATCH_NAMESPACE")
-		_, err := b.WaitForSecret(installNS, "cloud-credentials")
+		_, err := b.WaitForSecret(installNS, VeleroAWSSecretName)
 		if err != nil {
 			log.Error(err, "unable to fetch secert created by CCO")
 			return result, err
--- a/controllers/bucket_controller.go
+++ b/controllers/bucket_controller.go
@@ -119,7 +119,7 @@ func (b BucketReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	if common.CCOWorkflow() {
 		// wait for the credential request to be processed and the secret to be created
 		installNS := os.Getenv("WATCH_NAMESPACE")
-		_, err := b.WaitForSecret(installNS, VeleroAWSSecretName)
+		_, err = b.WaitForSecret(installNS, VeleroAWSSecretName)
 		if err != nil {
 			log.Error(err, "unable to fetch secert created by CCO")
 			return result, err
@@ -209,11 +209,12 @@ func (b *BucketReconciler) WaitForSecret(namespace, name string) (*corev1.Secret
 	timeout := 10 * time.Minute
 
 	secret := corev1.Secret{}
-	wait.PollImmediate(5*time.Second, timeout, func() (bool, error) {
-		key := types.NamespacedName{
-			Name:      name,
-			Namespace: namespace,
-		}
+	key := types.NamespacedName{
+		Name:      name,
+		Namespace: namespace,
+	}
+
+	err := wait.PollImmediate(5*time.Second, timeout, func() (bool, error) {
 
 		err := b.Client.Get(context.Background(), key, &secret)
 		if err != nil {
@@ -223,6 +224,10 @@ func (b *BucketReconciler) WaitForSecret(namespace, name string) (*corev1.Secret
 		}
 		return true, nil
 	})
+
+	if err != nil {
+		return nil, err
+	}
 
 	return &secret, nil
 }
--- a/controllers/bucket_controller.go
+++ b/controllers/bucket_controller.go
@@ -118,6 +118,7 @@ func (b BucketReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 	var ok bool
 	if common.CCOWorkflow() {
 		// wait for the credential request to be processed and the secret to be created
+		log.Info(fmt.Sprintf("Following standardized STS workflow, waiting for for the credential request to be processed and provision the secret"))
 		installNS := os.Getenv("WATCH_NAMESPACE")
 		_, err = b.WaitForSecret(installNS, VeleroAWSSecretName)
 		if err != nil {
--- a/controllers/bucket_controller.go
+++ b/controllers/bucket_controller.go
@@ -3,6 +3,11 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"github.com/openshift/oadp-operator/pkg/common"
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/wait"
+	"os"
 	"strconv"
 	"time"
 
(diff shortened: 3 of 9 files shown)


controllers/bucket_controller.go
```diff
@@ -188,3 +208,61 @@ func removeKey(slice []string, s string) []string {
 	}
 	return slice
 }
+
+func (b *BucketReconciler) WaitForSecret(namespace, name string) (*corev1.Secret, error) {
```
shawn-hurley, line 212: Hate to be that guy, but can you use something from: https://pkg.go.dev/k8s.io/apimachinery/pkg/util/wait to do this instead?

controllers/bucket_controller.go
```diff
@@ -188,3 +208,61 @@ func removeKey(slice []string, s string) []string {
...
+				// secret is found, return it
+				return &secret, nil
+			}
+		}
+	}
+}
+
+func (b *BucketReconciler) SharedCredentialsFileFromSecret(secret *corev1.Secret) (string, error) {
+	var data []byte
+	switch {
```
shawn-hurley, line 252: nit: if len is <= 0 error, and then use secret.Data["credentials"] instead of using a new byte array. I guess the bigger nit is I would not use a switch statement for this but this is a nit, the logic part is sound.

main.go
```diff
@@ -88,6 +91,52 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Get STS vars
+	// ROLEARN env var is set via operator subscription
+	roleARN := os.Getenv("ROLEARN")
+	setupLog.Info("getting role ARN", "role ARN =", roleARN)
+	// webIdentityTokenPath mount present on operator CSV
+	webIdentityTokenPath := "/var/run/secrets/openshift/serviceaccount/token"
```
shawn-hurley, line 99: Should this be a constant?

main.go
```diff
@@ -166,6 +215,15 @@ func main() {
 		setupLog.Error(err, "problem running manager")
 		os.Exit(1)
 	}
+
+	// create cred request
+	c := mgr.GetClient()
+	if err := c.Create(context.TODO(), credReq); err != nil {
```
shawn-hurley, line 221: Is it potentially the case this does not exist as an API for specific installs of openshift?

I am thinking about some of the edge scenarios, where we have way fewer operators installed AFAICT.

shawn-hurley (changes requested):
Most of the comments are nits, but I do think we need to make sure that we are able to use the minter API on all form factors.

main.go
```diff
@@ -166,6 +218,24 @@ func main() {
...
+	if err != nil {
+		setupLog.Error(err, "problem checking the existence of CredentialRequests CRD")
+		os.Exit(1)
+	}
+
+	if credReqCRDExists {
+		// create cred request
+		if err := c.Create(context.TODO(), credReq); err != nil {
+			if !errors.IsAlreadyExists(err) {
+				setupLog.Error(err, "unable to create credRequest")
```
kaovilai, line 234: This might be more reliable?
```
❯ oc describe co cloud-credential
Name:         cloud-credential
Namespace:    
Labels:       <none>
Annotations:  exclude.release.openshift.io/internal-openshift-hosted: true
              include.release.openshift.io/self-managed-high-availability: true
API Version:  config.openshift.io/v1
Kind:         ClusterOperator
Spec:
```

controllers/bucket_controller.go
shubham-pampattiwar: Done

main.go
shubham-pampattiwar: Done

controllers/bucket_controller.go
shubham-pampattiwar: Done

main.go
shubham-pampattiwar: Done

shubham-pampattiwar:
Updated PR to use raw yam for cred request creation. This helped in removal of extra deps that were pulled in because of the cloud credential request operator API import. 
Operator is getting built without any issues, installation is failing though with the error:
`the InstallPlan contains invalid data`

shubham-pampattiwar:
Removed usage of external yaml for credentialsrequest CR, operator bundle does not support this. Operator is getting built but its breaking for `credentialsrequest` create call:
```
2023-08-29T03:08:14Z    ERROR   setup   error aws credentials request   {"error": "CredentialsRequest.cloudcredential.openshift.io is forbidden: User \"system:serviceaccount:openshift-adp:openshift-adp-controller-manager\" cannot create resource \"CredentialsRequest\" in API group \"cloudcredential.openshift.io\" at the cluster scope"}
```
Added permissions in operator clusterrole as well as csv but still getting this error, needs more investigation.

controllers/bucket_controller.go
```diff
@@ -188,3 +209,42 @@ func removeKey(slice []string, s string) []string {
...
+			}
+		}
+		return true, nil
+	})
+
+	return &secret, nil
+}
+
+func (b *BucketReconciler) SharedCredentialsFileFromSecret(secret *corev1.Secret) (string, error) {
+	if len(secret.Data["credentials"]) < 0 {
```
mateusoliveira43, line 237: Can this be triggered? is not zero the lowest value possible of `len` function?

shubham-pampattiwar:
Automatic bucket creation successful:
```

1.6941277256321726e+09  DEBUG   events  Normal  {"object": {"kind":"CloudStorage","namespace":"openshift-adp","name":"example-44","uid":"5ba748cf-9240-4525-820c-2aab74a498a6","apiVersion":"oadp.openshift.io/v1alpha1","resourceVersion":"229836"}, "reason": "BucketCreated", "message": "bucket shubham-rocks has been created"}
```
BSL in available phase:
```
shubham@fedora ~/oadp-operator -  (aws-sts) $ velero get backup-location
NAME       PROVIDER   BUCKET/PREFIX   PHASE       LAST VALIDATED                  ACCESS MODE   DEFAULT
ts-dpa-1   aws        shubham-rocks   Available   2023-09-07 16:14:33 -0700 PDT   ReadWrite     true
```

shubham-pampattiwar:
Added test instructions in PR description.

tests/e2e/sample-applications/nginx/nginx-stateless-backup.yaml
```diff
@@ -6,5 +6,5 @@ metadata:
 spec:
   includedNamespaces:
   - nginx-example
-  storageLocation: dpa-sample-1
+  storageLocation: ts-dpa-1
```
mateusoliveira43, line 9: here why changed name?

main.go
```diff
@@ -88,6 +97,28 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Get STS vars
+	// ROLEARN env var is set via operator subscription
+	roleARN := os.Getenv("ROLEARN")
```
mateusoliveira43, line 102: should we check if ENV VAR is set or it will be always present?

main.go
```diff
@@ -218,3 +249,85 @@ func addPodSecurityPrivilegedLabels(watchNamespaceName string) error {
...
+
+// CreateCredRequest WITP : WebIdentityTokenPath
+func CreateCredRequest(roleARN string, WITP string, secretNS string) error {
+	cfg := config.GetConfigOrDie()
+	client, err := client2.New(cfg, client2.Options{})
+	if err != nil {
+		setupLog.Error(err, "unable to create client")
+	}
+
+	credRequest := &unstructured.Unstructured{
```
mateusoliveira43, line 292: could we import it from here https://pkg.go.dev/github.com/openshift/cloud-credential-operator/pkg/apis/cloudcredential/v1#CredentialsRequest or adds to much size to OADP?

tests/e2e/sample-applications/nginx/nginx-stateless-backup.yaml
shubham-pampattiwar: Done, updated

main.go
```diff
@@ -21,9 +21,15 @@ import (
...
 	"fmt"
 	monitor "github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring/v1"
+	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/kubernetes"
 	"os"
+	client2 "sigs.k8s.io/controller-runtime/pkg/client"
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated

main.go
shubham-pampattiwar: Done, updated

main.go
shubham-pampattiwar: So initially that is what I did but that added some additional updates to go.mod and started resulting in build failures (incompatible versions with velero deps), hence going via the unstructured route. Context comment: https://github.com/openshift/oadp-operator/pull/1106#issuecomment-1678186905

mateusoliveira43: alright, thanks for the info!

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -496,6 +500,18 @@ spec:
     spec:
       clusterPermissions:
       - rules:
+        - apiGroups:
+          - apiextensions.k8s.io
+          resources:
+          - customresourcedefinitions
+          verbs:
+          - '*'
```
shawn-hurley, line 508: I worry about this permission. Can we reduce it in some way? Maybe fewer verbs and specific resources?

config/rbac/role.yaml
```diff
@@ -6,6 +6,18 @@ metadata:
...
+  resources:
+  - customresourcedefinitions
+  verbs:
+  - '*'
+- apiGroups:
+  - cloudcredential.openshift.io
+  resources:
+  - '*'
+  verbs:
+  - '*'
```
shawn-hurley, line 20: Same here, I think that we really should only ask for GET and Update/Patch/Delete for our specific one? Can we have a standard name and we can reduce the scope?

main.go
```diff
@@ -88,6 +98,31 @@ func main() {
 		os.Exit(1)
 	}
 
+	// Get STS vars
+	// ROLEARN env var is set via operator subscription
+	var roleARN string
+	if common.CCOWorkflow() {
+		roleARN = os.Getenv("ROLEARN")
```
shawn-hurley, line 105: Do we want to add a check for empty role arn and expose that to the user for fixing? 

Is there a doc of where the CSV can be updated if you do mess this up when installing?

main.go
```diff
@@ -218,3 +253,85 @@ func addPodSecurityPrivilegedLabels(watchNamespaceName string) error {
 	}
 	return nil
 }
+
+func DoesCRDExist(CRDName string) (bool, error) {
```
shawn-hurley, line 257: Instead, can we use the discovery client to look for the cloud credential-specific resource? 

https://pkg.go.dev/k8s.io/client-go/discovery

shawn-hurley (commented):
Some questions and ideas just to reduce the permissions that this is asking for

kaovilai:
> Deploy a ROSA STS cluster (OCP 4.14+) by following the instructions given [here](https://mobb.ninja/docs/rosa/sts/)
Don't think that covers 4.14 yet :)

controllers/bucket_controller.go
```diff
@@ -111,6 +116,16 @@ func (b BucketReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl
 		}
 	}
 	var ok bool
+	if common.CCOWorkflow() {
+		// wait for the credential request to be processed and the secret to be created
+		installNS := os.Getenv("WATCH_NAMESPACE")
+		_, err := b.WaitForSecret(installNS, "cloud-credentials")
```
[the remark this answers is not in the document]
mateusoliveira43: It comes from here https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L39, right? Do not like this magic imports from Go :exploding_head:

main.go
shubham-pampattiwar: Role ARN is a compulsory field when the ROSA cluster is in  STS mode, you won't be able to install the operator without supplying a role arn during install time.

main.go
shubham-pampattiwar: Done, added a comment.

controllers/bucket_controller.go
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated the PR.

main.go
shubham-pampattiwar: Done, updated the PR.

config/rbac/role.yaml
shubham-pampattiwar: Done, reduced the scope and updated the perms.

bundle/manifests/oadp-operator.clusterserviceversion.yaml
shubham-pampattiwar: Done, reduced the scope and updated the perms.

controllers/bucket_controller.go
```diff
@@ -188,3 +203,26 @@ func removeKey(slice []string, s string) []string {
...
 	return slice
 }
+
+func (b *BucketReconciler) WaitForSecret(namespace, name string) (*corev1.Secret, error) {
+	// set a timeout of 10 minutes
+	timeout := 10 * time.Minute
+
+	secret := corev1.Secret{}
+	wait.PollImmediate(5*time.Second, timeout, func() (bool, error) {
+		key := types.NamespacedName{
```
mateusoliveira43, line 213: each 5 seconds this var is created? should we put it outside func body?

Should we check if `wait.PollImmediate` returned an error? The way it is, func always return `nil` as error, right? (or just return `*corev1.Secret` and not error as well?)

shubham-pampattiwar: Done, updated the PR.

main.go
shawn-hurley: Are there other situations where someone may have this cluster in STS mode, want to use this feature and not have this set? 

I wonder if it is worth the extra effort, just in case?

mateusoliveira43: The `common.CCOWorkflow` call checks if ENV VAR exists @shawn-hurley , so this only triggers if env var is set

was this your question?

shawn-hurley: It makes sense that we have empty roleARN sometimes. 

But there is also a distinct possibility that the cred-request will get triggered with an empty ARN:

https://github.com/openshift/oadp-operator/pull/1106/files/8f1850453cbdc824542133a6d0fdeddb0c10ed59#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R110-R124

I assume that this will lead to some issues. 

We need to make it clear a couple of things:

1. When a user should have put this value in here, but it is not, then we need to tell them. 
2. When a user should not have put this value in here, then we probably don't want to create the CredentialsRequerst correct?

shubham-pampattiwar: @shawn-hurley cred-request only gets created if the role arn is supplied https://github.com/openshift/oadp-operator/pull/1106/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R107-R127

shawn-hurley: something just changed, this was not the code I was looking at. why do you have this then?
https://github.com/openshift/oadp-operator/pull/1106/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R106

shubham-pampattiwar: yeah I can remove that var and update the PR, is there anything else that I need to update ?

mateusoliveira43 (approved):

main.go
shubham-pampattiwar: @shawn-hurley Done, added logs, updated the PR, PTAL thanks!

mateusoliveira43 (approved):

main.go
```diff
@@ -88,6 +101,33 @@ func main() {
...
+			os.Exit(1)
+		}
+
+		if credReqCRDExists {
+			// create cred request
+			setupLog.Info(fmt.Sprintf("Creating credentials request for role: %s, and WebIdentityTokenPath: %s", roleARN, WebIdentityTokenPath))
+			if err := CreateCredRequest(roleARN, WebIdentityTokenPath, watchNamespace); err != nil {
+				if !errors.IsAlreadyExists(err) {
+					setupLog.Error(err, "unable to create credRequest")
+					os.Exit(1)
```
kaovilai, line 125: We never really exit(1) in operators. Would this cause operator pod to crashloop backoff?

shubham-pampattiwar: yes we do exit(1) if the tasks are part of the operator init steps, yes this would cause the operator pod to crashloop backoff

shawn-hurley (approved):
looks good to me

kaovilai (approved):

Issue #1052, opened 2023-06-01
Design: TechPreview: Upstream datamover


weshayutin:
The context of this design is NOT if we should adopt or when, the context would be the technical requirements of including the native velero datamover in OADP.

Upstream Velero is releasing a native datamover in velero 1.12

https://github.com/vmware-tanzu/velero/blob/main/design/volume-snapshot-data-movement/volume-snapshot-data-movement.md
  
* DPA changes
  * pick a datamover  

* From Velero -  Remove OADP 1.2.0 datamover bits
* From OADP -  Remove OADP 1.2.0 datamover bits

* Need to compare OADP 1.2.0 data mover features / functions to native datamover
  *  address gaps
     * if native is missing a feature, create a design upstream
       *  oadp-1.2.0 backup and restore concurrency ( Scott ) 

* Design notes
  * We have one data mover at time??
    *  How is the OADP datamover integrated w/ upstream
      *   
  * Does our OADP datamover integrate into upstream datamover in 1.3.0

  To-do's
- [ ] From Velero - Remove OADP 1.2.0 datamover bits
- [ ] From OADP -  Remove OADP 1.2.0 datamover bits
- [ ] Remove validations for 1.2 DM
- [ ] Remove vsm controller
- [ ] Remove vsm-plugin
- [ ] remove 1.2 dpa parms, add built-in parms
- [ ] Rebase 
- [ ] Need to modify dm e2e test

TO-TEST
- [ ] provision eks in amazon,  
- [ ] do not add vsm plugin to dpa

To Evaluate
- [ ] basic backup / restore w/ csi volumes
- [ ] evaluate what artifacts are left on cluster ( hopefully none )
- [ ] various storage backends, aws, azure, odf, gcp 
- [ ] TTL, ensure velero artifacts are removed, and backups on s3
- [ ] Partner plugin - Shawn
- [ ] VM filesystem mode
- [ ] We suspect enabling kopia as fsb is not required.
- [ ] sts auth, wif auth

weshayutin:
related to #1057

shubham-pampattiwar:
Top of mind I can think of the following (if we want OADP datamover to co-exist with upstream datamover):
- DPA changes would be the configurations needed for DataUpload and DataDownload CRs 
- Integrating with Upstream pluggable datamover interface will involve:
    - Using DataUpload/DataDownload CRs to orchestrate Volsync processes instead of VSB/VSR.
[...]

PR #1107, opened 2023-08-01
Velero Built-in Data Mover design for OADP

sseago:
fixes https://github.com/openshift/oadp-operator/issues/1052

### DIFF
--- a/docs/design/upstream-datamover.md
+++ b/docs/design/upstream-datamover.md
@@ -1,0 +1,110 @@
+# Design for Integrating Velero Built-in Data Mover (VBDM) support into OADP
+
+
+## Abstract
+One of the primary new features of Velero 1.12 is the new Velero Built-in Data Mover (VBDM).
+The OADP Operator needs to add configuration (and e2e tests) to support this new datamover and remove any configuration and testing of the legacy OADP datamover.
+
+## Background
+Prior to the release of Velero 1.12, OADP 1.1 and 1.2 integrated an OADP-specific datamover based on Volsync into OADP and the OADP-installed Velero version.
+This took the form of some modifications to Velero core and the Velero CSI plugins, a new VSM plugin, and a separate VSM controller that used Volsync to store Volume data in the BackupStorageLocation.
+In addition to this, DPA and installation support for the OADP datamover was included in the OADP operator.
+
+In order to integrate VBDM into OADP, all of the above legacy OADP datamover parts will be removed from the OADP operator, the OADP Velero build, and the OADP build of the Velero Plugin for CSI.
+
+## Goals
+- Enable Velero Built-in Data Mover (VBDM) in OADP.
+- Remove support for the legacy Volsync-based OADP Data Mover.
+
+## Non Goals
+- Enabling restore of OADP 1.1 or 1.2 Data Mover backups in OADP with Velero Built-in Datamover enabled
+
+
+## High-Level Design
+Built-in Data Mover functionality will be available in Velero 1.12.
+This proposal will allow for enabling this new datamover for backup and restore in OADP.
+As part of implementation, any OADP support for the previous Volsync-based datamover will be removed.
+All OADP end-to-end tests for data movement will be updated to use the new Built-in Data Mover instead of the Volsync-based data mover.
+
+
+## Detailed Design
+
+### Volsync Datamover removal
+
+All Volsync Datamover-related code and configuration will be removed from oadp-operator.
+
+#### DPA CRD
+
+For the `DataProtectionApplication` CRD, remove the `DataMover` (and related) structs, including `RetainPolicy` and `VolumeOptions`.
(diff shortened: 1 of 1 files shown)


docs/design/upstream-datamover.md
```diff
@@ -0,0 +1,110 @@
...
+
+### Volsync Datamover removal
+
+All Volsync Datamover-related code and configuration will be removed from oadp-operator.
+
+#### DPA CRD
+
+For the `DataProtectionApplication` CRD, remove the `DataMover` (and related) structs, including `RetainPolicy` and `VolumeOptions`.
+Specifically, `DataMover` will be removed from the `Features` struct which is included in DPA spec.
+`Features` will be empty for now, but available for future tech preview features that need explicit configuration here.
```
shubham-pampattiwar, line 40: +1

docs/design/upstream-datamover.md
```diff
@@ -0,0 +1,110 @@
...
+One of the primary new features of Velero 1.12 is the new Velero Built-in Data Mover (VBDM).
+The OADP Operator needs to add configuration (and e2e tests) to support this new datamover and remove any configuration and testing of the legacy OADP datamover.
+
+## Background
+Prior to the release of Velero 1.12, OADP 1.1 and 1.2 integrated an OADP-specific datamover based on Volsync into OADP and the OADP-installed Velero version.
+This took the form of some modifications to Velero core and the Velero CSI plugins, a new VSM plugin, and a separate VSM controller that used Volsync to store Volume data in the BackupStorageLocation.
+In addition to this, DPA and installation support for the OADP datamover was included in the OADP operator.
+
+In order to integrate VBDM into OADP, all of the above legacy OADP datamover parts will be removed from the OADP operator, the OADP Velero build, and the OADP build of the Velero Plugin for CSI.
+
```
shubham-pampattiwar, line 14: Maybe also add a link regarding the Node Agent Configuration design as a dependency for this design ?

shubham-pampattiwar (approved):
/lgtm

docs/design/upstream-datamover.md
sseago: @shubham-pampattiwar Technically it's not a dependency. That's why I didn't really reference its contents. Everything in this design works just fine enabling the legacy restic config.

docs/design/upstream-datamover.md
```diff
@@ -0,0 +1,110 @@
...
+#### Velero code
+
+In addition to pulling in Velero 1.12 upstream changes into the OADP Velero fork, we will need to remove all of the Velero customizations in our velero and velero-plugin-for-csi repos that were added to support the Volsync data mover.
+
+### Velero Built-in Data Mover configuration
+
+No new DPA fields will be needed for VBDM -- to enable it, the user just needs to include the CSI plugin and enable the node agent. For 1.3, users should use the new `NodeAgentConfig` spec field for this, although the existing (deprecated in 1.3) `Restic` spec field will also work here..
+
+Note that unlike in the OADP Volsync data mover, with VBDM, to use the datamover for a specific backup, that backup must be created with the `--snapshot-move-data` flag to `velero backup create` or with `Spec.SnapshotMoveData` set to `true` on the Backup CR.
```
shawn-hurley, line 79: This is still a tech preview feature or are we considering this to be GA?

shawn-hurley: Having a field that is not hidden behind the "Feature" flag is what is what I am worried about. 

When we have Tech Preview, having explicit opt in is useful, if the field exists and someone starts using it they could never know that they are in an unsupported path. 

I think we have to determine this and we have to decide if we need to add something around this.

sseago: @shawn-hurley We can't hide it behind a feature flag. It's in the velero backup spec, *not* in the DPA. Anyone installing Velero 1.12 with csi plugin and kopia or restic enabled has datamover enabled.

sseago: The only way to *not* allow users to use datamover would be to either remove CSI or remove restic/kopia from the DPA.

shawn-hurley: First, can we agreement on the problem, and it is a real one that we want to add to the open question PR?

sseago: @shawn-hurley It's added to open questions. And I would agree that, in general, new features that are tech preview should be behind a feature flag, but I'm not seeing an implementation path here, since in Velero 1.12, every component needed to enable datamover backups is either 1) already available without a feature flag on the DPA (CSI plugin, restic/node-agent) or 2) not part of any OADP-specific APIs that we control (regular spec field on velero Backup). So for 1.3 and this datamover, I think the best we can do is either documentation, or we could add a validation that add a warning event if both CSI and restic/node-agent are enabled, warning that this combination will enable a tech preview feature, and that they should be aware that datamover is still tech preview.

PR #1108, opened 2023-08-02
updates to vbdm design

sseago:


weshayutin (commented):
I like!! ++  @shawn-hurley anything else required iyho?

shawn-hurley:
I believe the conversation around whether Data mover is GA or Tech Preview in 1.3 should be added. 

I usually like the UX of having to opt into tech preview features.

shubham-pampattiwar:
If we make the decision to consider VBDM to be a tech preview feature, it will be a bit tricky to provide an option to opt-in via DPA. The backup API spec gives an option to opt-in via the `snapshotMoveData` boolean, can we consider this is an opt-in mechanism ?

sseago:
@shubham-pampattiwar  The issue is that there's no real way to opt in/out. The only config changes required to enable datamover (add CSI plugin and enable restic/kopia) are already available as GA features. Basically even without any OADP/DPA changes for 1.3, users can use the new datamover by simply enabling restic and CSI plugin and setting the datamover spec field on backups.

sseago:
@shawn-hurley "I usually like the UX of having to opt into tech preview features." -- but there's nothing to opt-in. Everything needed on the DPA to enable this is already available as GA, and the field needed on the Velero Backup CR is something that the OADP operator doesn't get involved in.

sseago:
@shubham-pampattiwar 
"The backup API spec gives an option to opt-in via the snapshotMoveData boolean, can we consider this is an opt-in mechanism ?"
That's opt-in in the same way as default-volumes-to-restic is opt-in. It seems to have no bearing on the tech preview vs. GA question.

sseago:
In any case, I updated the "Open Questions" section with the GA vs. Tech Preview question.

weshayutin (approved):

sseago:
@shawn-hurley (repeating my response to your comment about tech-preview and feature flags on the now-closed PR, since there's more visibility here)

It's added to open questions. And I would agree that, in general, new features that are tech preview should be behind a feature flag, but I'm not seeing an implementation path here, since in Velero 1.12, every component needed to enable datamover backups is either 
1) already available without a feature flag on the DPA (CSI plugin, restic/node-agent) or 
2) not part of any OADP-specific APIs that we control (regular spec field on velero Backup). 

So for 1.3 and this datamover, I think the best we can do is either documentation, or we could add a validation that add a warning event if both CSI and restic/node-agent are enabled, warning that this combination will enable a tech preview feature, and that they should be aware that datamover is still tech preview.

Does the warning approach work? Similar to the warning we're adding for (supported-but-deprecated) restic config, we could add a warning that the particular combination of DPA options they have chosen enables a (not-fully-supported) tech preview feature.

shawn-hurley:
We COULD add an admission web hook that would deny backups if the field is set and the feature flag is not turned on.

Do we think that we may in the future need to enhance validations and/or do other things for resources, it is an option that we should consider.

sseago:
@shawn-hurley That would work too, although it would require users who are using CSI (without datamover) and restic to enable the datamover tech-preview feature in the DPA. Which could be fine if we document it. But it would be possibly more confusing than just doing the warning without the feature flag.

shawn-hurley:
I don't know if there are other parts, what it would allow for, is if velero doesn't want to add things like feature flags, then it allows us to have feature flags. 

I also wonder if, something was ever added that we didnt' want to support at all, we could deny it in the web hook. 

I don't know if we want to, but the question is open as it is a possible way to do this, and seems like a implementation that does not have huge drawbacks.

sseago:
@shawn-hurley I guess we can just leave the question open for now, until/unless we reach a consensus on the admission webhook approach or the dpa warn approach, we can ignore the issue for now. Adding it later won't be any harder than adding it initially, since it doesn't really interact with any of the other work needed to enable this.

shubham-pampattiwar (approved):

PR #1109, opened 2023-08-03
WIF support design

kaovilai:
Closes https://github.com/openshift/oadp-operator/issues/1047

Implementation: https://github.com/openshift/oadp-operator/pull/1112

PR #1112, opened 2023-08-04
implement GCP WIF support

kaovilai:
design #1109 

Target release: OADP 1.3

Features
- [x] Basic BSL WIF support
- [ ] Validation/Enhancements to support imagestream backup

#### Testing instructions
Prerequisites:
- [ocp setup instruction](https://github.com/openshift/oadp-operator/wiki/GCP-WIF-Authentication-on-OpenShift#apply-credentials-secret-to-openshift-adp-namespace). Substitute operator install steps for the following operator-sdk run command.

To try this patch, get [operator-sdk](https://sdk.operatorframework.io/docs/installation/) and run
```bash
export OADP_TEST_NS=openshift-adp
(oc create namespace $OADP_TEST_NS || true) && \
operator-sdk run bundle ghcr.io/kaovilai/oadp-operator:wif-support-bundle --namespace $OADP_TEST_NS
```

cleanup
```
operator-sdk cleanup oadp-operator --namespace $OADP_TEST_NS
```
After cleanup, you will have to create dpa again.

#### dev notes
bundle build command
```bash
IMG=ghcr.io/kaovilai/oadp-operator:wif-support-operator BUNDLE_IMG=ghcr.io/kaovilai/oadp-operator:wif-support-bundle make docker-build docker-push bundle bundle-build bundle-push
```
bundle run command + dpa
```bash
operator-sdk run bundle ghcr.io/kaovilai/oadp-operator:wif-support-bundle --namespace openshift-adp && echo 'apiVersion: oadp.openshift.io/v1alpha1
metadata:
  name: dpa-sample
  namespace: openshift-adp
spec:
  configuration:
    velero:
      defaultPlugins:
      - openshift
      - gcp
  backupLocations:
    - velero:
        provider: gcp
        default: true
        credential:
          key: service_account.json
          name: cloud-credentials-gcp 
        objectStorage:
          bucket: tkaovila-bucket
          prefix: velero
  # Temporary image override while https://github.com/vmware-tanzu/velero-plugin-for-gcp/pull/142 is in draft.
  unsupportedOverrides:
    gcpPluginImageFqin: ghcr.io/kaovilai/velero-plugin-for-gcp:file-wif
' | oc apply -f -
```

### DIFF
--- a/controllers/dpa_controller.go
+++ b/controllers/dpa_controller.go
@@ -31,6 +31,7 @@ import (
 
 	"github.com/go-logr/logr"
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
+	oadpClient "github.com/openshift/oadp-operator/pkg/client"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/record"
@@ -86,6 +87,9 @@ func (r *DPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		log.Error(err, "unable to fetch DataProtectionApplication CR")
 		return result, nil
 	}
+
+	// set client to pkg/client for use in non-reconcile functions
+	oadpClient.SetClient(r.Client)
 
 	_, err := ReconcileBatch(r.Log,
 		r.ValidateDataProtectionCR,
--- a/pkg/client/client.go
+++ b/pkg/client/client.go
@@ -1,0 +1,35 @@
+package client
+
+// Provides a common client for functions outside of controllers to use.
+
+import (
+	"context"
+
+	"k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+var commonClient client.Client
+
+func SetClient(c client.Client) {
+	commonClient = c
+}
+
+func NewClientFromConfig(cfg *rest.Config) (c client.Client, err error) {
+	commonClient, err = client.New(cfg, client.Options{})
+	return commonClient, err
+}
+
+func GetClient() client.Client {
+	return commonClient
+}
+
+func CreateOrUpdate(ctx context.Context, obj client.Object) error {
+	err := commonClient.Create(ctx, obj)
+	// if err is alreadyexists, try update
+	if errors.IsAlreadyExists(err) {
+		return commonClient.Update(ctx, obj)
+	}
+	return err
+}
--- a/pkg/credentials/gcp.go
+++ b/pkg/credentials/gcp.go
@@ -1,0 +1,32 @@
+package credentials
+
+import "encoding/json"
+
+type gcpCredAccountKeys string
+
+// From https://github.com/golang/oauth2/blob/d3ed0bb246c8d3c75b63937d9a5eecff9c74d7fe/google/google.go#L95
+const (
+	serviceAccountKey  gcpCredAccountKeys = "service_account"
+	externalAccountKey gcpCredAccountKeys = "external_account"
+)
+
+func getGCPSecretAccountTypeKey(secretByte []byte) (gcpCredAccountKeys, error) {
+	var f map[string]interface{}
+	if err := json.Unmarshal(secretByte, &f); err != nil {
+		return "", err
+	}
+	// following will panic if cannot cast to credAccountKeys
+	return gcpCredAccountKeys(f["type"].(string)), nil
+}
+
+func gcpSecretAccountTypeIsShortLived(secretName, secretKey, namespace string) (bool, error) {
+	secretBytes, err := GetDecodedSecretAsByte(secretName, secretKey, namespace)
+	if err != nil {
+		return false, err
+	}
+	credAccountTypeKey, err := getGCPSecretAccountTypeKey(secretBytes)
+	if err != nil {
+		return false, err
+	}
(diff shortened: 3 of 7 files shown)


kaovilai:
BSL is available as of [bff85f2](https://github.com/openshift/oadp-operator/commit/bff85f2ebc26a030ea427f576c31df25e15305db)

with the following dpa
```bash
echo 'apiVersion: oadp.openshift.io/v1alpha1
metadata:
  name: dpa-sample
  namespace: openshift-adp
spec:
  configuration:
    velero:
      defaultPlugins:
      - openshift
      - gcp
  backupLocations:
    - velero:
        provider: gcp
        default: true
        credential:
          key: service_account.json
          name: cloud-credentials-gcp 
        objectStorage:
          bucket: tkaovila-bucket
          prefix: velero
  # Temporary image override while https://github.com/vmware-tanzu/velero-plugin-for-gcp/pull/142 is in draft.
  unsupportedOverrides:
    gcpPluginImageFqin: ghcr.io/kaovilai/velero-plugin-for-gcp:file-wif
' | oc apply -f -
```
after following [wiki](https://github.com/openshift/oadp-operator/wiki/GCP-WIF-Authentication-on-OpenShift#apply-credentials-secret-to-openshift-adp-namespace)

kaovilai:
backup with just deployment ✅ 
backup containing buildConfig/imagestreamtag backups without imagebackup disabled ⚠️

pkg/credentials/gcp.go
```diff
@@ -0,0 +1,32 @@
+package credentials
+
+import "encoding/json"
+
+type gcpCredAccountKeys string
+
+// From https://github.com/golang/oauth2/blob/d3ed0bb246c8d3c75b63937d9a5eecff9c74d7fe/google/google.go#L95
```
mateusoliveira43, line 7: here could wee use it directly? i think we already use it as indirect dependecy https://github.com/openshift/oadp-operator/blob/master/go.mod#L136

pkg/client/client.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+var commonClient client.Client
+
+func SetClient(c client.Client) {
+	commonClient = c
+}
+
+func NewClientFromConfig(cfg *rest.Config) (c client.Client, err error) {
+	commonClient, err = client.New(cfg, client.Options{})
```
mateusoliveira43, line 20: here is just Go question: I had trouble using `:=` overriding vars I did not want to. how were you able to not use `:=` for `err`? (only `commonClient` is prior declared, right?)

sseago: `c` and `err` are declared as return variables at the top of the function, so they don't need to be declared in func body, and they're not needed in the return call. See https://go.dev/tour/basics/7 for example usage.

mateusoliveira43: nicee, Thankss!

sseago: But yes, `:=` in a block scope will override any variables declared outside that scope.
```
foo := 1
if err != nil {
    foo, bar := myFunc()
}
// foo is still 1
foo, bar := myFunc2()
// now foo is the first return value of `myFunc2` instead of 1, since := was used in the same scope as foo was previously defined
```

mateusoliveira43: yes, I had I lot of trouble with that in my first Go codes heheh

pkg/credentials/gcp.go
kaovilai: If a var or const is small case, it is not exported, not accessible by external package and modules like ours.

kaovilai:
moving back to draft.. forgot I still need to 
> Validation/Enhancements to support imagestream backup

pkg/credentials/gcp.go
mateusoliveira43: Oh yeah, forgot about that :man_facepalming:

kaovilai:
will follow up in another pr for
> backup containing buildConfig/imagestreamtag backups without imagebackup disabled ⚠️

might have to still fix unit tests

kaovilai:
> Add features.operators.openshift.io/token-auth-gcp: "true" annotation after rebasing on top of https://github.com/openshift/oadp-operator/pull/1155 ?

Nope, it's still manual. That annotation is for automated integration with CCO. CCO automation is only avail for AWS and Azure at this time, and @shubham-pampattiwar has only just implemented aws recently.

> Whether the operator supports configuration for tokenzied authentication with Google Cloud APIs via GCP Workload Identity Foundation (WIF) using the CloudCredentialOperator.

[src](https://docs.engineering.redhat.com/display/CFC/Best_Practices#Best_Practices-(New)RequiredInfrastructureAnnotations:~:text=Whether%20the%20operator%20supports%20configuration%20for%20tokenzied%20authentication%20with%20Google%20Cloud%20APIs%20via%20GCP%20Workload%20Identity%20Foundation%20(WIF)%20using%20the%20CloudCredentialOperator.)

sseago (approved):

controllers/dpa_controller.go
```diff
@@ -87,6 +88,9 @@ func (r *DPAReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 		return result, nil
 	}
 
+	// set client to pkg/client for use in non-reconcile functions
+	oadpClient.SetClient(r.Client)
```
shubham-pampattiwar, line 92: Do we need this ?

kaovilai: When I moved functions out of `controllers/` into `/pkg/` they lose access to clients.

I think this is a pretty clean way to make sure every `/pkg` can reuse clients.

shubham-pampattiwar (approved):

PR #1125, opened 2023-08-10
Update velero dep to 1.11.1

sseago:


shubham-pampattiwar (approved):

go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/operator-framework/operator-lib v0.9.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.51.2
 	github.com/sirupsen/logrus v1.8.1
-	github.com/vmware-tanzu/velero v1.10.0 // TODO: Update this to a pinned version
+	github.com/vmware-tanzu/velero v1.11.1 // TODO: Update this to a pinned version
```
mateusoliveira43, line 18: the TODO still applies?

kaovilai: don't think so.. it's replaced by https://github.com/openshift/oadp-operator/pull/1125/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R165 anyways

sseago: Yeah, I'm not sure what that TODO was really about. The replace below makes sure we're using our fork, but this line is updated to reflect what that fork maps to upstream.

kaovilai: it was v000.0000000 something IIRC

sseago: Probably should remove that comment on master. Not sure I'd worry about it here if tests pass.

PR #1137, opened 2023-08-23
OADP-2139: [oadp-1.2] remove old restic DaemonSet

mateusoliveira43:
Remove old restic DaemonSet (if user is coming from OADP 1.1) and documentation for needed manual steps.

## what was done to achieve the result

Studied the changes from OADP 1.1 to 1.2

## How to test

Create fake restic DaemonSet from
```yaml
apiVersion: v1
items:
- apiVersion: apps/v1
  kind: DaemonSet
  metadata:
    annotations:
      deprecated.daemonset.template.generation: "1"
    generation: 1
    labels:
      component: velero
    name: restic
    namespace: openshift-adp
    # ownerReferences:
    # - apiVersion: oadp.openshift.io/v1alpha1
    #   blockOwnerDeletion: true
    #   controller: true
    #   kind: DataProtectionApplication
    #   name: test-update-1-1
    #   uid: eb9d2796-599b-4a82-99fb-156ee58e276a
  spec:
    revisionHistoryLimit: 10
    selector:
      matchLabels:
        component: velero
        name: restic
    template:
      metadata:
        creationTimestamp: null
        labels:
          component: velero
          name: restic
      spec:
        containers:
        - args:
          - restic
          - server
          command:
          - /velero
          env:
          - name: NODE_NAME
            valueFrom:
              fieldRef:
                apiVersion: v1
                fieldPath: spec.nodeName
          - name: VELERO_NAMESPACE
            valueFrom:
              fieldRef:
                apiVersion: v1
                fieldPath: metadata.namespace
          - name: VELERO_SCRATCH_DIR
            value: /scratch
          image: registry.redhat.io/oadp/oadp-velero-rhel8@sha256:2c1c645410a77569705e5521f8abb652b2b46bcb6ee06b9650bde78fcf3e3061
          imagePullPolicy: Always
          name: restic
          resources:
            requests:
              cpu: 500m
              memory: 128Mi
          securityContext:
            privileged: true
          terminationMessagePath: /dev/termination-log
          terminationMessagePolicy: File
          volumeMounts:
          - mountPath: /host_pods
            mountPropagation: HostToContainer
            name: host-pods
          - mountPath: /scratch
            name: scratch
          - mountPath: /etc/ssl/certs
            name: certs
        dnsPolicy: ClusterFirst
        restartPolicy: Always
        schedulerName: default-scheduler
        securityContext:
          runAsUser: 0
        serviceAccount: velero
        serviceAccountName: velero
        terminationGracePeriodSeconds: 30
        volumes:
        - hostPath:
            path: /var/lib/kubelet/pods
            type: ""
          name: host-pods
        - emptyDir: {}
          name: scratch
        - emptyDir: {}
          name: certs
    updateStrategy:
      rollingUpdate:
        maxSurge: 0
        maxUnavailable: 1
      type: RollingUpdate
  status:
    currentNumberScheduled: 3
    desiredNumberScheduled: 3
    numberAvailable: 3
    numberMisscheduled: 0
    numberReady: 3
    observedGeneration: 1
    updatedNumberScheduled: 3
kind: List
metadata:
  resourceVersion: ""
  selfLink: ""
```
run `make deploy-olm` and create DPA. restic DaemonSet should be deleted.


controllers/validator.go
```diff
@@ -12,9 +12,40 @@ import (
...
+		Name: "restic", Namespace: r.NamespacedName.Namespace,
+	}})
+	resticRepositoriesResource := &unstructured.Unstructured{}
+	resticRepositoriesResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "velero.io",
+		Version: "v1",
+		Kind:    "CustomResourceDefinition",
+	})
+	resticRepositoriesResource.SetName("resticrepositories.velero.io")
+	r.Delete(r.Context, resticRepositoriesResource)
```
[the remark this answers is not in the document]
mateusoliveira43: fixed, thanks @shawn-hurley

kaovilai:
Might wanna be careful here.. removing CRD in essence = removing CRs as well.. so customers would lose history for debugging where things were for instance. I think the proper approach is warn rather than delete.. the CRD part at least.

shawn-hurley:
I believe that we already migrated this CRD to the renamed one (someone can correct me if I am wrong) 

We do need to make sure that the migration is complete, but once it is finished, nothing is keeping these in sync(AFAIK). I would instead remove the old one as it will cause less confusion when debugging.

What is the scenario you are worried about @kaovilai ? `User has backup X and it .....`. Trying to understand, because my understanding is that velero manages this CRD and that it will fully use the new one. Can you help me see where a user is losing context and info?

kaovilai:
It will fully use a new one for new backup.

I'm not so sure if velero have during backup sync automatically migrate.

Scenario: user used oadp 1.0 backup with restic, user upgraded to oadp 1.x with this CRD deleting code, does that backup from 1.0 does or does not still work as restore source on upgraded oadp 1.x

docs/upgrade1-1.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Uninstall the OADP operator
+
+Use the web console to uninstall the OADP operator by clicking on `Install Operators` under the `Operators` tab on the left-side menu. Then click on `OADP Operator`.
+
+After clicking on `OADP Operator` under `Installed Operators`, navigate to the right side of the page, where the `Actions` drop-down menu is. Click on that, and select `Uninstall Operator`.
+
+TODO add images?
+
+TODO CLI commands?
```
mateusoliveira43, line 51: need to add images and CLI commands to do the same?

docs/upgrade1-1.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Install OADP Operator 1.2.x
+
+Follow theses [basic install](../docs/install_olm.md) instructions to install the new OADP operator version, create DPA, and verify correct installation.
+
+TODO tell about update channel strategy? Ex.: update subscription stable-1.1 to stable-1.2
```
mateusoliveira43, line 57: Tell about the other way to update? (if user follows the one written, they do not hit the daemonset issue)

mateusoliveira43:
/hold
Discuss CRD deletion 
Wait TODOs discussions

mateusoliveira43:
/unhould
QE was not able to run upgrade suite from the PR, so we need to discuss if we want to delete it (and we can also merge with some doubt and check if it passes QE)

controllers/dpa_controller.go
```diff
@@ -64,10 +64,10 @@ var debugMode = os.Getenv("DEBUG") == "true"
 //+kubebuilder:rbac:groups=oadp.openshift.io,resources=dataprotectionapplications/status,verbs=get;update;patch
 //+kubebuilder:rbac:groups=oadp.openshift.io,resources=dataprotectionapplications/finalizers,verbs=update
```
kaovilai, line 65: do you wanna add delete permission as a kubebuilder tag here too?

kaovilai: for reference: [RBAC](https://book.kubebuilder.io/reference/markers/rbac) markers

kaovilai: If you add rbac markers I think when you run `make generate` it will add it to the role.yaml for you or something along those lines.

kaovilai: but doing it manually also works.

kaovilai: Pretty sure rbac markers are additive and only adds more to the existing role. It does not delete perms without markers that were added before.

controllers/validator.go
```diff
@@ -12,9 +12,34 @@ import (
...
+	metaV1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 )
 
 func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error) {
+	// When updating from OADP version 1.1.x to 1.2.x, delete
+	// daemonset.apps/restic (it was renamed to daemonset.apps/node-agent)
+	// resticrepositories.velero.io CRD (it was renamed to backuprepositories.velero.io)
+	if err := r.Delete(r.Context, &appsV1.DaemonSet{ObjectMeta: metaV1.ObjectMeta{
```
shubham-pampattiwar, line 27: I believe the Deletion of restic DS can be part of the `ReconcileNodeAgentDaemonset` function ?

controllers/validator.go
```diff
@@ -12,9 +12,34 @@ import (
...
+	}}); err != nil && !apiErrors.IsNotFound(err) {
+		return false, err
+	}
+	resticRepositoriesResource := &unstructured.Unstructured{}
+	resticRepositoriesResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apiextensions.k8s.io",
+		Version: "v1",
+		Kind:    "CustomResourceDefinition",
+	})
+	resticRepositoriesResource.SetName("resticrepositories.velero.io")
```
shubham-pampattiwar, line 38: Should we not make sure that there are no CRs present in the cluster before deleting the CRD ? Or Do we already do that ?

docs/upgrade1-1.md
```diff
@@ -0,0 +1,57 @@
...
+-        restic-timeout: 600
++        fs-backup-timeout: 600
+```
+
+### Uninstall the OADP operator
+
+Use the web console to uninstall the OADP operator by clicking on `Install Operators` under the `Operators` tab on the left-side menu. Then click on `OADP Operator`.
+
+After clicking on `OADP Operator` under `Installed Operators`, navigate to the right side of the page, where the `Actions` drop-down menu is. Click on that, and select `Uninstall Operator`.
+
```
shubham-pampattiwar, line 48: @mateusoliveira43 qq: Did you also need to remove old CRDs when you were dev testing ?

docs/upgrade1-1.md
```diff
@@ -0,0 +1,57 @@
...
+TODO add images?
```
shubham-pampattiwar, line 49: What images are we talking about here ?

docs/upgrade1-1.md
mateusoliveira43: No, no need (letting old CRD in cluster didn't seem to cause any trouble)
But I did not test with having an instance of the CRD (that would be the upgrade path)

controllers/validator.go
mateusoliveira43: Did not think about this :grimacing: 
I hope upgrade kills them for us

But will try to test tomorrow if I can see how it works (install stable 1.1 -> create DPA with restic -> upgrade to stable 1.2)

controllers/validator.go
mateusoliveira43: This one https://github.com/openshift/oadp-operator/blob/master/controllers/nodeagent.go#L72 ?

I believe is better for organization, yes :smile:

shubham-pampattiwar: lets do that then !

controllers/validator.go
shubham-pampattiwar: yeah let me know about this

docs/upgrade1-1.md
shubham-pampattiwar: I think we are fine without the images, you can remove the TODO if you agree.

shubham-pampattiwar (approved):

kaovilai (approved):
`OADP-2139: [oadp-1.2] remove old restic DaemonSet and old CRD`
`OADP-2139: [oadp-1.2] remove old restic DaemonSet and document old CRD to delete`

controllers/validator.go
mateusoliveira43: Tested @shubham-pampattiwar 
Upgrade does not kill CR instances of the `resticrepositories.velero.io` CRD

So we will probably need to kill them before killing CRD (but this is not in this PR anymore)

shubham-pampattiwar: ok, thank you! @mateusoliveira43 ++

sseago (approved):

PR #1139, opened 2023-08-25
fix: OADP-2688 Always check restore resource priorities

mateusoliveira43:
Address Tiger's comment https://github.com/openshift/oadp-operator/pull/1132#discussion_r1303365262


api/v1alpha1/oadp_types.go
```diff
@@ -406,8 +406,10 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
 		if pvOperationTimeout, err := time.ParseDuration(resticTimeout); err == nil && dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout == nil {
 			dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout = &pvOperationTimeout
 		}
-	}
-	if dpa.Spec.Configuration.Velero.Args != nil && dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities == "" {
+		if dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities == "" {
+			dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities = "securitycontextconstraints,customresourcedefinitions,namespaces,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io"
+		}
+	} else {
```
mateusoliveira43, line 412: That's what you wanted @kaovilai ?

sseago: So if args is nil, then the next line would result in a nil pointer crash since it would result in calling `nil.RestoreResourcePriorities`. Instead of this, above line 399, if `Args` is nil, then set it to a non-nil value. Then the rest of the logic should work.

mateusoliveira43: right, going to work on the fix, thanks for the review

mateusoliveira43: Applied fix to nil pointer. I am still not very familiar with Go, if you have a better approach in mind, please suggest it

api/v1alpha1/oadp_types.go
```diff
@@ -406,8 +406,11 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
 		if pvOperationTimeout, err := time.ParseDuration(resticTimeout); err == nil && dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout == nil {
 			dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout = &pvOperationTimeout
 		}
-	}
-	if dpa.Spec.Configuration.Velero.Args != nil && dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities == "" {
+		if dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities == "" {
```
shawn-hurley, line 409: I think you may still need to handle the Args nil check so that we don't have a nil pointer deference.

If it is nil, we are going to have to set it though.

api/v1alpha1/oadp_types.go
shawn-hurley: Yeah see my comment above, if you need help with the example let me know

api/v1alpha1/oadp_types.go
shawn-hurley: @mateusoliveira43 talked to me offline

api/v1alpha1/oadp_types.go
shawn-hurley: ignore my comment I can't review code :)

api/v1alpha1/oadp_types.go
```diff
@@ -406,8 +406,11 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
...
+			dpa.Spec.Configuration.Velero.Args.RestoreResourcePriorities = "securitycontextconstraints,customresourcedefinitions,namespaces,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io"
+		}
+	} else {
+		dpa.Spec.Configuration.Velero.Args = &server.Args{}
```
sseago, line 413: I think you want this line above line 399, because now it's not nil, which means the things done after line 399 are needed here. In fact, if you do that, you don't need the second set of RestoreResourcePriorities (which gives us 2 places to make changes when we update the list).
Basically above line 399, you want:
```
if dpa.Spec.Configuration.Velero.Args == nil {
	dpa.Spec.Configuration.Velero.Args = &server.Args{}
}
```
I think if you do this, the rest of this PR is unnecessary. Also, once that's done, you no longer need the nil check in line 399, so that code can move one indentation level to the left.

sseago (changes requested):

api/v1alpha1/oadp_types.go
mateusoliveira43: Done, @shawn-hurley suggested the same thing :)

sseago: @mateusoliveira43 Yes, this looks just right now.

sseago (dismissed):

mateusoliveira43:
Checked CI logs, for unit tests, the error seems to be the update adding the restore priority. Expected
```
"server", "--fs-backup-timeout=1h", "--log-level", "info", "--item-operation-sync-frequency=5m", "--default-item-operation-timeout=2h"
```
Got (but why 2 log level flags?)
```
"server", "--log-level=info", "--fs-backup-timeout=1h0m0s", "--restore-resource-priorities=securitycontextconstraints,customresourcedefinitions,namespaces,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io", "--log-level", "info", "--item-operation-sync-frequency=5m", "--default-item-operation-timeout=2h"
```

For E2E, I did not figured out the problem. 

Can we update the unit tests? (if yes, where do I update them?)

controllers/velero_test.go
```diff
@@ -712,7 +718,9 @@ func TestDPAReconciler_buildVeleroDeployment(t *testing.T) {
 									Command: []string{"/velero"},
 									Args: []string{
 										"server",
-										"--fs-backup-timeout=1h",
+										"--log-level=info",
```
mateusoliveira43, line 721: I did not like to change `--fs-backup-timeout=1h` to `--fs-backup-timeout=1h0m0s`, but I think is fine. But Isn't wrong to have to have to log level flags in some tests?

controllers/velero_test.go
```diff
@@ -2494,8 +2520,7 @@ func TestDPAReconciler_buildVeleroDeployment(t *testing.T) {
 				}
 			}
 			if !reflect.DeepEqual(tt.wantVeleroDeployment, tt.veleroDeployment) {
-				fmt.Println(cmp.Diff(tt.wantVeleroDeployment, tt.veleroDeployment))
-				t.Errorf("expected velero deployment spec to be \n%#v, got \n%#v", tt.wantVeleroDeployment, tt.veleroDeployment)
+				t.Errorf("expected velero deployment spec to be \n%#v, got \n%#v\nDIFF:%v", tt.wantVeleroDeployment, tt.veleroDeployment, cmp.Diff(tt.wantVeleroDeployment, tt.veleroDeployment))
```
mateusoliveira43, line 2523: Changed here because was easier to fix the tests passing the diff after

controllers/velero_test.go
sseago: I wonder whether you should remove the log-level without info below? default level is info in any case. I think this is probably fine as-is if tests pass, but it may be worth finding a way to remove the duplicate log-level param.

sseago (approved):

shubham-pampattiwar (approved):

controllers/velero_test.go
mateusoliveira43: Found out, here https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L308 (through here https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L437) was triggering the addition of "--log-level=info" (and also changing 1h to 1h0m0s), but here https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L317 does the exact same check of log, so it is safe to remove it from `StringArr` function, right?

api/v1alpha1/oadp_types.go
```diff
@@ -396,20 +396,23 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
...
-		resticTimeout := "1h"
-		if dpa.Spec.Configuration != nil && dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
-			resticTimeout = dpa.Spec.Configuration.Restic.Timeout
-		}
-		if pvOperationTimeout, err := time.ParseDuration(resticTimeout); err == nil && dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout == nil {
-			dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout = &pvOperationTimeout
-		}
+
+	if dpa.Spec.Configuration.Velero.Args == nil {
+		dpa.Spec.Configuration.Velero.Args = &server.Args{}
```
kaovilai, line 401:
```suggestion
```
Can we do this without this line?

This would cause https://github.com/openshift/oadp-operator/pull/1139/files#diff-992e40b76fc7cf12af63ec6c2cdcc70321669228469781c0d2bc527dc4c28627R436 path to run everytime in velero.go
```go
	// if server args is set, override the default server args
	if dpa.Spec.Configuration.Velero.Args != nil {
```

We only wanted to help fill dpa.Spec.Configuration.Velero.Args if user themselves have made velero args not nil, otherwise use default velero functions to fill velero container args.

As is, we would be overriding args defined earlier in velero.go such as
```
veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--fs-backup-timeout=%s", fsBackupTimeout))
```

kaovilai (changes requested):

api/v1alpha1/oadp_types.go
mateusoliveira43: I think not, as Scott pointed https://github.com/openshift/oadp-operator/pull/1139#discussion_r1305782591 without this, we can hit nil pointer

But yes, this is causing that to run everytime, I was even going to ask if we can change from
```go
        // Enable user to specify --fs-backup-timeout (defaults to 1h)
	fsBackupTimeout := "1h"
	if dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
		fsBackupTimeout = dpa.Spec.Configuration.Restic.Timeout
	}
	// Append restic timeout option manually. Not configurable via install package, missing from podTemplateConfig struct. See: https://github.com/vmware-tanzu/velero/blob/8d57215ded1aa91cdea2cf091d60e072ce3f340f/pkg/install/deployment.go#L34-L45
	veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--fs-backup-timeout=%s", fsBackupTimeout))

	setContainerDefaults(veleroContainer)
	// if server args is set, override the default server args
	if dpa.Spec.Configuration.Velero.Args != nil {
		var err error
		veleroContainer.Args, err = dpa.Spec.Configuration.Velero.Args.StringArr(dpa.Spec.Configuration.Velero.FeatureFlags)
		if err != nil {
			return err
		}
	}
	return nil
```

to
```go

	setContainerDefaults(veleroContainer)
	var err error
	veleroContainer.Args, err = dpa.Spec.Configuration.Velero.Args.StringArr(dpa.Spec.Configuration.Velero.FeatureFlags)
	if err != nil {
		return err
	}
	return nil
```

mateusoliveira43:
/hold
Apply Tiger's changes
Wait Scott's PR

api/v1alpha1/oadp_types.go
mateusoliveira43: Going to fix this in next commit

controllers/velero_test.go
mateusoliveira43: going to ignore refactor needs for now

sseago (dismissed):

weshayutin:
@mateusoliveira43 updated the oadp ticket to a cloned jira issue targeted for 1.3.0 
 since this is merging to master.  https://issues.redhat.com/browse/OADP-2688

shubham-pampattiwar (dismissed):

kaovilai (approved):

weshayutin (approved):

PR #1146, opened 2023-08-30
fix: [oadp-1.2] OADP-2094 Always check restore resource priorities

mateusoliveira43:
Address Tiger's comment https://github.com/openshift/oadp-operator/pull/1132#discussion_r1303365262


tests/e2e/must-gather_suite_test.go
mateusoliveira43, line 1: this file and `tests/e2e/utils/common_utils.go` were mofified by `go fmt` I think, after running `make test`. I can remove them, if you want

shubham-pampattiwar (approved):

PR #1147, opened 2023-09-01
OADP-2505: Add additional types to default restore priorities

sseago:


shubham-pampattiwar (approved):

pkg/common/common.go
```diff
@@ -3,25 +3,58 @@ package common
...
+	DataMoverController          = "data-mover-controller"
+	DataMoverControllerContainer = "data-mover-controller-container"
+	OADPOperatorServiceAccount   = "openshift-adp-controller-manager"
+	VolSyncDeploymentName        = "volsync-controller-manager"
+	VolSyncDeploymentNamespace   = "openshift-operators"
+	VolsyncPrivilegedAnnotation  = "volsync.backube/privileged-movers"
 )
 
+var DefaultRestoreResourcePriorities = restore.Priorities{
+	HighPriorities: []string{
```
mateusoliveira43, line 26: niceeeee :rocket: much better to read

sseago: @mateusoliveira43 yes -- upstream provides the restore.Priorities struct. I realized that I should just use that struct and let it convert to string with the `.String()` method provided. Still need the messy list in test expectations, cli docs, etc. though

mateusoliveira43: yeah, if we could dynamically get that in docs (CRD description) would be perfect

mateusoliveira43 (approved):

pkg/common/common.go
sseago: @mateusoliveira43 We'd probably have to do that with a make task somehow, since it's in a golang comment, we can't exactly run golang code there. But yeah, something to think about at some point.

pkg/common/common.go
```diff
@@ -3,25 +3,58 @@ package common
...
+		"namespaces",
+		"managedcluster.cluster.open-cluster-management.io",
+		"managedcluster.clusterview.open-cluster-management.io",
+		"klusterletaddonconfig.agent.open-cluster-management.io",
+		"managedclusteraddon.addon.open-cluster-management.io",
+		"storageclasses",
+		"volumesnapshotbackups.datamover.oadp.openshift.io",
+		"volumesnapshotclass.snapshot.storage.k8s.io",
+		"volumesnapshotcontents.snapshot.storage.k8s.io",
+		"volumesnapshots.snapshot.storage.k8s.io",
```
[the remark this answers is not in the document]
sseago: @mateusoliveira43 it doesn't exist in OADP 1.2/Velero 1.11

pkg/velero/server/config.go
```diff
@@ -38,7 +38,7 @@ type ServerConfig struct {
...
+	// Desired order of resource restores, the priority list contains two parts which are split by "-" element. The resources before "-" element are restored first as high priorities, the resources after "-" element are restored last as low priorities, and any resource not in the list will be restored alphabetically between the high and low priorities. (default securitycontextconstraints,customresourcedefinitions,namespaces,managedcluster.cluster.open-cluster-management.io,managedcluster.clusterview.open-cluster-management.io,klusterletaddonconfig.agent.open-cluster-management.io,managedclusteraddon.addon.open-cluster-management.io,storageclasses,volumesnapshotbackups.datamover.oadp.openshift.io,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io)
 	// +optional
 	RestoreResourcePriorities string `json:"restore-resource-priorities,omitempty"`
```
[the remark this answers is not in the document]
sseago: @kaovilai There's code elsewhere that deals with actually applying the default. This line just updates the list to reflect the current values.

shawn-hurley (approved):
Except for the nit, I think this makes sense. I have not tested it though and I don't see a test case or cases here with it. 

Is there some test already that when this list is not applied correctly, will fail, but now will pass? If not can we add something to track that work? 

WDYT?

sseago:
@shawn-hurley These types are needed in the context of ACM restore, so you'd need an ACM cluster with OADP to test.

PR #1159, opened 2023-09-20
OADP-1737: Fix defaultl credentials logic

mpryc:
A logic which was looping over BSL and VSL then setting requirement for default credentials was broken as the loop was overriding need for those credentials finally setting it to false rather then requireing it on a first need.

This fixes [OADP-1737](https://issues.redhat.com//browse/OADP-1737) in which the BSL was going to an unavailable phase when VSL credentials were specified as there were no credentials mounted to the velero pod nor the AWS_SHARED_CREDENTIALS_FILE env specified for that pod.

########## Testing:
create DPA with snapshotLocation that uses credentials and check if the default credentials are used in the velero pod by checking at the env. values for that pod or by checking existence of the `/credentials/` directory within that pod e.g.:
```
[...]
  snapshotLocations:
    - velero:
        config:
          profile: default
          region: us-west-2
        credential:
          key: cloud
          name: cloud-credentials
        provider: aws
```

### DIFF
--- a/controllers/velero.go
+++ b/controllers/velero.go
@@ -712,7 +712,7 @@ func (r DPAReconciler) noDefaultCredentials(dpa oadpv1alpha1.DataProtectionAppli
 			}
 			if bsl.Velero != nil && bsl.Velero.Credential != nil {
 				bslProvider := strings.TrimPrefix(bsl.Velero.Provider, veleroIOPrefix)
-				if found := providerNeedsDefaultCreds[bslProvider]; !found {
+				if _, found := providerNeedsDefaultCreds[bslProvider]; !found {
 					providerNeedsDefaultCreds[bslProvider] = false
 				}
 			}
@@ -736,7 +736,9 @@ func (r DPAReconciler) noDefaultCredentials(dpa oadpv1alpha1.DataProtectionAppli
 			// Bucket credentials via configuration. Only AWS is supported
 			provider := strings.TrimPrefix(vsl.Velero.Provider, veleroIOPrefix)
 			if vsl.Velero.Credential != nil || provider == string(oadpv1alpha1.AWSBucketProvider) && hasCloudStorage {
-				providerNeedsDefaultCreds[provider] = false
+				if _, found := providerNeedsDefaultCreds[provider]; !found {
+					providerNeedsDefaultCreds[provider] = false
+				}
 			} else {
 				providerNeedsDefaultCreds[provider] = true
 			}


mateusoliveira43:
I would remove `[oadp-1.2]` from PR title, gives me impression it is 1.2 PR, but the target is master

sseago:
It looks like unit tests are expecting to *not* find credentials file in places where it's now shown. Not sure if that means there's a bug in the code or whether the tests were written to the wrong expectations, so if you can determine that the test expectations are wrong -- that the credentials *should* be there -- then the failing tests can be updated.

sseago:
@mpryc ok, so I figured out why tests are failing. We made an improper assumption that "the value is false if it's not set to true" -- it's a bool map, so if it's not set to anything, it's not false, it's not in the map. So I think we want to revert the changes here and instead, in the two places where the old code sets the value to false, we only set it to false if it's not currently set.

i.e. instead of:
```
					providerNeedsDefaultCreds[bslProvider] = false
```
We'd want:
```
					_, ok := providerNeedsDefaultCreds[bslProvider]
					if !ok {
						providerNeedsDefaultCreds[bslProvider] = false
					}
```

controllers/velero.go
```diff
@@ -712,7 +712,7 @@ func (r DPAReconciler) noDefaultCredentials(dpa oadpv1alpha1.DataProtectionAppli
 			}
 			if bsl.Velero != nil && bsl.Velero.Credential != nil {
 				bslProvider := strings.TrimPrefix(bsl.Velero.Provider, veleroIOPrefix)
-				if found := providerNeedsDefaultCreds[bslProvider]; !found {
+				if _, found := providerNeedsDefaultCreds[bslProvider]; !found {
```
mateusoliveira43, line 715: I would use only the first return value here

Since  `providerNeedsDefaultCreds` is a map that returns `bool`, if the key doesn't exist, the zero value of `bool` is false, which gives the same result

mpryc: This is *really* equivalent and the same there is a benefit of using `_, found :=` over just `found :=`  in clarity. The use of _ in this context signals the intention to only check for existence and not use the value itself.

sseago: @mpryc using just `found :=` would not be equivalent -- that would grab the value of the field, not a bool saying whether the key existed.

mateusoliveira43: got it, thanks for the explanation @sseago :smile:

sseago: (although I guess even though the logic isn't equivalent, the effect would be the same).
Basically if you use `_, found` then you're setting it to false only if it's not set, leaving prior true or false alone. If you use just `found` then you're setting it to false if it's either not set or false, only leaving prior true alone -- end result is the same, you're just setting false more often than needed. But I think using the "was the key present?" second value is clearer as to the purpose.

mateusoliveira43: right

As a suggestion, since it caused a little confusion, can we initialize `providerNeedsDefaultCreds` with all the keys false (instead of an empty map) and remove the not found checks? (I do not know if we know all the keys name when we initialize it)

sseago: If we did that, we'd be losing information. "This provider isn't used", "This provider needs default credentials", and "This provider does not need default credentials". Maybe just a comment above the if block indicating the distinction?

mpryc: I feel like we should not be refactoring code if there is no need for a simple bug fix. Otherwise we may have higher probability of introducing new issues.

mateusoliveira43:
GCP we disabled, right? need to override I think

shubham-pampattiwar (approved):

weshayutin:
ci/prow/4.12-operator-e2e-aws failure is a deprovision flake, can be ignored:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1159/pull-ci-openshift-oadp-operator-master-4.12-operator-e2e-aws/1706285807045709824/artifacts/operator-e2e-aws/e2e/build-log.txt

PR #1164, opened 2023-09-27
fix: [oadp-1.2] Remove 1.1 CRD

mateusoliveira43:
Follow up PR for https://github.com/openshift/oadp-operator/pull/1137

Removes instances and CRD of 1.1 leftover.

### DIFF
--- a/controllers/restic.go
+++ b/controllers/restic.go
@@ -17,6 +17,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	"k8s.io/utils/pointer"
@@ -68,9 +70,20 @@ func getNodeAgentObjectMeta(r *DPAReconciler) metav1.ObjectMeta {
 func (r *DPAReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error) {
 	// When updating from OADP version 1.1.x to 1.2.x, delete
 	// daemonset.apps/restic (it was renamed to daemonset.apps/node-agent)
+	// resticrepositories.velero.io CRD (it was renamed to backuprepositories.velero.io)
 	if err := r.Delete(r.Context, &appsV1.DaemonSet{ObjectMeta: metaV1.ObjectMeta{
 		Name: "restic", Namespace: r.NamespacedName.Namespace,
 	}}); err != nil && !apiErrors.IsNotFound(err) {
+		return false, err
+	}
+	resticRepositoriesResource := &unstructured.Unstructured{}
+	resticRepositoriesResource.SetGroupVersionKind(schema.GroupVersionKind{
+		Group:   "apiextensions.k8s.io",
+		Version: "v1",
+		Kind:    "CustomResourceDefinition",
+	})
+	resticRepositoriesResource.SetName("resticrepositories.velero.io")
+	if err := r.Delete(r.Context, resticRepositoriesResource); err != nil && !apiErrors.IsNotFound(err) {
 		return false, err
 	}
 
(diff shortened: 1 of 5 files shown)


mateusoliveira43:
/hold
wait 1.2.2 release

controllers/restic.go
```diff
@@ -68,11 +70,22 @@ func getNodeAgentObjectMeta(r *DPAReconciler) metav1.ObjectMeta {
 func (r *DPAReconciler) ReconcileResticDaemonset(log logr.Logger) (bool, error) {
 	// When updating from OADP version 1.1.x to 1.2.x, delete
 	// daemonset.apps/restic (it was renamed to daemonset.apps/node-agent)
+	// resticrepositories.velero.io CRD (it was renamed to backuprepositories.velero.io)
 	if err := r.Delete(r.Context, &appsV1.DaemonSet{ObjectMeta: metaV1.ObjectMeta{
 		Name: "restic", Namespace: r.NamespacedName.Namespace,
 	}}); err != nil && !apiErrors.IsNotFound(err) {
 		return false, err
 	}
+	resticRepositoriesResource := &unstructured.Unstructured{}
```
mateusoliveira43, line 79: 2 questions:
- right place? this only gets created (instances) in 1.1 if user enables restic? But even if user does not create an instance, the CRD will always be there
- should we delete instances prior to deleting CRD?

kaovilai: Deleting CRD will automatically delete instances as long as there's no finalizers.

kaovilai: > this only gets created (instances) in 1.1 if user enables restic?

Yes.

> even if user does not create an instance, the CRD will always be there

Yes. CRD installation handled by OLM, and is defined by what CRD is in the `bundle/` and so it is not dynamically created when needed by oadp reconciliation of dpa.

mateusoliveira43: I will check if there are finalizers

mateusoliveira43: about location I was worried that if user creates DPA without restic, it won't trigger this function, but it was triggered, right?

kaovilai: This [runs](https://github.com/openshift/oadp-operator/blob/b570aa4012b9acce7e48bdd6c13ab4b9bdd7cd02/controllers/dpa_controller.go#L102) regardless of dpa settings.

sseago: @mateusoliveira43 Checking for finalizers means listing and iterating over every ResticRepository in the namespace. That may be overkill, since Velero never adds finalizers to these CRs.

mateusoliveira43: oh I meant manually deleting CRD in my cluster with instances, and see if any problem occurs hehehehe Just to check if I do not have to add delete instances code

kaovilai: > Checking for finalizers means listing and iterating over every ResticRepository

Actually I think it would show up in CRD status in some form that there exists CRs with finalizers.. so we can work off that rather than querying ourselves each item.

sseago: @kaovilai I don't think that's true. For example, velero Restores have a finalizer, but if you get the restore CRD, there's nothing in status about restore CRs or their finalizers.

kaovilai: > velero Restores have a finalizer, but if you get the restore CRD, there's nothing in status about restore CRs or their finalizers.

I don't mean this scenario. I meant during deletion.

See [example CRD status here](https://stackoverflow.com/questions/52009124/not-able-to-completely-remove-kubernetes-customresource#:~:text=Finally%2C%20here%20is%20the%20output%20of%2C)

```
Status:
  Conditions:
    Last Transition Time:  2018-08-23T13:29:45Z
    Message:               CustomResource deletion is in progress
    Reason:                InstanceDeletionInProgress
    Status:                True
    Type:                  Terminating
Events:                    <none>
```

mateusoliveira43:
from this discussion https://github.com/openshift/oadp-operator/pull/1171#discussion_r1370588067

should we have the same approach here? (do not know if resticRepositories instances have finalizers) meaning this PR should be closed without merging

kaovilai:
> should we have the same approach here?

We have VSB/VSR in oadp-1.1 as well so yes.

>if resticRepositories instances have finalizers

No, they do not have finalizers.

mateusoliveira43:
@kaovilai what I meant about same approach is not deleting the CRD resticrepository

kaovilai:
> what I meant about same approach is not deleting the CRD resticrepository

sounds good.

weshayutin (approved):

PR #1165, opened 2023-09-28
oadp-1.1: Fix oadp operator image tag

dmrost:
This fixes "unknown flag: --fs-backup-timeout error"

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -159,7 +159,7 @@ metadata:
     capabilities: Seamless Upgrades
     categories: OpenShift Optional
     certified: "false"
-    containerImage: quay.io/konveyor/oadp-operator:latest
+    containerImage: quay.io/konveyor/oadp-operator:oadp-1.1
     createdAt: "2020-09-08T12:21:00Z"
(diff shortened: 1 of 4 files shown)


kaovilai (dismissed):
/lgtm

shubham-pampattiwar (dismissed):

mateusoliveira43:
/hold
Should we update the config files and then run `make bundle` to create the bundle files?

kaovilai:
> Should we update the config files and then run make bundle to create the bundle files?

ya. will push here in a bit.

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mateusoliveira43, line 1: Thanks for the PR @dmrost 

The files under `bundle/` folder are generated running `make bundle`. Can you please edit the files that hold the change, run `make bundle` and commit it? (I think @kaovilai can do this for you)

I believe the files you want to modify are https://github.com/openshift/oadp-operator/blob/oadp-1.1/Makefile#L117 and https://github.com/openshift/oadp-operator/blob/oadp-1.1/config/manifests/bases/oadp-operator.clusterserviceversion.yaml#L8

mateusoliveira43 (approved):

PR #1169, opened 2023-10-02
OADP-659 Remove HTTP/HTTPS port numbers from AWS S3 URLs.

mrnold:
Address [OADP-659](https://issues.redhat.com//browse/OADP-659) by removing explicit port numbers from HTTP and HTTPS S3 URLs. The AWS SDK removes these before calculating a signature for blob access, and some S3-compatible servers do not expect this, causing a signature mismatch. Actual alternate port numbers (anything but 80 or 443) are unaffected.

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/go-logr/logr"
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
+	"github.com/openshift/oadp-operator/pkg/bucket"
 	"github.com/openshift/oadp-operator/pkg/common"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -219,6 +220,18 @@ func (r *DPAReconciler) updateBSLFromSpec(bsl *velerov1.BackupStorageLocation, d
 	if bslSpec.Provider == "azure" {
 		if len(bslSpec.Config["storageAccountKeyEnvVar"]) == 0 {
 			registryDeployment = "False"
+		}
+	}
+	// The AWS SDK expects the server providing S3 blobs to remove default ports
+	// (80 for HTTP and 443 for HTTPS) before calculating a signature, and not
+	// all S3-compatible services do this. Remove the ports here to avoid 403
+	// errors from mismatched signatures.
+	if bslSpec.Provider == "aws" {
+		s3Url := bslSpec.Config["s3Url"]
+		if len(s3Url) > 0 {
+			if s3Url, err = bucket.StripDefaultPorts(s3Url); err == nil {
+				bslSpec.Config["s3Url"] = s3Url
+			}
 		}
 	}
 	bsl.Labels = map[string]string{
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/go-logr/logr"
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
-	"github.com/openshift/oadp-operator/pkg/bucket"
 	"github.com/openshift/oadp-operator/pkg/common"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -229,7 +228,7 @@ func (r *DPAReconciler) updateBSLFromSpec(bsl *velerov1.BackupStorageLocation, d
 	if bslSpec.Provider == "aws" {
 		s3Url := bslSpec.Config["s3Url"]
 		if len(s3Url) > 0 {
-			if s3Url, err = bucket.StripDefaultPorts(s3Url); err == nil {
+			if s3Url, err = common.StripDefaultPorts(s3Url); err == nil {
 				bslSpec.Config["s3Url"] = s3Url
 			}
 		}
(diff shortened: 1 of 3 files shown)


controllers/bsl.go
```diff
@@ -221,6 +222,18 @@ func (r *DPAReconciler) updateBSLFromSpec(bsl *velerov1.BackupStorageLocation, d
 			registryDeployment = "False"
 		}
 	}
+	// The AWS SDK expects the server providing S3 blobs to remove default ports
+	// (80 for HTTP and 443 for HTTPS) before calculating a signature, and not
+	// all S3-compatible services do this. Remove the ports here to avoid 403
+	// errors from mismatched signatures.
+	if bslSpec.Provider == "aws" {
```
shubham-pampattiwar, line 229: This code will get executed for minio as well as noobaa, right ? Any idea if we need this port stripping for them too ?

mrnold: Yes, it needs to run for anything that goes through the AWS SDK. I don't know if minio already deals with this, but I will try it and see. I think this change shouldn't cause any existing working connection types to stop working.

kaovilai:
master do not support 4.10 anyways :)
/override ci/prow/4.10-operator-e2e-aws

mrnold:
On further testing, I found that you can configure a self-signed certificate that is only valid for the URL with the port number. I think removing this port number directly from the BSL will make it (the BSL) unusable. I don't know if it's a common practice to create certificates this way, but it is something to consider before merging this.

pkg/bucket/aws.go
```diff
@@ -170,3 +173,18 @@ func (a awsBucketClient) Delete() (bool, error) {
...
+	if err != nil {
+		return "", err
+	}
+	r := http.Request{
+		URL: u,
+	}
+	request.SanitizeHostForHeader(&r)
+	r.URL.Host = r.Host
+	return r.URL.String(), nil
+}
```
kaovilai, line 190: LGTM but I would move this to more generic place.. maybe util.go you never know when this might be needed for non aws.

kaovilai (approved):

kaovilai:
> I don't know if it's a common practice to create certificates this way, but it is something to consider before merging this.

Could be a flag to disable the port stripping and we would note that this is not supported workflow for imagestream backup.

shawn-hurley:
Can we quickly verify that the openshift router does not generate certs that way?

pkg/bucket/aws.go
mrnold: Okay, I tried moving this to pkg/common.

mrnold:
> Can we quickly verify that the openshift router does not generate certs that way?

I don't think it does, although I am not 100% sure how to verify this. I opened Networking->Routes in the OpenShift console, clicked a few of the externally-accessible links, and looked at the certificates each one provided. Does that cover what you were hoping to see? I didn't see any :80s or :443s attached to DNS names, and I think it would be kind of weird if there were any.

kaovilai:
> I didn't see any :80s or :443s attached to DNS names

then I think we're clear here. If you could point me to where in aws sdk is doing this and why aws BSL for normal backup do not hit this would be great as well.

shubham-pampattiwar:
@kaovilai Is this what you wanted for reference? https://github.com/aws/aws-sdk-go/blob/v1.45.20/aws/request/request.go#L717

kaovilai:
> wanted

yup. Guess now my question is why BSL doesn't break.. but imagestream copy does.

kaovilai:
/override ci/prow/4.10-operator-e2e-aws
/override ci/prow/4.10-operator-e2e-azure

4.10 ci no longer works on master. should be removed soon.

pkg/common/common.go
```diff
@@ -198,3 +201,18 @@ func CCOWorkflow() bool {
 	}
 	return false
 }
+
+// StripDefaultPorts removes port 79 from HTTP URLs and 443 from HTTPS URLs.
```
kaovilai, line 205:
```suggestion
// StripDefaultPorts removes port 80 from HTTP URLs and 443 from HTTPS URLs.
```

mrnold: I have no idea how this happened.

kaovilai (approved):

shubham-pampattiwar (approved):

pkg/common/common.go
```diff
@@ -198,3 +201,18 @@ func CCOWorkflow() bool {
 	}
 	return false
 }
+
+// StripDefaultPorts removes port 80 from HTTP URLs and 443 from HTTPS URLs.
+// Defer to the actual AWS SDK implementation to match its behavior exactly.
+func StripDefaultPorts(fromUrl string) (string, error) {
```
mateusoliveira43, line 207: here I ran a test focusing on IBM and was not successful

```go
func TestStripDefaultPorts(t *testing.T) {
	base := "https://s3.region.cloud-object-storage.appdomain.cloud/bucket-name"
	want := "https://s3.region.cloud-object-storage.appdomain.cloud/bucket-name"
	t.Run("original URL is returned", func(t *testing.T) {
		got, err := StripDefaultPorts(base)
		if err != nil {
			t.Errorf("An error occurred: %v", err)
		}
		if !reflect.DeepEqual(got, want) {
			t.Errorf("StripDefaultPorts() = %v, want %v", got, want)
		}
	})
}
```

kaovilai: Add a test case and fix?

mrnold: I think this is a new test case, right Mateus? Thanks for adding this. In this test, StripDefaultPorts correctly extracts the host name from the URL, but then overwrites it with a blank string. That blank string comes from an HTTP request field that is not blank when running against a real AWS cluster. I will add a fix that should get this test working.

mrnold: This should fix it: https://github.com/openshift/oadp-operator/pull/1177

PR #1170, opened 2023-10-02
OADP-2747: Re-enable DataMover E2E tests

mateusoliveira43:
Update DataMover E2E tests to run against Velero builtin DataMover.

## How to test

Read Testing documentation and run `make test-e2e`. Check if DataMover was invoked (`DataUploads` and `DataDownloads` instances were created).

PR #1171, opened 2023-10-03
OADP-2767: Upgrade: Remove unbreaking references to VSM/Volsync DataMover

mateusoliveira43:
Remove unbreaking references to old (VSM/Volsync) DataMover (allowing users to still automatically upgrade to OADP 1.3) and add instructions on how to use new DataMover when upgrading from OADP 1.2.

What is left of references to old (VSM/Volsync) DataMover, will be deleted in OADP 1.4 (https://github.com/openshift/oadp-operator/issues/1192).

## What was done to achieve the goal

To find occurrences
```sh
grep -Iinr 'volsync' . --exclude-dir=tests
grep -Iinr 'vsm' . --exclude-dir=tests
grep -Iinr 'datamover' . --exclude-dir=tests
grep -Iinr 'mover' . --exclude-dir=tests
grep -Iinr 'dummy' . --exclude-dir=tests
```
(excluding `tests` folder, because this should be https://github.com/openshift/oadp-operator/pull/1170 responsibility)

To apply changes
```sh
make generate
make bundle
```

### How to test

To test an upgrade from 1.2 to 1.3, from this PR branch run
```sh
make catalog-test-upgrade
```
This will create `oadp-operator-catalog-test-upgrade` catalog in your cluster (OperatorHub in the UI).
Screenshot from 2023-10-23 14-08-46

Select the new catalog in the UI and install selecting `stable 1.2` update channel. This will install version 1.2.2 of OADP operator (from `oadp-1.2` branch, not 1.2.2 from marketplace).
Screenshot from 2023-10-23 14-09-11

Create a DPA with DataMover using the newly installed operator.

Finally, change channel of the installed operator to `stable` (version 99.0.0) in Subscription tab.
Screenshot from 2023-10-23 14-10-26


api/v1alpha1/oadp_types.go
```diff
@@ -231,97 +223,8 @@ type SnapshotLocation struct {
...
-	CacheAccessMode string `json:"cacheAccessMode,omitempty"`
-}
-
 // Features defines the configuration for the DPA to enable the tech preview features
-type Features struct {
-	// Contains data mover specific configurations
-	// +optional
-	DataMover *DataMover `json:"dataMover,omitempty"`
-}
+type Features struct{}
```
mateusoliveira43, line 227: I think we discussed this: left it as an empty struct, right?

config/manager/manager.yaml
```diff
@@ -50,10 +50,6 @@ spec:
             value: quay.io/konveyor/velero-plugin-for-gcp:latest
           - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_CSI
             value: quay.io/konveyor/velero-plugin-for-csi:latest
-          - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_VSM
-            value: quay.io/konveyor/velero-plugin-for-vsm:latest
-          - name: RELATED_IMAGE_VOLUME_SNAPSHOT_MOVER
-            value: quay.io/konveyor/volume-snapshot-mover:latest
           - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
             value: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
           - name: RELATED_IMAGE_DATA_MOVER_DUMMY_POD
```
mateusoliveira43, line 55: this one should be removes as well?

controllers/validator.go
```diff
@@ -63,32 +60,6 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 		}
 	}
```
mateusoliveira43, line 62: Should we add some validation for new dataMover?

config/manager/manager.yaml
kaovilai: +1 yes. Only for oadp 1.2 data mover.

controllers/validator.go
kaovilai: probably.

api/v1alpha1/oadp_types.go
kaovilai: I think empty struct would still cause breakage.

We would want to in this types.go, whereever this is referenced to omit empty  in the jsontag for at leats one minor version ie. oadp 1.3.z

then we warn/notify user to remove so they know about velero upstream data mover.

sseago: The reason for the empty struct is that we may (in the future) have additional Features to add.

config/manager/manager.yaml
sseago: Yes, remove both VSM plugin and VSM controller.

controllers/validator.go
sseago: There's not really any config for datamover, so I don't know that we have any datamover-specific validation. To enable DataMover, the user just needs to add CSI plugin and enable node agent. Both of these two things can also be used independently from each other for non-datamover cases.

api/v1alpha1/oadp_types.go
```diff
@@ -34,22 +34,16 @@ const ReconcileCompleteMessage = "Reconcile complete"
 
 const OadpOperatorLabel = "openshift.io/oadp"
 const RegistryDeploymentLabel = "openshift.io/oadp-registry"
-const DataMoverDeploymentLabel = "openshift.io/oadp-data-mover"
```
sseago, line 37: We may still need this if we have code to delete the 1.2 datamover deployment on upgrade to 1.3

pkg/common/common.go
```diff
@@ -60,12 +54,9 @@ const (
 	AzurePluginImage     = "quay.io/konveyor/velero-plugin-for-microsoft-azure:latest"
 	GCPPluginImage       = "quay.io/konveyor/velero-plugin-for-gcp:latest"
 	CSIPluginImage       = "quay.io/konveyor/velero-plugin-for-csi:latest"
-	VSMPluginImage       = "quay.io/konveyor/velero-plugin-for-vsm:latest"
-	// DataMoverImage is the data mover controller for data mover CRs - VolumeSnapshotBackup and VolumeSnapshotRestore
-	DataMoverImage      = "quay.io/konveyor/volume-snapshot-mover:latest"
-	DummyPodImage       = "quay.io/konveyor/rsync-transfer:latest"
-	RegistryImage       = "quay.io/konveyor/registry:latest"
-	KubeVirtPluginImage = "quay.io/konveyor/kubevirt-velero-plugin:v0.2.0"
+	DummyPodImage        = "quay.io/konveyor/rsync-transfer:latest"
```
sseago, line 57: DummyPodImage is also for 1.2 datamover, so you should remove all references to it as well.

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -827,10 +809,6 @@ spec:
                   value: quay.io/konveyor/velero-plugin-for-gcp:latest
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_CSI
                   value: quay.io/konveyor/velero-plugin-for-csi:latest
-                - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_VSM
-                  value: quay.io/konveyor/velero-plugin-for-vsm:latest
-                - name: RELATED_IMAGE_VOLUME_SNAPSHOT_MOVER
-                  value: quay.io/konveyor/volume-snapshot-mover:latest
                 - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
                   value: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
                 - name: RELATED_IMAGE_DATA_MOVER_DUMMY_POD
```
sseago, line 814: Dummy pod is also for 1.2 datamover, so you should remove all references to it as well.

config/manager/manager.yaml
sseago: (and dummy pod)

mateusoliveira43: Going to remove dummy pod references in next commit, thanks!

controllers/validator.go
mateusoliveira43: Yeah, only thing I thought about was user adding `defaultSnapshotMoveData` in DPA and then we check if both node agent and csi are present

api/v1alpha1/oadp_types.go
mateusoliveira43: resolving this conversation as we agree

config/rbac/role.yaml
```diff
@@ -24,18 +24,6 @@ rules:
...
-  verbs:
-  - '*'
-- apiGroups:
-  - volsync.backube
-  resources:
-  - '*'
-  verbs:
-  - '*'
 - apiGroups:
   - snapshot.storage.k8s.io
```
shubham-pampattiwar, line 28: We might not need the storage API group permissions as well.

mateusoliveira43: will investigate

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,52 @@
...
+    From this update, OADP now uses Velero Built-in DataMover instead of VSM/Volsync DataMover. This change the following:
+
+    - `spec.features.dataMover` was removed (you can delete secret for `spec.features.dataMover.credentialName` from your cluster, if you want)
+
+    - `vsm` plugin was removed
+
+    - `volsync` operator is not necessary anymore (you can uninstall it from your cluster, if you want)
+
+    - `volumesnapshotbackups.datamover.oadp.openshift.io` and `volumesnapshotrestores.datamover.oadp.openshift.io` CustomResourceDefinitions are not necessary anymore (you can delete `volumesnapshotbackups.datamover.oadp.openshift.io` and `volumesnapshotrestores.datamover.oadp.openshift.io` CRD from your cluster, if you want)
+
```
mateusoliveira43, line 17: I need to test if this is all. Do you remember anything? There is anything in especial you want to add the responsibility to OADP code to handle?

mateusoliveira43:
I think I touched every file that had reference to DataMover. Please review if my changes in documentation files make sense

controllers/validator.go
sseago: @mateusoliveira43 Oh, actually, yes, that would be something to validate. If they set that to true, then it's only useful if node agent (or the deprecated restic config) and CSI plugin are present.

docs/upgrade_1-2_to_1-3.md
sseago: I guess we could have OADP remove these VSB/VSR CRs. Or we could just leave this as a doc for users to clean up if they want to -- leaving them around won't break anything.

kaovilai: +1 with user initiating clean up. We should automate deletion as minimally required IMO in case of downgrade.

mateusoliveira43: Also talk about restic to nodeAgent change

mateusoliveira43:
/hold
address PR comments

kaovilai:
infra failed to install
/test 4.12-operator-e2e-azure

kaovilai:
```
INFO[2023-10-05T15:15:04Z] Step operator-e2e-azure-e2e succeeded after 54m44s. 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1171/pull-ci-openshift-oadp-operator-master-4.11-operator-e2e-azure/1709919661857968128#1:build-log.txt%3A79

flake after e2e
/override ci/prow/4.11-operator-e2e-azure

config/manifests/bases/oadp-operator.clusterserviceversion.yaml
```diff
@@ -327,7 +327,7 @@ spec:
     an archive file on object storage. OADP backs up persistent volumes
     (PVs) by creating snapshots with the native cloud snapshot API or with
     the Container Storage Interface (CSI). For cloud providers that do not
-    support snapshots, OADP backs up resources and PV data with Restic.
+    support snapshots, OADP backs up resources and PV data with Restic or Kopia.
```
mateusoliveira43, line 330: If you want to update description for 1.3, I think that's the time

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,81 @@
...
++      uploaderType: restic
+```
+
+> **Note:** OADP will be favoring Kopia over Restic in the near future.
+
+## Upgrade steps
+
+### Create new backup
+
+If you are using DataMover, you need to create backup before upgrade, because they 1.2 DataMover backups will not work after upgrade. We suggest to create Restic backups.
```
mateusoliveira43, line 47: I wasn't able to test that, still TODO

config/rbac/role.yaml
mateusoliveira43: I was not able to test yet, but I was wondering if that is still needed for CSI path

mateusoliveira43: I think you are right @shubham-pampattiwar 

I ran the 3 CSI and 2 DataMover E2E tests removing lines 27 through 62 of this file (removing `snapshot.storage.k8s.io` and `storage.k8s.io` rules) and the tests passed (this is enough to confirm that they are not necessary, right?)

My question: is why we do not need it anymore?

shubham-pampattiwar: We needed them for VSM controller, we created clones of VS and VSC during the VSM datamover backup process.

mateusoliveira43: alright, will remove in following commit
Thanks!

controllers/validator.go
mateusoliveira43: we will have a future PR that introduces that to DPA, right? So, can I resolve this conversation and bring it back in that PR?

kaovilai: +1 separate add PR.

mateusoliveira43: Alright, resolving, thanks!

docs/upgrade_1-2_to_1-3.md
mateusoliveira43: tried to run an upgrade test, but it did not work :( during upgrade got this error

error validating existing CRs against new CRD's schema for "dataprotectionapplications.oadp.openshift.io": error validating custom resource against new schema for DataProtectionApplication openshift-adp/test-upgrade: [].spec.configuration.velero.defaultPlugins[3]: Unsupported value: "vsm": supported values: "aws", "gcp", "azure", "csi", "openshift", "kubevirt"

to test it locally, run `make catalog-test-upgrade` and run the `oc` command required after that

Is this a problem or we should require users to stop running instance of DPA before update? (instead of changing channel, do the steps from `docs/upgrade_1-2_to_1-3.md`)

docs/upgrade_1-2_to_1-3.md
mateusoliveira43: Tested with mysql-persistent, and actually worked...

I ran csi, datamover and restic backups in stable-1.2 and restored with `make deploy-olm`, all """worked"""

All of them were **PartiallyFailed** with the same errors:

>mysql-persistent:  error getting resource from lister for rolebindings.authorization.openshift.io, mysql-persistent/system:deployers: rolebindings.authorization.openshift.io "system:deployers" not found
>                       error getting resource from lister for rolebindings.authorization.openshift.io, mysql-persistent/system:image-builders: rolebindings.authorization.openshift.io "system:image-builders" not found
>                       error getting resource from lister for rolebindings.authorization.openshift.io, mysql-persistent/system:image-pullers: rolebindings.authorization.openshift.io "system:image-pullers" not found

but when checking applications, they were running and were restored correctly. Can someone also run some backups to double check this behavior?

docs/upgrade_1-2_to_1-3.md
kaovilai: Relevant issue: https://issues.redhat.com/browse/OCPBUGS-15819
Once upgrade started, fixing dpa won't resolve this. You need to uninstall then reinstall.

docs/upgrade_1-2_to_1-3.md
mateusoliveira43: we will test this in a specific day :)

api/v1alpha1/oadp_types.go
mateusoliveira43: labels volume-snapshot-mover deployment leftover has
- app.kubernetes.io/component: volume-snapshot-mover
- app.kubernetes.io/instance: volume-snapshot-mover
- app.kubernetes.io/managed-by: oadp-operator
- app.kubernetes.io/name: oadp-operator-velero
- openshift.io/oadp: 'True'
- openshift.io/oadp-data-mover: 'True'

Best one to use is `openshift.io/oadp-data-mover`?

docs/upgrade_1-2_to_1-3.md
mateusoliveira43: Alright, fixed the automatically upgrade bug in following commit

I think for leftover we have 
- volume-snapshot-mover deployment (will write code to delete it)
- Volsync operator (I would not remove it, since its cluster wide, maybe there are users that use it for other things?)
- and CRDs

Do you agree with volsync approach? How should we delete old CRDs? Can just delete, first check if tehre no running instances? if there are: tell users to delete them themselves?

shubham-pampattiwar: Yes please do not add logic to remove volsync operator, volsync controller, deployment and its CRDs. We do not own these, volsync operator and its resources may/can exist independently without any correlation with OADP operator.

Makefile
```diff
@@ -423,6 +423,44 @@ catalog-build-replaces: opm ## Build a catalog image using replace mode
...
+		sed -i '/replaces:/d' ./bundle/manifests/oadp-operator.clusterserviceversion.yaml && \
+		IMG=$(PREVIOUS_OPERATOR_IMAGE) BUNDLE_IMG=$(PREVIOUS_BUNDLE_IMAGE) \
+		make docker-build docker-push bundle-build bundle-push && cd -
+	cd $(TEMP)/current && IMG=$(THIS_OPERATOR_IMAGE) BUNDLE_IMG=$(THIS_BUNDLE_IMAGE) make bundle && \
+		sed -i '/replaces:/d' ./bundle/manifests/oadp-operator.clusterserviceversion.yaml && \
+		IMG=$(THIS_OPERATOR_IMAGE) BUNDLE_IMG=$(THIS_BUNDLE_IMAGE) \
+		make docker-build docker-push bundle-build bundle-push && cd -
+	$(OPM) index add --container-tool docker --bundles $(PREVIOUS_BUNDLE_IMAGE),$(THIS_BUNDLE_IMAGE) --tag $(CATALOG_IMAGE)
+	docker push $(CATALOG_IMAGE)
+	# Run the following command to create the catalog
```
mateusoliveira43, line 449: I tried to make this part of the command, but make command and EOF were not working together :cry: 

If someone kwons how to automate this part

api/v1alpha1/oadp_types.go
```diff
@@ -318,7 +311,9 @@ type VolumeOptions struct {
 
 // Features defines the configuration for the DPA to enable the tech preview features
 type Features struct {
-	// Contains data mover specific configurations
+	// (do not use warning) Contains data mover specific configurations
```
mateusoliveira43, line 314: here is not a deprecating, because user can longer use the field, so no need to use something like this, right? 
https://github.com/openshift/oadp-operator/blob/master/controllers/nodeagent.go#L94-L99

Makefile
kaovilai: EOF don't work here. You can create a .txt or .yaml for this :)

Alternatively if really needed, this could be hacked into a variable but not worth it.

mateusoliveira43: thanks, created a yaml, but had to use an ugly syntax of echo, but at least it is fully automated now

controllers/validator.go
```diff
@@ -64,29 +61,14 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
-		if !VSMPluginPresent {
-			return false, errors.New("datamover is enabled, specify vsm as a default plugin")
+			return false, errors.New("Delete vsm from spec.configuration.velero.defaultPlugins, it is not needed for DataMover usage in OADP 1.3")
 		}
 	}
 
-	if !r.checkIfDataMoverIsEnabled(&dpa) && VSMPluginPresent {
-		return false, errors.New("datamover is disabled, remove vsm as a default plugin")
+	if dpa.Spec.Features != nil && dpa.Spec.Features.DataMover != nil {
+		return false, errors.New("Delete dataMover object from spec.features, it is not needed for DataMover usage in OADP 1.3")
```
weshayutin, line 71: It would be great to raise this reconcile error to the scrum team for review

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,79 @@
...
+-    restic:
+-      enable: true
++    nodeAgent:
++      enable: true
++      uploaderType: restic
+```
+
+> **Note:** OADP will be favoring Kopia over Restic in the near future.
+
+## Upgrade steps
```
weshayutin, line 41: same thing here, lets merge as is. Follow up w/ some doc changes. :) THANK YOU for including it

Makefile
```diff
@@ -423,6 +423,36 @@ catalog-build-replaces: opm ## Build a catalog image using replace mode
...
+		IMG=$(PREVIOUS_OPERATOR_IMAGE) BUNDLE_IMG=$(PREVIOUS_BUNDLE_IMAGE) \
+		make docker-build docker-push bundle-build bundle-push && cd -
+	cd $(TEMP)/current && IMG=$(THIS_OPERATOR_IMAGE) BUNDLE_IMG=$(THIS_BUNDLE_IMAGE) make bundle && \
+		sed -i '/replaces:/d' ./bundle/manifests/oadp-operator.clusterserviceversion.yaml && \
+		IMG=$(THIS_OPERATOR_IMAGE) BUNDLE_IMG=$(THIS_BUNDLE_IMAGE) \
+		make docker-build docker-push bundle-build bundle-push && cd -
+	$(OPM) index add --container-tool docker --bundles $(PREVIOUS_BUNDLE_IMAGE),$(THIS_BUNDLE_IMAGE) --tag $(CATALOG_IMAGE)
+	docker push $(CATALOG_IMAGE)
+	echo -e "apiVersion: operators.coreos.com/v1alpha1\nkind: CatalogSource\nmetadata:\n  name: oadp-operator-catalog-test-upgrade\n  namespace: openshift-marketplace\nspec:\n  sourceType: grpc\n  image: $(CATALOG_IMAGE)" > catalog-test-upgrade.yaml
+	$(OC_CLI) create -f catalog-test-upgrade.yaml
```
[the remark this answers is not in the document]
mateusoliveira43: tested and works :) going to add this in following commit, thanks!!

config/manifests/bases/oadp-operator.clusterserviceversion.yaml
mateusoliveira43: Can be done in another PR

docs/upgrade_1-2_to_1-3.md
mateusoliveira43: wrote the deletion code in `controllers/datamover.go`, it looks like this
```go
package controllers

import (
	"github.com/go-logr/logr"
	appsv1 "k8s.io/api/apps/v1"
	"k8s.io/apimachinery/pkg/api/errors"
	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
	"k8s.io/apimachinery/pkg/runtime/schema"
)

func (r *DPAReconciler) ReconcileDataMoverController(log logr.Logger) (bool, error) {
	// When updating from OADP version 1.2.x to 1.3.x, delete
	//   volume-snapshot-mover deployment
	//   volumesnapshotbackups.datamover.oadp.openshift.io CRD
	//   volumesnapshotrestores.datamover.oadp.openshift.io CRD
	// because OADP now uses Velero builtin DataMover instead of VSM/Volsync DataMover
	if err := r.Delete(r.Context, &appsv1.Deployment{ObjectMeta: metav1.ObjectMeta{
		Name:      "volume-snapshot-mover",
		Labels:    map[string]string{"openshift.io/oadp-data-mover": "True"},
	}}); err != nil && !errors.IsNotFound(err) {
		return false, err
	}

	volumeSnapshotBackupsResource := &unstructured.Unstructured{}
	volumeSnapshotBackupsResource.SetGroupVersionKind(schema.GroupVersionKind{
		Group:   "apiextensions.k8s.io",
		Version: "v1",
		Kind:    "CustomResourceDefinition",
	})
	volumeSnapshotBackupsResource.SetName("volumesnapshotbackups.datamover.oadp.openshift.io")
	if err := r.Delete(r.Context, volumeSnapshotBackupsResource); err != nil && !errors.IsNotFound(err) {
		return false, err
	}

	volumeSnapshotRestoresResource := &unstructured.Unstructured{}
	volumeSnapshotRestoresResource.SetGroupVersionKind(schema.GroupVersionKind{
		Group:   "apiextensions.k8s.io",
		Version: "v1",
		Kind:    "CustomResourceDefinition",
	})
	volumeSnapshotRestoresResource.SetName("volumesnapshotrestores.datamover.oadp.openshift.io")
	if err := r.Delete(r.Context, volumeSnapshotRestoresResource); err != nil && !errors.IsNotFound(err) {
		return false, err
	}

	return true, nil
}
```

plus adding this to `config/rbac/role.yaml`
```yaml
- apiGroups:
  - apiextensions.k8s.io
  verbs:
  - delete
  resources:
  - customresourcedefinitions
  resourceNames:
  - volumesnapshotbackups.datamover.oadp.openshift.io
  - volumesnapshotrestores.datamover.oadp.openshift.io
```

Can I commit this?

controllers/validator.go
mateusoliveira43: How url looks like in UI
Screenshot from 2023-10-24 11-38-33

docs/upgrade_1-2_to_1-3.md
kaovilai: We add finalizer to VSB/VSR [here](https://github.com/search?q=repo%3Amigtools%2Fvolume-snapshot-mover%20dmFinalizer&type=code)

kaovilai: We remove finalizer [here](https://github.com/search?q=repo%3Amigtools%2Fvolume-snapshot-mover%20RemoveFinalizer&type=code)

kaovilai: Haven't found where we call delete for the user yet.

mateusoliveira43: As talked in scrum, we will not delete CRDs because failed backup/restores can have VSB/VSR leftovers that have finalizers, and that can be trouble

We will add a `oc` CLI commands to users that want to delete them, in another PR

resolving conversation

controllers/validator.go
mateusoliveira43: Going to update message to have both information in next commit

weshayutin:
ci/prow/4.12-operator-e2e-azure may be suspect flaking

api/v1alpha1/oadp_types.go
shubham-pampattiwar: @mateusoliveira43 Why are we not removing the `DataMover` configuration ?

api/v1alpha1/oadp_types.go
```diff
@@ -339,10 +334,11 @@ type DataProtectionApplicationSpec struct {
 	//   - azurePluginImageFqin
 	//   - gcpPluginImageFqin
 	//   - csiPluginImageFqin
-	//   - dataMoverImageFqin
 	//   - resticRestoreImageFqin
 	//   - kubevirtPluginImageFqin
+	//   - operator-type
```
shubham-pampattiwar, line 339: Why do we need this `operator-type` key here ? What does it mean ?

controllers/validator.go
```diff
@@ -63,30 +60,14 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
-		}
-
-		if !VSMPluginPresent {
-			return false, errors.New("datamover is enabled, specify vsm as a default plugin")
-		}
-	}
-
-	if !r.checkIfDataMoverIsEnabled(&dpa) && VSMPluginPresent {
-		return false, errors.New("datamover is disabled, remove vsm as a default plugin")
+	if dpa.Spec.Features != nil && dpa.Spec.Features.DataMover != nil {
```
shubham-pampattiwar, line 69: This is not needed if we directly remover `Datamover` from `Features` , right ?

controllers/datamover.go
```diff
@@ -1,1073 +1,23 @@
...
-				log.Info("Found immutable selector from previous deployment, recreating DataMover Deployment")
-				err := r.Delete(r.Context, dataMoverDeployment)
-				if err != nil {
-					return false, err
-				}
-				return r.ReconcileDataMoverController(log)
-			}
-		}
-
+	// When updating from OADP version 1.2.x to 1.3.x, delete
```
shubham-pampattiwar, line 11: IMO if we are adding code to delete the volume-snapshot-mover controller deployment then we should also add logic to delete:
- the datamover restic secret created per BSL configured 
- the datamover volume options configmap

shubham-pampattiwar (commented):
Added a few comments

api/v1alpha1/oadp_types.go
mateusoliveira43: Because then user upgrading directly would not complete the upgrade (old DPA would not be valid and upgrade gets stuck)

api/v1alpha1/oadp_types.go
mateusoliveira43: just added the options this field has. The options it has https://github.com/openshift/oadp-operator/blob/master/api/v1alpha1/oadp_types.go#L61-L73

controllers/datamover.go
mateusoliveira43: Did not know about the configmap, can you explain to me how it gets created?

Will check if I can delete the secret

controllers/validator.go
mateusoliveira43: yes, but as we discussed, it better to left it there and tell user to remove it during DPA reconcile, otherwise, upgrade can break and get stuck 

reason https://github.com/openshift/oadp-operator/pull/1171#discussion_r1362431555

controllers/datamover.go
mateusoliveira43: checked and the created secrets by code are left in the cluster. Going to write code to remove them

The configmaps I did not see them yet. can you describe how they are created and how they look like?

mateusoliveira43: these configmaps, right?
https://github.com/openshift/oadp-operator/blob/master/controllers/datamover.go#L234-L244

mateusoliveira43: added code. can you test if the configMaps are deleted? I think they only get created if using `VolumeOptionsForStorageClasses` field in features.DataMover

kaovilai: You can add unit tests using https://book.kubebuilder.io/reference/envtest :)
[Example](https://github.com/mateusoliveira43/oadp-operator/blob/c5d06a9f4dfbe43010493b0b35e63e1cdebd5140/pkg/credentials/credentials_test.go#L502)

sseago (approved):

shubham-pampattiwar (approved):

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,79 @@
+# Upgrading from OADP 1.2
+
+> **NOTE:** Always upgrade to next minor version, do NOT skip versions. To update to higher version, do one upgrade at a time. Example: to upgrade from 1.1 to 1.3, upgrade first to 1.2, then to 1.3.
+## Changes from OADP 1.2 to 1.3
+
+- Velero was updated from version 1.11 to 1.12 (Changes reference: https://velero.io/docs/v1.12/upgrade-to-1.12/#upgrade-from-v110-or-higher)
+
+    From this update, OADP now uses Velero Built-in DataMover instead of VSM/Volsync DataMover. This change the following:
```
kaovilai, line 8:
```suggestion
    From this update, OADP now uses Velero Built-in DataMover instead of VSM/Volsync DataMover. This changes the following:
```

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,79 @@
...
+
+    - `spec.features.dataMover` (you can delete secret for `spec.features.dataMover.credentialName` from your cluster, if you want) and `vsm` plugin are not necessary anymore and you need to remove them from your DPA
```
kaovilai, line 10:
```suggestion
    - `spec.features.dataMover` (you can delete secret for `spec.features.dataMover.credentialName` from your cluster, if you want) and `vsm` plugin are not compatible anymore and you need to remove them from your DPA
```

docs/upgrade_1-2_to_1-3.md
```diff
@@ -0,0 +1,79 @@
...
+
+If you are using DataMover, you need to create new backups before upgrade, because the 1.2 DataMover backups will not work after upgrade. We suggest to create Restic backups.
+
+### Copy old DPA
+
+Save your current DataProtectionApplication (DPA) CustomResource config, be sure to remember the values.
+
+### Convert your DPA to the new version
+
+If you are using DataMover, you need to update with the new configuration. Example
```
kaovilai, line 53: It's unclear if this is example of new or old configuration from the wording

kaovilai (approved):
/lgtm

Nit for follow up

PR #1173, opened 2023-10-04
ci: Add new owners

mateusoliveira43:
Add new owners.

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -9,9 +9,17 @@ approvers:
   - kaovilai
   - eemcmullan
   - savitharaghunathan
+  - mateusoliveira43
+  - mrnold
+  - mpryc
+  - weshayutin
 reviewers:
   - sseago
   - shubham-pampattiwar
   - kaovilai
   - eemcmullan
   - savitharaghunathan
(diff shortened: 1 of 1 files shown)


OWNERS
```diff
@@ -9,9 +9,17 @@ approvers:
...
 reviewers:
   - sseago
   - shubham-pampattiwar
   - kaovilai
   - eemcmullan
   - savitharaghunathan
+  - mateusoliveira43
+  - mrnold
+  - mpryc
+  - weshayutin
```
mateusoliveira43, line 25: Did I forget any one? There is also an OWNERS_ALIASES file, need to update that as well?

kaovilai (approved):

weshayutin:
/hold 
for discussion

OWNERS
kaovilai: unless you have or will change username, this should be ok as is.. perhaps also discuss if someone would like to be removed to not be pinged about this.

mateusoliveira43:
Created issue about it https://github.com/openshift/oadp-operator/issues/1258
Please do not reopen this PR

PR #1174, opened 2023-10-05
OADP-2746: Update oadp-operator master branch for RESTIC && KOPIA

mpryc:
Adds kopia to the e2e tests and inclues some renaming to be consistent with nodeagent convention.

Some of the left over functions will be deprecated in 1.4.


tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -703,7 +711,14 @@ var _ = Describe("Configuration testing for DPA Custom Resource", func() {
 			if dpa.Spec.Configuration.Restic != nil && dpa.Spec.Configuration.Restic.PodConfig != nil {
```
mateusoliveira43, line 711: In this file, there are still old style restic configuration for DPA (`grep -Iinr '\.restic' tests`). I would update them to use nodeAgent syntax

tests/e2e/lib/dpa_helpers.go
```diff
@@ -72,7 +73,7 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) error {
 					DefaultPlugins: v.CustomResource.Spec.Configuration.Velero.DefaultPlugins,
 				},
 				NodeAgent: &oadpv1alpha1.NodeAgentConfig{
-					UploaderType: "restic",
+					UploaderType: string(backupRestoreType),
```
mateusoliveira43, line 76: This will not break? the accepted values are restic and kopia but the type can be csi, csi-datamover, restic or kopia

tests/e2e/dpa_deployment_suite_test.go
mpryc: Yes. I left the above intentionally. We are removing the old configuration in 1.4, so this change is not about removing it. We should still have test to cover what should be functional and it's going to be removed later.

tests/e2e/lib/apps.go
```diff
@@ -521,7 +521,7 @@ func VerifyBackupRestoreData(artifact_dir string, namespace string, routeName st
 	}
 
 	//Verifying backup-restore data only for CSI as of now.
-	if backupRestoretype == CSI || backupRestoretype == CSIDataMover || backupRestoretype == RESTIC {
+	if backupRestoretype == CSI || backupRestoretype == CSIDataMover || backupRestoretype == RESTIC || backupRestoretype == KOPIA {
```
mateusoliveira43, line 524: Not direct related to the PR, but I would remove the comment "Verifying backup-restore data only for CSI as of now" (it does not make sense to me) and remove the if, as it has all possible backupRestoretype

tests/e2e/lib/dpa_helpers.go
```diff
@@ -120,6 +121,12 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) error {
 		delete(defaultPlugins, oadpv1alpha1.DefaultPluginCSI)
 		delete(veleroFeatureFlags, "EnableCSI")
 		dpaInstance.Spec.SnapshotLocations = nil
+	case KOPIA:
```
mateusoliveira43, line 124: Can we merge this two cases and use your idea of `string(backupRestoreType)` for UploaderType?

tests/e2e/dpa_deployment_suite_test.go
mateusoliveira43: Agree
I would just update some logs that still say restic instead of Node Agent then  (`grep -Iinr 'log.Printf("Waiting for restic' tests`)

tests/e2e/lib/apps.go
mpryc: NIT, should remove in another PR as currently PROW is having hard time and don't want to have all the jobs restarted.

docs/TROUBLESHOOTING.md
```diff
@@ -214,13 +214,13 @@ This section includes how to debug a failed restore. For more specific issues re
 
 <hr style="height:1px;border:none;color:#333;"> 
 
-<h3 align="center">Issue with Backup/Restore of DeploymentConfig using Restic or restore hooks<a id="deployconfig"></a></h3>
+<h3 align="center">Issue with Backup/Restore of DeploymentConfig using Restic/Kopia or restore hooks<a id="deployconfig"></a></h3>
```
sseago, line 217: For 1.3, this also applies to datamover backups. Based on the updated logic in the controller, I would use this text:
"Issue with Backup/Restore of DeploymentConfig with volumes or restore hooks"

docs/TROUBLESHOOTING.md
```diff
@@ -214,13 +214,13 @@ This section includes how to debug a failed restore. For more specific issues re
...
 
--  (OADP 1.1+) **Error:** `DeploymentConfigs restore with spec.Replicas==0 or DC pods fail to restart if they crash if using Restic or restore hooks`
+-  (OADP 1.1+) **Error:** `DeploymentConfigs restore with spec.Replicas==0 or DC pods fail to restart if they crash if using Restic/Kopia or restore hooks`
```
sseago, line 219: For OADP 1.1-1.2, it applies to DCs with restic restores or restore hooks. For OADP 1.3+, it applies to DCs with volumes (regardless of backup method) or restore hooks

docs/TROUBLESHOOTING.md
```diff
@@ -214,13 +214,13 @@ This section includes how to debug a failed restore. For more specific issues re
...
 
     **Solution:**
     
-    This is expected behavior on restore if you are restoring DeploymentConfigs and are either using Restic for volume restore or you have post-restore hooks. The pod and DC plugins make these modifications to ensure that Restic and hooks work properly, and [dc-restic-post-restore.sh](../docs/scripts/dc-restic-post-restore.sh) should have been run immediately after a successful restore. Usage for this script is `dc-restic-post-restore.sh <restore-name>`
+    This is expected behavior on restore if you are restoring DeploymentConfigs and are either using Restic or Kopia for volume restore or you have post-restore hooks. The pod and DC plugins make these modifications to ensure that Restic or Kopia and hooks work properly, and [dc-fs-post-restore.sh](../docs/scripts/dc-fs-post-restore.sh) should have been run immediately after a successful restore. Usage for this script is `dc-fs-post-restore.sh <restore-name>`
```
sseago, line 223: "...if you are restoring DeploymentConfigs and have either volumes or post-restore hooks..."

docs/scripts/dc-fs-post-restore.sh
sseago, line 1: This should probably be called "dc-post-restore.sh" since it applies to more scenarios than just fs backup.

tests/e2e/backup_restore_suite_test.go
```diff
@@ -240,10 +240,10 @@ var _ = Describe("AWS backup restore tests", func() {
 			Expect(err).ToNot(HaveOccurred())
 			Expect(succeeded).To(Equal(true))
 
-			if brCase.BackupRestoreType == RESTIC && nsRequiresResticDCWorkaround {
-				// run the restic post restore script if restore type is RESTIC
-				log.Printf("Running restic post restore script for case %s", brCase.Name)
-				err = RunResticPostRestoreScript(restoreName)
+			if brCase.BackupRestoreType == RESTIC || brCase.BackupRestoreType == KOPIA && nsRequiresResticDCWorkaround {
+				// run the restic /kopia post restore script if restore type is RESTIC or KOPIA
```
sseago, line 244: This needs to be run any time you have DCs with volumes attached, regardless of B/R type.

sseago (changes requested):

docs/TROUBLESHOOTING.md
mpryc: Is this the actual error from the OCP object itself ? I am asking, because if yes I need to match what's in the logs/error.

sseago: It's not an "error" in the sense of a message/status. It's an error in the sense that behavior doesn't match expectations. For DCs with either volumes or hooks, our restore plugin disconnects pod from DC, scaling down DC and restoring bare pod. This means post-restore, application is running, data is restored, but they can't redeploy via DC. What the script does is delete these bare pods and scales DC back up to the original replicas value.

tests/e2e/backup_restore_suite_test.go
mpryc: @sseago could you please be more precise here? My understanding from your comment is that I should remove check for the BackupRestoreType and extend nsRequiresResticDCWorkaround to include ones that has volumes attached only? What about the backwards compatibility, or it doesn't matter as it won't be backported to 1.2 ?

sseago: This is a change in behavior for 1.3 vs. 1.2. For e2e tests, we're only ever working with backups from the same velero version, so backwards compatibility isn't a concern within the test. As for the plugin changes, the functionality change involves setting additional annotations on backup, so the restore plugin uses the presence of those annotations to trigger the changed behavior, so it should all be backwards-compatible with restoring 1.2 backups.

tests/e2e/lib/apps.go
mpryc: I will leave backupRestoretype here as it may be some null etc.. so prefer to be on a safe side

mpryc: Removed the comment and updated as well the dpa_deployment_suite_test.go to address #1176. Still some restic in logs is left, but only in the places where restic is directly consumed.

docs/TROUBLESHOOTING.md
mpryc: Changed.

docs/TROUBLESHOOTING.md
mpryc: I think that's sorted. I have changed this doc and added OADP 1.3+ which have reference to previous solution with different pre-condition.

docs/TROUBLESHOOTING.md
mpryc: changed.

docs/scripts/dc-fs-post-restore.sh
mpryc: Renamed and left the symlink to the original filename so other documents outside of this project are also valid.

tests/e2e/backup_restore_suite_test.go
```diff
@@ -192,14 +192,13 @@ var _ = Describe("AWS backup restore tests", func() {
 			err = brCase.PreBackupVerify(dpaCR.Client, brCase.ApplicationNamespace)
 			Expect(err).ToNot(HaveOccurred())
 
-			nsRequiresResticDCWorkaround, err := NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.ApplicationNamespace)
-			Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
mpryc: From our discussions this script should work regardless of it's conditions, so I don't think so. Error from the script run should catch issues here in a first place and there is a condition for a script run to expect no errors.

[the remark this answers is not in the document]
kaovilai: would be a good test to make sure the script doesn't break non restic dc case I guess.

tests/e2e/must-gather_suite_test.go
```diff
@@ -159,14 +159,13 @@ var _ = Describe("Must-gather backup restore tests", func() {
 			err = brCase.PreBackupVerify(dpaCR.Client, brCase.ApplicationNamespace)
 			Expect(err).ToNot(HaveOccurred())
 
-			nsRequiresResticDCWorkaround, err := NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.ApplicationNamespace)
-			Expect(err).ToNot(HaveOccurred())
+			nsRequiresResticDCWorkaround, _ := NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.ApplicationNamespace)
```
mateusoliveira43, line 162: do not change, to let it equal to `tests/e2e/backup_restore_suite_test.go`

mateusoliveira43:
Did not run locally, but it seems all the errors in running the tests were "Velero restore error: rpc error: code = Unknown desc = *pod.BackupPlugin is not a restore item action"

mpryc:
> Did not run locally, but it seems all the errors in running the tests were "Velero restore error: rpc error: code = Unknown desc = *pod.BackupPlugin is not a restore item action"

@mateusoliveira43 I think it's unrelated to this PR, please check my other PR which also fails in this:
   https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1193/pull-ci-openshift-oadp-operator-master-4.13-operator-e2e-aws/1717279805713420288

mateusoliveira43 (approved):
Ran the new backup/restore tests locally and the 2 vol that is failing CI (probably timeout)

docs/TROUBLESHOOTING.md
```diff
@@ -214,13 +214,19 @@ This section includes how to debug a failed restore. For more specific issues re
 
 <hr style="height:1px;border:none;color:#333;"> 
 
-<h3 align="center">Issue with Backup/Restore of DeploymentConfig using Restic or restore hooks<a id="deployconfig"></a></h3>
+<h3 align="center">Issue with Backup/Restore of DeploymentConfig with volumes or restore hooks<a id="deployconfig"></a></h3>
```
kaovilai, line 217:
```suggestion
<h3 align="center">Issue with Backup/Restore of DeploymentConfig with volumes via file system backup or restore hooks<a id="deployconfig"></a></h3>
```

sseago: @kaovilai It's not specific to fs backup anymore -- that's why I suggested removing the restic/kopia reference. This is now required for any DCs with volumes, since datamover also needs the pods restored.

sseago (approved):

weshayutin:
Check the results here: https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1174/pull-ci-openshift-oadp-operator-master-4.13-operator-e2e-aws/1717646231989129216/artifacts/operator-e2e-aws/e2e/build-log.txt

LGTM

Unrelated to this PR, but our velero describe / log debugging is no longer working.  https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1174/pull-ci-openshift-oadp-operator-master-4.13-operator-e2e-aws/1717646231989129216/artifacts/operator-e2e-aws/e2e/artifacts/mongo-datamover-e2e/must-gather/quay-io-konveyor-oadp-must-gather-sha256-17c270de0bed794ec43a7108b646e70e199c08b37e49a663e4f1e43cde587c55/must-gather/clusters/4c7bcc45/namespaces/openshift-adp/oadp.openshift.io/dpa-ts-velero-test/

This update looks great through.. MERGE IT

weshayutin (approved):

PR #1177, opened 2023-10-10
OADP-659 Avoid overwriting good hostname.

mrnold:
Fix an issue with StripDefaultPorts that breaks a new test case being added for https://github.com/openshift/oadp-operator/pull/1169.

### DIFF
--- a/pkg/common/common.go
+++ b/pkg/common/common.go
@@ -213,6 +213,8 @@ func StripDefaultPorts(fromUrl string) (string, error) {
 		URL: u,
 	}
 	request.SanitizeHostForHeader(&r)
-	r.URL.Host = r.Host
+	if r.Host != "" {
+		r.URL.Host = r.Host
+	}
 	return r.URL.String(), nil
 }


pkg/common/common.go
```diff
@@ -213,6 +213,8 @@ func StripDefaultPorts(fromUrl string) (string, error) {
 		URL: u,
 	}
 	request.SanitizeHostForHeader(&r)
-	r.URL.Host = r.Host
+	if r.Host != "" {
```
mateusoliveira43, line 216: I would add unit tests to some scenarios (IBM, Noobaa, etc) here https://github.com/openshift/oadp-operator/blob/master/pkg/common/common_test.go to avoid regressions

kaovilai:
```
INFO[2023-10-10T15:20:47Z] Running step operator-e2e-aws-e2e.           
INFO[2023-10-10T16:16:45Z] Step operator-e2e-aws-e2e succeeded after 55m58s. 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1177/pull-ci-openshift-oadp-operator-master-4.12-operator-e2e-aws/1711749018561810432#1:build-log.txt%3A76
/override ci/prow/4.12-operator-e2e-aws

/test 4.13-operator-e2e-azure

pkg/common/common.go
mateusoliveira43: Tested, and this fixes my use case :smile: 

Only concern I have with follow up PR is that we need this in 1.2 branch right? then will be 3 cherry-picks

shubham-pampattiwar (approved):

pkg/common/common.go
kaovilai: 3 cherry-picks to oadp-1.2 and oadp 1.1 and what's the other one?

But yeah follow up pr is fine.

kaovilai (approved):
/lgtm

pkg/common/common.go
kaovilai: - [ ] add unit tests to some scenarios (IBM, Noobaa, etc) here https://github.com/openshift/oadp-operator/blob/master/pkg/common/common_test.go to avoid regressions

mateusoliveira43: oh no, I think no need to 1.1, only for 1.2

it would be form this PR, the previous (https://github.com/openshift/oadp-operator/pull/1169) and the follow up (3, but for the same branch)

sseago (approved):

kaovilai:
@mrnold cherrypick #1169 and #1177 into oadp-1.2 :) and move https://issues.redhat.com/browse/OADP-659 status accordingly.

PR #1179, opened 2023-10-10
OADP-2686: add ACM resources to restore priority

sseago:


shubham-pampattiwar (approved):

mateusoliveira43 (approved):

pkg/velero/server/config.go
```diff
@@ -38,7 +38,7 @@ type ServerConfig struct {
...
+	// Desired order of resource restores, the priority list contains two parts which are split by "-" element. The resources before "-" element are restored first as high priorities, the resources after "-" element are restored last as low priorities, and any resource not in the list will be restored alphabetically between the high and low priorities. (default securitycontextconstraints,customresourcedefinitions,namespaces,managedcluster.cluster.open-cluster-management.io,managedcluster.clusterview.open-cluster-management.io,klusterletaddonconfig.agent.open-cluster-management.io,managedclusteraddon.addon.open-cluster-management.io,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io)
```
kaovilai, line 41: Any thoughts putting this into oadp_types.go using kubebuilder validation markers

[`// +kubebuilder:default:=<any>`](https://book.kubebuilder.io/reference/markers/crd-validation.html#:~:text=//%20%2Bkubebuilder%3Adefault)

mateusoliveira43: Wouldn't we need to change the code logic for that? https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L440

kaovilai: meh. they could co-exist. but yes the go code to add default would be obsolete by the api-server doing the work for us if empty.

sseago: Seems like it might be confusing to have 2 functional default lists in different formats (although we almost have that now, it's just that the above list is only in documentation). I've lost track to some degree of the flow here as to how this particular field's value interacts with the default setting in L440 linked above, though. Would adding a kubebuilder default mean that the velero controller line above (and the structured default list) should be removed entirely?

sseago: Also, since this is a bugfix, if we want to refactor the way we handle defaults for this, that should probably be a separate PR, and this one should just add the new types, that makes the history easier to follow in terms of noting that we added types here. If we refactor at the same time, the fact that new types were added will be obscured.

mateusoliveira43:
I think CI was flake https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1179/pull-ci-openshift-oadp-operator-master-4.11-operator-e2e-aws/1712080406334935040/artifacts/operator-e2e-aws/e2e/build-log.txt (at the end, says success) If someone wants to override

kaovilai:
```
INFO[2023-10-11T14:01:41Z] Step operator-e2e-aws-e2e succeeded after 56m49s. 
```
https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1179/pull-ci-openshift-oadp-operator-master-4.11-operator-e2e-aws/1712080406334935040#1:build-log.txt%3A77
/override ci/prow/4.11-operator-e2e-aws

PR #1197, opened 2023-10-24
Update velero dep after rebase 2023-10-24

sseago:


weshayutin (dismissed):

kaovilai (approved):

mateusoliveira43 (changes requested):
Changing the error to warning did not fix E2E tests, because we do a regex filter by the word error (here https://github.com/openshift/oadp-operator/blob/master/tests/e2e/lib/velero_helpers.go#L194) and the warning message has error on it.

if we can change the regex to something like
```go
errorRegex, err := regexp.Compile("level=error") // only check for error logs
```
would be more appropriate

kaovilai (approved):
/lgtm

PR #1205, opened 2023-10-28
fix: One kubernetes client per E2E suite run

mateusoliveira43:
After talking to @shawn-hurley, we should just have one kubernetes client per suite run, but we were creating more than one per test. With this PR, all necessary clients are created in TOP-LEVEL before-suite and passed along where necessary. Refactor can be done, but not the goal of this PR.

This is important, because with each client created,  a new cache would also be created. This could create memory problems.

With only one client, we could have out of date cache, but if this happens, we can disable cache from that client.

Also, increased the default QPS (20) to 50 and Burst (30) to 100 in client configuration.

### Other changes

Change all `context.TODO()` in tests code for `context.Background()`. This should not change the behavior of tests, since it does not run in parallel today.

Make colored output an option in suite run, so logs in PROW are easier to read.

### DIFF
--- a/docs/developer/testing/TESTING.md
+++ b/docs/developer/testing/TESTING.md
@@ -20,6 +20,7 @@ To get started, you need to provide the following **required** environment varia
 | `VSL_REGION` | The region of snapshotLocations | - | true |
 | `BSL_REGION` | The region of backupLocations | `us-east-1` | false |
 | `OADP_TEST_NAMESPACE` | The namespace where OADP operator is installed | `openshift-adp` | false |
+| `OPENSHIFT_CI` | Disable colored output from tests suite run | `true` | false |
 
 The expected format for `OADP_CRED_FILE` and `CI_CRED_FILE` files is:
 ```
@@ -38,6 +39,7 @@ export VSL_REGION=<snapshotLocations_region>
 # non required
 export BSL_REGION=<backupLocations_region>
 export OADP_TEST_NAMESPACE=<test_namespace>
+export OPENSHIFT_CI=false
 ```
 
 It is also possible to set the environment variables during make command call. Example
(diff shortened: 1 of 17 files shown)


kaovilai (commented):
Why not have all the helper functions use a client getter?

mateusoliveira43:
@kaovilai I wanted to do the fastest approach possible, so did the most ugly way

I think your idea would be best, like we do here https://github.com/openshift/oadp-operator/blob/c87d3d3ceba46b55b6e2864fb9764360d1fc71bc/tests/e2e/lib/dpa_helpers.go#L266-L284

docs/developer/testing/TESTING.md
```diff
@@ -20,6 +20,7 @@ To get started, you need to provide the following **required** environment varia
 | `VSL_REGION` | The region of snapshotLocations | - | true |
 | `BSL_REGION` | The region of backupLocations | `us-east-1` | false |
 | `OADP_TEST_NAMESPACE` | The namespace where OADP operator is installed | `openshift-adp` | false |
+| `OPENSHIFT_CI` | Disable colored output from tests suite run | `true` | false |
```
mrnold, line 23: It would be helpful to change this name to something that matches what it actually does, "OPENSHIFT_CI" sounds kind of vague. The name would make more sense if there were other options that it controlled, to indicate the tests are running in some kind of "OpenShift-CI mode". But if it only does this one thing, it will be less confusing to call it something else.

mateusoliveira43: The only other place it is used is here https://github.com/openshift/oadp-operator/blob/master/tests/e2e/scripts/azure_settings.sh#L9-L11(did not use azure yet, I do not understand what this does)

I can create a separate environment variable for color

weshayutin:
@kaovilai if you have a second to do one last review that would be ++ thank you!

kaovilai (approved):

sseago (approved):

shubham-pampattiwar (approved):
/lgtm

PR #1209, opened 2023-10-30
fix: dc post restore script logic to delete disconnected pods

mateusoliveira43:
Add `--all-namespaces` flag to delete command. This should make CI more stable.


sseago (approved):

shubham-pampattiwar (approved):

mpryc:
> We changed the name of the script, but not the prow config... can someone trigger e2e run for me in this PR?

There is symlink to the old one to ensure it's compatible, because it's mentioned in many places in docs even outside of OADP repo.

Issue #1211, opened 2023-11-02
VSB and VSR need to be removed from the master branch, they are unnecessary in OADP 1.3+.


mpryc:
VSB and VSR need to be removed from the master branch. They are unnecessary in OADP 1.3+, but they also don't pose any harm. In the absence of the VSB and VSR in the CRD's / OCP APIs, the description will not be displayed, so adding extra description won't create any issues. For OADP 1.2, this will be a straightforward backport from the master branch because it contains VSB and VSR, which are required in OADP 1.2.

_Originally posted by @mpryc in https://github.com/openshift/oadp-operator/issues/1200#issuecomment-1790860811_

mateusoliveira43:
please close this issue, it is already fixed

PR #1212, opened 2023-11-03
Remove VSB and VSR from OADP 1.3+

mpryc:
Since VSB and VSR is not needed in OADP 1.3+ removing it from the CSV.

Updated API_ref and script to not show errors. E.g. when the ocp is missing those APIs, however left the functions in the datamover_resources.sh as one may want to run it against earlier OADP versions.


kaovilai (approved):

mateusoliveira43:
fixes https://github.com/openshift/oadp-operator/issues/1211

mateusoliveira43 (approved):

shubham-pampattiwar:
Shouldn't we also remove it from bundle/manifests ?

shubham-pampattiwar:
if that's already done or you want to do it in a follow up PR, feel free to merge this existing PR.

mateusoliveira43:
@shubham-pampattiwar thats already done, otherwise unit test would have failed

I think because that CRD does not exist in master, it does not go to bundle even running `make bundle`

shubham-pampattiwar:
Thanks @mate usoliveira43

PR #1217, opened 2023-11-06
fix: [oadp-1.2] Remove unused code in OADP 1.2

mateusoliveira43:
Remove references to 1.3 CRDs in 1.2 branch.


rayfordj:
@mateusoliveira43 , should this PR also contain these changes to the CSV generated by executing `make bundle`?

mateusoliveira43:
@rayfordj usually yes, but since those CRDs do not exist on this branch, `make bundle` does not change `bundle` folder with entries related to them

One way to assure us that, is the unit test jobs, that check if bundle is up to date

rayfordj:
> @rayfordj usually yes, but since those CRDs do not exist on this branch, `make bundle` does not change `bundle` folder with entries related to them
> 
> One way to assure us that, is the unit test jobs, that check if bundle is up to date

:sweat_smile:  Thanks for explanation.  I'm so used to seeing new things and it caught me off that didn't see it change too.

/lgtm
/approve

rayfordj (approved):

kaovilai (approved):

PR #1223, opened 2023-11-10
fix: Refactor and removal of duplications in E2E tests

mateusoliveira43:
Improve E2E tests code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -514,6 +514,7 @@ test-e2e: test-e2e-setup install-ginkgo
 	-timeout_multiplier=$(E2E_TIMEOUT_MULTIPLIER) \
 	-artifact_dir=$(ARTIFACT_DIR) \
 	-oc_cli=$(OC_CLI) \
+	--ginkgo.vv \
 	--ginkgo.no-color=$(OPENSHIFT_CI) \
 	--ginkgo.label-filter="$(TEST_FILTER)" \
 	--ginkgo.timeout=2h
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -11,7 +11,6 @@ import (
 
 	"github.com/google/uuid"
 	. "github.com/onsi/ginkgo/v2"
-	"github.com/onsi/ginkgo/v2/types"
 	. "github.com/onsi/gomega"
 	. "github.com/openshift/oadp-operator/tests/e2e/lib"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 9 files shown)


tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -465,10 +619,7 @@ var _ = Describe("Configuration testing for DPA Custom Resource", func() {
 			},
 			WantError: false,
 		}, nil),
-	}
-
-	awsTests := []TableEntry{
-		Entry("AWS Without Region No S3ForcePathStyle with BackupImages false should succeed", Label("aws"), InstallCase{
+		Entry("AWS Without Region No S3ForcePathStyle with BackupImages false should succeed", Label("aws", "ibmcloud"), InstallCase{
```
mateusoliveira43, line 622: TODO test if developers in ibm environment can really run these tests

mateusoliveira43: BSL will not be available, because no s3 url is passed, but that not a problem, we check only DPA in these tests

Makefile
```diff
@@ -514,6 +514,7 @@ test-e2e: test-e2e-setup install-ginkgo
 	-timeout_multiplier=$(E2E_TIMEOUT_MULTIPLIER) \
 	-artifact_dir=$(ARTIFACT_DIR) \
 	-oc_cli=$(OC_CLI) \
+	--ginkgo.vv \
```
kaovilai, line 517:
```
  --vv 
    If set, emits with maximal verbosity - includes skipped and pending tests.
```

tests/e2e/backup_restore_suite_test.go
```diff
@@ -56,245 +56,235 @@ func mysqlReady(preBackupState bool, twoVol bool, backupRestoreType BackupRestor
...
-		ApplicationTemplate  string
-		PvcSuffixName        string
-		ApplicationNamespace string
-		Name                 string
-		BackupRestoreType    BackupRestoreType
-		PreBackupVerify      VerificationFunction
-		PostRestoreVerify    VerificationFunction
-		AppReadyDelay        time.Duration
-		MaxK8SVersion        *K8sVersion
-		MinK8SVersion        *K8sVersion
```
kaovilai, line 76: These two are removed from struct. I would add this back so don't have to reengineer this in the future

kaovilai (changes requested):

Makefile
mateusoliveira43: There are some drawbacks of using `vv`, but I felt is better. Without it, is hard to see in CI logs where a test begins and ends. But I am open for discussion

tests/e2e/backup_restore_suite_test.go
mateusoliveira43: I think this is still used in older versions of OCP (like 4.9 in 1.1 branch). But I will not cherry-pick these changes to any branch, so I think its safe to remove from master (which lowest OCP version is 4.11/4.12). But I can add back as well

kaovilai: Safe to remove yup.. and yeah can always look to prior commits to add it back in the future if needed.. resolving.

kaovilai (approved):

PR #1235, opened 2023-11-22
OADP-5158: Do not allow restic config use in OADP 1.5+; Deprecate restic for file system backups

mateusoliveira43:
## Description

This PR removes logic of using old restic field (it should use nodeAgent field instead) from 1.5+ code. But it does not remove old restic field from DPA (which should only be removed when DPA version is updated), for a safe upgrade and better user experience for the users.

Also, log a Deprecation Warning when using restic in NodeAgent config.

### OADP 1.5

run `make deploy-olm` and test that
- DPA is Reconciled false if old restic field (`spec.configuration.restic`) is present
- OADP manager pod has 1 warning log and OADP namespace has 1 warning event if DPA `spec.configuration.nodeAgent.uploaderType` is set to `restic`

### Simulate a upgrade

Simulate a upgrade from OADP 1.4 to 1.5 (this PR branch), by 
- running `make catalog-test-upgrade`
- installing OADP 1.4
- creating a DPA with old restic field (wait until it is reconciled true)
- and update to OADP 1.5

[More info](https://github.com/openshift/oadp-operator/blob/master/docs/developer/testing/test_oadp_version_upgrade.md)

Update should finish successfully and DPA should:
- be reconciled false, with the error message that restic field is not allowed
- restic fields should be still in DPA

### DIFF
--- a/api/v1alpha1/dataprotectionapplication_types.go
+++ b/api/v1alpha1/dataprotectionapplication_types.go
@@ -374,10 +374,8 @@ type ResticConfig struct {
 // ApplicationConfig defines the configuration for the Data Protection Application
 type ApplicationConfig struct {
 	Velero *VeleroConfig `json:"velero,omitempty"`
-	// (deprecation warning) ResticConfig is the configuration for restic DaemonSet.
-	// restic is for backwards compatibility and is replaced by the nodeAgent
-	// restic will be removed in the future
-	// +kubebuilder:deprecatedversion:warning=1.3
+	// (do not use warning) restic field is for backwards compatibility and
+	// will be removed in the future. Use nodeAgent field instead
 	// +optional
 	Restic *ResticConfig `json:"restic,omitempty"`
 
@@ -761,8 +759,6 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
 		if dpa.Spec.Configuration != nil {
 			if dpa.Spec.Configuration.NodeAgent != nil && len(dpa.Spec.Configuration.NodeAgent.Timeout) > 0 {
 				fsBackupTimeout = dpa.Spec.Configuration.NodeAgent.Timeout
-			} else if dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
-				fsBackupTimeout = dpa.Spec.Configuration.Restic.Timeout
 			}
 		}
 		if pvOperationTimeout, err := time.ParseDuration(fsBackupTimeout); err == nil && dpa.Spec.Configuration.Velero.Args.PodVolumeOperationTimeout == nil {
--- a/config/samples/kustomization.yaml
+++ b/config/samples/kustomization.yaml
@@ -8,7 +8,6 @@ resources:
 - downloadrequest.yaml
 - podvolumebackup.yaml
 - podvolumerestore.yaml
-- resticrepository.yaml
 - backup.yaml
 - restore.yaml
 - schedule.yaml
--- a/internal/controller/nodeagent.go
+++ b/internal/controller/nodeagent.go
@@ -371,9 +371,8 @@ func (r *DataProtectionApplicationReconciler) customizeNodeAgentDaemonset(ds *ap
 		return nil, err
 	}
 
-	if err := credentials.AppendCloudProviderVolumes(dpa, ds, providerNeedsDefaultCreds, hasCloudStorage); err != nil {
-		return nil, err
-	}
+	credentials.AppendCloudProviderVolumes(dpa, ds, providerNeedsDefaultCreds, hasCloudStorage)
+
 	setPodTemplateSpecDefaults(&ds.Spec.Template)
 	if ds.Spec.UpdateStrategy.Type == appsv1.RollingUpdateDaemonSetStrategyType {
 		ds.Spec.UpdateStrategy.RollingUpdate = &appsv1.RollingUpdateDaemonSet{
--- a/internal/controller/nodeagent.go
+++ b/internal/controller/nodeagent.go
@@ -16,7 +16,6 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/utils/pointer"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
@@ -27,7 +26,6 @@ import (
 )
 
 const (
-	ResticRestoreHelperCM   = "restic-restore-action-config"
 	FsRestoreHelperCM       = "fs-restore-action-config"
 	HostPods                = "host-pods"
 	HostPlugins             = "host-plugins"
@@ -37,7 +35,6 @@ const (
 	IBMCloudPVHostPath      = "/var/data/kubelet/pods"
 	GenericPluginsHostPath  = "/var/lib/kubelet/plugins"
 	IBMCloudPluginsHostPath = "/var/data/kubelet/plugins"
-	ResticPVHostPathEnvVar  = "RESTIC_PV_HOSTPATH"
 	FSPVHostPathEnvVar      = "FS_PV_HOSTPATH"
 	PluginsHostPathEnvVar   = "PLUGINS_HOSTPATH"
 )
@@ -61,10 +58,6 @@ func getFsPvHostPath(platformType string) string {
 	// Check if environment variables are set for host paths
 	if envFs := os.Getenv(FSPVHostPathEnvVar); envFs != "" {
 		return envFs
-	}
-
-	if env := os.Getenv(ResticPVHostPathEnvVar); env != "" {
-		return env
 	}
 
 	// Return platform-specific host paths
@@ -108,16 +101,7 @@ func (r *DataProtectionApplicationReconciler) ReconcileNodeAgentDaemonset(log lo
 		ObjectMeta: getNodeAgentObjectMeta(r),
 	}
 
-	if dpa.Spec.Configuration.Restic != nil {
-		// V(-1) corresponds to the warn level
-		var deprecationMsg string = "(Deprecation Warning) Use nodeAgent instead of restic, which is deprecated and will be removed in the future"
-		log.V(-1).Info(deprecationMsg)
-		r.EventRecorder.Event(dpa, corev1.EventTypeWarning, "DeprecationResticConfig", deprecationMsg)
-	}
-
-	if dpa.Spec.Configuration.Restic != nil && dpa.Spec.Configuration.Restic.Enable != nil && *dpa.Spec.Configuration.Restic.Enable {
-		deleteDaemonSet = false
-	} else if dpa.Spec.Configuration.NodeAgent != nil && dpa.Spec.Configuration.NodeAgent.Enable != nil && *dpa.Spec.Configuration.NodeAgent.Enable {
+	if dpa.Spec.Configuration.NodeAgent != nil && dpa.Spec.Configuration.NodeAgent.Enable != nil && *dpa.Spec.Configuration.NodeAgent.Enable {
 		deleteDaemonSet = false
 	}
 
@@ -132,11 +116,9 @@ func (r *DataProtectionApplicationReconciler) ReconcileNodeAgentDaemonset(log lo
 			}
 			return false, err
 		}
-		// no errors means there already is an existing DaeMonset.
+		// no errors means there is already an existing DaemonSet.
 		// TODO: Check if NodeAgent is in use, a backup is running, so don't blindly delete NodeAgent.
-		// If dpa.Spec.Configuration.NodeAgent enable exists and is false, attempt to delete.
-		deleteOptionPropagationForeground := metav1.DeletePropagationForeground
-		if err := r.Delete(deleteContext, ds, &client.DeleteOptions{PropagationPolicy: &deleteOptionPropagationForeground}); err != nil {
+		if err := r.Delete(deleteContext, ds, &client.DeleteOptions{PropagationPolicy: ptr.To(metav1.DeletePropagationForeground)}); err != nil {
 			// TODO: Come back and fix event recording to be consistent
 			r.EventRecorder.Event(ds, corev1.EventTypeNormal, "DeleteDaemonSetFailed", "Got DaemonSet to delete but could not delete err:"+err.Error())
 			return false, err
@@ -214,24 +196,16 @@ func (r *DataProtectionApplicationReconciler) ReconcileNodeAgentDaemonset(log lo
  */
 func (r *DataProtectionApplicationReconciler) buildNodeAgentDaemonset(ds *appsv1.DaemonSet) (*appsv1.DaemonSet, error) {
 	dpa := r.dpa
-	if dpa == nil {
-		return nil, fmt.Errorf("dpa cannot be nil")
-	}
+
 	if ds == nil {
-		return nil, fmt.Errorf("ds cannot be nil")
+		return nil, fmt.Errorf("DaemonSet cannot be nil")
 	}
 
 	var nodeAgentResourceReqs corev1.ResourceRequirements
 
 	// get resource requirements for nodeAgent ds
 	// ignoring err here as it is checked in validator.go
-	if dpa.Spec.Configuration.Restic != nil {
-		nodeAgentResourceReqs, _ = getResticResourceReqs(dpa)
-	} else if dpa.Spec.Configuration.NodeAgent != nil {
-		nodeAgentResourceReqs, _ = getNodeAgentResourceReqs(dpa)
-	} else {
-		return nil, fmt.Errorf("NodeAgent or Restic configuration cannot be nil")
-	}
+	nodeAgentResourceReqs, _ = getNodeAgentResourceReqs(dpa)
 
 	installDs := install.DaemonSet(ds.Namespace,
 		install.WithResources(nodeAgentResourceReqs),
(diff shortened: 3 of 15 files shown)


Makefile
```diff
@@ -331,7 +331,6 @@ nullables:
 nullable-crds-bundle: DPA_SPEC_CONFIG_PROP = .spec.versions.0.schema.openAPIV3Schema.properties.spec.properties.configuration.properties
 nullable-crds-bundle: PROP_RESOURCE_ALLOC = properties.podConfig.properties.resourceAllocations
 nullable-crds-bundle: VELERO_RESOURCE_ALLOC = $(DPA_SPEC_CONFIG_PROP).velero.$(PROP_RESOURCE_ALLOC)
-nullable-crds-bundle: RESTIC_RESOURCE_ALLOC = $(DPA_SPEC_CONFIG_PROP).restic.$(PROP_RESOURCE_ALLOC)
```
kaovilai, line 334: do we need nodeagent version of this to replace it?

mateusoliveira43: maybe yes, we do not have that for node agent today

mateusoliveira43: looking the code, I would delete things from makefile related to nullables and add to code a duplication of the object we use in `oadp_types.go`, with the necessary modifications

example:
```go
type OADPResourceRequirements struct {
	// Limits describes the maximum amount of compute resources allowed.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
	// +optional
       // +nullable
	Limits corev1.ResourceList `json:"limits,omitempty" protobuf:"bytes,1,rep,name=limits,casttype=ResourceList,castkey=ResourceName"`
	// Requests describes the minimum amount of compute resources required.
	// If Requests is omitted for a container, it defaults to Limits if that is explicitly specified,
	// otherwise to an implementation-defined value.
	// More info: https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/
	// +optional
       // +nullable
	Requests corev1.ResourceList `json:"requests,omitempty" protobuf:"bytes,2,rep,name=requests,casttype=ResourceList,castkey=ResourceName"`
}
```

what you think?

kaovilai: Works for me.. as long as we keep up with changes to resourceRequirement types in k8s if it gets updated.

mpryc (commented):
Overall ok, however is there any deprecation error when the user tries to use unsupported old configuration?

What happens if one upgrades OADP from 1.3 to that change. Shoud this be in a release notes or an informative error on the OCP object that it's misconfigured. I am less worried about new config options as they won't be allowed by the OCP itself, right?

controllers/validator_test.go
```diff
@@ -271,7 +271,7 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
 								oadpv1alpha1.DefaultPluginAWS,
 							},
 							PodConfig: &oadpv1alpha1.PodConfig{
-								ResourceAllocations: corev1.ResourceRequirements{
```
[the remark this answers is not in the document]
mateusoliveira43: discussion here https://github.com/openshift/oadp-operator/pull/1235#discussion_r1406576752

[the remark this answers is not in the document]
mpryc: Ok makes sense.

mateusoliveira43:
@mpryc I think I need to add an error like we did with feature.datamover to 1.4 updates not break. Will test this later

mateusoliveira43:
@mpryc tested upgrade from 1.3 to master (1.4) using `make catalog-test-upgrade` and it did not break the upgrade if I use resticConfig in 1.3 (do not know why, maybe because of this https://github.com/openshift/oadp-operator/blob/oadp-1.3/api/v1alpha1/oadp_types.go#L165 ?)

But it does remove resticConfig from DPA after upgrade, removing nodeAgent pods as well, for example. Is this ok? We have added deprecated notice in 1.3 about this field, so user may have seen it, or may not. If user does not check things after upgrade, it may get an error trying to do a restic backup afterwards, as his DPA is not confugured for filesystem backup anymore.

@shawn-hurley you helped us with similar problem about upgrading from 1.2 to 1.3 last time, what you think?

kaovilai:
@shawn-hurley the deprecation warning should've already been added in oadp-1.3 per [design](https://github.com/openshift/oadp-operator/pull/1101/files#diff-ca18723cc9071a694da8da994ecf31cd090d740d469ae80a0ba22f4908101631R89), this PR is for 1.4 removal.

See also: https://github.com/openshift/oadp-operator/pull/1174#discussion_r1347374829

kaovilai:
warning implemented in https://github.com/openshift/oadp-operator/pull/1102/files#diff-e186f96e971d9f53a2614ad247e98c7313a2fcc086cb9f6243d776e1e540fe73R173

mateusoliveira43:
Going to undo removal of fields from yaml

we have something similar to old datamover in 1.3: restic field is available in 1.4, but DPA does not reconcile if it is present, so upgrades are safe, and in 1.5, we do remove the field from yaml

everyone alright with this approach?

shawn-hurley:
I don't know if it is ever safe to remove fields, but you can make them useless. 

I usually try to follow these guides as these folks have lots of experience making api changes:
https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#incompatible-api-changes

shawn-hurley:
One other way to make this is adding some CEL validation that will make it so that Rustic Config must always be nil.

kaovilai:
Had to look up CEL.. I think you mean https://kubernetes.io/docs/reference/using-api/cel/

mateusoliveira43:
/remove-lifecycle stale 
We need this in 1.4, otherwise upgrades may break

mateusoliveira43:
/remove-lifecycle stale
discuss in next scrum

api/v1alpha1/oadp_types.go
mateusoliveira43, line 1: run something like `grep -Iinr 'restic' . --exclude-dir={.go,docs,bundle,must-gather,config,blogs}` to ensure everything was removed

controllers/nodeagent.go
```diff
@@ -111,16 +105,7 @@ func (r *DPAReconciler) ReconcileNodeAgentDaemonset(log logr.Logger) (bool, erro
 		ObjectMeta: getNodeAgentObjectMeta(r),
 	}
 
-	if dpa.Spec.Configuration.Restic != nil {
-		// V(-1) corresponds to the warn level
-		var deprecationMsg string = "(Deprecation Warning) Use nodeAgent instead of restic, which is deprecated and will be removed in the future"
-		log.V(-1).Info(deprecationMsg)
```
mateusoliveira43, line 117: add Deprecation Warning when using restic in Node agent

controllers/validator.go
```diff
@@ -48,11 +44,25 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
 	}
 
+	// ENSURE UPGRADES --------------------------------------------------------
 	// check for VSM/Volsync DataMover (OADP 1.2 or below) syntax
 	if dpa.Spec.Features != nil && dpa.Spec.Features.DataMover != nil {
 		return false, errors.New("Delete vsm from spec.configuration.velero.defaultPlugins and dataMover object from spec.features. Use Velero Built-in Data Mover instead")
 	}
 
+	// check for ResticConfig (OADP 1.4 or below) syntax
+	if dpa.Spec.Configuration.Restic != nil {
```
mateusoliveira43, line 54: no need to be soft here, right?

Like, checking if `dpa.Spec.Configuration.Restic.Enable` is `true` before error out

controllers/velero.go
```diff
@@ -174,8 +174,6 @@ func (r *DPAReconciler) buildVeleroDeployment(veleroDeployment *appsv1.Deploymen
 		return fmt.Errorf("error appending pod annotations: %v", err)
 	}
 
-	// Since `restic` can be still be used and it default's to an empty string, we can't just
-	// pass the dpa.Spec.Configuration.NodeAgent.UploaderType directly
 	uploaderType := ""
 	if dpa.Spec.Configuration.NodeAgent != nil && len(dpa.Spec.Configuration.NodeAgent.UploaderType) > 0 {
 		uploaderType = dpa.Spec.Configuration.NodeAgent.UploaderType
```
mateusoliveira43, line 179: change default?

controllers/nodeagent.go
```diff
@@ -62,10 +60,6 @@ func getFsPvHostPath(platformType string) string {
 		return envFs
 	}
 
-	if env := os.Getenv(ResticPVHostPathEnvVar); env != "" {
```
mateusoliveira43, line 65: still not sure if I should delete this

controllers/validator.go
```diff
@@ -48,11 +45,27 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
+		return false, errors.New("Delete restic object from spec.configuration, use spec.configuration.nodeAgent instead")
+	}
+	// ENSURE UPGRADES --------------------------------------------------------
+
+	// DEPRECATIONS -----------------------------------------------------------
+	if dpa.Spec.Configuration.NodeAgent != nil && dpa.Spec.Configuration.NodeAgent.UploaderType == "restic" {
+		deprecationWarning := "(Deprecation Warning) Use kopia instead of restic in spec.configuration.nodeAgent.uploaderType, which is deprecated and will be removed in the future"
```
mateusoliveira43, line 62: add more info here, like restore will still be supported?

controllers/validator.go
```diff
@@ -48,11 +45,27 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
 	}
 
+	// check for ResticConfig (OADP 1.4 or below) syntax
+	if dpa.Spec.Configuration.Restic != nil {
+		return false, errors.New("Delete restic object from spec.configuration, use spec.configuration.nodeAgent instead")
+	}
+	// ENSURE UPGRADES --------------------------------------------------------
+
+	// DEPRECATIONS -----------------------------------------------------------
+	if dpa.Spec.Configuration.NodeAgent != nil && dpa.Spec.Configuration.NodeAgent.UploaderType == "restic" {
```
mateusoliveira43, line 61: putting it here logs and creates an event every minute. Is this ok?

weshayutin:
In the last scrum we all agreed on this PR.  Let's rebase and get it merged.  @mateusoliveira43 @shubham-pampattiwar.   Thank you at @mateusoliveira43

api/v1alpha1/oadp_types.go
```diff
@@ -177,10 +177,8 @@ type ResticConfig struct {
 // ApplicationConfig defines the configuration for the Data Protection Application
 type ApplicationConfig struct {
 	Velero *VeleroConfig `json:"velero,omitempty"`
-	// (deprecation warning) ResticConfig is the configuration for restic DaemonSet.
-	// restic is for backwards compatibility and is replaced by the nodeAgent
-	// restic will be removed in the future
-	// +kubebuilder:deprecatedversion:warning=1.3
+	// (do not use warning) restic field is for backwards compatibility and
```
shubham-pampattiwar, line 180: What does (do not use warning) mean here ?

shubham-pampattiwar: I think the intend here should be that "If a user uses restic config then the operator should do a No-Op and should add a warning on DPA that use node-agent config instead of restic"

config/samples/kustomization.yaml
```diff
@@ -6,7 +6,6 @@ resources:
 - downloadrequest.yaml
 - podvolumebackup.yaml
 - podvolumerestore.yaml
-- resticrepository.yaml
```
shubham-pampattiwar, line 9: Shouldn't we keep this for backwards compatibility ?

controllers/validator.go
shubham-pampattiwar: Its fine, you do not need to treat errors/warning as a substitute for documentation. Support policy is best explained via docs IMHO

controllers/validator.go
shubham-pampattiwar: Hmm. Wont't just one event be sufficient ? Along with log wanr and DPA warning ?

mateusoliveira43: I think so, wanted to move code to here, as it seems more appropriate place to do deprecation validation, but it has this drawback :( 

(also, only event is sufficient? or its relevant to also log the info?)

api/v1alpha1/oadp_types.go
mateusoliveira43: So, the current idea is that user can only use restic through node-agent config, this is a way to tell user not to use restic config

If user tries to use, DPA want reconcile telling user to use node-agent config instead

config/samples/kustomization.yaml
mateusoliveira43: I believe there is no need, will test to confirm

But worst case scenario (upgrade), what I think will happen is: on operator UI, `resticrepository` will not be shown, but CRD will not be deleted from cluster

controllers/validator.go
shubham-pampattiwar: Lets do both

config/samples/kustomization.yaml
mateusoliveira43: samples are used to generate examples for users
https://github.com/openshift/oadp-operator/blob/4291e6113c1ce130016190db8f548eccfafc3009/bundle/manifests/oadp-operator.clusterserviceversion.yaml#L5

This change is harmless, even though file exists in `config` folder, it was removed from `bundle` folder (which is what is shipped for users) in this [PR](https://github.com/openshift/oadp-operator/pull/809/files#diff-069ea2f824de6770c7836f0e18755a37580805128fe91bc747c41152efcea852) (which should have removed this reference and old file)

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -370,10 +370,8 @@ type ResticConfig struct {
 // ApplicationConfig defines the configuration for the Data Protection Application
 type ApplicationConfig struct {
 	Velero *VeleroConfig `json:"velero,omitempty"`
-	// (deprecation warning) ResticConfig is the configuration for restic DaemonSet.
-	// restic is for backwards compatibility and is replaced by the nodeAgent
-	// restic will be removed in the future
-	// +kubebuilder:deprecatedversion:warning=1.3
+	// (do not use warning) restic field is for backwards compatibility and
```
shubham-pampattiwar, line 373: `s/field/spec` at both places

internal/controller/nodeagent.go
```diff
@@ -37,7 +35,6 @@ const (
 	IBMCloudPVHostPath      = "/var/data/kubelet/pods"
 	GenericPluginsHostPath  = "/var/lib/kubelet/plugins"
 	IBMCloudPluginsHostPath = "/var/data/kubelet/plugins"
-	ResticPVHostPathEnvVar  = "RESTIC_PV_HOSTPATH"
```
shubham-pampattiwar, line 40: I think you will also need to remove `RESTIC_PV_HOSTPATH` from manager.yaml and operator csv

internal/controller/validator.go
```diff
@@ -56,11 +55,32 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
...
 
+	// ENSURE UPGRADES --------------------------------------------------------
 	// check for VSM/Volsync DataMover (OADP 1.2 or below) syntax
 	if r.dpa.Spec.Features != nil && r.dpa.Spec.Features.DataMover != nil {
 		return false, errors.New("Delete vsm from spec.configuration.velero.defaultPlugins and dataMover object from spec.features. Use Velero Built-in Data Mover instead")
 	}
 
+	// check for ResticConfig (OADP 1.4 or below) syntax
+	if r.dpa.Spec.Configuration.Restic != nil {
+		return false, errors.New("Delete restic object from spec.configuration, use spec.configuration.nodeAgent instead")
```
shubham-pampattiwar, line 66: `s/object/spec`

internal/controller/validator.go
```diff
@@ -16,6 +17,8 @@ import (
 	"github.com/openshift/oadp-operator/pkg/credentials"
 )
 
+var wasRestic bool
```
shubham-pampattiwar, line 20: Should we explicitly set this as false ? (I think default is false but still)

kaovilai: If we do, it would fail https://revive.run/r#var-declaration linter that was recently merged https://github.com/vmware-tanzu/velero/pull/8636/files#diff-9917ddc9f1c3304218f7269265b746d997c5c0615478177b5fceecd33ef47cb5
specifically because we are [declaring "zeroLiteral"](https://github.com/mgechev/revive/blob/d9c61c1518f16418ad94d92b2d5976c266e0ad6e/rule/var_declarations.go#L13-L26)
```go
var zeroLiteral = map[string]bool{
	"false": true, // bool
	// runes
	`'\x00'`: true,
	`'\000'`: true,
	// strings
	`""`: true,
	"``": true,
	// numerics
	"0":   true,
	"0.":  true,
	"0.0": true,
	"0i":  true,
}
```

internal/controller/validator.go
mateusoliveira43: not sold on this. Changing to spec would not pass sense that restic is an CR instead of part of DPA spec?

internal/controller/nodeagent.go
mateusoliveira43: good catch, will update

weshayutin:
I started w/ the just the most basic test..  created dpa w/ uploaderType: kopia and then switched it to restic. 
It reconciled.. :/

```
metadata:
  creationTimestamp: "2025-03-06T19:28:31Z"
  generation: 2
  name: dpa-sample
  namespace: openshift-adp
  resourceVersion: "18101553"
  uid: 65267f86-8146-4b24-8e20-b80429eb84d6
spec:
  backupLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      credential:
        key: cloud
        name: cloud-credentials
      default: true
      objectStorage:
        bucket: cvpbucketuswest2
        prefix: velero
      provider: aws
  configuration:
    nodeAgent:
      enable: true
      uploaderType: restic
    velero:
      defaultPlugins:
      - kubevirt
      - csi
      - openshift
      - aws
  snapshotLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      provider: aws
status:
  conditions:
  - lastTransitionTime: "2025-03-06T19:29:35Z"
    message: Reconcile complete
    reason: Complete
    status: "True"
    type: Reconciled
```
```
git branch
* fix/remove-restic-config-master
```

weshayutin:
toying around adding and removing the config.restic.enabled true/false and clearing the whole thing.
LGTM.   I think we can save upgrade testing for QE
```
status:
  conditions:
    - lastTransitionTime: '2025-03-06T21:21:17Z'
      message: 'Delete restic object from spec.configuration, use spec.configuration.nodeAgent instead'
      reason: Error
      status: 'False'
      type: Reconciled
```

weshayutin (approved):
/LGTM

api/v1alpha1/dataprotectionapplication_types.go
kaovilai: 🤷  I like field. alternatively spec field.

kaovilai: or better yet.. 
```suggestion
	// (do not use warning) restic config is for backwards compatibility and
```

shubham-pampattiwar: +1 @kaovilai

PR #1242, opened 2023-11-27
Indicate to user a new vsc def is required

weshayutin:
Consultant pointed out these docs are not quite clear. We need to indicate a new vsc must be created.

### DIFF
--- a/docs/examples/CSI/csi_example.md
+++ b/docs/examples/CSI/csi_example.md
@@ -57,10 +57,14 @@
 
 ### Create a StorageClass and VolumeShapshotClass:
 
-- Exisiting VolumeSnapshotClass definitions can not be modified, a new VolumeSnapshotClass must be created.
-
 - A `StorageClass` and a `VolumeSnapshotClass` are needed before the Mssql application
 is created. The app will map to the `StorageClass`, which contains information about the CSI driver.
+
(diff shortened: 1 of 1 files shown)


shubham-pampattiwar (commented):
We can always annotate and add labels to vs class but cannot change the spec I believe

docs/examples/CSI/csi_example.md
```diff
@@ -57,6 +57,8 @@
 
 ### Create a StorageClass and VolumeShapshotClass:
 
+- Exisiting VolumeSnapshotClass definitions can not be modified, a new VolumeSnapshotClass must be created.
```
kaovilai, line 60: Only true for ones managed by an operator/controller.. which is most of cloud provider defaults

weshayutin: thanks, updated!

docs/examples/CSI/csi_example.md
```diff
@@ -60,6 +60,12 @@
 - A `StorageClass` and a `VolumeSnapshotClass` are needed before the Mssql application
```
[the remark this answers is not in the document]
weshayutin: addressed in latest push, still self reviewing the entire doc.  Thank you!

shubham-pampattiwar (approved):

PR #1250, opened 2023-12-01
feat: Update Operator SDK version strategy

mateusoliveira43:
The goal of this PR is to create a strategy (document or script) for updating Operator SDK version in the project.

My goal was to have all operator generated files in a separate folder (inspired by https://github.com/dora-metrics/pelorus/pull/1045), but helm and Go Operator SDK structures are different, its not the best idea to try to put Go generated files in a separate folder.

The idea is: in temporaries folders, create Operator SDK structure using the current Operator SDK version and the version which we want to upgrade it to, following these steps:
`operator-sdk init --project-name=oadp-operator --repo=github.com/openshift/oadp-operator --domain=openshift.io`
`operator-sdk create api --group oadp --version v1alpha1 --kind DataProtectionApplication --resource --controller`
`operator-sdk create api --group oadp --version v1alpha1 --kind CloudStorage --resource --controller`

Then generate a `diff` file and apply changes to repo code.  
It seems to work https://github.com/operator-framework/operator-sdk/issues/6628

There are still a few points to adjust, like: when do we update boilerplate year? What to update in `go.mod`? Create another Makefile? ( the example added can be run with `make -f oadp.mk help`)


op.diff
mateusoliveira43, line 1: example of diff file between current and next operator sdk version structures

mateusoliveira43:
Should we update `hack/boilerplate.go.txt` to match this?
https://github.com/openshift/oadp-operator/blob/0347669ab8b5f051d815a59d6f45183742380720/LICENSE#L179-L191

controllers/cloudstorage_controller.go
```diff
@@ -25,36 +28,25 @@ import (
...
+type CloudStorageReconciler struct {
 	Client        client.Client
 	Scheme        *runtime.Scheme
 	Log           logr.Logger
 	EventRecorder record.EventRecorder
 }
 
-//TODO!!! FIX THIS!!!!
+//TODO!!! FIX THIS!!!!?
+// TODO change buckets to cloudstorages?
```
mateusoliveira43, line 40: this a leases role need updating

mateusoliveira43:
also run `operator-sdk create webhook` command after `operator-sdk create api`?

api/v1alpha1/groupversion_info.go
```diff
@@ -28,8 +28,6 @@ var (
 	// GroupVersion is group version used to register these objects
 	GroupVersion = schema.GroupVersion{Group: "oadp.openshift.io", Version: "v1alpha1"}
 
-	Kind = "DataProtectionApplication"
```
kaovilai, line 31: Do we wanna add kind back and add to schemebuilder line 32?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -317,16 +308,10 @@ spec:
...
-        recovery and migrate Kubernetes cluster resources and persistent volumes.
-      displayName: DataProtectionApplication
+    - description: DataProtectionApplication is the Schema for the dpa API
+      displayName: Data Protection Application
       kind: DataProtectionApplication
       name: dataprotectionapplications.oadp.openshift.io
-      statusDescriptors:
-      - description: Conditions defines the observed state of DataProtectionApplication
-        displayName: Conditions
-        path: conditions
```
kaovilai, line 329: Where possible, please add this back programmatically using operatorsdk's API Markers

read: https://sdk.operatorframework.io/docs/building-operators/golang/references/markers/
https://github.com/openshift/console/blob/master/frontend/packages/operator-lifecycle-manager/src/components/descriptors/reference/reference.md

mpryc:
The idea is really nice, may be tricky to maintain with more diffs. 

We have something like that in other project, right? 
   https://github.com/dora-metrics/pelorus/tree/master/scripts/pelorus-operator-patches

The benefit is that we can always re-generate operator starting from scratch.

This is fine for operator where there is not a huge number of changes to the initial files generated by the `operator-sdk` however in the case of oadp-operator where majority of the code is implemented within that generated initial files this may be very hard to maintain, especially when we are talking later about backports to older releases.

What about using kustomize files instead of patches whenever applicable ?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mateusoliveira43: done :smile:

mateusoliveira43:
@mpryc similar yes, but we will not have patch files here. Its terrible developer experience (editing a diff file instead of a yaml file, for example). 

We would re-generate files only when we want to update opeartor-sdk version, then we would create a diff file (not tracked by git), and then apply it and then open a PR for that action.

kustomize files can be an option, but not for all things I think. Makefile may be updated, for example. And that is the thing I think we need to be more careful: editing generated files. So, when we update then, it is a smooth experience. Maybe just adding new things to the end of those files can be enough to avoid patching errors

mateusoliveira43:
/remove-lifecycle stale
Still want to work on this

kaovilai:
unstale?

mateusoliveira43:
@kaovilai going to close this one and create small PRs that together do the same

PR #1253, opened 2023-12-08
Allow to use side container to connect to running service inside cluster

mpryc:
This PR have the following changes:

- A number of utility functions for interacting with k8s/OC APIs and utility functions to check connections to the pod exposed services were written or moved to the `utils/common_utils.go` and `utils/common_k8s_utils.go`.
- To talk to the exposed DB port an enhanced way was implemented, which uses structure `RequestParameters` that simplifies and unifies interaction with any http endpoint in the cluster and it's included in the logic to:
  - Try to connect using external route associated with the pod's service endpoint
  - If the above fails tries to use sidecar container `curl-tool`, which is associated with the DB pod. This container has `curl` and `nc` to interact with pod's service endpoints using `http` requests `POST` or `GET`.
- Removed unnecessary parameters in the mongoready and mysqlReady
- Renamed `common_utils.go` to `common_helpers.go` to be alighed with `k8s_common_helpers.go` and `ocp_common_helpers.go`.


tests/e2e/lib/apps.go
```diff
@@ -457,22 +457,156 @@ func makeRequest(request string, api string, todo string) {
...
-	backupFile := artifact_dir + "/backup-data.txt"
-	routev1.AddToScheme(clientv1.Scheme())
-	err := clientv1.Get(context.Background(), client.ObjectKey{
-		Namespace: namespace,
-		Name:      routeName,
-	}, appRoute)
+
+	apiEndpoint, err := getApiEndpoint(clientv1, namespace, serviceName, routeName)
 	if err != nil {
-		return err
```
mateusoliveira43, line 471: I think this logic is incorrect, if error happens, it will try later to make request with empty `apiEndpoint`

tests/e2e/lib/apps.go
```diff
@@ -457,22 +457,156 @@ func makeRequest(request string, api string, todo string) {
...
+		fmt.Println("Error:", err)
```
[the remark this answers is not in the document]
mpryc: That was meant to be `fmt.Errorf()` do we have equivalent in `fmt.*` to distinguish between standard print and error... ?

mpryc:
@mateusoliveira43 I believe this is ready to review. It solves the problem with connection to the pod service, however it does not fully fix CI issue, where the pod is unhealthy after restore - in such situation the pod won't expose service so it won't be available for connection.

I've traced some error for the above issue (which is separate to this PR) and comes from the test app itself. The sql pod seems to be running just fine:

```
time="2023-12-22T22:59:31Z" level=info msg="Success: Attached volume and redirected logs to /tmp/log/todoapp/app.log" func=main.main file="/build/todolist.go:179"
time="2023-12-22T22:59:31Z" level=info msg="A running instance of the db: todolist not found, creating" func=main.main file="/build/todolist.go:185"
panic: sql: database is closed

goroutine 1 [running]:
github.com/jinzhu/gorm.mysql.HasTable({{{0x827460, 0xc00025fc70}, {}}}, {0xc00031e1a0, 0x10})
	/go/pkg/mod/github.com/jinzhu/gorm@v1.9.16/dialect_mysql.go:180 +0x296
github.com/jinzhu/gorm.(*DB).HasTable(0xc000310dd0, {0x734440, 0xc00032a100})
	/go/pkg/mod/github.com/jinzhu/gorm@v1.9.16/main.go:680 +0x89
github.com/jinzhu/gorm.(*DB).DropTableIfExists(0xc00025fd40, {0xc0002a3ea0, 0x1, 0x1})
	/go/pkg/mod/github.com/jinzhu/gorm@v1.9.16/main.go:660 +0x8b
main.main()
	/build/todolist.go:186 +0x39c
```

tests/e2e/lib/apps.go
```diff
@@ -22,9 +18,9 @@ import (
 	"github.com/onsi/gomega"
 	ocpappsv1 "github.com/openshift/api/apps/v1"
 	buildv1 "github.com/openshift/api/build/v1"
-	routev1 "github.com/openshift/api/route/v1"
 	security "github.com/openshift/api/security/v1"
 	templatev1 "github.com/openshift/api/template/v1"
+	utils "github.com/openshift/oadp-operator/tests/e2e/utils"
```
mateusoliveira43, line 23: unnecessary import alias

tests/e2e/lib/apps.go
```diff
@@ -36,8 +32,9 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/serializer/yaml"
 	"k8s.io/apimachinery/pkg/selection"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
+	apiwait "k8s.io/apimachinery/pkg/util/wait"
```
mateusoliveira43, line 35: Please undo this change, I believe it is unnecessary

tests/e2e/utils/common_utils.go
```diff
@@ -1,16 +1,33 @@
...
 	"io/ioutil"
+	"log"
+	"net/http"
+	"net/url"
+	"os"
 	"os/exec"
 	"strings"
 )
 
+const (
```
mateusoliveira43, line 15: create a type and use in `MakeHTTPRequest` function

tests/e2e/utils/common_utils.go
```diff
@@ -1,16 +1,33 @@
...
+	GET  string = "GET"
+	POST string = "POST"
+)
+
 func ReadFile(path string) ([]byte, error) {
 	file, err := ioutil.ReadFile(path)
 	return file, err
 }
 
+func CleanupFileIfExists(filePath string) {
```
mateusoliveira43, line 25: clean up seems a strange name here, would change to delete or remove

tests/e2e/utils/common_utils.go
mpryc: Ah I didn't include that in MakeHTTPRequest, good catch

tests/e2e/lib/apps.go
mpryc: Correct. I did change it as I previously had also my own wait for service function and did not want to mix two different wait names, however it's not required now. Will undo.

tests/e2e/lib/apps.go
mpryc: Correct. Will change.

tests/e2e/utils/common_k8s_utils.go
```diff
@@ -0,0 +1,399 @@
...
+//	(*corev1.Pod, error): A pointer to the first pod matching the label selector, or an error if any.
+//
+// Example:
+//
+//	pod, err := GetPodByLabel(kubeClient, "mongo-persistent", "curl-tool=true")
+//	if err != nil {
+//	    log.Println(err)
+//	}
+//	log.Printf("Pod found: %s\n", pod.Name)
+func GetPodByLabel(kubeClient *kubernetes.Clientset, namespace, labelSelector string) (*corev1.Pod, error) {
```
kaovilai, line 75:
```suggestion
func GetFirstPodByLabel(kubeClient *kubernetes.Clientset, namespace, labelSelector string) (*corev1.Pod, error) {
```
This func return the first one pod by label.. I think this makes it clearer.

tests/e2e/utils/common_utils.go
```diff
@@ -21,3 +40,149 @@ func ReplaceSecretDataNewLineWithCarriageReturn(data []byte) []byte {
...
+			return "", "", fmt.Errorf("Error reading response body: %v", err)
+		}
+		return string(body), "", nil
+	}
+
+	// The response status code indicates an error
+	// Read the error response body
+	responseBody, responseErr := ioutil.ReadAll(resp.Body)
+	if responseErr != nil {
+		return "", "", fmt.Errorf("HTTP request failed with status code %d: %v", resp.StatusCode, responseErr)
```
kaovilai, line 184:
```suggestion
		return "", string(responseBody), fmt.Errorf("HTTP request failed with status code %d: %v", resp.StatusCode, responseErr)
```

mpryc:
This is the random failure, subsequent runs are PASSING or FAILING, randomly.
Similarly without using this fix, however this fix allowed to at least get to the point where data was retrieved even without external route.

Including test run (with Focus, so only one test is executed) where the route can not be reached and the proxy pod is used.
[test_run_with_proxy_pod.txt](https://github.com/openshift/oadp-operator/files/13830238/test_run_with_proxy_pod.txt)

What happened here:

 1. During backup operation the route was not yet accessible so the test used proxy pod to access TODOLIST service endpoint and got the data:
```
2024/01/04 14:21:18 Using Proxy pod container: mysql-f8c8cfdfc-fgxpj
2024/01/04 14:21:18 Provided command: curl -X GET --silent --show-error http://todolist.mysql-persistent.svc.cluster.local:8000/todo-incomplete
[...]

2024/01/04 14:21:19 stdOutput:
[{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:21:18.25146434 +0100 CET m=+39.065424202","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]
```

2. After restore the route was accessible:
```
2024/01/04 14:22:58 Fallback verifying if the service is accessible via: http://todolist-route-mysql-persistent.apps.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com
2024/01/04 14:22:58 Data came from backup-file
[{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:21:18.25146434 +0100 CET m=+39.065424202","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]
```

We are missing some debug information of what the data was after restore, will add that so we know what's going on.

tests/e2e/utils/common_k8s_utils.go
mpryc: Will change.

mpryc:
Ok something wrong in the compare function:

```
2024/01/04 14:57:22 Data came from backup-file
 [{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:55:32.099347111 +0100 CET m=+32.691216556","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]

2024/01/04 14:57:22 Data came from response
 [{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:55:32.099347111 +0100 CET m=+32.691216556","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]

  [FAILED] Unexpected error:
      <*errors.errorString | 0xc0006b04e0>: 
      Backup and Restore Data are not the same
      {
          s: "Backup and Restore Data are not the same",
      }
  occurred

```

mpryc:
> Ok something wrong in the compare function:
> 
> ```
> 2024/01/04 14:57:22 Data came from backup-file
>  [{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:55:32.099347111 +0100 CET m=+32.691216556","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]
> 
> 2024/01/04 14:57:22 Data came from response
>  [{"Id":1,"Description":"time to make the donuts","Completed":false},{"Id":3,"Description":"2024-01-04 14:55:32.099347111 +0100 CET m=+32.691216556","Completed":false},{"Id":4,"Description":"Thursday","Completed":false}]
> 
>   [FAILED] Unexpected error:
>       <*errors.errorString | 0xc0006b04e0>: 
>       Backup and Restore Data are not the same
>       {
>           s: "Backup and Restore Data are not the same",
>       }
>   occurred
> ```

The error is with the newline character being different, will update the PR. So far I was testing either pod proxy or standard route, so the error was not visible, here I have mix where it's either accessible via pod proxy or standard route, whetever is first.

This of course is not fixing root cause where the race condition happens, but that's something I'll fix in another PR (also adding rediness and liveness probes to the db services).

tests/e2e/utils/common_utils.go
mpryc: changed. thanks

tests/e2e/lib/apps.go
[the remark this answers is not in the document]
mpryc: fixed.

tests/e2e/utils/common_utils.go
```diff
@@ -21,3 +40,149 @@ func ReplaceSecretDataNewLineWithCarriageReturn(data []byte) []byte {
...
+
+	if err != nil {
+		return "", "", err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode >= 200 && resp.StatusCode < 300 {
+		body, err := ioutil.ReadAll(resp.Body)
+		if err != nil {
+			return "", "", fmt.Errorf("Error reading response body: %v", err)
```
kaovilai, line 175:
```suggestion
			return "", string(body), fmt.Errorf("Error reading response body: %v", err)
```

mpryc: Thanks changed as suggested

mateusoliveira43:
I would break this PR in at least 3 smaller PRs
- liveness, readiness and startup probes
- change route to curl tool (I think there is no necessity to try 3 different things, just adds complexity)
- refactoring

mpryc:
> I would break this PR in at least 3 smaller PRs
> 
> * liveness, readiness and startup probes
> * change route to curl tool (I think there is no necessity to try 3 different things, just adds complexity)
> * refactoring

It's really hard to do so, because one is in conflict with another and only the full set of changes gives *some* passing rate of CI. Otherwise it's always failing, meaning it will never be green on the CI tests.

tests/e2e/sample-applications/mongo-persistent/mongo-persistent-block-git.yaml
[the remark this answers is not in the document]
mpryc: Until the https://github.com/vmware-tanzu/velero/issues/5856 is fixed we should not run e2e tests which are using source code builds as they are creating ImageStreams and based on that there are s3 failures.

The `-git.yaml` files are worth preserving when the #5856 is fixed we can create new type of test or enable one or two based on the source code build so also the backup&restore of such type of deployment is executed.

Until then to reduce number of flakes we should switch to non source code e2e scenarios.

Also I really don't want to make this PR too complex, it's already way over what it should be, but it's touching one problem (flakes in CI), which imo should be addressed as quickly as possible.

We still have some flakes, but at least ratio of failing ones is reduced.

mpryc:
FYI @shubham-pampattiwar updated PR comment.

shubham-pampattiwar:
Thank you ! @mpryc 😊

mpryc:
RE failures:

1. `ci/prow/4.14-e2e-test-azure` and `ci/prow/4.14-e2e-test-aws`
    -  Some new unseen previously flake, which is during PVC attach. The Rediness probe fails then then pod is being restarted and the Liveness probe also fails, but we know now it's during the restore of DB. The DB fails to restore because of timeout. There are logs which causes PVC to be failing to be attached, but issue may be from somewhere else. 
 
2. `ci/prow/4.13-e2e-test-aws`
    -  Already known as CI flake during Backup phase. "Failed to check and update snapshot content". This may be our code problem during tests or velero one.

weshayutin (dismissed):
nice work

kaovilai (dismissed):

shubham-pampattiwar (dismissed):

tests/e2e/backup_restore_suite_test.go
```diff
@@ -24,7 +24,7 @@ type VerificationFunction func(client.Client, string) error
 type appVerificationFunction func(bool, bool, BackupRestoreType) VerificationFunction
 
 // TODO duplications with mongoready
-func mongoready(preBackupState bool, twoVol bool, backupRestoreType BackupRestoreType) VerificationFunction {
+func mongoready(preBackupState bool, twoVol bool) VerificationFunction {
 	return VerificationFunction(func(ocClient client.Client, namespace string) error {
 		Eventually(IsDCReady(ocClient, namespace, "todolist"), timeoutMultiplier*time.Minute*10, time.Second*10).Should(BeTrue())
 		exists, err := DoesSCCExist(ocClient, "mongo-persistent-scc")
```
mateusoliveira43, line 30: yaml files for mongo/mysql where modified (added liveness probe and changed deploy method), but the way we check if apps are ready was not modified. Shouldn't it be also modified?

mpryc: This can be done in separate PR. The way everything is UP should not be changed, just different parts will now exit in appropriate times.

weshayutin:
I think azure flaked due to
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1253/pull-ci-openshift-oadp-operator-master-4.12-e2e-test-azure/1746908722602446848/artifacts/e2e-test-azure/e2e/artifacts/MySQL%20application%20two%20Vol%20CSI/openshift-adp/velero-78bbb97cdb-zlq6f/velero.log
```
error getting storage account properties: DefaultAzureCredential authentication failed\nPOST https://login.microsoftonline.com/6047c7e9-b2ad-488d-a54e-dc3f6be6a7ee/oauth2/v2.0/token\
```

mateusoliveira43 (changes requested):
I tested running e2e locally, and my 2 runs from the PR failed. Running from master I did not have problems. @weshayutin @kaovilai @shubham-pampattiwar have you tested the PR locally?

AGAIN I really think this PR should be break down in smaller ones, starting with one about liveness probe

weshayutin:
> I tested running e2e locally, and my 2 runs from the PR failed. Running from master I did not have problems. @weshayutin @kaovilai @shubham-pampattiwar have you tested the PR locally?
> 
Starting w/ master is the right thing.  OADP-1.3 would be the only other branch of concern, did you have errors and what were they on that branch?  If you are going to post you had errors, please be more specific.

> AGAIN I really think this PR should be break down in smaller ones, starting with one about liveness probe
I like the concern, perhaps we can wait on this for Michal to return.

mpryc:
> I tested running e2e locally, and my 2 runs from the PR failed. Running from master I did not have problems. @weshayutin @kaovilai @shubham-pampattiwar have you tested the PR locally?
> 
> AGAIN I really think this PR should be break down in smaller ones, starting with one about liveness probe

@mateusoliveira43 please provide some more logs when you are writing "PR failed". It's hard to check if the failure is caused by this PR or other flake if no data was provided. What are the logs from the failures ?

mateusoliveira43:
Logs from my two fails running `make test-e2e` from the PR branch

```
  > Enter [It] Mongo application KOPIA - /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:314 @ 01/15/24 14:41:19.19
2024/01/15 14:41:19 Waiting for velero pod to be running
2024/01/15 14:41:19 velero pods not found
2024/01/15 14:41:24 pod: velero-cf667848b-s6lbf is not yet running: phase is Pending
2024/01/15 14:41:29 velero pods are running
2024/01/15 14:41:29 Waiting for Node Agent pods to be running
2024/01/15 14:41:29 Checking for correct number of running Node Agent pods...
2024/01/15 14:41:29 Installing application for case mongo-kopia-e2e
  ...
2024/01/15 14:41:29 updating resource: security.openshift.io/v1, Kind=SecurityContextConstraints; name: mongo-persistent-scc
2024/01/15 14:41:30 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:41:35 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:41:40 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:41:46 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:41:51 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:41:56 Pod mongo-7664879648-2jv4m not yet succeeded: phase is Pending
2024/01/15 14:42:01 Running pre-backup function for case mongo-kopia-e2e
  [FAILED] Timed out after 600.001s.
  Expected
      <bool>: false
  to be true
  In [It] at: /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:29 @ 01/15/24 14:52:01.187
  < Exit [It] Mongo application KOPIA - /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:314 @ 01/15/24 14:52:01.187 (10m41.997s)

# second run
        
  > Enter [It] MySQL application KOPIA - /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:322 @ 01/15/24 16:31:33.341
2024/01/15 16:31:33 Waiting for velero pod to be running
2024/01/15 16:31:33 velero pods not found
2024/01/15 16:31:38 pod: node-agent-tscnd is not yet running: phase is Pending
2024/01/15 16:31:43 velero pods are running
2024/01/15 16:31:43 Waiting for Node Agent pods to be running
2024/01/15 16:31:43 Checking for correct number of running Node Agent pods...
2024/01/15 16:31:43 Installing application for case mysql-kopia-e2e
...
2024/01/15 16:31:44 updating resource: security.openshift.io/v1, Kind=SecurityContextConstraints; name: mysql-persistent-scc
2024/01/15 16:31:44 Pod mysql-77c6bb8c9-xtd9k not yet succeeded: phase is Pending
2024/01/15 16:31:50 Pod mysql-77c6bb8c9-xtd9k not yet succeeded: phase is Pending
2024/01/15 16:31:55 Pod mysql-77c6bb8c9-xtd9k not yet succeeded: phase is Pending
2024/01/15 16:32:00 Pod mysql-77c6bb8c9-xtd9k not yet succeeded: phase is Pending
2024/01/15 16:32:05 Pod mysql-77c6bb8c9-xtd9k not yet succeeded: phase is Pending
2024/01/15 16:32:10 Running pre-backup function for case mysql-kopia-e2e
2024/01/15 16:32:10 checking for the NAMESPACE: mysql-persistent
  deployment not available with condition: Deployment does not have minimum availability.
  deployment not available with condition: ReplicaSet "mysql-77c6bb8c9" is progressing.
2024/01/15 16:32:20 Verifying backup/restore data of todolist
2024/01/15 16:32:20 Verifying if the service is accessible: http://todolist.mysql-persistent.svc.cluster.local:8000
2024/01/15 16:32:20 Failed to use Service endpoint for service todolist in namespace mysql-persistent: Failed to connect to the service: Get "http://todolist.mysql-persistent.svc.cluster.local:8000": dial tcp: lookup todolist.mysql-persistent.svc.cluster.local: Temporary failure in name resolution
2024/01/15 16:32:20 Fallback verifying if the service is accessible via route
2024/01/15 16:32:20 Fallback verifying if the service is accessible via: http://todolist-route-mysql-persistent.mateus01022023-br-sao-3626522b15eedb880d7b99992e225c1b-0000.br-sao.containers.appdomain.cloud
2024/01/15 16:32:20 Can not connect to the application endpoint with route or service name: Failed to get both service and route endpoints: Route endpoint not accessible: HTTP request failed with status code: 503
2024/01/15 16:32:20 Trying to get to the service via proxy POD
2024/01/15 16:32:20 Using Proxy pod container: mysql-77c6bb8c9-xtd9k
2024/01/15 16:32:20 Provided command: curl -X GET --silent --show-error http://todolist.mysql-persistent.svc.cluster.local:8000/todo-incomplete
2024/01/15 16:32:20 Command will run in the pod: mysql-77c6bb8c9-xtd9k, container: curl-tool in the namespace: mysql-persistent
2024/01/15 16:32:20 Full command URL: https://c104-e.br-sao.containers.cloud.ibm.com:30014/api/v1/namespaces/mysql-persistent/pods/mysql-77c6bb8c9-xtd9k/exec?command=curl&command=-X&command=GET&command=--silent&command=--show-error&command=http%3A%2F%2Ftodolist.mysql-persistent.svc.cluster.local%3A8000%2Ftodo-incomplete&container=curl-tool&stderr=true&stdout=true&tty=true
2024/01/15 16:32:22 stdOutput: curl: (7) Failed to connect to todolist.mysql-persistent.svc.cluster.local port 8000 after 1038 ms: Couldn't connect to server

2024/01/15 16:32:22 Error while streaming command output: command terminated with exit code 7
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc0006968b0>: 
      Error while streaming command output: command terminated with exit code 7
      {
          s: "Error while streaming command output: command terminated with exit code 7",
      }
  occurred
  In [It] at: /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:141 @ 01/15/24 16:32:22.322
  < Exit [It] MySQL application KOPIA - /path/mateus_oadp/tests/e2e/backup_restore_suite_test.go:322 @ 01/15/24 16:32:22.322 (48.98s)
```

mpryc:
As a start I've created #1270  #1271 #1273  #1272. Of course that's not the end of split and I am questioning it, but will leave the decision to you, which is better (many small ones, or one covering a bit wider scope) in this case.

mpryc:
@mateusoliveira43 did you observe this randomly/always/one time? I've also seen it randomly in the CI, but that imo is actually proper as we know what exactly caused the failure (DB was not able to come up and the tests failed immediately), but as always I may be wrong. I will increase time in which db is expected to be up. In my opinion we were balancing on the edge and now with side container in a slower systems it may be just over. On a 15 runs on AWS I have experienced 0 errors like this:
```
$ for i in {1..15}; do make test-e2e > /tmp/run_e2e_$i.txt 2>&1; done

$ grep "Timed out after" /tmp/run_*
```

mateusoliveira43:
can we close this one @mpryc ? some parts were already implemented by your previous PRs

One part that was not implemented yet is the removal BackupRestoreType from PreBackupVerify/PostRestoreVerify in backup/restore/must-gather tests, but I think it is easier to open a new PR for that

mpryc:
@mateusoliveira43 no, please don't close this one. I am doing rebase and will still want to have side container proxy to be in place, so we have much better view of why the test is failing in case the external route is not available.

mpryc:
@mateusoliveira43 here is example of the log where we clearly see how the communication of our application with route service is done:

```
024/01/26 13:36:14 Verifying if the service is accessible: http://todolist.mysql-persistent.svc.cluster.local:8000
2024/01/26 13:36:14 Failed to use Service endpoint for service todolist in namespace mysql-persistent: Failed to connect to the service: Get "http://todolist.mysql-persistent.svc.cluster.local:8000": dial tcp: lookup todolist.mysql-persistent.svc.cluster.local: Temporary failure in name resolution
2024/01/26 13:36:14 Fallback verifying if the service is accessible via route
2024/01/26 13:36:14 Fallback verifying if the service is accessible via: http://todolist-route-mysql-persistent.apps.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com
2024/01/26 13:36:14 Data before the curl request: 
 [{"Id":1,"Description":"time to make the donuts","Completed":false}]

2024/01/26 13:36:14 Payload encoded parameters: description=2024-01-26+13%3A36%3A14.703745158+%2B0100+CET+m%3D%2B53.732085706
2024/01/26 13:36:14 Payload encoded parameters: description=Friday

```

mateusoliveira43:
@mpryc still not sold on adding more ways of checking route (I did not see any of the new methods work, direct through service or with side containers), but would be ok with better logs

mpryc:
> @mpryc still not sold on adding more ways of checking route (I did not see any of the new methods work, direct through service or with side containers), but would be ok with better logs

1. More logs - we can get that once we find out if the internal/external route is accessible and actually try to connect to it, so we can tell in the logs what was possible what was not.
2. If you run from your local box is different then from the CI and the more access to the endpoints makes it more important.
3. The external route way is cluster type specific an it may be not there if your cluster is misconfigured or configured in a way you do not want to have external dns to it, then our tests still should pass.

tests/e2e/backup_restore_suite_test.go
mpryc: not relevant in this PR

mpryc:
/unhold

I am unholding this PR as I believe it's ready and we can discuss now. The PR was reworked and number of changes pulled out from it, so it's smaller, however I think it's important for us.

tests/e2e/utils/common_k8s_utils.go
mateusoliveira43, line 1: would delete this file. Why not to use `tests/e2e/lib/kube_helpers.go`?

tests/e2e/lib/apps.go
```diff
@@ -449,105 +446,138 @@ func RunMustGather(oc_cli string, artifact_dir string) error {
...
 	if err != nil {
 		return err
 	}
-	appApi := "http://" + appRoute.Spec.Host
-	appEndpoint := appApi + "/todo-incomplete"
-	volumeEndpoint := appApi + "/log"
 
-	//if this is prebackstate = true, add items via makeRequest function. We only want to make request before backup
-	//and ignore post restore checks.
-	log.Printf("PrebackState: %t\n", prebackupState)
```
mateusoliveira43, line 487: would not remove this from logs, I think it is good for debugging them later

tests/e2e/utils/common_k8s_utils.go
mpryc: We could move the logic to `kube_helpers.go`, but also then move the `common_utils.go` into the `tests/e2e/lib/`.

Also I think it's worth keeping separate utils for the ones that interact with `ocClient client.Client` and the ones which are using only `kubeClient *kubernetes.Clientset`, this way we know the line which are using pure k8s vs additional ocp clients.

tests/e2e/lib/apps.go
mpryc: I think that is really useless information in here. Every test consists of always two calls, one `PreBackupVerify` and another `PostRestoreVerify`, it's clear from the logs in which we are in without having additional `PrebackState` entry.

Here is log example. I think line first and last are duplications.
```
2024/01/25 20:46:02 Running pre-backup function for case mongo-csi-e2e
2024/01/25 20:46:02 Verifying backup/restore data of todolist
2024/01/25 20:46:02 PrebackState: true
```

tests/e2e/utils/common_k8s_utils.go
mateusoliveira43: I like the idea

tests/e2e/lib/apps.go
mateusoliveira43: yeah, agreed (my concern is calling function with wrong bool, but that should be easy to catch as well without this log)

mpryc: That should really not happen, we are not setting the bool dynamically it's within Entry() that has true/false passed as a single value, so very limited place for mistake.

Example:
https://github.com/openshift/oadp-operator/blob/master/tests/e2e/backup_restore_suite_test.go#L321-L322

tests/e2e/utils/common_k8s_utils.go
mpryc: So something like:

```
	rename:    tests/e2e/utils/common_utils.go -> tests/e2e/lib/common_helpers.go
	rename:    tests/e2e/utils/common_k8s_utils.go ->tests/e2e/lib/ocp_common_helpers.go
        rename:    tests/e2e/lib/kube_helpers.go -> tests/e2e/lib/k8s_common_helpers.go


```

mpryc:
Few test failures:

3 x `Mongo application DATAMOVER` is not known yet, but it was also observed in other PR e.g. https://github.com/openshift/oadp-operator/pull/1257#issuecomment-1912728524

1 x `MySQL application DATAMOVER`, which may be very similar to above. The pods are not yet ready and times out. There are some failures to provision storage and update endpoint for mysql-persistent/mysql.

1 x `MySQL application two Vol CSI`, which is not known yet, but may be just previous flake with CSI not being able to came up, however the logs are slightly different and test is failing on the try to connect attempt, so flake logic may require update.

1 x `MySQL application two Vol CSI`, which failed with known flake and retried, then it failed with unknown problem, possible same as above.

mpryc:
Another round:

2 x `Mongo application DATAMOVER`, same as previous round
1 x `MySQL application two Vol CSI`, same as previous round
1 x `Mongo application CSI`, same as previous round

So all the flakes from the above comes to the error after restore:
`Timed out after 540.001s.` to have DB pod up and running.

mpryc:
`MySQL application two Vol CSI` with one restart (found known Flake) and the second attempt was a failure due to another issue which is not known Flake.

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi.yaml
```diff
@@ -173,7 +177,7 @@ items:
         spec:
           containers:
           - name: todolist
-            image: quay.io/konveyor/todolist-mariadb-go:v2_4
+            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:latest
```
mateusoliveira43, line 180: why this change?

mateusoliveira43 (changes requested):
Checked all 6 ci jobs that passed, and none of them used the 2 new ways of checking route (direct through service or curl-tool. search for "Verifying if the service is accessible" in jobs logs)

Also, still missing files renaming

tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi.yaml
mpryc: To keep all our images in one place which we have access to, also built those apps using newest go version.

tests/e2e/utils/common_k8s_utils.go
mpryc: Done.

mpryc:
> Checked all 6 ci jobs that passed, and none of them used the 2 new ways of checking route (direct through service or curl-tool. search for "Verifying if the service is accessible" in jobs logs)
> 
> Also, still missing files renaming

I believe that change to not use `git` build made this more reliable, but I still think it's worth keeping it as we know exactly that the route was there, if it was accessible. One scenarios where this may be important is misconfigured Ingress controller on the cluster, or the one that does not allow to create external routes.

Below is a broken URL for that Ingress route where our test still works fine:
```
2024/01/29 13:38:22 Running pre-backup function for case mongo-csi-e2e
2024/01/29 13:38:22 Verifying backup/restore data of todolist
2024/01/29 13:38:22 Verifying if the service is accessible: http://todolist.mongo-persistent.svc.cluster.local:8000
2024/01/29 13:38:22 Failed to use Service endpoint for service todolist in namespace mongo-persistent: Failed to connect to the service: Get "http://todolist.mongo-persistent.svc.cluster.local:8000": dial tcp: lookup todolist.mongo-persistent.svc.cluster.local: Temporary failure in name resolution
2024/01/29 13:38:22 Fallback verifying if the service is accessible via route
2024/01/29 13:38:22 Fallback verifying if the service is accessible via: http://todolist-broken.misconfigureddomain.com
2024/01/29 13:38:22 Can not connect to the application endpoint with route or service name: Failed to get both service and route endpoints: Route endpoint not accessible: Get "http://todolist-broken.misconfigureddomain.com": dial tcp: lookup todolist-broken.misconfigureddomain.com: no such host
2024/01/29 13:38:22 Trying to get to the service via proxy POD
2024/01/29 13:38:22 Using Proxy pod container: mongo-5fb49f6768-j9r8j
2024/01/29 13:38:22 Provided command: curl -X GET --silent --show-error http://todolist.mongo-persistent.svc.cluster.local:8000/todo-incomplete
2024/01/29 13:38:22 Command will run in the pod: mongo-5fb49f6768-j9r8j, container: curl-tool in the namespace: mongo-persistent
2024/01/29 13:38:22 Full command URL: https://api.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com:6443/api/v1/namespaces/mongo-persistent/pods/mongo-5fb49f6768-j9r8j/exec?command=curl&command=-X&command=GET&command=--silent&command=--show-error&command=http%3A%2F%2Ftodolist.mongo-persistent.svc.cluster.local%3A8000%2Ftodo-incomplete&container=curl-tool&stderr=true&stdout=true&tty=true
2024/01/29 13:38:23 stdOutput: [{"Id":"65b79c3c2a2b1e6efd0b787d","Description":"time","Completed":false}]

2024/01/29 13:38:23 Data before the curl request: 
 [{"Id":"65b79c3c2a2b1e6efd0b787d","Description":"time","Completed":false}]

2024/01/29 13:38:23 Payload encoded parameters: description=2024-01-29+13%3A38%3A23.43816687+%2B0100+CET+m%3D%2B63.756468130
2024/01/29 13:38:23 Provided command: curl -X POST -d description=2024-01-29+13%3A38%3A23.43816687+%2B0100+CET+m%3D%2B63.756468130 --silent --show-error http://todolist.mongo-persistent.svc.cluster.local:8000/todo
2024/01/29 13:38:23 Command will run in the pod: mongo-5fb49f6768-j9r8j, container: curl-tool in the namespace: mongo-persistent
2024/01/29 13:38:23 Full command URL: https://api.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com:6443/api/v1/namespaces/mongo-persistent/pods/mongo-5fb49f6768-j9r8j/exec?command=curl&command=-X&command=POST&command=-d&command=description%3D2024-01-29%2B13%253A38%253A23.43816687%2B%252B0100%2BCET%2Bm%253D%252B63.756468130&command=--silent&command=--show-error&command=http%3A%2F%2Ftodolist.mongo-persistent.svc.cluster.local%3A8000%2Ftodo&container=curl-tool&stderr=true&stdout=true&tty=true
2024/01/29 13:38:23 stdOutput: {"Id":"65b79c3f2a2b1e6efd0b787e"}

2024/01/29 13:38:23 Payload encoded parameters: description=Monday
2024/01/29 13:38:23 Provided command: curl -X POST -d description=Monday --silent --show-error http://todolist.mongo-persistent.svc.cluster.local:8000/todo
2024/01/29 13:38:23 Command will run in the pod: mongo-5fb49f6768-j9r8j, container: curl-tool in the namespace: mongo-persistent
2024/01/29 13:38:23 Full command URL: https://api.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com:6443/api/v1/namespaces/mongo-persistent/pods/mongo-5fb49f6768-j9r8j/exec?command=curl&command=-X&command=POST&command=-d&command=description%3DMonday&command=--silent&command=--show-error&command=http%3A%2F%2Ftodolist.mongo-persistent.svc.cluster.local%3A8000%2Ftodo&container=curl-tool&stderr=true&stdout=true&tty=true
2024/01/29 13:38:24 stdOutput: {"Id":"65b79c402a2b1e6efd0b787f"}

2024/01/29 13:38:24 Using Proxy pod container: mongo-5fb49f6768-j9r8j
2024/01/29 13:38:24 Provided command: curl -X GET --silent --show-error http://todolist.mongo-persistent.svc.cluster.local:8000/todo-incomplete
2024/01/29 13:38:24 Command will run in the pod: mongo-5fb49f6768-j9r8j, container: curl-tool in the namespace: mongo-persistent
2024/01/29 13:38:24 Full command URL: https://api.cluster-mpryc1502ocp412x.mpryc1502ocp412x.mg.dog8code.com:6443/api/v1/namespaces/mongo-persistent/pods/mongo-5fb49f6768-j9r8j/exec?command=curl&command=-X&command=GET&command=--silent&command=--show-error&command=http%3A%2F%2Ftodolist.mongo-persistent.svc.cluster.local%3A8000%2Ftodo-incomplete&container=curl-tool&stderr=true&stdout=true&tty=true
2024/01/29 13:38:24 stdOutput: [{"Id":"65b79c3c2a2b1e6efd0b787d","Description":"time","Completed":false},{"Id":"65b79c3f2a2b1e6efd0b787e","Description":"2024-01-29 13:38:23.43816687 +0100 CET m=+63.756468130","Completed":false},{"Id":"65b79c402a2b1e6efd0b787f","Description":"Monday","Completed":false}]
```

The scenario which is used to connect via internal route may and will be really cool once we run gingko in a pod on the same cluster where tests are running. In prow the runner and executor are on other clusters at the moment.

mpryc:
Today Prow has some outages.

mpryc:
@mateusoliveira43 an ideal log to present failure and issue where the pod did not come up at all, even the proxy pod was not able to help here. In such case we know it's not the route problem:
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1253/pull-ci-openshift-oadp-operator-master-4.14-e2e-test-aws/1751987830927659008/artifacts/e2e-test-aws/e2e/build-log.txt

mpryc:
@mateusoliveira43 I've removed part which is trying to connect to use  the internal route together with the functions associated with that `dead code`.

Also attaching diff to what was removed for reference and eventual future use:
```diff
diff --git a/tests/e2e/lib/apps.go b/tests/e2e/lib/apps.go
index b10280d..a801c94 100755
--- a/tests/e2e/lib/apps.go
+++ b/tests/e2e/lib/apps.go
@@ -524,10 +524,10 @@ func getRequestParameters(url string, proxyPodParams *ProxyPodParameters, method
 }
 
 func getAppEndpointURLAndProxyParams(ocClient client.Client, kubeClient *kubernetes.Clientset, kubeConfig *rest.Config, namespace, serviceName, routeName string) (string, *ProxyPodParameters, error) {
-	appEndpointURL, err := GetServiceEndpointURL(ocClient, namespace, serviceName, routeName)
+	appEndpointURL, err := GetRouteEndpointURL(ocClient, namespace, routeName)
 	// Something wrong with standard endpoint, try with proxy pod.
 	if err != nil {
-		log.Println("Can not connect to the application endpoint with route or service name:", err)
+		log.Println("Can not connect to the application endpoint with route:", err)
 		log.Println("Trying to get to the service via proxy POD")
 
 		pod, podErr := GetFirstPodByLabel(kubeClient, namespace, "curl-tool=true")
diff --git a/tests/e2e/lib/common_helpers.go b/tests/e2e/lib/common_helpers.go
index 1aff04a..2cc91bf 100644
--- a/tests/e2e/lib/common_helpers.go
+++ b/tests/e2e/lib/common_helpers.go
@@ -153,46 +153,6 @@ func GetInternalServiceEndpointURL(namespace, serviceName string, servicePort ..
 	return fmt.Sprintf("http://%s.%s.svc.cluster.local:%d", serviceName, namespace, port)
 }
 
-// GetReachableInternalServiceEndpointURL verifies the accessibility of a Kubernetes service within the same cluster
-//
-// Parameters:
-//   - namespace:      The Kubernetes namespace in which the service is located.
-//   - serviceName:    The name of the Kubernetes service.
-//   - port:           The port on which service should be accessible (optional)
-//
-// Returns:
-//   - string:         The full service endpoint URL (including the specified or default port) if the service is accessible.
-//   - error:          An error message if the service is not accessible or if there is an issue with the HTTP request.
-//
-// Example Usage:
-//
-//	serviceEndpoint, err := GetReachableInternalServiceEndpointURL("mongo-persistent", "todolist")
-//	if err != nil {
-//	    log.Println("Error:", err)
-//	} else {
-//	    log.Println("Service todolist is accessible at:", serviceEndpoint)
-//	}
-func GetReachableInternalServiceEndpointURL(namespace, serviceName string, port ...int) (string, error) {
-
-	var serviceEndpoint string
-
-	if len(port) > 0 {
-		serviceEndpoint = GetInternalServiceEndpointURL(namespace, serviceName, port[0])
-	} else {
-		serviceEndpoint = GetInternalServiceEndpointURL(namespace, serviceName)
-	}
-
-	log.Printf("Verifying if the service is accessible: %s", serviceEndpoint)
-
-	resp, err := IsURLReachable(serviceEndpoint)
-
-	if err != nil || resp == false {
-		return "", fmt.Errorf("Failed to connect to the service: %v", err)
-	}
-
-	return serviceEndpoint, nil
-}
-
 // ConvertJsonStringToURLParams takes a JSON string as input and converts it to URL-encoded parameters.
 // It returns a string containing the URL-encoded parameters.
 //
diff --git a/tests/e2e/lib/ocp_common_helpers.go b/tests/e2e/lib/ocp_common_helpers.go
index 3ae1c09..e09efdd 100644
--- a/tests/e2e/lib/ocp_common_helpers.go
+++ b/tests/e2e/lib/ocp_common_helpers.go
@@ -10,50 +10,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-// GetServiceEndpointURL retrieves and verifies the accessibility of a Kubernetes service or route endpoint.
-//
-// Parameters:
-//   - ocClient:     An instance of the OpenShift client.
-//   - namespace:    The Kubernetes namespace in which the service or route is located.
-//   - serviceName:  The name of the Kubernetes service (optional, routeName, serviceName or both must be provided).
-//   - routeName:    The name of the Kubernetes service route (optional, routeName, serviceName or both must be provided).
-//
-// Returns:
-//   - string:       The full service or route endpoint URL if accessible.
-//   - error:        An error message if both service and route endpoints are not accessible, if neither serviceName nor routeName is provided, or if there is an issue with the HTTP request.
-//
-// Example Usage:
-//
-//	appEndpoint, err := GetServiceEndpointURL(ocClient, "mongo-persistent", "todolist", todolist-route")
-//	if err != nil {
-//	    log.Println("Error:", err)
-//	} else {
-//	    log.Println("URL endpoint is accessible at:", appEndpoint)
-//	}
-func GetServiceEndpointURL(ocClient client.Client, namespace, serviceName, routeName string) (string, error) {
-	if serviceName == "" && routeName == "" {
-		return "", fmt.Errorf("Either serviceName or routeName must be provided")
-	}
-
-	if serviceName != "" {
-		internalServiceEndpoint, err := GetReachableInternalServiceEndpointURL(namespace, serviceName)
-		if err == nil {
-			return internalServiceEndpoint, nil
-		}
-		log.Printf("Failed to use Service endpoint for service %s in namespace %s: %v", serviceName, namespace, err)
-	}
-
-	if routeName != "" {
-		routeEndpoint, err := getRouteEndpointURL(ocClient, namespace, routeName)
-		if err != nil {
-			return "", fmt.Errorf("Failed to get both service and route endpoints: %v", err)
-		}
-		return routeEndpoint, nil
-	}
-
-	return "", fmt.Errorf("serviceName or routeName must be provided")
-}
-
 // getRouteEndpointURL retrieves and verifies the accessibility of a Kubernetes route HOST endpoint
 //
 // Parameters:
@@ -73,8 +29,8 @@ func GetServiceEndpointURL(ocClient client.Client, namespace, serviceName, route
 //	} else {
 //	    log.Println("Service route todolist-route is accessible at:", routeEndpoint)
 //	}
-func getRouteEndpointURL(ocClient client.Client, namespace, routeName string) (string, error) {
-	log.Println("Fallback verifying if the service is accessible via route")
+func GetRouteEndpointURL(ocClient client.Client, namespace, routeName string) (string, error) {
+	log.Println("Verifying if the service is accessible via route")
 	route := &routev1.Route{}
 	routev1.AddToScheme(ocClient.Scheme())
 	err := ocClient.Get(context.Background(), client.ObjectKey{Namespace: namespace, Name: routeName}, route)
@@ -88,7 +44,7 @@ func getRouteEndpointURL(ocClient client.Client, namespace, routeName string) (s
 	routeEndpoint := "http://" + route.Spec.Host
 
 	// Check if the route is accessible
-	log.Printf("Fallback verifying if the service is accessible via: %s", routeEndpoint)
+	log.Printf("Verifying if the service is accessible via: %s", routeEndpoint)
 	resp, err := IsURLReachable(routeEndpoint)
 	if err != nil || resp == false {
 		return "", fmt.Errorf("Route endpoint not accessible: %v", err)
```

tests/e2e/lib/apps.go
```diff
@@ -449,105 +445,138 @@ func RunMustGather(oc_cli string, artifact_dir string) error {
...
-		if _, err := os.Stat(backupFile); err == nil {
-			os.Remove(backupFile)
-		}
-		//data before curl request
-		dataBeforeCurl, err := getResponseData(appEndpoint)
+		// Clean up existing backup file
+		RemoveFileIfExists(artifactDir + "/backup-data.txt")
+
+		// Make requests and update data before backup
+		dataBeforeCurl, _, err := MakeRequest(*requestParamsTodoIncomplete)
```
mateusoliveira43, line 464: All places that call `MakeRequest` (to check it `grep -Inr 'MakeRequest' tests/e2e`) ignore second return variable. Use it or remove from function

tests/e2e/lib/k8s_common_helpers.go
```diff
@@ -0,0 +1,279 @@
...
+		LabelSelector: labelSelector,
+	}
+
+	podList, err := clientset.CoreV1().Pods(namespace).List(context.Background(), podOptions)
+
+	if err != nil {
+		return nil, err
+	}
+
+	if len(podList.Items) == 0 {
```
mateusoliveira43, line 215: should not the check be for length greater then 1?

tests/e2e/lib/k8s_common_helpers.go
```diff
@@ -0,0 +1,279 @@
...
+//
+// Parameters:
+//   - clientset: A pointer to the Kubernetes client (*kubernetes.Clientset).
+//   - namespace: The namespace in which to search for the pod.
+//   - labelSelector: The label selector to filter pods.
+//
+// Returns:
+//   - (*corev1.Pod, error): A pointer to the first pod matching the label selector, or an error if any.
+//
+// Example:
```
mateusoliveira43, line 196: I would remove examples from Doc comments. I think the code itself acts as example, but it harmless to let them

mpryc: Not everywhere, I would leave it for example here:

```
// ConvertJsonStringToURLParams takes a JSON string as input and converts it to URL-encoded parameters.
// It returns a string containing the URL-encoded parameters.
//
// Parameters:
//   - payload (string): The JSON string to be converted.
//
// Returns:
//   - string: The URL-encoded parameters.
//   - error: An error, if any, during the conversion process.
//
// Example:
//
//	input:  `{"name": "John", "age": 30, "city": "New York"}`
//	output: `&{name=John&age=30&city=New+York}`

```

tests/e2e/lib/k8s_common_helpers.go
mpryc: Why?
See the error for that if (no pods, so 0):
```
		return nil, fmt.Errorf("no pods found with label selector: %s", labelSelector)
```

tests/e2e/lib/apps.go
mpryc: Ok I've made that change. We are interested in that message and separate this from other errors, because we need to know if it was transport or for example api issue.

tests/e2e/lib/k8s_common_helpers.go
mpryc: Removed from the examples how to use functions, but left the above one, because it's important to know what example input/output is expected as payload.

tests/e2e/lib/k8s_common_helpers.go
mateusoliveira43: If you get more than 1 result, command may end up executing in different pods during runs, which can create unreproducible behaviors?

mpryc: No, that is using route from any container within cluster that is labeled, we don't care which one. The function name is pretty precise and I have change the function name to address comment from @kaovilai, because at first was not clear that's it's first matching pod.


```
// GetFirstPodByLabel retrieves a first found pod in the specified namespace based on the given label selector.
// It uses the provided Kubernetes client to interact with the Kubernetes API.
//

// Returns:
//   - (*corev1.Pod, error): A pointer to the first pod matching the label selector, or an error if any.
func GetFirstPodByLabel
```

mpryc:
The failing test is `VM backup and restore tests`, @mrnold does those tests also require flake:

https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1253/pull-ci-openshift-oadp-operator-master-4.13-e2e-test-aws/1752981029745856512/artifacts/e2e-test-aws/e2e/build-log.txt

```
2024/02/01 11:06:35 HCO health status is: healthy
  [FAILED] Expected
      <*errors.errorString | 0xc0001cff70>: 
      timed out waiting for the condition
      {
          s: "timed out waiting for the condition",
      }
```

mpryc:
> This failure happened on cleanup, when it was waiting to remove the HCO instance. I don't think this should need a flake, since the cleanup is not really the point of the test. Maybe HCO removal can be a best-effort kind of thing that does not cause the whole test to fail? I can take out the Expect that triggered this.

We do remove/cleanup when the backup is done before restore and after the test ran. I agree that cleanup should not fail the test, unless it may introduce issues while restoring (first case just after backup) ?.

mrnold:
I added a new pull request to remove the Expect: https://github.com/openshift/oadp-operator/pull/1303

tests/e2e/lib/k8s_common_helpers.go
kaovilai: > it will use route from any container within cluster that is labeled as a proxy

Do you mean it will only get pods with label curltool which is a deployment we control so we would limit replicas to 1 anyway? Or if they are multiple replicas, the application would be mounting the same PVC thus returning the same data for volume persistence tests.

mpryc: @kaovilai yes we do control deployment and we limit to 1 anyway, but imo it's always better in a function act to return only what is needed that's why there is limit of 1. To be fair even if you have multiple deployments in one namespace and for example left overs from other jobs or additionally deployed apps it will still work. Within the same namespace we really don't care who will act as a proxy as it will always find it's way via internal routing to the service within the same namespace.

https://github.com/openshift/oadp-operator/pull/1253/files#diff-454849abc186dbe835e4d394e7142c24d4afcc776d63d556bae8df350db2d723R66

If the user has access to other namespaces the above statement will be valid across multiple namespaces as we try to reach the endpoint using it's service name and internal hostname:

https://github.com/openshift/oadp-operator/pull/1253/files#diff-f56b125140f13a9f4b7a6959d7a2224a1368223762f064fe7c1490729a490f0eR147

kaovilai (approved):

weshayutin (approved):

PR #1257, opened 2023-12-12
CI: install virtualization operator to enable VM backup/restore testing

mrnold:
This pull request adds a place to put VM backup/restore testing, and adds a helper library for installing the Openshift Virtualization operator automatically. The only new test included is a check to see that the installation was successful, actual VM testing will be added in subsequent pull requests.


tests/e2e/scripts/virt_install.sh
mateusoliveira43, line 1: the scripts in this folder are all bash, but this one seems a little more complex. Can it be written in go?

mrnold: Yes, I have a Go version partially written. I can finish that for the final pull request.

weshayutin:
```
Checking CNV CSV...
Error from server (NotFound): clusterserviceversions.operators.coreos.com "kubevirt-hyperconverged-operator.v4.13.6" not found
CSV phase is 
Failed installing CSV after two minutes!
```

mrnold:
No more failures installing the virtualization operator, that's good. Hopefully the Mongo datamover failures are unrelated.

mrnold:
I am only seeing consistent virtualization installation failures on 4.12, other failures look like unrelated tests that are also failing in other pull requests. I will look into the 4.12 failures, but otherwise this can be reviewed.

Makefile
```diff
@@ -492,6 +493,8 @@ test-e2e-setup: login-required
 	VSL_REGION="$(VSL_REGION)" \
 	BSL_REGION="$(BSL_REGION)" \
 	BSL_AWS_PROFILE="$(BSL_AWS_PROFILE)" \
+	OC_CLI="$(OC_CLI)" \
```
mateusoliveira43, line 496: These env vars are not needed anymore here, right?

mrnold: Right, the *_settings.sh scripts don't need either of these.

tests/e2e/e2e_suite_test.go
```diff
@@ -38,6 +38,7 @@ func init() {
 	flag.StringVar(&oc_cli, "oc_cli", "oc", "OC CLI Client")
 	flag.StringVar(&stream, "stream", "up", "[up, down] upstream or downstream")
 	flag.Int64Var(&timeoutMultiplierInput, "timeout_multiplier", 1, "Customize timeout multiplier from default (1)")
+	flag.BoolVar(&virtTestingEnabled, "test_virt", false, "Flag to enable or disable virtualization testing, true or false")
```
mateusoliveira43, line 41: would put the same as in docs: Enable Virtual Machine backup/restore testing (default false)

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,44 @@
...
+	"os"
+	"strings"
+	"time"
+
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	. "github.com/openshift/oadp-operator/tests/e2e/lib"
+)
+
+var virtTestingEnabled bool
```
mateusoliveira43, line 13: I would declare the variable and also move the os.Getenv logic to `tests/e2e/e2e_suite_test.go`

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,44 @@
...
+	if strings.ToLower(os.Getenv("TEST_VIRT")) == "true" {
+		virtTestingEnabled = true
+	}
+}
+
+var _ = Describe("VM backup and restore tests", Ordered, func() {
+	var v *VirtOperator
+	var err error
+
+	var _ = BeforeAll(func() {
```
mateusoliveira43, line 25: should we also uninstall it?

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+// GetVirtOperator fills out a new VirtOperator
+func GetVirtOperator(client client.Client, clientset *kubernetes.Clientset, stream string) (*VirtOperator, error) {
+	operatorType := DownstreamOperator
+	namespace := "openshift-cnv"
+	if stream == "up" {
+		operatorType = CommunityOperator
+		namespace = "kubevirt-hyperconverged"
+	}
+
+	config, err := ctrl.GetConfig()
```
mateusoliveira43, line 47: here I think you can use `kubeConf` from `tests/e2e/e2e_suite_test.go`

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+	if err != nil {
+		log.Printf("Failed to get dynamic client config: %v", err)
+		return nil, err
+	}
+
+	dynamicClient, err := dynamic.NewForConfig(config)
```
mateusoliveira43, line 53: would declare this client in `tests/e2e/e2e_suite_test.go`, for standardization

tests/e2e/lib/subscription_helpers.go
```diff
@@ -39,6 +30,33 @@ func (d *DpaCustomResource) GetOperatorSubscription(c client.Client, stream stri
...
+		label = "operators.coreos.com/oadp-operator." + d.Namespace
+	}
+	if stream == "down" {
+		label = "operators.coreos.com/redhat-oadp-operator." + d.Namespace
+	}
+	return GetOperatorSubscription(c, d.Namespace, label)
+}
+
+func (v *VirtOperator) GetOperatorSubscription(stream string) (*Subscription, error) {
+	label := ""
```
mateusoliveira43, line 50: are both streams necessary? are we interested in testing both?
Screenshot from 2023-12-28 14-20-35

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+	err := v.Client.Get(context.TODO(), client.ObjectKey{Namespace: v.Namespace, Name: "hco-operatorhub"}, &subscription)
+	if err != nil {
+		return false
+	}
+	return true
+}
+
+// Checks if the ClusterServiceVersion status has changed to ready
+func (v *VirtOperator) checkCsv() bool {
+	subscription, err := v.GetOperatorSubscription("down")
```
mateusoliveira43, line 153: instead of hardcoding this, can we just test one?

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+	}
+
+	unstructuredHco := unstructured.Unstructured{
+		Object: map[string]interface{}{
+			"apiVersion": "hco.kubevirt.io/v1beta1",
+			"kind":       "HyperConverged",
+			"metadata": map[string]interface{}{
+				"name":      "kubevirt-hyperconverged",
+				"namespace": v.Namespace,
+				"spec":      "{}",
```
mateusoliveira43, line 268: spec outside metadata? https://docs.openshift.com/container-platform/4.12/virt/install/installing-virt-cli.html#virt-deploying-operator-cli_installing-virt-cli

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+	log.Println("Created subscription")
+
+	log.Printf("Waiting for ClusterServiceVersion")
+	if err := v.ensureCsv(5 * time.Minute); err != nil {
+		return err
+	}
+	log.Println("CSV ready")
+
+	log.Printf("Creating hyperconverged operator")
+	if err := v.ensureHco(15 * time.Minute); err != nil {
```
mateusoliveira43, line 411: 5 should be enough? 15 is too much

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,417 @@
...
+		log.Printf("Error getting HCO: %v", err)
+		return false
+	}
+
+	health, ok, err := unstructured.NestedString(unstructuredHco.UnstructuredContent(), "status", "systemHealthStatus")
+	if err != nil {
+		log.Printf("Error getting HCO health: %v", err)
+		return false
+	}
+	if !ok {
```
mateusoliveira43, line 184: 4.12 jobs are failing here, maybe field does not exist there? https://github.com/kubevirt/hyperconverged-cluster-operator/commit/bab7b17a146b31ffbc4b749f6c70189245de64e8

I have 4.12 cluster, can try to debug (did not install kubevirt before)

mateusoliveira43 (changes requested):
fix 4.12 installations

tests/e2e/virt_backup_restore_suite_test.go
mrnold: I tried it this way first, but it got messy dealing with uninstalls stuck on finalizers and trying to only uninstall the operator if the test was the one to install it. It is possible, but I'm not sure it's worth the effort yet.

tests/e2e/lib/subscription_helpers.go
mrnold: I don't really know, I just saw that there was existing subscription-related code that did the same thing and copied it. I took this out to make it simpler for now.

tests/e2e/lib/virt_helpers.go
mrnold: Yes, that was the problem. I could probably create my own check that does the same thing, but it was easier just to skip these tests on versions older than 4.13.

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43: can we just add in the docs how to unistall it then? so someone running locally the test can return the cluster to original state after running the tests

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,41 @@
...
+		if !virtTestingEnabled {
+			Skip("Virtualization testing is disabled, skipping tests")
+		}
+
+		Expect(runTimeClientForSuiteRun).ToNot(BeNil())
```
mateusoliveira43, line 21: in `e2e_suite_test.go`we check if no error occurred when creating each client, so I think we do need to check if they are not nil here

tests/e2e/e2e_suite_test.go
```diff
@@ -92,6 +99,7 @@ var kubernetesClientForSuiteRun *kubernetes.Clientset
 var runTimeClientForSuiteRun client.Client
 var veleroClientForSuiteRun veleroClientset.Interface
 var csiClientForSuiteRun *snapshotv1client.Clientset
+var dynamicClient dynamic.Interface
```
mateusoliveira43, line 102: for standardization, add `ForSuiteRun` suffix to variable name

tests/e2e/lib/virt_helpers.go
mateusoliveira43: here I think `spec` is still inside `metadata`. It should be in the same level of apiVersion, Kind, and metadata, no?

tests/e2e/lib/virt_helpers.go
mateusoliveira43: just to we not forget, would create an issue to remove this code when we update CI and minimal version becomes 4.13

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Was this "do not need to check"? I took these checks out, but I can add checks for every client instead if I read this the wrong way.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Sure, I will see what I can do that is convenient for local tests.

tests/e2e/lib/virt_helpers.go
mrnold: Okay, if this code actually gets merged I will create an issue.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+	err := v.Client.Create(context.Background(), &corev1.Namespace{ObjectMeta: metav1.ObjectMeta{Name: v.Namespace}})
+	if err != nil {
+		log.Printf("Failed to create namespace %s: %v", v.Namespace, err)
+		return err
+	}
+	return nil
+}
+
+// Creates the virtualization operator group
+func (v *VirtOperator) installOperatorGroup() error {
```
mpryc, line 196: Same comment. I like simplicity of installOperatorGroup function, but possibly it should be just calling onother common method with name/namespace. Those functions are accessible within `&operatorsv1.*` but to simplify our code would be great to have such call in one place and the `installOperatorGroup` just a minimum required metadata to install such group. We may us then such call for other operatorgroups.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+		if !strings.Contains(err.Error(), "already exists") {
+			log.Printf("Failed to create operator group: %v", err)
+			return err
+		}
+	}
+	return nil
+}
+
+// Creates the subscription, which triggers creation of the ClusterServiceVersion.
+func (v *VirtOperator) installSubscription() error {
```
mpryc, line 219: Similar to above comment.

Our CatalogSource, CatalogSourceNamespace could be another type, this way we can have easy way of installing operators from there? Just a thought, which may not be worth changing.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+		return err
+	}
+
+	return nil
+}
+
+// Creates a HyperConverged Operator instance. Another dynamic client to avoid
+// bringing in the KubeVirt APIs for now.
+func (v *VirtOperator) installHco() error {
+	resourceId := schema.GroupVersionResource{
```
mpryc, line 246: Could this schema be defined outside of the code or at least somewhere common together with other schema attributes which are not really functional.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+	_, err := v.Dynamic.Resource(resourceId).Namespace(v.Namespace).Create(context.Background(), &unstructuredHco, v1.CreateOptions{})
+	if err != nil {
+		log.Printf("Error creating HCO: %v", err)
+		return err
+	}
+
+	return nil
+}
+
+// Helper function to poll status with a timeout
```
mpryc, line 273: I don't understand what it waits for? When reading the code I got it, just should be more straight forward from the desc.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+			return err
+		}
+	} else {
+		log.Printf("Subscription already created, no action required")
+	}
+
+	return nil
+}
+
+// Waits for the ClusterServiceVersion to go to ready, triggered by subscription
```
mpryc, line 338: Is that required? Why not to wait for csv when the csv is being created?

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+func (v *VirtOperator) ensureCsv(timeout time.Duration) error {
+	return waitForCheck(timeout, 5*time.Second, v.checkCsv, "timed out waiting for CSV to become ready")
+}
+
+// Creates HCO instance if needed, and waits for it to go healthy
```
mpryc, line 343: I would use full  HyperConverged Operator in the desc as I had to go back to remember what the HCO stands for :)

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+		log.Printf("HCO already created, no action required")
+	}
+
+	return nil
+}
+
+// IsVirtInstalled returns whether or not the OpenShift Virtualization operator
+// is installed and ready, by checking for a HyperConverged operator resource.
+func (v *VirtOperator) IsVirtInstalled() bool {
+	if !v.checkNamespace() {
```
mpryc, line 362: Should this be part of v.checkHco? the v knows the namespace already and the checkHco() may simply return error if it's not installed within that NS. I don't think the ns check should be separate from checkHco() as one needs to be for other.

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,39 @@
...
+		Expect(err).To(BeNil())
+		Expect(v).ToNot(BeNil())
+
+		minimum, err := version.ParseGeneric("4.13")
+		Expect(err).To(BeNil())
+		if !v.Version.AtLeast(minimum) {
+			Skip("Skipping virtualization testing on cluster version " + v.Version.String() + ", minimum required is " + minimum.String())
+		}
+
+		err = v.EnsureVirtInstallation(5 * time.Minute)
```
mpryc, line 31: The errors from many functions underneath are just as the Expect, should this be as well presented using log?

mpryc (commented):
I was unable to test this on my cluster as I am still on 4.12.x and those tests are required to be ran on 4.13+. Some of the comments within the code itself.

```
VM backup and restore tests
/home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:12
  should verify virt installation
  /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:35
  > Enter [BeforeAll] VM backup and restore tests - /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:16 @ 01/12/24 11:41:58.07
2024/01/12 11:41:58 Getting packagemanifest...
2024/01/12 11:41:58 Extracting channels...
2024/01/12 11:41:58 Current CSV is: kubevirt-hyperconverged-operator.v4.12.8
2024/01/12 11:41:58 Current operator version is: 4.12.8
  [SKIPPED] Skipping virtualization testing on cluster version 4.12.8, minimum required is 4.13
  In [BeforeAll] at: /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:28 @ 01/12/24 11:41:58.6
  < Exit [BeforeAll] VM backup and restore tests - /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:16 @ 01/12/24 11:41:58.6 (530ms)
S [SKIPPED] [0.530 seconds]
VM backup and restore tests [BeforeAll]
/home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:16
  should verify virt installation
  /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/virt_backup_restore_suite_test.go:35
------------------------------
[AfterSuite] 
/home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:159
  > Enter [AfterSuite] TOP-LEVEL - /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:159 @ 01/12/24 11:41:58.6
2024/01/12 11:41:58 Deleting Velero CR
  [FAILED] Unexpected error:
      <*errors.errorString | 0xc00054f390>: 
      resource name may not be empty
      {
          s: "resource name may not be empty",
      }
  occurred
  In [AfterSuite] at: /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:168 @ 01/12/24 11:41:58.737
  < Exit [AfterSuite] TOP-LEVEL - /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:159 @ 01/12/24 11:41:58.737 (137ms)
[AfterSuite] [FAILED] [0.137 seconds]
[AfterSuite] 
/home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:159
------------------------------

Summarizing 1 Failure:
  [FAIL] [AfterSuite] 
  /home/migi/Development/Upstream/OADP/oadp-operator-virt-12.01.2024/tests/e2e/e2e_suite_test.go:168

Ran 0 of 32 Specs in 1.212 seconds
FAIL! -- 0 Passed | 0 Failed | 0 Pending | 32 Skipped
--- FAIL: TestOADPE2E (1.22s)
FAIL

Ginkgo ran 1 suite in 13.779622014s

Test Suite Failed

```

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43: yes yes, forget to add "not", thanks!

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+		log.Printf("HCO health field not populated yet")
+		return false
+	}
+	log.Printf("HCO health status is: %s", health)
+
+	return health == "healthy"
+}
+
+// Creates the target virtualization namespace, likely openshift-cnv or kubevirt-hyperconverged
+func (v *VirtOperator) installNamespace() error {
```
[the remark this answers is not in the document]
mateusoliveira43: I would not do this kind of refactor in this PR. That can create questions as why not to use `tests/e2e/lib/kube_helpers.go` for this? It can be follow up PR

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+	Client    client.Client
+	Clientset *kubernetes.Clientset
+	Dynamic   dynamic.Interface
+	Namespace string
+	Csv       string
+	Version   *version.Version
+}
+
+// GetVirtOperator fills out a new VirtOperator
+func GetVirtOperator(client client.Client, clientset *kubernetes.Clientset, dynamicClient dynamic.Interface) (*VirtOperator, error) {
```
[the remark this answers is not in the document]
mrnold: In this file, I think GetVirtOperator, IsVirtInstalled, and EnsureVirtInstallation need to stay exported. They are called from virt_backup_restore_suite_test.go.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+		return "", nil, err
+	}
+
+	return csv, version, nil
+}
+
+// Checks the existence of the operator's target namespace
+func (v *VirtOperator) checkNamespace() bool {
+	// First check that the namespace exists
+	exists, _ := DoesNamespaceExist(v.Clientset, v.Namespace)
```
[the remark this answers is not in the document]
mrnold: Where is this note? I don't see it in the current version: https://github.com/openshift/oadp-operator/blob/master/tests/e2e/lib/kube_helpers.go#L19

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,41 @@
...
+
+		Expect(runTimeClientForSuiteRun).ToNot(BeNil())
+		Expect(kubernetesClientForSuiteRun).ToNot(BeNil())
+		v, err = GetVirtOperator(runTimeClientForSuiteRun, kubernetesClientForSuiteRun, dynamicClient)
+		Expect(err).To(BeNil())
+		Expect(v).ToNot(BeNil())
+
+		minimum, err := version.ParseGeneric("4.13")
+		Expect(err).To(BeNil())
+		if !v.Version.AtLeast(minimum) {
```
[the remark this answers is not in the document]
mrnold: I added a nil check for v.Version, if that's what you were asking for?

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -0,0 +1,39 @@
...
+			Skip("Skipping virtualization testing on cluster version " + v.Version.String() + ", minimum required is " + minimum.String())
+		}
+
+		err = v.EnsureVirtInstallation(5 * time.Minute)
+		Expect(err).To(BeNil())
+	})
+
+	It("should verify virt installation", func() {
```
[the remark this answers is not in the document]
mrnold: I think this is just because it is a standalone test, not in a table of tests yet. I am sure there will be a table eventually, and this will be an Entry (or more likely get removed altogether).

tests/e2e/lib/subscription_helpers.go
```diff
@@ -39,6 +30,27 @@ func (d *DpaCustomResource) GetOperatorSubscription(c client.Client, stream stri
...
+	if stream == "up" {
+		label = "operators.coreos.com/oadp-operator." + d.Namespace
+	}
+	if stream == "down" {
+		label = "operators.coreos.com/redhat-oadp-operator." + d.Namespace
+	}
+	return GetOperatorSubscription(c, d.Namespace, label)
+}
+
+func (v *VirtOperator) GetOperatorSubscription() (*Subscription, error) {
```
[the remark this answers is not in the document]
mrnold: I think this only works because virt_helpers and subscription_helpers are in the same package? I am pretty sure Go doesn't otherwise let you add arbitrary functions to types like this, so it might not be applicable everywhere.

tests/e2e/lib/virt_helpers.go
```diff
@@ -0,0 +1,408 @@
...
+	return exists
+}
+
+// Checks for the existence of the virtualization operator group
+func (v *VirtOperator) checkOperatorGroup() bool {
```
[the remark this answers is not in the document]
mrnold: These ones (check*, install*, ensure*) are definitely not meant to be called from outside. GetOperatorSubscription is capitalized just to match the existing subscription helper code, but lower case does make more sense because it is only used privately.

The others that are still capitalized need to stay that way so they can be called from the test suite code.

tests/e2e/lib/virt_helpers.go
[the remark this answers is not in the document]
mrnold: When common_k8s_utils is merged, installNamespace would be a good candidate to put there. But probably named CreateNamespace or something. Maybe installOperatorGroup too, but probably not the others.

tests/e2e/lib/virt_helpers.go
mrnold: Yes, having this defer to a common CreateOperatorGroup would be nice.

tests/e2e/lib/virt_helpers.go
mrnold: This one I am less sure about, because it would likely look pretty much the same. It's pretty much just passing eight arguments in a Subscription to Create, and checking the result. It would be nice to have a whole factored-out Operator installation module, but I don't know that it's worth building that until we have more operators to install.

tests/e2e/lib/virt_helpers.go
mrnold: I'm not totally sure what you are asking for here, but I moved the GVRs together to the top of the file.

tests/e2e/lib/virt_helpers.go
mrnold: I added more words, and moved it to common_utils.

tests/e2e/lib/virt_helpers.go
mrnold: I don't know what is really required or not here. It's just the way I found to get the operator installed, directly translated from the command line version.

mpryc: I am fine waiting here, just move it maybe inside the subscription function in it's body rather then separate function outside of the subscription. This way user will always wait for it to be ready. Otherwise consumer of function which creates subscription will need to remember to also wait for csv.

tests/e2e/lib/virt_helpers.go
mrnold: Right, the HCO check is the final thing that this uses to determine whether virtualization is good to go, and it depends on the namespace existing. But the public VirtOperator interface is only IsVirtInstalled and EnsureVirtInstallation, checkHco is an internal piece that is used correctly by both already.

The private functions are organized into check*, install*, and ensure*, where each * deals with just one thing. Then the public functions are responsible for using the private ones in the right order, which includes a namespace check (or creation) before doing anything with the HCO instance. So I don't see what would be the benefit of an extra namespace check in checkHco, unless we also want every other check function to also check all of its down dependencies.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: The virt_helpers stuff is reasonably well logged, and I was thinking ginkgo logs any error from this Expect nicer than I would anyway. Is there something else that needs to be logged?

tests/e2e/lib/virt_helpers.go
mrnold: I think that would be a good plan for a generic operator installer module, but here the only public interface is IsVirtInstalled and EnsureVirtInstallation. A consumer of these should not have to think about subscriptions, CSVs, or HCOs at all.

These could certainly be combined, but I am slightly more in favor of keeping it split into smaller chunks. If for whatever reason the Subscription creation failed, having them combined would mean you would have to wait for the longer timeout (5 minutes vs. 10 seconds) to fail to test. That's a small and unlikely failure mode, but enough of a reason to demand further convincing in favor of combining them.

tests/e2e/lib/virt_helpers.go
mateusoliveira43: I think this function can be deleted, ginkgo has a built in function that does a very similar job

example on how to use it https://github.com/openshift/oadp-operator/blob/master/tests/e2e/lib/apps.go#L580

mateusoliveira43 (dismissed):
Just need to check if WaitForCheck function can be replaced by eventually and add docs on how to uninstall hco operator

tests/e2e/virt_backup_restore_suite_test.go
mpryc: That's fine. thanks for clarification.

tests/e2e/virt_backup_restore_suite_test.go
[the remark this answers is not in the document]
mpryc: Thanks.

tests/e2e/lib/virt_helpers.go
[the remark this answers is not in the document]
mpryc: You are right, don't know why I've checked so old version of this file:

https://github.com/openshift/oadp-operator/commit/a2d85f4f00d70db4d9a0555266713bd8785deb17#diff-06a8d2a8ef8552713ec00f3c9653d7c9727cf45a1b0bfa9c47a24929ba0e94e6L127-R78

tests/e2e/lib/virt_helpers.go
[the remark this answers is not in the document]
mpryc: ok.

tests/e2e/lib/virt_helpers.go
mpryc: Thanks, makes sense.

tests/e2e/lib/virt_helpers.go
mpryc: If we have long running concurrent operations then it makes sense to first ask for all of them and then check if we got them independently, so we save some blocking calls. But I am fine having a split here as well.

tests/e2e/lib/virt_helpers.go
mrnold: I didn't see ginkgo/gomega imported in any of the other *_helpers files, so I thought maybe it wasn't supposed to be used there for some reason. I did find (via nodeagent_helpers.go) a Kubernetes API machinery module called wait, with a PollImmediate function that does pretty much the same thing. I would rather use this one than the one from gomega, but either option is definitely better than writing a new one.

mrnold: Okay, replaced this with PollImmediate from apimachinery.

mateusoliveira43 (approved):
just missing docs on how to uninstall hco operator

mrnold:
> just missing docs on how to uninstall hco operator

I added some code to do the removal, just not automatically. Does this look any better than a doc? https://github.com/openshift/oadp-operator/pull/1257/commits/6db4abb185684b034785acb0aa33dc9864eed267

Makefile
```diff
@@ -525,3 +527,11 @@ test-e2e-cleanup: login-required
 	$(OC_CLI) delete restore -n $(OADP_TEST_NAMESPACE) --all --wait=false
 	for restore_name in $(shell $(OC_CLI) get restore -n $(OADP_TEST_NAMESPACE) -o name);do $(OC_CLI) patch "$$restore_name" -n $(OADP_TEST_NAMESPACE) -p '{"metadata":{"finalizers":null}}' --type=merge;done
 	rm -rf $(SETTINGS_TMP)
+
+.PHONY: install-virt
```
mateusoliveira43, line 531: I would undo this change. We used a similar approach with volsync in the past

https://github.com/openshift/oadp-operator/commit/b73b5c1207a1e199403b192c23ed49f76c1f450d#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L467-L470

https://github.com/openshift/oadp-operator/commit/a76c5dc2e2ef3706e9827933f45faddf0aae2c17#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52L473-L493

but it did not install through test run, only through makefile. This way, we would have two options (test run and makefile). Would stick only with test run, removing this and `deploy/virt-operator.go` script. Since uninstall logic is already written, would add it to afterAll of virt test or add it in a follow up PR, after this one merges (PREFERRED)

mrnold: Sure, I put an AfterAll to do this now instead, with a quick check to only remove it if it was the test that installed it in the first place.

deploy/virt-operator.go
mateusoliveira43, line 1: this file is still used?

mateusoliveira43 (dismissed):

deploy/virt-operator.go
mrnold: Nope, I just forgot to remove it in the last commit.

mpryc:
The `MySQL application two Vol CSI` found out flake, however 3 attempts of that test were not enough to run it.

The `Mongo application DATAMOVER` failure is not know flake, however it times out and more insight is needed to find out why.

kaovilai (approved):
Per PR description, the changes here LGTM.

PR #1261, opened 2024-01-05
feat: Nullable logic in code

mateusoliveira43:
Remove logic for adding nullable fields from Makefile and add it to code.


go.mod
```diff
@@ -15,7 +15,7 @@ require (
 	github.com/operator-framework/operator-lib v0.9.0
 	github.com/prometheus-operator/prometheus-operator/pkg/apis/monitoring v0.51.2
 	github.com/sirupsen/logrus v1.9.0
-	k8s.io/api v0.25.6
+	k8s.io/api v0.25.6 // must update OADPResourceRequirements in api/v1alpha1/oadp_types.go when package updates
```
kaovilai, line 18: Can we create a unit test that turns resoureRequirements into unstructured and run deepEqual against current expected value? That way we know when change is required.

mateusoliveira43: nice idea!

Going to try to do that, thanks!

mateusoliveira43: did not find a way to converted to unstructured. How can I do it?

did this
```go
func TestEqual(t *testing.T) {
	created := oadpv1alpha1.OADPResourceRequirements{}
	real := corev1.ResourceRequirements{}

	createdConverted := corev1.ResourceRequirements(created)
	realCoverted := oadpv1alpha1.OADPResourceRequirements(real)

	fmt.Printf("Equal: %#v\n", reflect.DeepEqual(created, realCoverted))
	fmt.Printf("Equal: %#v", reflect.DeepEqual(createdConverted, real))
}
```

But I am not satisfied with the test, basically the difference is the nullable comment, so we should check that is the only difference between the structs is that comment (get go comments in structs, does not seem to be easy, from what I searched)

Another question: do we need these fields to allow to be nullable?

kaovilai: Re: how to convert to unstructured... that's what velero does :)

Here's an example: https://github.com/vmware-tanzu/velero/blob/2caba3efb93c37edc73aac1301f68ba91a670a80/pkg/backup/item_backupper_test.go#L49

```go
content, _ := runtime.DefaultUnstructuredConverter.ToUnstructured(tt.resource)
```

kaovilai: https://pkg.go.dev/k8s.io/apimachinery@v0.29.0/pkg/runtime#DefaultUnstructuredConverter

kaovilai: it may not help with nullables but it should help track field changes.

mateusoliveira43:
This implementation still does not add all nullable keys to YAML

Giving up on it since from my search, solution would be to remove this hack from our codebase, but needs to be done in a new API version, to be sure to not break users' upgrade

Motivation and implementations
https://issues.redhat.com/browse/OADP-535
https://github.com/openshift/oadp-operator/pull/711
https://github.com/openshift/oadp-operator/pull/796
https://github.com/openshift/oadp-operator/pull/846

PR #1265, opened 2024-01-08
OADP-3260: Change operator image pull policy to IfNotPresent

shubham-pampattiwar:


mateusoliveira43:
The related issue is about performance right? Changing the value of the field can speed up some scenarios?

My concern is if can break anything? (ideally no, if we always tag images with different tags in each release, right? but on master we use latest, so probably better to just apply this change to release branches?)

kaovilai:
> My concern is if can break anything? (ideally no, if we always tag images with different tags in each release, right? but on master we use latest, so probably better to just apply this change to release branches?)

Or address this in downstream bundle automation.

shubham-pampattiwar:
As discussed in scrum we will be going ahead with the changes only in downstream csv.

PR #1215, opened 2023-11-03
OADP-3041 Update dpa args fs-backup-timeout default to match velero

kaovilai:
Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

PR #1266, opened 2024-01-12
feat: Add editorconfig to project

mateusoliveira43:
## Description

Add [editorconfig](https://editorconfig.org/) to project. This will check all project's file format, by checking:
- if there is no trailing whitespace in them
- if there is a empty newline at the end of them
- if they use the correct indent style (tab or space)
- if they have correct indent size

## Motivation

Basically is standardization.

## How to test

Run `make ec`


weshayutin:
I like the idea! Bring that python juju to golang

Makefile
```diff
@@ -525,3 +526,19 @@ test-e2e-cleanup: login-required
...
 	rm -rf $(SETTINGS_TMP)
+
+EC ?= $(shell pwd)/bin/ec-$(shell go env GOOS)-$(shell go env GOARCH)
+EC_VERSION ?= 2.7.2
+
+.PHONY: editorconfig
+editorconfig:
+	@[ -f $(EC) ] || { \
+	set -e ;\
+	curl -sSLo $(EC).tar.gz https://github.com/editorconfig-checker/editorconfig-checker/releases/download/$(EC_VERSION)/ec-$(shell go env GOOS)-$(shell go env GOARCH).tar.gz ;\
```
kaovilai, line 537: I would make this work offline. Something like https://github.com/openshift/oadp-operator/pull/1215/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R313

kaovilai: The command linked does a path check that would only invoke curl if version is mismatched.
```
	if [ ! -f $(OPERATOR_SDK) ]; then \
```

You could totally run this offline (ie. not invoking curl) after you run it the first time.

mateusoliveira43: alright, understood. I think it already does that with first if, but will check again

kaovilai: ok you're right.. it just maybe an incorrect version installed.

also instead of version checking like I did, maybe append version number to bin would be simpler.. then you'd only path check the `bin/EC-vXYZ` and good to go.

weshayutin:
@mateusoliveira43 most of this is great, thank you for doing it.  You did end up changing must-gather a bit, can you please post a manual test of oadp-must-gather. I know it's some what tested via e2e but good to recheck imho

must-gather/collection-scripts/_metrics_chart_template_part_1.html
```diff
@@ -95,4 +95,4 @@ <h1>Metrics Chart</h1>
 		};
 
 		// Browsers/CORS blocks loading local files, embeding all chart data directly into the HTML file
-		window.dataset = prepareDataset('
\ No newline at end of file
+		window.dataset = prepareDataset('
```
mateusoliveira43, line 98: I think I will need to rework this part

mateusoliveira43:
/hold
must-gather Dockerfile in master/1.3 is broken

must-gather/collection-scripts/_metrics_chart_template_part_1.html
mateusoliveira43: will add this
```editorconfig
[must-gather/collection-scripts/_metrics_chart_template_part_1.html]
insert_final_newline = false
```

to editorconfig file

mateusoliveira43:
still think this is important, but it took to long to merge

will try again with quality downstream repository

PR #1267, opened 2024-01-15
fix: Remove unused field from e2e config

mateusoliveira43:
## Description

Remove unused field (credential) from e2e config. That field is not read by the code.

## Other problems

Today, azure and gcp create non default secrets and do not add that info to SnapshotLocations in DPA, but nevertheless, this test do not break

https://github.com/openshift/oadp-operator/blob/a663e781b8d3f7912838dd9a362309e1b8dc3b22/tests/e2e/dpa_deployment_suite_test.go#L412

We should validate BSL/VSL in our tests

For consistency, SnapshotLocations build code should be updated to have similar logic to BackupLocations one

https://github.com/openshift/oadp-operator/blob/a663e781b8d3f7912838dd9a362309e1b8dc3b22/tests/e2e/lib/dpa_helpers.go#L82

https://github.com/openshift/oadp-operator/blob/a663e781b8d3f7912838dd9a362309e1b8dc3b22/tests/e2e/lib/dpa_helpers.go#L89-L94

Also, this option should be removed, as developer can not select BSL secret name for e2e run

https://github.com/openshift/oadp-operator/blob/a663e781b8d3f7912838dd9a362309e1b8dc3b22/tests/e2e/e2e_suite_test.go#L33


kaovilai (approved):

shubham-pampattiwar:
Velero BSL API does allow the user to specify a custome secret name for BSL: https://velero.io/docs/v1.6/api-types/backupstoragelocation/

mateusoliveira43:
@shubham-pampattiwar yes, agree

Sorry for the misunderstanding in the PR description. what I wanted to say by "user can not select BSL secret name" is on e2e run: a user (a developer running `make test-e2e`) can not select BSL secret name (but can select VSL secret name). So for consistency, both secret names should be generated by test run

Updated description to try to make it clearer

mateusoliveira43:
4.13 azure fail was the known route timeout post restore problem

/test 4.13-e2e-test-azure

kaovilai (approved):

weshayutin (approved):

PR #1268, opened 2024-01-15
OADP-3165: Update default restore resource priorities

shubham-pampattiwar:


mateusoliveira43 (changes requested):
Update default in documentation (I think is here https://github.com/openshift/oadp-operator/blob/master/pkg/velero/server/config.go#L41) and can you also add an example on how to reproduce the error of the issue so we can test it out?

shubham-pampattiwar:
@mateusoliveira43 Slack thread confirming the change from the RFE reporter and tester: https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1705593931640569?thread_ts=1701439906.337739&cid=C0144ECKUJ0

shubham-pampattiwar:
> right @shubham-pampattiwar , thanks
> 
> just missing the documentation update in DPA field now

Done, updated the PR ! PTAL !

mateusoliveira43 (approved):

kaovilai (approved):

Issue #1241, opened 2023-11-27
Configure liveness, readiness and startup probes for sample applications


mateusoliveira43:
Remove arbitrary waits from E2E code, by configuring liveness, readiness and startup probes for sample applications.

Example of arbitrary wait:
https://github.com/openshift/oadp-operator/blob/78f6b7b3d26feeafd28b6dd0a164657435181168/tests/e2e/backup_restore_suite_test.go#L146-L148

Reference: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/

mateusoliveira43:
Closing this because the only 2 sleeps currently in code have message explaining why they are needed

PR #1270, opened 2024-01-16
Add livenessProbe and readinessProbe to the sample-applications

mpryc:
Add livenessProbe and startupProbe to the sample-applications, also slightly adjusts how the application startup process is considered done from the go code.
    
This change adds the liveness and startup probe to the mongo and mysql applications within the sample applications that are used during CI runs.
    
The livenessProbe checks if the relevant port is open within the container run, where the startupProbe checks if it's possible to actually connect to the DB endpoint from the container in which database is running.
    
It also checks for the correct condition of the running pod and not only for the state in case probe fails.
    
Increased timeout in the AreApplicationPodsRunning which is ran via Eventually to 19minuts which should be more then enough time for even complex scenarios.
    
Removed sleep as it should not be needed anymore as we are checking for the startup probe now.

### DIFF
--- a/tests/e2e/sample-applications/mongo-persistent/mongo-persistent-block.yaml
+++ b/tests/e2e/sample-applications/mongo-persistent/mongo-persistent-block.yaml
@@ -129,6 +129,25 @@ items:
               volumeDevices:
                 - name:  block-volume-pv
                   devicePath: /dev/xvdx
+              livenessProbe:
+                tcpSocket:
+                  port: mongo
+                initialDelaySeconds: 5
+                periodSeconds: 10
+              startupProbe:
+                exec:
+                  command:
+                  - mongosh
+                  - admin
+                  - -u $(MONGO_INITDB_ROOT_USERNAME)
+                  - -p $(MONGO_INITDB_ROOT_PASSWORD)
+                  - --eval 'printjson(db.getCollectionNames())'
+                initialDelaySeconds: 5
+                periodSeconds: 30
+                timeoutSeconds: 2
+                successThreshold: 1
+                failureThreshold: 40 # 40x30sec before restart pod
+              restartPolicy: Always
           volumes:
           - name: block-volume-pv
             persistentVolumeClaim:
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -64,7 +64,6 @@ type BackupRestoreCase struct {
 	BackupRestoreType            BackupRestoreType
 	PreBackupVerify              VerificationFunction
 	PostRestoreVerify            VerificationFunction
-	AppReadyDelay                time.Duration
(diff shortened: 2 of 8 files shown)


tests/e2e/sample-applications/mongo-persistent/mongo-persistent-block.yaml
mateusoliveira43, line 1: still reviewing

still understand probes, one first question is: startup probes can not be used to avoid init containers https://github.com/openshift/oadp-operator/pull/1272 ?

please also check this issue https://github.com/openshift/oadp-operator/issues/1241 and relate it to PR

mpryc: RE first question: Eventually they could be used to avoid init containers, however I would like to make sure we have one way of testing connection between application and DB engine. That is served well using `curl` and `nc` tools, which are not available within the sample app image.

Secondly init container starts before main application starts, so we know exactly when the problem happens and it's disconnected from the main application troubles.

Readiness probes are better suited when the DB can already accept traffic where using an init container allows  to have a separate container dedicated to the initialization tasks without affecting the readiness status of the main sample application.

RE: #1241 That's another PR to address it. As you requested I am splitting now into smaller chunks so don't want to mess with them. Current `initialDelaySeconds`, `periodSeconds` and `timeoutSeconds` are my estimates for DB starts and can be adjusted.

weshayutin (approved):

kaovilai (approved):

tests/e2e/sample-applications/mongo-persistent/mongo-persistent-block.yaml
```diff
@@ -129,6 +129,25 @@ items:
...
+                  - mongosh
+                  - admin
+                  - -u $(MONGO_INITDB_ROOT_USERNAME)
+                  - -p $(MONGO_INITDB_ROOT_PASSWORD)
+                  - --eval 'printjson(db.getCollectionNames())'
+                initialDelaySeconds: 5
+                periodSeconds: 30
+                timeoutSeconds: 2
+                successThreshold: 1
+                failureThreshold: 40 # 40x30sec before restart pod
```
mateusoliveira43, line 149: this big failure threshold is to avoid it to fail? just not setting it wouldn't do the trick?

mpryc: So we want the timeout set on Gingko for example [here](https://github.com/openshift/oadp-operator/blob/master/tests/e2e/backup_restore_suite_test.go#L136)  to fail before the pod timeout is up, so gingko reports it as failed. 20min is in the situation where we want to really wait 19min on the gingko side as but I don't have clear number of how much wait we should really wait.

mateusoliveira43:
as discussed, without delay, the tests will not pass

should we close this one, since the changes are already present in https://github.com/openshift/oadp-operator/pull/1277 ?

mateusoliveira43:
Closing as the other PR already have the same changes

PR #1271, opened 2024-01-16
Use full URI as the mongodb container image path.

mpryc:
Use of mongodb container path should use full URI which will limit chances of failures while trying to pull it from other sources.


mateusoliveira43 (approved):
Tested locally

weshayutin (approved):
/lgtm

kaovilai (approved):

PR #1272, opened 2024-01-16
Use more reliable way for checking if DB is alive from sample app.

mpryc:
Change which modifies the way sample application checks if DB engine is alive before continuing to start.

The more reliable way is to find out if the port is actually open and not only if the host is resolvable. For the MongoDB check is over standard HTTP request, while for the MySQL nc is being used as MySQL does not understand HTTP requests.


weshayutin (approved):

mpryc:
@weshayutin @kaovilai can this be merged?

PR #1232, opened 2023-11-17
fix: Remove 1.3 code from master branch

mateusoliveira43:
Now that 1.3 branch is created, remove code related to it form master branch.

### Related issues

Fixes https://github.com/openshift/oadp-operator/issues/1195
Fixes https://github.com/openshift/oadp-operator/issues/1192

PR #1275, opened 2024-01-20
fix: Add old datamover fields back to DPA

mateusoliveira43:
## Description

Revert removal of fields from DPA from https://github.com/openshift/oadp-operator/pull/1232, following the project deprecation process (only remove fields when API version updates). Still not allow users to use those fields.

## How to test

Create a DPA with vsm plugin or old datamover feature field. DPA should not reconcile and tell user to use Velero built-in datamover.


controllers/validator.go
```diff
@@ -78,6 +78,16 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 		}
 	}
 
+	// check for VSM/Volsync DataMover (OADP 1.2 or below) syntax
+	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
+		if plugin == oadpv1alpha1.DefaultPluginVSM {
+			return false, errors.New("Delete vsm from spec.configuration.velero.defaultPlugins and dataMover object from spec.features. Use Velero Built-in Data Mover instead")
+		}
+	}
```
kaovilai, line 86: Move this into [`func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {`](https://github.com/openshift/oadp-operator/blob/db116623f3e66ad9763514d562b688624cc44d00/controllers/validator.go#L121) ?

kaovilai: there's already a for loop there.

mateusoliveira43: going to check that out

mateusoliveira43:
/remove-lifecycle stale
Need this in 1.4, otherwise upgrades may break

controllers/validator.go
mateusoliveira43: done

kaovilai (approved):

PR #1278, opened 2024-01-23
Fix OADP-3044: DPA doesnt report missing region error on adding profile: default

shubham-pampattiwar:


controllers/bsl.go
```diff
@@ -284,7 +284,7 @@ func (r *DPAReconciler) validateAWSBackupStorageLocation(bslSpec velerov1.Backup
 		return fmt.Errorf("prefix for AWS backupstoragelocation object storage cannot be empty. It is required for backing up images")
 	}
 	// BSL region is required when s3ForcePathStyle is true AND BackupImages is true
-	if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0 && bslSpec.Config[S3ForcePathStyle] == "true") && dpa.BackupImages() {
+	if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0 || (len(bslSpec.Config[Region]) == 0 && bslSpec.Config[S3ForcePathStyle] == "true")) && dpa.BackupImages() {
```
kaovilai, line 287:
```
len(bslSpec.Config[Region]) == 0
 || 
(len(bslSpec.Config[Region]) == 0
```

wait what does this do?

kaovilai: ahh.... nvm I see..
```
(len(bslSpec.Config[Region]) == 0 && bslSpec.Config[S3ForcePathStyle] == "true")
```

kaovilai (approved):

controllers/bsl.go
mateusoliveira43: from the error message, I think this is wrong

example: if I set backup images to false, this if is never triggered, right?

to make sure, can you add one more test with `BackupImages` false but with `S3ForcePathStyle` true and no region (only one is this I think https://github.com/openshift/oadp-operator/blob/master/controllers/bsl_test.go#L1141)

shubham-pampattiwar: yes if `backupImages` is false, we will not get the error.

shubham-pampattiwar: updated validation and added more unit tests

controllers/bsl_test.go
```diff
@@ -1174,6 +1174,201 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 			want:    true,
 			wantErr: false,
 		},
+		{
+			name: "BSL without config section for aws provider and default backupImages is true behaviour",
```
mateusoliveira43, line 1178: NIT: behavior

controllers/bsl.go
mateusoliveira43: can we simplify the second if? something like (merging the 2 together)
```go
       // BSL region is required when s3ForcePathStyle is true or BackupImages is true
	if len(bslSpec.Config[Region]) == 0 {
		if bslSpec.Config[S3ForcePathStyle] == "true" {
			return fmt.Errorf("region for AWS backupstoragelocation cannot be empty when s3ForcePathStyle is true")
		}
		if dpa.BackupImages() {
			return fmt.Errorf("region for AWS backupstoragelocation cannot be empty when BackupImages is true")
		}
	}
```

shubham-pampattiwar: This would not cover the case when region is empty and `S3ForcePathStyle` is `true` alongwith `backupImages` is enabled.

controllers/bsl_test.go
shubham-pampattiwar: Done, updated (not a typo according to UK english though :) )

controllers/bsl.go
mateusoliveira43: The error message would not tell all info, but it would be covered by `s3ForcePathStyle=true` case, right?

shubham-pampattiwar: current PR covers the case

sseago (approved):

controllers/bsl.go
```diff
@@ -283,8 +283,9 @@ func (r *DPAReconciler) validateAWSBackupStorageLocation(bslSpec velerov1.Backup
 	if len(bslSpec.StorageType.ObjectStorage.Prefix) == 0 && dpa.BackupImages() {
 		return fmt.Errorf("prefix for AWS backupstoragelocation object storage cannot be empty. It is required for backing up images")
 	}
+
 	// BSL region is required when s3ForcePathStyle is true AND BackupImages is true
```
mateusoliveira43, line 287: change comment AND to or to match if behavior? Add TODO (or create a GitHub issue, no JIRA, just for tracking) about need to change when https://github.com/openshift/openshift-velero-plugin/pull/223 merges?

kaovilai (dismissed):

controllers/bsl.go
shubham-pampattiwar: Done

sseago (approved):

mateusoliveira43 (approved):

shubham-pampattiwar:
Is it ok to overrride this test ?

kaovilai:
Yeah azure's failing in multiple PRs..

PR #1280, opened 2024-01-24
OADP-2866: Add DPA support for default-volumes-to-fs-backup and snapshot-move-data

shubham-pampattiwar:


controllers/velero.go
```diff
@@ -447,6 +461,30 @@ func getFsBackupTimeout(dpa *oadpv1alpha1.DataProtectionApplication) string {
 	return defaultFsBackupTimeout
 }
 
+func getSnapshotMoveDataValue(dpa *oadpv1alpha1.DataProtectionApplication) string {
+	if dpa.Spec.Configuration.Velero != nil && boolptr.IsSetToTrue(dpa.Spec.Configuration.Velero.SnapshotMoveData) {
```
mateusoliveira43, line 465: In this 2 functions, could we simplify like this?
```go
if dpa.Spec.Configuration.Velero != nil {
    if boolptr.IsSetToTrue(dpa.Spec.Configuration.Velero.<field>) {
        return TrueVal
    }
    // since the field is typed as bool in CRD, there is no other option for the field
    return FalseVal
}
return ""
```

controllers/velero.go
```diff
@@ -357,6 +361,16 @@ func (r *DPAReconciler) customizeVeleroDeployment(dpa *oadpv1alpha1.DataProtecti
 		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--resource-timeout=%v", resourceTimeoutString))
 	}
 
+	// check for default-snapshot-move-data parameter
+	if len(getSnapshotMoveDataValue(dpa)) > 0 {
```
mateusoliveira43, line 365: create variable prior to the `if`s to avoid calling function twice?

controllers/velero.go
shubham-pampattiwar: The field is a bool pointer so in Go it can have three values - `true`, `false` and `nil` That's why I have done the explicit assignment.

mateusoliveira43: you are right, we need to check for `nil`

so, would suggest something like this
```go
if dpa.Spec.Configuration.Velero != nil && value := dpa.Spec.Configuration.Velero.<field>; value != nil {
    if boolptr.IsSetToTrue(value) {
        return TrueVal
    }
    return FalseVal
}
return ""
```

shubham-pampattiwar: `boolptr.IsSetToTrue(value)` already check for nil.

controllers/velero.go
shubham-pampattiwar: Done, updated

mateusoliveira43 (approved):

sseago (approved):

controllers/velero.go
```diff
@@ -357,6 +361,18 @@ func (r *DPAReconciler) customizeVeleroDeployment(dpa *oadpv1alpha1.DataProtecti
...
+	snapshotMoveData := getSnapshotMoveDataValue(dpa)
+	if len(snapshotMoveData) > 0 {
+		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--snapshot-move-data=%s", snapshotMoveData))
+	}
+
+	// check for default-volumes-to-fs-backup
+	defaultVolumesToFSBackup := getDefaultVolumesToFSBackup(dpa)
+	if len(defaultVolumesToFSBackup) > 0 {
+		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--default-volumes-to-fs-backup=%s", defaultVolumesToFSBackup))
+	}
```
kaovilai, line 374:
```diff
	installDeployment := install.Deployment(veleroDeployment.Namespace,
		install.WithResources(veleroResourceReqs),
		install.WithImage(getVeleroImage(dpa)),
		install.WithAnnotations(podAnnotations),
		install.WithFeatures(dpa.Spec.Configuration.Velero.FeatureFlags),
		install.WithUploaderType(uploaderType),
		// last label overrides previous ones
		install.WithLabels(veleroDeployment.Labels),
		// use WithSecret false even if we have secret because we use a different VolumeMounts and EnvVars
		// see: https://github.com/vmware-tanzu/velero/blob/ed5809b7fc22f3661eeef10bdcb63f0d74472b76/pkg/install/deployment.go#L223-L261
		// our secrets are appended to containers/volumeMounts in credentials.AppendPluginSpecificSpecs function
		install.WithSecret(false),
		install.WithServiceAccountName(common.Velero),
+		install.WithDefaultSnapshotMoveData(dpa.spec...),
+		install.WithDefaultVolumesToFsBackup(dpa.spec...),
	)
```

was gonna say we try to use something like above.. but seems like those don't accept bool.. so the only way to use podTemplateOption is to create an array then pass that in, requiring exporting [PodTemplateOption](https://github.com/vmware-tanzu/velero/blob/main/pkg/install/deployment.go#L32C6-L32C23) in velero. Not ideal.

Another would've been to add new podTemplateOption func called `func (xxx) WithDefaultSnapshotMoveDataBool(option bool) podTemplateOption`.. also not ideal.

So all things considered this is ok if we're rushing and/or moving away from using podTemplateOptions alltogether..

kaovilai (approved):

PR #1283, opened 2024-01-26
fix: Remove unused files from e2e folder

mateusoliveira43:
## Description

Remove files not used by end-to-end tests code.

## TODO

Check documentation, some of the files may be still referenced there.


weshayutin:
So I am probably to blame for a lot of these files. Happy to figure out a reorg / rewrite with in the context of symlinks to developer doc, and downstream bvt automation. Said a different way many of these files are not directly used in the e2e, however they are used elsewhere.

PR #1287, opened 2024-01-27
fix: Better description to wait for two vol test

mateusoliveira43:
Add better description for why it is needed to have a wait in two volumes test case.

### DIFF
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -63,10 +63,7 @@ type BackupRestoreCase struct {
 	BackupRestoreType lib.BackupRestoreType
 	PreBackupVerify   VerificationFunction
 	PostRestoreVerify VerificationFunction
-	ReadyDelay        time.Duration
-	SkipVerifyLogs    bool
-	SnapshotVolumes   bool
-	RestorePVs        bool
+	SkipVerifyLogs    bool // TODO remove
 }
 
 type ApplicationBackupRestoreCase struct {
@@ -189,12 +186,16 @@ func runBackup(brCase BackupRestoreCase, backupName string) bool {
 	nsRequiresResticDCWorkaround, err := lib.NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.Namespace)
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-	// TODO this should be a function, not an arbitrary sleep
-	log.Printf("Sleeping for %v to allow application to be ready for case %s", brCase.ReadyDelay, brCase.Name)
-	time.Sleep(brCase.ReadyDelay)
+	if strings.Contains(brCase.Name, "twovol") {
+		volumeSyncDelay := 30 * time.Second
+		log.Printf("Sleeping for %v to allow volume to be in sync with /tmp/log/ for case %s", volumeSyncDelay, brCase.Name)
+		// TODO this should be a function, not an arbitrary sleep
+		time.Sleep(volumeSyncDelay)
+	}
+
 	// create backup
 	log.Printf("Creating backup %s for case %s", backupName, brCase.Name)
-	backup, err := lib.CreateBackupForNamespaces(dpaCR.Client, namespace, backupName, []string{brCase.Namespace}, brCase.BackupRestoreType == lib.RESTIC || brCase.BackupRestoreType == lib.KOPIA, brCase.BackupRestoreType == lib.CSIDataMover, brCase.SnapshotVolumes)
+	backup, err := lib.CreateBackupForNamespaces(dpaCR.Client, namespace, backupName, []string{brCase.Namespace}, brCase.BackupRestoreType == lib.RESTIC || brCase.BackupRestoreType == lib.KOPIA, brCase.BackupRestoreType == lib.CSIDataMover)
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
 	// wait for backup to not be running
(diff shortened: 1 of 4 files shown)


tests/e2e/backup_restore_suite_test.go
```diff
@@ -143,9 +142,12 @@ func runBackupAndRestore(brCase BackupRestoreCase, expectedErr error, updateLast
 	nsRequiresResticDCWorkaround, err := NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.ApplicationNamespace)
 	Expect(err).ToNot(HaveOccurred())
 
-	// TODO this should be a function, not an arbitrary sleep
-	log.Printf("Sleeping for %v to allow application to be ready for case %s", brCase.AppReadyDelay, brCase.Name)
-	time.Sleep(brCase.AppReadyDelay)
+	if strings.Contains(brCase.Name, "twovol") {
```
mpryc, line 145: I like this improvement of removing it from function, but the proper solution would be to something less "around x seconds", how about putting `readinessProbe` in the `tests/e2e/sample-applications/mysql-persistent/mysql-persistent-twovol-csi.yaml` ? 

```
            volumeMounts:
            - name: applog
              mountPath: /tmp/log/todoapp/
            readinessProbe:
              exec:
                command:
                - sh
                - -c
                - '[ -f /tmp/log/todoapp/app.log ]'
              initialDelaySeconds: 1    # Delay before the first probe
              periodSeconds: 5          # Probe interval
              timeoutSeconds: 1         # Maximum time for the probe to complete, should be immediate as it's just file check
              successThreshold: 1       # One time is enough to check for file 
              failureThreshold: 30      # Number of consecutive failed probes to be considered failed   
```

mateusoliveira43: That would check, until it succeeds, if the file exists, right?

I think that would not be enough, the problem we face is not that the file does not exist, it is that volume does not have all logs yet

example

first log is saying application started (this one creates files, if not present, I assume)
second log is saying 2 things were added to database (this one, without hardcoded timeout, is not present in backup)

mpryc: So we are checking for the actual item to be in database as well? That could be also part of the check here, just slightly different logic. I am thinking here to solve the issue once and for all to not come back to this problem and actually remove the logic from the go code as that should not be mixed too much. Once the container is ready (whatever we do to make it ready via init-container or probe) then the test should continue. The test itself should not have too much knowledge about application itself.

mateusoliveira43: would not that create a cyclic dependency? 
1. we need to check if application is running (also check probes)
2. then we interact with application (creating TO DO items)
3. then we backup

are you not asking to check 2 inside 1, which is impossible?

weshayutin (dismissed):

tests/e2e/backup_restore_suite_test.go
kaovilai: 2 is not inside 1. 1. only indicates app is awake and ready to accept commands.

kaovilai (dismissed):

tests/e2e/lib/backup.go
```diff
@@ -25,9 +24,6 @@ func CreateBackupForNamespaces(ocClient client.Client, veleroNamespace, backupNa
 			SnapshotMoveData:         &snapshotMoveData,
 		},
 	}
-	if snapshotVolumes {
```
mateusoliveira43, line 28: @mrnold I reverted some changes from your PR that I think were not necessary (this option for example, is enabled by default)

Checking if the kubevirt test pass is enough to assure I did not break anything there?

mrnold: Yes, that's fine. I wasn't 100% sure whether or not this was needed.

mateusoliveira43:
13 azure, and 14 azure and aws had subscription test error
/retest

mateusoliveira43:
/unhold
rebased in top of master (with fixed CI)

mpryc (approved):

tests/e2e/lib/backup.go
```diff
@@ -9,11 +9,10 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/utils/pointer"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
-func CreateBackupForNamespaces(ocClient client.Client, veleroNamespace, backupName string, namespaces []string, defaultVolumesToFsBackup bool, snapshotMoveData, snapshotVolumes bool) (velero.Backup, error) {
+func CreateBackupForNamespaces(ocClient client.Client, veleroNamespace, backupName string, namespaces []string, defaultVolumesToFsBackup bool, snapshotMoveData bool) (velero.Backup, error) {
```
mpryc, line 15: Maybe simplify:

```
func CreateBackupForNamespaces(ocClient client.Client, veleroNamespace, backupName string, namespaces []string, defaultVolumesToFsBackup, snapshotMoveData bool) (velero.Backup, error) {}
```

tests/e2e/backup_restore_suite_test.go
```diff
@@ -189,12 +186,16 @@ func runBackup(brCase BackupRestoreCase, backupName string) bool {
 	nsRequiresResticDCWorkaround, err := lib.NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.Namespace)
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
 
-	// TODO this should be a function, not an arbitrary sleep
-	log.Printf("Sleeping for %v to allow application to be ready for case %s", brCase.ReadyDelay, brCase.Name)
-	time.Sleep(brCase.ReadyDelay)
+	if strings.Contains(brCase.Name, "twovol") {
+		volumeSyncDelay := 30 * time.Second
```
mpryc, line 190: The 30 seconds is a guess ?

tests/e2e/lib/backup.go
mateusoliveira43: I hope this will be catch by future linter tasks

tests/e2e/backup_restore_suite_test.go
mateusoliveira43: just used what was being used already

tests/e2e/lib/backup.go
mateusoliveira43: will update in linter task, not in this PR

mpryc (approved):

mateusoliveira43:
restore succeeds but application seems to unavailable after flake

PR #1289, opened 2024-01-27
OADP-3326: Add DPA support for disable-informer-cache velero config flag

shubham-pampattiwar:


api/v1alpha1/oadp_types.go
```diff
@@ -97,6 +97,9 @@ type VeleroConfig struct {
 	// Specify whether CSI snapshot data should be moved to backup storage by default
 	// +optional
 	SnapshotMoveData *bool `json:"snapshotMoveData,omitempty"`
+	// Disable informer cache for Get calls on restore. With this enabled, it will speed up restore in cases where there are backup resources which already exist in the cluster, but for very large clusters this will increase velero memory usage. Default is false.
```
mateusoliveira43, line 100: flag name is disable but explanation says if enabled. would change "enabled" to "true" or "false" (if description in velero also says disable and enable, then better to let equal for consistency)

api/v1alpha1/oadp_types.go
```diff
@@ -306,6 +309,14 @@ func (dpa *DataProtectionApplication) BackupImages() bool {
 	return dpa.Spec.BackupImages == nil || *dpa.Spec.BackupImages
 }
 
+// Default DisableInformerCache behavior when nil to false
+func (dpa *DataProtectionApplication) DisableInformerCache() bool {
```
mateusoliveira43, line 313: would change function name to something like `GetDisableInformerCache`. Otherwise, seems this functions disables it instead of getting the bool value from it

controllers/velero_test.go
```diff
@@ -1264,6 +1273,552 @@ func TestDPAReconciler_buildVeleroDeployment(t *testing.T) {
...
+									Args: []string{
+										"server",
+										defaultFileSystemBackupTimeout,
+										defaultRestoreResourcePriorities,
+										"--log-level",
+										logrus.InfoLevel.String(),
+										"--item-operation-sync-frequency=5m",
+										"--default-item-operation-timeout=2h",
+										"--default-volumes-to-fs-backup=false",
+										"--disable-informer-cache=false",
```
mateusoliveira43, line 1821: some places do not use variable `defaultDisableInformerCache`

api/v1alpha1/oadp_types.go
shubham-pampattiwar: yeah grabbed the description from upstream velero: https://github.com/vmware-tanzu/velero/blob/1034d6aee0b32b6ec0dbe4203902cd1d465ec950/pkg/cmd/server/server.go#L239

api/v1alpha1/oadp_types.go
shubham-pampattiwar: Done, updated

controllers/velero_test.go
shubham-pampattiwar: Done, updated

kaovilai (approved):

mateusoliveira43 (approved):

PR #1296, opened 2024-01-31
OADP-3379: fix wrong descriptions on time fields

mateusoliveira43:
## Description

All velero args time fields have wrong description, saying they are string, but actually are integer.

A better solution would be to change fields from integer nanoseconds to time syntax strings, but this is dangerous, adding possibility of breaking updates to users. Changing the unit of measure is also dangerous, as can change a value from minutes to years, when user updates OADP.

## How to test

Check that description of the fields are correct now, by creating a DPA like this
```yaml
metadata:
  name: test-time-field
  namespace: openshift-adp
spec:
  configuration:
    velero:
      args:
        backup-sync-period: 1
        default-backup-ttl: 1
        default-item-operation-timeout: 1
        default-repo-maintain-frequency: 1
        fs-backup-timeout: 1
        garbage-collection-frequency: 1
        item-operation-sync-frequency: 1
        resource-timeout: 1
        store-validation-frequency: 1
        terminating-resource-timeout: 1
      defaultPlugins:
      - openshift
      - aws
  backupLocations:
    - velero:
        provider: aws
        config:
            region: #
            profile: default
            s3ForcePathStyle: 'true'
            s3Url: #
        default: true
        credential:
            name: cloud-credentials
            key: cloud
        objectStorage:
            bucket: #
            prefix: velero
```
and check Velero deployment pod args.

### DIFF
--- a/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
+++ b/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
@@ -493,7 +493,7 @@ spec:
                               description: log to standard error as well as files (no effect when -logtostderr=true)
                               type: boolean
                             backup-sync-period:
-                              description: How often (in minutes) to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
+                              description: How often (in nanoseconds) to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
                               format: int64
                               type: integer
                             client-burst:
@@ -509,15 +509,15 @@ spec:
                               description: Show colored output in TTY
                               type: boolean
                             default-backup-ttl:
-                              description: ' default 720h0m0s'
+                              description: How long (in nanoseconds) to wait by default before backups can be garbage collected. (default is 720 hours)
                               format: int64
                               type: integer
                             default-item-operation-timeout:
-                              description: ' How long to wait on asynchronous BackupItemActions and RestoreItemActions to complete before timing out. (default 1h0m0s)'
+                              description: How long (in nanoseconds) to wait on asynchronous BackupItemActions and RestoreItemActions to complete before timing out. (default is 1 hour)
                               format: int64
                               type: integer
                             default-repo-maintain-frequency:
-                              description: ' How often ''maintain'' is run for backup repositories by default.'
+                              description: How often (in nanoseconds) 'maintain' is run for backup repositories by default.
                               format: int64
                               type: integer
                             default-volumes-to-fs-backup:
@@ -542,15 +542,15 @@ spec:
                                 type: string
                               type: array
                             fs-backup-timeout:
-                              description: ' How long pod volume file system backups/restores should be allowed to run before timing out. (default 4h0m0s)'
+                              description: How long (in nanoseconds) pod volume file system backups/restores should be allowed to run before timing out. (default is 4 hours)
                               format: int64
                               type: integer
                             garbage-collection-frequency:
-                              description: ' How long to wait by default before backups can be garbage collected. (default 720h0m0s)'
+                              description: How long (in nanoseconds) to wait by default before backups can be garbage collected. (default is 720 hours)
                               format: int64
                               type: integer
                             item-operation-sync-frequency:
-                              description: ' How often to check status on backup/restore operations after backup/restore processing.'
+                              description: How often (in nanoseconds) to check status on backup/restore operations after backup/restore processing.
                               format: int64
                               type: integer
                             log-format:
@@ -589,7 +589,7 @@ spec:
                               description: The address to expose the pprof profiler.
                               type: string
                             resource-timeout:
-                              description: ' How long to wait for resource processes which are not covered by other specific timeout parameters. Default is 10 minutes. (default 10m0s)'
+                              description: How long (in nanoseconds) to wait for resource processes which are not covered by other specific timeout parameters. (default is 10 minutes)
                               format: int64
                               type: integer
                             restore-resource-priorities:
@@ -605,11 +605,11 @@ spec:
                               description: logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=false) (default 2)
                               type: integer
                             store-validation-frequency:
-                              description: ' How often to verify if the storage is valid. Optional. Set this to `0s` to disable sync. Default 1 minute.'
+                              description: How often (in nanoseconds) to verify if the storage is valid. Optional. Set this to `0` to disable sync. (default is 1 minute)
                               format: int64
                               type: integer
                             terminating-resource-timeout:
-                              description: ' How long to wait on persistent volumes and namespaces to terminate during a restore before timing out.'
+                              description: How long (in nanoseconds) to wait on persistent volumes and namespaces to terminate during a restore before timing out.
                               format: int64
                               type: integer
                             v:
--- a/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
+++ b/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
@@ -493,7 +493,7 @@ spec:
                               description: log to standard error as well as files (no effect when -logtostderr=true)
                               type: boolean
                             backup-sync-period:
-                              description: How often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
+                              description: How often (in minutes) to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
                               format: int64
                               type: integer
                             client-burst:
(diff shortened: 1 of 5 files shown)


pkg/velero/server/args.go
```diff
@@ -47,7 +48,7 @@ func (a Args) StringArr(dpaFeatureFlags []string, logLevel string) ([]string, er
 		args = append(args, fmt.Sprintf("--log-level=%s", logrusLevel.String()))
 	}
 	if a.BackupSyncPeriod != nil {
-		args = append(args, fmt.Sprintf("--backup-sync-period=%s", a.BackupSyncPeriod.String())) // duration
+		args = append(args, fmt.Sprintf("--backup-sync-period=%s", (time.Duration(*a.BackupSyncPeriod)*time.Minute).String())) // duration
```
mateusoliveira43, line 51: This changes the field behavior, but I do not know if anyone ever used them. Without this, to pass a minute, user would input `60000000000` right? Prior to making this changing to all fields, I want the approval of the team

pkg/velero/server/config.go
```diff
@@ -23,19 +23,19 @@ type ServerConfig struct {
 	// defaultBackupLocation will be defined outside of server config in DataProtectionApplication
 	// defaultBackupLocation                        string
 
-	// How often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
+	// How often (in minutes) to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location.
 	// +optional
-	BackupSyncPeriod *time.Duration `json:"backup-sync-period,omitempty"`
-	// How long pod volume file system backups/restores should be allowed to run before timing out. (default 4h0m0s)
+	BackupSyncPeriod *int64 `json:"backup-sync-period,omitempty"`
+	//  How long pod volume file system backups/restores should be allowed to run before timing out. (default 4h0m0s)
```
shubham-pampattiwar, line 29: I would remove this extra space added in the description strins right at the start, I believe this is causing the insert of single quotes in CRD descriptions. Was this a desirable change ?

mateusoliveira43: this is for me to remember, all those field have wrong description, and I need to fix them. BUT prior to that, I want us to accept a solution

pkg/velero/server/args.go
mateusoliveira43: Checked and all these time fields are in nano seconds

```
args:
  - server
  - '--backup-sync-period=1ns'
  - '--default-backup-ttl=1ns'
  - '--default-item-operation-timeout=1ns'
  - '--resource-timeout=1ns'
  - '--default-repo-maintain-frequency=1ns'
  - '--garbage-collection-frequency=1ns'
  - '--fs-backup-timeout=1ns'
  - '--item-operation-sync-frequency=1ns'
  - >-
    --restore-resource-priorities=securitycontextconstraints,customresourcedefinitions,namespaces,managedcluster.cluster.open-cluster-management.io,managedcluster.clusterview.open-cluster-management.io,klusterletaddonconfig.agent.open-cluster-management.io,managedclusteraddon.addon.open-cluster-management.io,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io
  - '--store-validation-frequency=1ns'
  - '--terminating-resource-timeout=1ns'
```  

My question is: should I update the unit of measure of this fields?

kaovilai: > Without this, to pass a minute, user would input 60000000000 right?

No. User would put into their yaml `4h` for example.
```
	// How long pod volume file system backups/restores should be allowed to run before timing out. (default 4h0m0s)
	// +optional
	PodVolumeOperationTimeout *time.Duration `json:"fs-backup-timeout,omitempty"`
```

Per the example.

Somewhere there is a text during parsing from string into time.Duration for you.

kaovilai: The unmarshal function https://pkg.go.dev/time#Time.UnmarshalJSON for time package will be used during yaml unmarshalling via https://pkg.go.dev/sigs.k8s.io/yaml

> In short, this library first converts YAML to JSON using go-yaml and then uses json.Marshal and json.Unmarshal to convert to or from the struct. This means that it effectively reuses the JSON struct tags as well as the custom JSON methods MarshalJSON and UnmarshalJSON unlike go-yaml. For a detailed overview of the rationale behind this method, [see this blog post](http://web.archive.org/web/20190603050330/http://ghodss.com/2014/the-right-way-to-handle-yaml-in-golang/).

I would not make this change.

kaovilai: time.Duration.UnmarshalJSON function is used and can accept string in the yaml.
> UnmarshalJSON implements the json.Unmarshaler interface. The time must be a quoted string in the [RFC 3339](https://rfc-editor.org/rfc/rfc3339.html) format.

mateusoliveira43: Screenshot from 2024-02-01 14-57-02

Its not possible (1.3 example), I can not even create the DPA using hour syntax

kaovilai: Ok I can't find the specific place where this is implemented.. but point is it can decode duration.

Here's the unit test https://github.com/kubernetes-sigs/yaml/blob/c3772b51db126345efe2dfe4ff8dac83b8141684/goyaml.v3/decode_test.go#L926-L931
```
func (s *S) TestUnmarshalDurationInt(c *C) {
	// Don't accept plain ints as durations as it's unclear (issue #200).
	var d time.Duration
	err := yaml.Unmarshal([]byte("123"), &d)
	c.Assert(err, ErrorMatches, "(?s).* line 1: cannot unmarshal !!int `123` into time.Duration")
}
```

mateusoliveira43: How would we change this validation without changing the CRD?

Here an example which accepts time syntax https://github.com/openshift/oadp-operator/blob/master/api/v1alpha1/oadp_types.go#L202 https://github.com/openshift/oadp-operator/blob/master/bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml#L46-L49 but to work like this, CRD change would be necessary, no?

kaovilai: If it really was broken and no one could use it then CRD change would be ok if we can assume no one used it.

mateusoliveira43: I think so, I do not see people passing nano seconds to these fields

I believe there are 2 options:

- just update CRD descriptions telling the unit of measure (and then, without changing CRD, we can change the unit of measure from nano seconds, to minutes, for example)
- (CHANGES CRD) fix field to accept time syntax string (better, but more dangerous)

kaovilai: Ok I found the issue.. have to convert all the time.Duration to metav1.Duration. Then all is good.

mateusoliveira43: that works, but would change the CRD (type integer to string) (so follows to second option)

kaovilai: PR: https://github.com/openshift/oadp-operator/pull/1306

mateusoliveira43: resolving as we have agreed on just fix fields descriptions, without changing their unit of measure

bundle/manifests/oadp.openshift.io_dataprotectionapplications.yaml
```diff
@@ -542,15 +542,15 @@ spec:
                                 type: string
                               type: array
                             fs-backup-timeout:
```
mateusoliveira43, line 544: @weshayutin wanted to add info that this field can be specified in Backup as well, but the fields I found taht can be added to backup are TTL and item-operation-timeout. Ref: `go run cmd/velero/velero.go backup create --help | grep duration`

weshayutin:
seems to work :)
note: ItemOperationTimeout:  1ns

```
whayutin@thinkdoe:~/OADP/SETUP/OADP_1.3$ remote_velero backup describe mysql-persistent --details
Name:         mysql-persistent
Namespace:    openshift-adp
Labels:       velero.io/storage-location=dpa-sample-1
Annotations:  velero.io/resource-timeout=1ns
              velero.io/source-cluster-k8s-gitversion=v1.27.8+4fab27b
              velero.io/source-cluster-k8s-major-version=1
              velero.io/source-cluster-k8s-minor-version=27

Phase:  Failed (run `velero backup logs mysql-persistent` for more information)


Namespaces:
  Included:  mysql-persistent
  Excluded:  <none>

Resources:
  Included:        *
  Excluded:        <none>
  Cluster-scoped:  auto

Label selector:  <none>

Or label selector:  <none>

Storage Location:  dpa-sample-1

Velero-Native Snapshot PVs:  auto
Snapshot Move Data:          false
Data Mover:                  velero

TTL:  720h0m0s

CSISnapshotTimeout:    10m0s
ItemOperationTimeout:  1ns

Hooks:  <none>

```
```
metadata:
  creationTimestamp: "2024-02-08T21:11:11Z"
  generation: 3
  name: dpa-sample
  namespace: openshift-adp
  resourceVersion: "14383360"
  uid: c3c408d3-ec11-4285-bc71-f5f2c81c39af
spec:
  backupLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      credential:
        key: cloud
        name: cloud-credentials
      default: true
      objectStorage:
        bucket: cvpbucketuswest2
        prefix: velero
      provider: aws
  configuration:
    nodeAgent:
      enable: true
      uploaderType: kopia
    velero:
      args:
        backup-sync-period: 1
        default-backup-ttl: 1
        default-item-operation-timeout: 1
        default-repo-maintain-frequency: 1
        fs-backup-timeout: 1
        garbage-collection-frequency: 1
        item-operation-sync-frequency: 1
        resource-timeout: 1
        store-validation-frequency: 1
        terminating-resource-timeout: 1
      defaultPlugins:
      - openshift
      - aws
      - csi
      featureFlags:
      - EnableCSI
  snapshotLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      provider: aws
status:
  conditions:
  - lastTransitionTime: "2024-02-08T21:11:11Z"
    message: Reconcile complete
    reason: Complete
    status: "True"
    type: Reconciled
```

time="2024-02-08T21:18:18Z" level=error msg="backup failed" backuprequest=openshift-adp/mysql-persistent controller=backup error="timed out waiting for caches to sync" logSource="pkg/controller/backup_controller.go:292"

```

weshayutin (approved):

PR #1300, opened 2024-01-31
Attempt to make common parts of the test app common.

mpryc:


mpryc:
@mateusoliveira43 please let me know what do you think about such change and if that's good, then we can have similar with mysql, otherwise abandon as it's just a simple quick rework without any additional functionality, but thought it's worth having one copy of common snippets.

tests/e2e/sample-applications/mongo-persistent/mongo-common-steps.yaml
mateusoliveira43, line 1: I do not think this is the best approach. These things being YAML instead of in code are for us deploy them in cluster. Today we need to deploy application and PVC. With this change, we would need to deploy application, common and PVC. What we should do in my opinion to avoid duplication and be easy to be deployed with `oc` command outside of tests, is to create a template (with variables that change field for backup type and PVC provider)

mpryc: YAML itself does not have native support for conditional statements controlled via variables. That's where Helm or Kustomize may come to play, but I don't think it's worth adding another dependency to process it.

Things such as CONTAINER_IMAGE or some parts of YAML could be moved to variables, but this will not reduce duplications which we currently have, because the objects will be in exactly the same structure, just with different parameters. Deployment types which we are changing are changing the structure of the YAML file.

I am fine keeping it in one place, just thought it's better to remove duplication and keep common parts in one place, the drawback is that we would need to run `oc apply -f` 3 times instead of 2x.

mateusoliveira43: this is inside OpenShift https://docs.openshift.com/container-platform/4.14/openshift_images/using-templates.html no new dependencies

mpryc:
I am closing this PR as the approach here isn't best and there isn't really desire to fix it this way.

PR #1294, opened 2024-01-30
fix: Refactor no default BSL code validation

mateusoliveira43:
## Description

Refactor validation of no default BSL and also check error messages of validations.

## How to test

Run `make test`.

PR #1302, opened 2024-02-01
BSL/VSL validation code cleanup

### LINKED ISSUE #1129: Clean up duplicate logic for validating backupstoragelocations and volumesnapshotlocations in dpa
https://github.com/kaovilai/oadp-operator/blob/27cccabbd2e845da21ef49489e7506f10c30696d/controllers/validator.go#L19

stillalearner:
Fixes #1129

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -15,22 +15,23 @@ import (
 )
 
 func (r *DPAReconciler) ValidateBackupStorageLocations(dpa oadpv1alpha1.DataProtectionApplication) (bool, error) {
-	if dpa.Spec.Configuration == nil || dpa.Spec.Configuration.Velero == nil {
-		return false, errors.New("DPA CR Velero configuration cannot be nil")
-	}
-	// Ensure we have a BSL or user has specified NoDefaultBackupLocation install
-	if len(dpa.Spec.BackupLocations) == 0 && !dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
-		return false, errors.New("no backupstoragelocations configured, ensure a backupstoragelocation has been configured")
-	}
-
-	if err := r.ensureBackupLocationHasVeleroOrCloudStorage(&dpa); err != nil {
-		return false, err
-	}
-
 	// Ensure BSL is a valid configuration
 	// First, check for provider and then call functions based on the cloud provider for each backupstoragelocation configured
 	numDefaultLocations := 0
 	for _, bslSpec := range dpa.Spec.BackupLocations {
+
+		if err := r.ensureBackupLocationHasVeleroOrCloudStorage(&bslSpec); err != nil {
+			return false, err
+		}
+
+		if err := r.ensurePrefixWhenBackupImages(&dpa, &bslSpec); err != nil {
+			return false, err
+		}
+
+		if err := r.ensureSecretDataExists(&dpa, &bslSpec); err != nil {
+			return false, err
+		}
+
 		if bslSpec.Velero != nil {
 			if bslSpec.Velero.Default {
 				numDefaultLocations++
@@ -377,14 +378,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
 	return nil
 }
 
-func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(dpa *oadpv1alpha1.DataProtectionApplication) error {
-	for _, bsl := range dpa.Spec.BackupLocations {
-		if bsl.CloudStorage == nil && bsl.Velero == nil {
-			return fmt.Errorf("no bucket CloudStorage or velero BackupStorageLocation provided for backupLocations")
-		}
+func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(bsl *oadpv1alpha1.BackupLocation) error {
+	if bsl.CloudStorage == nil && bsl.Velero == nil {
+		return fmt.Errorf("BackupLocation must have velero or bucket configuration")
+	}
 
-		if bsl.CloudStorage != nil && bsl.Velero != nil {
-			return fmt.Errorf("cannot have both backupstoragelocations and bucket provided for a single StorageLocation")
+	if bsl.CloudStorage != nil && bsl.Velero != nil {
+		return fmt.Errorf("cannot have both backupstoragelocations and bucket provided for a single StorageLocation")
+	}
+	return nil
+}
+
+func (r *DPAReconciler) ensurePrefixWhenBackupImages(dpa *oadpv1alpha1.DataProtectionApplication, bsl *oadpv1alpha1.BackupLocation) error {
+
+	if bsl.Velero != nil && bsl.Velero.ObjectStorage != nil && bsl.Velero.ObjectStorage.Prefix == "" && dpa.BackupImages() {
+		return fmt.Errorf("BackupLocation must have velero prefix when backupImages is not set to false")
+	}
+
+	if bsl.CloudStorage != nil && bsl.CloudStorage.Prefix == "" && dpa.BackupImages() {
+		return fmt.Errorf("BackupLocation must have cloud storage prefix when backupImages is not set to false")
+	}
+
+	return nil
+}
+
+func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, bsl *oadpv1alpha1.BackupLocation) error {
+	// Check if the Velero feature flag 'no-secret' is not set
+	if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret")) {
+		// Check if the user specified credential under velero
+		if bsl.Velero != nil && bsl.Velero.Credential != nil {
+			// Check if user specified empty credential key
+			if bsl.Velero.Credential.Key == "" {
+				return fmt.Errorf("Secret key specified in BackupLocation %s cannot be empty", bsl.Name)
+			}
+			// Check if user specified empty credential name
+			if bsl.Velero.Credential.Name == "" {
+				return fmt.Errorf("Secret name specified in BackupLocation %s cannot be empty", bsl.Name)
+			}
+		}
+		// Check if the BSL secret key configured in the DPA exists with a secret data
+		secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
+		bslSecret, err := r.getProviderSecret(secretName)
+		if err != nil {
+			return err
+		}
+		data, foundKey := bslSecret.Data[secretKey]
+		if !foundKey || len(data) == 0 {
+			return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
 		}
 	}
 	return nil
--- a/controllers/bsl_test.go
+++ b/controllers/bsl_test.go
@@ -61,7 +61,12 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 					Name:      "foo",
 					Namespace: "test-ns",
 				},
-				Spec: oadpv1alpha1.DataProtectionApplicationSpec{},
+				Spec: oadpv1alpha1.DataProtectionApplicationSpec{
+					Configuration: &oadpv1alpha1.ApplicationConfig{
+						Velero: &oadpv1alpha1.VeleroConfig{},
+					},
+					BackupLocations: []oadpv1alpha1.BackupLocation{},
+				},
 			},
 			want:    false,
 			wantErr: true,
@@ -189,6 +194,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 					Name:      "cloud-credentials",
 					Namespace: "test-ns",
 				},
+				Data: map[string][]byte{"cloud": []byte("dummy_data")},
 			},
 		},
 		{
@@ -376,6 +382,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 									LocalObjectReference: corev1.LocalObjectReference{
 										Name: "cloud-credentials",
 									},
+									Key: "cloud",
 								},
 							},
 						},
@@ -389,6 +396,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 					Name:      "cloud-credentials",
 					Namespace: "test-ns",
 				},
+				Data: map[string][]byte{"cloud": []byte("dummy_data")},
 			},
 		},
 		{
@@ -419,6 +427,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 									LocalObjectReference: corev1.LocalObjectReference{
 										Name: "cloud-credentials",
 									},
+									Key: "cloud",
 								},
 								Default: true,
 							},
@@ -433,6 +442,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 					Name:      "cloud-credentials",
 					Namespace: "test-ns",
 				},
+				Data: map[string][]byte{"cloud": []byte("dummy_data")},
 			},
 		},
 		{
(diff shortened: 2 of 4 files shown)


controllers/bsl.go
```diff
@@ -15,9 +15,6 @@ import (
 )
 
 func (r *DPAReconciler) ValidateBackupStorageLocations(dpa oadpv1alpha1.DataProtectionApplication) (bool, error) {
-	if dpa.Spec.Configuration == nil || dpa.Spec.Configuration.Velero == nil {
-		return false, errors.New("DPA CR Velero configuration cannot be nil")
-	}
```
kaovilai, line 20: Note to self: This is already in validator.go

controllers/vsl.go
```diff
@@ -105,13 +105,10 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
 }
 
 func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataProtectionApplication) (bool, error) {
-	if dpa.Spec.Configuration == nil {
-		return false, errors.New("application configuration not found")
-	}
-	if dpa.Spec.Configuration.Velero == nil {
-		return false, errors.New("velero configuration not found")
-	}
```
kaovilai, line 113: Note to self: This is already in validator.go

kaovilai:
oh and unit tests failing, can assist with that if needed.

controllers/bsl.go
```diff
@@ -27,6 +24,14 @@ func (r *DPAReconciler) ValidateBackupStorageLocations(dpa oadpv1alpha1.DataProt
 		return false, err
 	}
 
+	if err := r.ensurePrefixWhenBackupImages(&dpa); err != nil {
+		return false, err
+	}
+
+	if err:= r.ensureSecretDataExists(&dpa); err!=nil {
+		return false, err
+	}
```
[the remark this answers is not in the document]
stillalearner: done

controllers/bsl_test.go
```diff
@@ -1095,6 +1122,7 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 					Name:      "cloud-credentials",
 					Namespace: "test-ns",
 				},
+				Data: map[string][]byte{"cloud": []byte("dummy_data"), "testing": []byte("dummy_data2")},
```
kaovilai, line 1125: What's the `testing` key here for?

stillalearner: haha, can be removed, took this yaml from validator_test file. let me make it "cloud"

stillalearner: fixed.

kaovilai: ahh sorry. didn't notice the keys were specified in cloudstorage

kaovilai: before was probably better if there's a check for `dummy_data2` being used correctly..

kaovilai: but since there's not, up to you.

kaovilai (approved):

kaovilai (approved):

mateusoliveira43:
can we hold this one? I think after this ones merges https://github.com/openshift/oadp-operator/pull/1294, this will need rebase (or vice versa)

controllers/bsl.go
```diff
@@ -377,14 +384,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
...
 }
 
-func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(dpa *oadpv1alpha1.DataProtectionApplication) error {
-	for _, bsl := range dpa.Spec.BackupLocations {
-		if bsl.CloudStorage == nil && bsl.Velero == nil {
-			return fmt.Errorf("no bucket CloudStorage or velero BackupStorageLocation provided for backupLocations")
-		}
+func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(bsl *oadpv1alpha1.BackupLocation) error {
+	if bsl.CloudStorage == nil && bsl.Velero == nil {
+		return fmt.Errorf("no bucket CloudStorage or velero BackupStorageLocation provided for backupLocations")
```
mateusoliveira43, line 389: For consistency, I would change message here to **"BackupLocation must have velero or bucket configuration"**

Since this check was in validator.go, it would run prior to this. So, I think this message was never presented to the user

controllers/bsl.go
```diff
@@ -15,16 +15,23 @@ import (
...
-		return false, errors.New("DPA CR Velero configuration cannot be nil")
-	}
 	// Ensure we have a BSL or user has specified NoDefaultBackupLocation install
 	if len(dpa.Spec.BackupLocations) == 0 && !dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
 		return false, errors.New("no backupstoragelocations configured, ensure a backupstoragelocation has been configured")
 	}
 
-	if err := r.ensureBackupLocationHasVeleroOrCloudStorage(&dpa); err != nil {
-		return false, err
+	for _, bsl := range dpa.Spec.BackupLocations {
```
mateusoliveira43, line 23: can this for be merged with the following?

something like
```go
       // Ensure BSL is a valid configuration
	// First, check for provider and then call functions based on the cloud provider for each backupstoragelocation configured
	numDefaultLocations := 0
	for _, bslSpec := range dpa.Spec.BackupLocations {
           	if err := r.ensureBackupLocationHasVeleroOrCloudStorage(&bsl); err != nil {
			return false, err
		}

		if err := r.ensurePrefixWhenBackupImages(&dpa, &bsl); err != nil {
			return false, err
		}

		if err := r.ensureSecretDataExists(&dpa, &bsl); err != nil {
			return false, err
		}
             if bslSpec.Velero != nil {
                ....
```

controllers/bsl.go
```diff
@@ -377,14 +384,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
...
+	if bsl.CloudStorage != nil && bsl.CloudStorage.Prefix == "" && dpa.BackupImages() {
+		return fmt.Errorf("BackupLocation must have cloud storage prefix when backupImages is not set to false")
+	}
+
+	return nil
+}
+
+func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, bsl *oadpv1alpha1.BackupLocation) error {
+	// Check the Velero flags 'no-secret' or 'no-default-backup-location' are not set
+	if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret") || dpa.Spec.Configuration.Velero.NoDefaultBackupLocation) {
```
mateusoliveira43, line 413: after rebase on top of https://github.com/openshift/oadp-operator/pull/1294, please change lines 412 and 413 to 
```go
// Check if the Velero feature flag 'no-secret' is not set
if !dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret") {
```

controllers/bsl_test.go
```diff
@@ -1666,9 +1697,183 @@ func TestDPAReconciler_ensureBackupLocationHasVeleroOrCloudStorage(t *testing.T)
...
+
+		})
+	}
+}
+
+func TestDPAReconciler_ensurePrefixWhenBackupImages(t *testing.T) {
+	tests := []struct {
+		name    string
+		dpa     *oadpv1alpha1.DataProtectionApplication
+		wantErr bool
```
mateusoliveira43, line 1714: would also add the error message as well here; then, in cases where we want error, we would check if the message is the expected one

controllers/bsl.go
stillalearner: fixed.

controllers/bsl.go
stillalearner: yes, i was thinking of that earlier but kept it separate for better readability. But as we have proper functions now, it looks readable like this as well.

stillalearner: fixed

controllers/bsl_test.go
stillalearner: fixed. Thank you!

controllers/bsl.go
```diff
@@ -15,22 +15,28 @@ import (
 )
 
 func (r *DPAReconciler) ValidateBackupStorageLocations(dpa oadpv1alpha1.DataProtectionApplication) (bool, error) {
-	if dpa.Spec.Configuration == nil || dpa.Spec.Configuration.Velero == nil {
-		return false, errors.New("DPA CR Velero configuration cannot be nil")
-	}
 	// Ensure we have a BSL or user has specified NoDefaultBackupLocation install
 	if len(dpa.Spec.BackupLocations) == 0 && !dpa.Spec.Configuration.Velero.NoDefaultBackupLocation {
```
mateusoliveira43, line 19: This one we can remove as well, it is duplication of https://github.com/openshift/oadp-operator/blob/8890679c9b2c6866860badbd3c870ff3392a5ec8/controllers/validator.go#L37-L41

Maybe is even better to bring this part of code https://github.com/openshift/oadp-operator/blob/8890679c9b2c6866860badbd3c870ff3392a5ec8/controllers/validator.go#L30-L41
to here (the other PR was merged @stillalearner :smile: you can rebase now)

mateusoliveira43: for maintainers: should we move this code https://github.com/openshift/oadp-operator/blob/8890679c9b2c6866860badbd3c870ff3392a5ec8/controllers/validator.go#L109-L114

up in the validation chain? to do not change the order it was previously?

mateusoliveira43 (dismissed):
Just rebase and check my last comment

Thanks for the PR @stillalearner ! :rocket:

controllers/bsl.go
stillalearner: @mateusoliveira43 you are right, actually this backupLocation for loop and volumeSnapshot for loop in the validator.go is also duplicate as it is being checked already in the new functions ensureDataSecretExists and the VSl part also. removing it.

stillalearner: done. I also think that bsl and vsl validation should be done first, as the duplicate code we are removing from there was actually doing the same previously, which is now included in ValidateBSL and ValidateVSL functions.

kaovilai (approved):

controllers/bsl.go
```diff
@@ -377,14 +378,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
...
 }
 
-func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(dpa *oadpv1alpha1.DataProtectionApplication) error {
-	for _, bsl := range dpa.Spec.BackupLocations {
-		if bsl.CloudStorage == nil && bsl.Velero == nil {
-			return fmt.Errorf("no bucket CloudStorage or velero BackupStorageLocation provided for backupLocations")
-		}
+func (r *DPAReconciler) ensureBackupLocationHasVeleroOrCloudStorage(bsl *oadpv1alpha1.BackupLocation) error {
+	if bsl.CloudStorage == nil && bsl.Velero == nil {
+		return fmt.Errorf("BackupLocation must have velero or bucket configuration")
```
mpryc, line 383: Should we add on top of returning error logging error in the function like this? We could have log.error in the caller, but in my opinion it would make sense to add such error or debug to the function which actually returns the error itself, then we don't need to take care of that logging in any future caller of this func ? This would apply to other parts of this PR.

controllers/bsl.go
```diff
@@ -377,14 +378,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
...
+	if bsl.CloudStorage != nil && bsl.CloudStorage.Prefix == "" && dpa.BackupImages() {
+		return fmt.Errorf("BackupLocation must have cloud storage prefix when backupImages is not set to false")
+	}
+
+	return nil
+}
+
+func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, bsl *oadpv1alpha1.BackupLocation) error {
+	// Check if the Velero feature flag 'no-secret' is not set
+	if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret")) {
```
mpryc, line 407: Is there a risk of null pointer in that check? / Is there guarantee that the HasFeatureFlag always is there?

e.g.
```
if dpa.Spec != nil && dpa.Spec.Configuration != nil && !dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret") {

```

controllers/bsl.go
```diff
@@ -377,14 +378,53 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
...
+			// Check if user specified empty credential name
+			if bsl.Velero.Credential.Name == "" {
+				return fmt.Errorf("Secret name specified in BackupLocation %s cannot be empty", bsl.Name)
+			}
+		}
+		// Check if the BSL secret key configured in the DPA exists with a secret data
+		secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
+		bslSecret, err := r.getProviderSecret(secretName)
+		if err != nil {
+			return err
```
mpryc, line 423: Any type of log would be nice, so we know in the run what's wrong.

controllers/bsl_test.go
```diff
@@ -1666,9 +1697,191 @@ func TestDPAReconciler_ensureBackupLocationHasVeleroOrCloudStorage(t *testing.T)
...
+					Configuration: &oadpv1alpha1.ApplicationConfig{
+						Velero: &oadpv1alpha1.VeleroConfig{
+							DefaultPlugins: []oadpv1alpha1.DefaultPlugin{},
+						},
+					},
+					BackupImages: pointer.Bool(true),
+				},
+			},
+			wantErr:     true,
+			expectedErr: "BackupLocation must have cloud storage prefix when backupImages is not set to false",
```
mpryc, line 1744: Would it make sense to create type of error within controllers/*.go where the strings used in errors are stored? This way we don't have to compare against string here, but the string taken from that type? In case string message changes we don't have to update it in few places.

controllers/bsl.go
stillalearner: should not be an issue, as dpa.Spec.Configuration and dpa.Spec.Configuration.Velero nil check are already covered in validator.go

controllers/bsl.go
stillalearner: I have seen this error, it is pretty straightforward when it fails, complaining that "secret <name> not found"

controllers/bsl_test.go
stillalearner: ya this one makes sense to all the messages strings in all the test files, to be converted to Types.
We can have a separate PR for this one. I agree. Let me know if you want me to create one.

controllers/bsl.go
stillalearner: ya, I think this can also be part of the same PR where we can add Types for all these error strings
like,
ErrBucketConfigurationMissing = "BackupLocation must have...."

and then, adding log something like this for all functions?

`if bsl.Velero != nil && bsl.Velero.ObjectStorage != nil && bsl.Velero.ObjectStorage.Prefix == "" && dpa.BackupImages() {
		err:= fmt.Errorf("BackupLocation must have velero prefix when backupImages is not set to false")
		
		r.Log.Error(err, "Validation Failed")

		return err
}`

controllers/bsl_test.go
mpryc: Ok so let's not this be part of current PR, but I think it makes sense to keep all the error types / strings separate. Thinking here of something like:

```
type OadpSeverity string

const (
    Error       OadpSeverity = "error"
    Warning     OadpSeverity = "warning"
    Info        OadpSeverity = "info"
    Debug       OadpSeverity = "debug"
    Deprecation OadpSeverity = "deprecation"
)

type CustomMessage struct {
    OadpMessageID   string // Unique ID to identify message
    Message         string
    Severity        Severity
    Info            map[string]interface{} // Additional information for logging or debugging
}
```

controllers/bsl.go
mpryc: Ok worth noting, so it's not lost, but not within current PR.

PR #1306, opened 2024-02-01
Velero Args yaml can accept time as string.

kaovilai:
Related to #1307


kaovilai:
oadp-1.2 will need a separate PR

mateusoliveira43:
/hold

This change is good both  for user and developer experience, but can possibly break updates! Let's discuss prior to merging

kaovilai:
I believe it is possible to make this a field that would work with prior int64 format and the new string. We'd parse as one or the other and convert types during querying.

PR #1310, opened 2024-02-02
Do not error test on uninstall/delete

mpryc:
The test should not fail when uninstall/delete operation is failing, this is not something OADP e2e should be testing, just a wrapper around actual project test.

Of course it's not ideal when there are some objects left after test ran, but they should not interfere with the test result. If there are left overs after test there is a bug, but in testing code/framework and not OADP itself.

### DIFF
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -176,8 +176,7 @@ func runBackupAndRestore(brCase BackupRestoreCase, expectedErr error, updateLast
 
 	// uninstall app
 	log.Printf("Uninstalling app for case %s", brCase.Name)
-	err = UninstallApplication(dpaCR.Client, brCase.ApplicationTemplate)
-	Expect(err).ToNot(HaveOccurred())
+	UninstallApplication(dpaCR.Client, brCase.ApplicationTemplate)
 
 	// Wait for namespace to be deleted
 	Eventually(IsNamespaceDeleted(kubernetesClientForSuiteRun, brCase.ApplicationNamespace), timeoutMultiplier*time.Minute*4, time.Second*5).Should(BeTrue())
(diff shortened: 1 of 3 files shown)


tests/e2e/backup_restore_suite_test.go
```diff
@@ -176,8 +176,7 @@ func runBackupAndRestore(brCase BackupRestoreCase, expectedErr error, updateLast
 
 	// uninstall app
 	log.Printf("Uninstalling app for case %s", brCase.Name)
-	err = UninstallApplication(dpaCR.Client, brCase.ApplicationTemplate)
-	Expect(err).ToNot(HaveOccurred())
+	UninstallApplication(dpaCR.Client, brCase.ApplicationTemplate)
```
mateusoliveira43, line 179: I do not like this, if we do not break if something goes wrong here, are we not changing the behavior of the test?

mpryc: The test is about backup and restore and not removal of namespace or resources using OpenShift go library/API. We of course should know why it failed to remove (finalizer??) that's why there is log, but imo failing each time finalizer is not done is something outside of oadp interest.

With my personal dev env I am using the following script to clean up OADP left overs, see that I have to clean finalizer on the restore object as it's always preventing that object to be properly removed. Maybe there is bug here preventing restore object to be properly removed, but I don't think it should stop every single PR from passing or introducing flakes.

```bash
#!/bin/bash

velero delete restore -n openshift-adp --all --confirm
velero delete backup -n openshift-adp --all --confirm

aws s3 rm s3://mprycoadp/ --recursive

# Patch restore objects with empty finalizer
for restore_object in $(oc get restore -n openshift-adp | grep -v NAME | awk '{print $1}'); do
    oc patch restore "$restore_object" -n openshift-adp --type merge --patch '
metadata:
  finalizers:
';
done

# Create an array of objects to be removed
ocp_object_types=("vsc" "backup" "backuprepository" "restore" "deletebackuprequest" "dataupload" "datadownload" "downloadrequest" "podvolumerestore")

for ocp_object in "${ocp_object_types[@]}"; do
    for ocp_object_instance in $(oc get "$ocp_object" -n openshift-adp |grep -v NAME|awk '{print $1}'); do
        oc delete "$ocp_object" "$ocp_object_instance" -n openshift-adp;
    done
done
```

mateusoliveira43: Restore does have finalizers, we have a test clean script https://github.com/openshift/oadp-operator/blob/master/Makefile#L516C1-L528C24

But I did not see problems with deleting applications and their namespaces in tests run yet (and if a problem occurs, when we restore, it can in the end doing nothing, as everything is still there. I think this IS NOT WHAT WE WANT)

I would hold this for discussion, at least check how many times this is erroring. I remember to see this in afterEach only once in all logs I saw

mpryc: On my system I see it 100% times after running e2e tests that's why I've added this to my local cleanup script.

mateusoliveira43: restores are not deleted after test suite run (for debuging), that's why we have the `make test-e2e-cleanup` command, to remove OADP stuff after a test. 

If you are seeing a problem with that script, that is another problem, not related to applications to backup/restore. 

I, for example, do not need to run `make test-e2e-cleanup` before running a test suite locally again, the leftovers from previous runs, do not interfere with the following run

Issue #1311, opened 2024-02-02
Update DPA API version


mateusoliveira43:
This issue to track improvements taht could be done if DPA API versions updates. We can not change fields (except their descriptions) or remove fields within same API version, or we will be creating problems to our users when they update OADP versions.

- https://github.com/openshift/oadp-operator/issues/1295
- https://github.com/openshift/oadp-operator/issues/1276
- https://github.com/openshift/oadp-operator/issues/1379
- https://github.com/openshift/oadp-operator/issues/1528
- https://github.com/openshift/oadp-operator/issues/1534

kaovilai:
We can add new versions and auto upgrade using https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#upgrade-existing-objects-to-a-new-stored-version

kaovilai:
v1alpha2 might work well

PR #1313, opened 2024-02-02
OADP-2866: change snapshotMoveData to defaultSnapshotMoveData

weshayutin:
fix https://issues.redhat.com/browse/OADP-2866. 
velero server flag --snapshot-move-data -> --default-snapshot-move-data

Both variables can be set to true:
    -  defaultSnapshotMoveData: true
    -  defaultVolumesToFSBackup: true

In a little local testing on a very basic backup spec, the defaultVolumesToFSBackup var wins.  If defaultVolumesToFSBackup is set to false on a very basic spec defaultSnapshotMoveData worked and created a dataupload.

example dpa:
```
  spec:
    configuration:
      nodeAgent:
        enable: true
        uploaderType: kopia
      velero:
        defaultPlugins:
        - openshift
        - aws
        - csi
        defaultSnapshotMoveData: true
        defaultVolumesToFSBackup: false
```

example backup cr:
```
apiVersion: velero.io/v1
kind: Backup
metadata:
  name: mysql-persistent-test-defaults-3
  namespace: openshift-adp
spec:
  includedNamespaces:
  - mysql-persistent
  storageLocation: dpa-sample-1
  ttl: 720h0m0s

```


api/v1alpha1/oadp_types.go
```diff
@@ -97,7 +97,7 @@ type VeleroConfig struct {
 	DefaultVolumesToFSBackup *bool `json:"defaultVolumesToFSBackup,omitempty"`
 	// Specify whether CSI snapshot data should be moved to backup storage by default
 	// +optional
-	SnapshotMoveData *bool `json:"snapshotMoveData,omitempty"`
+	DefaultSnapshotMoveData *bool `json:"defaultSnapshotMoveData,omitempty"`
```
mateusoliveira43, line 100: This change can break updates

Would suggest adding this to this list https://github.com/openshift/oadp-operator/issues/1311

I believe the harmless way to do this, would be create the new field (`DefaultSnapshotMoveData`) and deprecate the older one (`SnapshotMoveData`)

weshayutin: or we can just change the server flag I suppose so it's correct.  
that's a one liner.

controllers/velero.go
```diff
@@ -362,14 +363,18 @@ func (r *DPAReconciler) customizeVeleroDeployment(dpa *oadpv1alpha1.DataProtecti
...
 	// check for default-volumes-to-fs-backup
 	defaultVolumesToFSBackup := getDefaultVolumesToFSBackup(dpa)
-	if len(defaultVolumesToFSBackup) > 0 {
+	// FIXME both default-snapshot and default-volumes can not be true at the same time.
+	if defaultSnapshotMoveData == "true" && defaultVolumesToFSBackup == "true" {
+		return fmt.Errorf("default-snapshot-move-data and default-volumes-to-fs-backup can not be true at the same time")
+	}
+
+	if len(defaultSnapshotMoveData) > 0 && len(defaultVolumesToFSBackup) == 0 {
+		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--default-snapshot-move-data=%s", defaultSnapshotMoveData))
```
weshayutin, line 375: @mateusoliveira43 so I'm referring to JUST changing the one line here. line 375

api/v1alpha1/oadp_types.go
kaovilai: You've become the anti crd change man after encountering the enums once 😅.

If this field was guaranteed to crash velero pod, which it was, it meant no one was using it in production. This I believe this field is safe to change. We can also just keep crd and only make minimal code change

mateusoliveira43: hmm I think this case you can change the field name, because this was recently added https://github.com/openshift/oadp-operator/pull/1280/files (so, we did not release a DPA with this field yet, so, we can its name)

weshayutin:
I need a reconcile check, 
I can add:
```
  configuration:
    nodeAgent:
      enable: true
      uploaderType: kopia
    velero:
      defaultPlugins:
        - openshift
        - aws
        - csi
      defaultSnapshotMoveData: true
      defaultVolumesToFSBackup: true
      featureFlags:
        - EnableCSI
```
The dpa will reconcile, but velero pod doesn't change.  So not done yet

sseago:
@weshayutin "also do not allow default-fs-copy and default-snapshot-move-data at the same time."

Why not? There's no upstream restriction on it, and it needlessly limits OADP flexibility. There is no logical conflict between these options. Specifying both simply means this:
1) "use fs backup unless specified otherwise for this volume"
2) "if this volume uses snapshots, use datamover to store it in s3"

If a customer has mostly non-CSI volumes but a few CSI volumes, they probably want to use the above options together.

weshayutin:
Oh, that's interesting TIL. I think I'll have another question about this. Thanks for the review.

controllers/velero.go
```diff
@@ -362,13 +363,16 @@ func (r *DPAReconciler) customizeVeleroDeployment(dpa *oadpv1alpha1.DataProtecti
...
 
 	// check for default-snapshot-move-data parameter
-	snapshotMoveData := getSnapshotMoveDataValue(dpa)
-	if len(snapshotMoveData) > 0 {
-		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--snapshot-move-data=%s", snapshotMoveData))
+	defaultSnapshotMoveData := getDefaultSnapshotMoveDataValue(dpa)
+	// check for default-volumes-to-fs-backup
+	defaultVolumesToFSBackup := getDefaultVolumesToFSBackup(dpa)
+
+	// check for default-volumes-to-fs-backup
```
kaovilai, line 370:
```suggestion
	// check for default-snapshot-move-data
```

tests/e2e/sample-applications/mysql-persistent/mysql-backup.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: velero.io/v1
 kind: Backup
 metadata:
-  name: mysql-persistent
+  name: mysql-persistent-test-defaults-3
```
kaovilai, line 4: Is this change intended?

weshayutin:
thanks for the review Tiger

kaovilai (approved):

kaovilai:
[temporary CSI flake error](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1313/pull-ci-openshift-oadp-operator-master-4.13-e2e-test-aws/1757141787274645504#1:build-log.txt%3A3465)

/override 
ci/prow/4.13-e2e-test-aws
--


<br class="Apple-interchange-newline">

PR #1323, opened 2024-02-07
CI: Add filter for virtualization testing.

mrnold:
Replace current virtualization test skipping mechanism with a ginkgo test filter.

We will create a new prow job to run virtualization tests separately from the other E2E tests.
It will set `TEST_VIRT` to `true` which will be added to the make command in the new job.
When TEST_VIRT is true, "virt" filter will be added to the test-e2e ginkgo command line, and all current virtualization tests are updated to have this label.

### DIFF
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -6,7 +6,6 @@ import (
 	"log"
 	"os"
 	"strconv"
-	"strings"
 	"testing"
 	"time"
 
@@ -26,7 +25,6 @@ import (
 var bslCredFile, namespace, credSecretRef, instanceName, provider, vslCredFile, settings, artifact_dir, oc_cli, stream string
 var timeoutMultiplierInput, flakeAttempts int64
 var timeoutMultiplier time.Duration
-var virtTestingEnabled bool
 
 func init() {
 	// TODO better descriptions to flags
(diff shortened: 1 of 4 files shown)


mrnold:
This might be a little hasty. I should probably make sure all the tests have good labels, and see if the virtualization tests actually need to be combined with the cluster type.

The filter mechanism itself is good though. Maybe it would be better to give prow more control over the filter, instead of just setting a virtualization-only test flag?

kaovilai:
infra issue
/retest

tests/e2e/e2e_suite_test.go
```diff
@@ -42,7 +40,6 @@ func init() {
 	flag.StringVar(&oc_cli, "oc_cli", "oc", "OC CLI Client")
 	flag.StringVar(&stream, "stream", "up", "[up, down] upstream or downstream")
 	flag.Int64Var(&timeoutMultiplierInput, "timeout_multiplier", 1, "Customize timeout multiplier from default (1)")
-	flag.BoolVar(&virtTestingEnabled, "test_virt", false, "Enable Virtual Machine backup/restore testing (default false)")
```
kaovilai, line 45: So this flag doesn't apply anymore?

kaovilai: I see.. it's in the It label
```go
	It("should upload a data volume successfully", Label("virt"), func() {
```

kaovilai (approved):
/lgtm

kaovilai:
ordered by the commander

weshayutin (approved):

tests/e2e/e2e_suite_test.go
mrnold: Right, I am trying to reduce this part to ginkgo labels, so all the virtualization testing is triggered by the "virt" test filter.

kaovilai:
> well.. if the change doesn't in fact get tested by the tests.. who cares

due to the new tests being behind a label, the existing tests do not apply.

/override ci/prow/4.12-ci-index
/override ci/prow/4.12-images
/override ci/prow/4.13-ci-index
/override ci/prow/4.13-images
/override ci/prow/4.14-ci-index
/override ci/prow/4.14-e2e-test-azure
/override ci/prow/4.14-images
/override ci/prow/images
/override ci/prow/unit-test

mateusoliveira43:
why have we disabled virt tests?

mrnold:
> why have we disabled virt tests?

There is a new prow job for virtualization tests: https://github.com/openshift/release/pull/48623

The new job sets TEST_VIRT, which now sets up ginkgo labels to only run the virtualization tests and nothing else. The labels may need to be adjusted when there are actual VM tests to run. I took out the code that figured out when to skip the tests, so now it defers to prow.

PR #1216, opened 2023-11-05
Fix DPA reconcile on empty BSL secret name and key

sbahar619:
This PR addresses: 
https://issues.redhat.com/browse/OADP-2940 

I added coverage for fail DPA reconcile on empty BSL credential name and key.
Also, I fixed the logic condition which was not accurate.
I added 4 more unit tests for more coverage.

PR #1326, opened 2024-02-09
OADP-3052: Fix VSL secret data validation

stillalearner:
Added VSL secret Validation.

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -121,8 +121,14 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool, 
 		//	 1. oadpApi.OadpOperatorLabel: "True"
 		// 	 2. dataprotectionapplication.name: <name>
 		// which in turn will be used in the label handler to trigger the reconciliation loop
+		var secretName string
+		if bslSpec.CloudStorage != nil {
+			secretName, _, _ = r.getSecretNameAndKeyFromCloudStorage(bslSpec.CloudStorage)
+		}
 
-		secretName, _ := r.getSecretNameAndKeyforBackupLocation(bslSpec)
+		if bslSpec.Velero != nil {
+			secretName, _, _ = r.getSecretNameAndKey(bslSpec.Velero.Config, bslSpec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Velero.Provider))
+		}
 		_, err := r.UpdateCredentialsSecretLabels(secretName, dpa.Namespace, dpa.Name)
 		if err != nil {
 			return false, err
@@ -381,7 +387,7 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
 		r.Log.Info(fmt.Sprintf("%s backupstoragelocation is configured but velero plugin for %s is not present", bslSpec.Provider, bslSpec.Provider))
 		//TODO: set warning condition on Velero CR
 	}
-	secretName, _ := r.getSecretNameAndKey(&bslSpec, oadpv1alpha1.DefaultPlugin(bslSpec.Provider))
+	secretName, _, _ := r.getSecretNameAndKey(bslSpec.Config, bslSpec.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Provider))
 
 	_, err := r.getProviderSecret(secretName)
 
@@ -430,16 +436,23 @@ func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionA
 				return fmt.Errorf("Secret name specified in BackupLocation %s cannot be empty", bsl.Name)
 			}
 		}
+
 		// Check if the BSL secret key configured in the DPA exists with a secret data
-		secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
-		bslSecret, err := r.getProviderSecret(secretName)
-		if err != nil {
-			return err
+
+		if bsl.CloudStorage != nil {
+			_, _, err := r.getSecretNameAndKeyFromCloudStorage(bsl.CloudStorage)
+			if err != nil {
+				return err
+			}
 		}
-		data, foundKey := bslSecret.Data[secretKey]
-		if !foundKey || len(data) == 0 {
-			return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
+
+		if bsl.Velero != nil {
+			_, _, err := r.getSecretNameAndKey(bsl.Velero.Config, bsl.Velero.Credential, oadpv1alpha1.DefaultPlugin(bsl.Velero.Provider))
+			if err != nil {
+				return err
+			}
 		}
+
 	}
 	return nil
 }
--- a/controllers/registry.go
+++ b/controllers/registry.go
@@ -242,33 +242,35 @@ func replaceCarriageReturn(data map[string][]byte, logger logr.Logger) map[strin
 	return data
 }
 
-func (r *DPAReconciler) getSecretNameAndKeyforBackupLocation(bslspec oadpv1alpha1.BackupLocation) (string, string) {
-
-	if bslspec.CloudStorage != nil {
-		if bslspec.CloudStorage.Credential != nil {
-			return bslspec.CloudStorage.Credential.Name, bslspec.CloudStorage.Credential.Key
+func (r *DPAReconciler) getSecretNameAndKeyFromCloudStorage(cloudStorage *oadpv1alpha1.CloudStorageLocation) (string, string, error) {
+	if cloudStorage.Credential != nil {
+		// Check if user specified empty credential key
+		if cloudStorage.Credential.Key == "" {
+			return "", "", fmt.Errorf("Secret key specified in CloudStorage cannot be empty")
 		}
+		// Check if user specified empty credential name
+		if cloudStorage.Credential.Name == "" {
+			return "", "", fmt.Errorf("Secret name specified in CloudStorage cannot be empty")
+		}
+		secretName, secretKey := cloudStorage.Credential.Name, cloudStorage.Credential.Key
+		err := r.verifySecretContent(secretName, secretKey)
+		return secretName, secretKey, err
 	}
-	if bslspec.Velero != nil {
-		return r.getSecretNameAndKey(bslspec.Velero, oadpv1alpha1.DefaultPlugin(bslspec.Velero.Provider))
-	}
-
-	return "", ""
+	return "", "", nil
 }
 
-func (r *DPAReconciler) getSecretNameAndKey(bslSpec *velerov1.BackupStorageLocationSpec, plugin oadpv1alpha1.DefaultPlugin) (string, string) {
+func (r *DPAReconciler) getSecretNameAndKey(config map[string]string, credential *corev1.SecretKeySelector, plugin oadpv1alpha1.DefaultPlugin) (string, string, error) {
 	// Assume default values unless user has overriden them
 	secretName := credentials.PluginSpecificFields[plugin].SecretName
 	secretKey := credentials.PluginSpecificFields[plugin].PluginSecretKey
-	if _, ok := bslSpec.Config["credentialsFile"]; ok {
+	if _, ok := config["credentialsFile"]; ok {
 		if secretName, secretKey, err :=
-			credentials.GetSecretNameKeyFromCredentialsFileConfigString(bslSpec.Config["credentialsFile"]); err == nil {
+			credentials.GetSecretNameKeyFromCredentialsFileConfigString(config["credentialsFile"]); err == nil {
 			r.Log.Info(fmt.Sprintf("credentialsFile secret: %s, key: %s", secretName, secretKey))
-			return secretName, secretKey
+			return secretName, secretKey, nil
 		}
 	}
 	// check if user specified the Credential Name and Key
-	credential := bslSpec.Credential
 	if credential != nil {
 		if len(credential.Name) > 0 {
 			secretName = credential.Name
@@ -278,7 +280,12 @@ func (r *DPAReconciler) getSecretNameAndKey(bslSpec *velerov1.BackupStorageLocat
 		}
 	}
 
-	return secretName, secretKey
+	err := r.verifySecretContent(secretName, secretKey)
+	if err != nil {
+		return secretName, secretKey, err
+	}
+
+	return secretName, secretKey, nil
 }
 
 func (r *DPAReconciler) parseAWSSecret(secret corev1.Secret, secretKey string, matchProfile string) (string, string, error) {
@@ -711,7 +718,7 @@ func (r *DPAReconciler) patchRegistrySecret(secret *corev1.Secret, bsl *velerov1
 
 func (r *DPAReconciler) populateAWSRegistrySecret(bsl *velerov1.BackupStorageLocation, registrySecret *corev1.Secret) error {
 	// Check for secret name
-	secretName, secretKey := r.getSecretNameAndKey(&bsl.Spec, oadpv1alpha1.DefaultPluginAWS)
+	secretName, secretKey, _ := r.getSecretNameAndKey(bsl.Spec.Config, bsl.Spec.Credential, oadpv1alpha1.DefaultPluginAWS)
 
 	// fetch secret and error
 	secret, err := r.getProviderSecret(secretName)
@@ -740,7 +747,7 @@ func (r *DPAReconciler) populateAWSRegistrySecret(bsl *velerov1.BackupStorageLoc
 
 func (r *DPAReconciler) populateAzureRegistrySecret(bsl *velerov1.BackupStorageLocation, registrySecret *corev1.Secret) error {
 	// Check for secret name
-	secretName, secretKey := r.getSecretNameAndKey(&bsl.Spec, oadpv1alpha1.DefaultPluginMicrosoftAzure)
+	secretName, secretKey, _ := r.getSecretNameAndKey(bsl.Spec.Config, bsl.Spec.Credential, oadpv1alpha1.DefaultPluginMicrosoftAzure)
 
 	// fetch secret and error
 	secret, err := r.getProviderSecret(secretName)
@@ -780,5 +787,17 @@ func (r *DPAReconciler) populateAzureRegistrySecret(bsl *velerov1.BackupStorageL
 		"resource_group_key":  []byte(azcreds.resourceGroup),
 	}
 
+	return nil
+}
+
+func (r *DPAReconciler) verifySecretContent(secretName string, secretKey string) error {
+	secret, err := r.getProviderSecret(secretName)
+	if err != nil {
+		return err
+	}
+	data, foundKey := secret.Data[secretKey]
+	if !foundKey || len(data) == 0 {
+		return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
+	}
 	return nil
 }
--- a/controllers/registry_test.go
+++ b/controllers/registry_test.go
@@ -151,7 +151,7 @@ var testAWSEnvVar = cloudProviderEnvVarMap["aws"]
 var testAzureEnvVar = cloudProviderEnvVarMap["azure"]
 var testGCPEnvVar = cloudProviderEnvVarMap["gcp"]
 
-func TestDPAReconciler_getSecretNameAndKeyforBackupLocation(t *testing.T) {
+func TestDPAReconciler_getSecretNameAndKey(t *testing.T) {
 	tests := []struct {
 		name           string
 		bsl            *oadpv1alpha1.BackupLocation
@@ -210,6 +210,51 @@ func TestDPAReconciler_getSecretNameAndKeyforBackupLocation(t *testing.T) {
 
 			wantProfile: "aws-provider-no-cred",
 		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			fakeClient, err := getFakeClientFromObjects(tt.secret)
+			if err != nil {
+				t.Errorf("error in creating fake client, likely programmer error")
+			}
+			r := &DPAReconciler{
+				Client:        fakeClient,
+				Scheme:        fakeClient.Scheme(),
+				Log:           logr.Discard(),
+				Context:       newContextForTest(tt.name),
+				EventRecorder: record.NewFakeRecorder(10),
+			}
+
+			if tt.wantProfile == "aws-provider" {
+				tt.wantSecretKey = "cloud"
+				tt.wantSecretName = "cloud-credentials-aws"
+			}
+			if tt.wantProfile == "aws-provider-no-cred" {
+				tt.wantSecretKey = "cloud"
+				tt.wantSecretName = "cloud-credentials"
+			}
+
+			gotName, gotKey, _ := r.getSecretNameAndKey(tt.bsl.Velero.Config, tt.bsl.Velero.Credential, oadpv1alpha1.DefaultPlugin(tt.bsl.Velero.Provider))
+
+			if !reflect.DeepEqual(tt.wantSecretName, gotName) {
+				t.Errorf("expected registry container env var to be %#v, got %#v", tt.wantSecretName, gotName)
+			}
+			if !reflect.DeepEqual(tt.wantSecretKey, gotKey) {
+				t.Errorf("expected registry container env var to be %#v, got %#v", tt.wantSecretKey, gotKey)
+			}
+		})
+	}
+}
+
+func TestDPAReconciler_getSecretNameAndKeyFromCloudStorage(t *testing.T) {
+	tests := []struct {
+		name           string
+		bsl            *oadpv1alpha1.BackupLocation
+		secret         *corev1.Secret
+		wantProfile    string
+		wantSecretName string
+		wantSecretKey  string
+	}{
 		{
 			name: "given cloud storage secret, appropriate secret name and key are returned",
 			bsl: &oadpv1alpha1.BackupLocation{
@@ -269,14 +314,6 @@ func TestDPAReconciler_getSecretNameAndKeyforBackupLocation(t *testing.T) {
 				EventRecorder: record.NewFakeRecorder(10),
 			}
 
-			if tt.wantProfile == "aws-provider" {
-				tt.wantSecretKey = "cloud"
-				tt.wantSecretName = "cloud-credentials-aws"
-			}
-			if tt.wantProfile == "aws-provider-no-cred" {
-				tt.wantSecretKey = "cloud"
-				tt.wantSecretName = "cloud-credentials"
-			}
 			if tt.wantProfile == "aws-cloud-cred" {
 				tt.wantSecretKey = "cloud"
 				tt.wantSecretName = "cloud-credentials-aws"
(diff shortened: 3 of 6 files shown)


stillalearner:
Some of the code here is similar to the BSL secret validation.
Should this be merged?

Everything I believe cannot be merged due to some differences between the BackupStorageLocationSpec & VolumeSnapshotLocationSpec, like the CloudStorage key, and others.

But the function getSecretNameAndKey in registry.go, should that be converted to a common function somehow, like by passing interface?

would like to know your views @kaovilai @shubham-pampattiwar

mateusoliveira43:
Yeah @stillalearner , there is some very similar parts (like https://github.com/openshift/oadp-operator/pull/1216). I believe a common function should be created and used in both validations

stillalearner:
@mateusoliveira43 So, i have created this common function getSecretNameAndKeyFinal which can be used for both the validations. Keeping some parts separate due to the differences in VSL and BSL specs. Does it look good?

PS: Please ignore the comments / names, will fix in final PR.

stillalearner:
Updated. Created a common function. 

Please review @kaovilai , @mateusoliveira43 , @shubham-pampattiwar

controllers/registry.go
```diff
@@ -251,25 +251,58 @@ func (r *DPAReconciler) getSecretNameAndKeyforBackupLocation(bslspec oadpv1alpha
...
+		}
+		if len(credential.Key) > 0 {
+			secretKey = credential.Key
+		}
+	}
+
+	return secretName, secretKey
+}
+
+func (r *DPAReconciler) getSecretNameAndKeyFromSnapshotLocation(vslspec *velerov1.VolumeSnapshotLocationSpec, plugin oadpv1alpha1.DefaultPlugin) (string, string) {
```
kaovilai, line 293: Where is this func used?

stillalearner: oops, nowhere, not needed anymore, common code is being used,  removed it. Thanks!

kaovilai (approved):

kaovilai:
one failure

```
      <[]string | len:4, cap:4>: [
          "time=\"2024-02-13T23:04:56Z\" level=warning msg=\"VolumeSnapshot has a temporary error Failed to check and update snapshot content: failed to remove VolumeSnapshotBeingCreated annotation on the content snapcontent-ce80b4e3-2026-471f-9a1e-b0f578427909: \\\"snapshot controller failed to update snapcontent-ce80b4e3-2026-471f-9a1e-b0f578427909 on API server: Operation cannot be fulfilled on volumesnapshotcontents.snapshot.storage.k8s.io \\\\\\\"snapcontent-ce80b4e3-2026-471f-9a1e-b0f578427909\\\\\\\": the object has been modified; please apply your changes to the latest version and try again\\\". Snapshot controller will retry later.\" backup=openshift-adp/mysql-twovol-csi-e2e-14e1486f-cac4-11ee-84a4-0a580a81e561 cmd=/plugins/velero-plugin-for-csi logSource=\"/opt/app-root/src/velero-plugin-for-csi/internal/backup/volumesnapshot_action.go:250\" pluginName=velero-plugin-for-csi",
          "time=\"2024-02-13T23:04:56Z\" level=warning msg=\"VolumeSnapshot has a temporary error Failed to create snapshot: failed to remove VolumeSnapshotBeingCreated annotation on the content snapcontent-b18c34f7-50ed-43f4-981c-4cc4613fb5d0: \\\"snapshot controller failed to update snapcontent-b18c34f7-50ed-43f4-981c-4cc4613fb5d0 on API server: Operation cannot be fulfilled on volumesnapshotcontents.snapshot.storage.k8s.io \\\\\\\"snapcontent-b18c34f7-50ed-43f4-981c-4cc4613fb5d0\\\\\\\": the object has been modified; please apply your changes to the latest version and try again\\\". Snapshot controller will retry later.\" backup=openshift-adp/mysql-twovol-csi-e2e-14e1486f-cac4-11ee-84a4-0a580a81e561 cmd=/plugins/velero-plugin-for-csi logSource=\"/opt/app-root/src/velero-plugin-for-csi/internal/backup/volumesnapshot_action.go:250\" pluginName=velero-plugin-for-csi",
          "time=\"2024-02-13T23:04:56Z\" level=error msg=0 backup=openshift-adp/mysql-twovol-csi-e2e-14e1486f-cac4-11ee-84a4-0a580a81e561 logSource=\"pkg/controller/backup_controller.go:729\"",
          "time=\"2024-02-13T23:04:56Z\" level=error msg=1 backup=openshift-adp/mysql-twovol-csi-e2e-14e1486f-cac4-11ee-84a4-0a580a81e561 logSource=\"pkg/controller/backup_controller.go:729\"",
      ]
  to equal
      <[]string | len:0, cap:0>: []
```

https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1326/pull-ci-openshift-oadp-operator-master-4.14-e2e-test-aws/1757519239977111552#1:build-log.txt%3A2223

kaovilai:
one more review :)
/cc @shubham-pampattiwar @mpryc

controllers/vsl.go
```diff
@@ -261,3 +266,33 @@ func containsPlugin(d []oadpv1alpha1.DefaultPlugin, value string) bool {
 	}
 	return false
 }
+
+func (r *DPAReconciler) ensureVslSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, vsl *oadpv1alpha1.SnapshotLocation) error {
```
mateusoliveira43, line 270: this is a duplication of  `ensureSecretDataExists` function in `controllers/bsl.go`, but I was not able to remove it, compiler always complained about my ideas. Is it possible to have a single function for both bsl and vsl?

stillalearner: I also tried it actually, but there are some differences between bsl and vsl specs, so it was making it more complex,  also considering the future use , I thought keeping it separate will be better.

Do you have some other ideas? @mateusoliveira43

mpryc: How about something like this, it compiles but didn't test in the OCP with reconcile nor e2e tests, just an idea:

```diff
diff --git a/controllers/bsl.go b/controllers/bsl.go
index 005b851..2140184 100644
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -403,29 +403,9 @@ func (r *DPAReconciler) ensurePrefixWhenBackupImages(dpa *oadpv1alpha1.DataProte
 }
 
 func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, bsl *oadpv1alpha1.BackupLocation) error {
-       // Check if the Velero feature flag 'no-secret' is not set
-       if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret")) {
-               // Check if the user specified credential under velero
-               if bsl.Velero != nil && bsl.Velero.Credential != nil {
-                       // Check if user specified empty credential key
-                       if bsl.Velero.Credential.Key == "" {
-                               return fmt.Errorf("Secret key specified in BackupLocation %s cannot be empty", bsl.Name)
-                       }
-                       // Check if user specified empty credential name
-                       if bsl.Velero.Credential.Name == "" {
-                               return fmt.Errorf("Secret name specified in BackupLocation %s cannot be empty", bsl.Name)
-                       }
-               }
-               // Check if the BSL secret key configured in the DPA exists with a secret data
-               secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
-               bslSecret, err := r.getProviderSecret(secretName)
-               if err != nil {
-                       return err
-               }
-               data, foundKey := bslSecret.Data[secretKey]
-               if !foundKey || len(data) == 0 {
-                       return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
-               }
+       getLocationNameAndKeyFunc := func(loc interface{}) (string, string) {
+               bsl := loc.(*oadpv1alpha1.BackupLocation)
+               return r.getSecretNameAndKeyforBackupLocation(*bsl)
        }
-       return nil
+       return r.ensureSecretDataExistsCommon(dpa, bsl, getLocationNameAndKeyFunc)
 }
diff --git a/controllers/common.go b/controllers/common.go
index 80f5eeb..77aee99 100644
--- a/controllers/common.go
+++ b/controllers/common.go
@@ -1,11 +1,54 @@
 package controllers
 
 import (
+       "fmt"
+       oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
        "github.com/openshift/oadp-operator/pkg/common"
        corev1 "k8s.io/api/core/v1"
        "k8s.io/utils/pointer"
+       "reflect"
 )
 
+func (r *DPAReconciler) ensureSecretDataExistsCommon(dpa *oadpv1alpha1.DataProtectionApplication, location interface{}, getLocationNameAndKeyFunc func(interface{}) (string, string)) error {
+       // Check if the Velero feature flag 'no-secret' is not set
+       if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret")) {
+               // Check if the user specified credential under velero
+               var cred *corev1.SecretKeySelector
+               var name string
+               switch loc := location.(type) {
+               case *oadpv1alpha1.BackupLocation:
+                       cred = loc.Velero.Credential
+                       name = cred.Name
+               case *oadpv1alpha1.SnapshotLocation:
+                       cred = loc.Velero.Credential
+                       name = cred.Name
+               default:
+                       return fmt.Errorf("unknown location type")
+               }
+               if cred != nil {
+                       // Check if user specified empty credential key
+                       if cred.Key == "" {
+                               return fmt.Errorf("Secret key specified in %s %s cannot be empty", reflect.TypeOf(location).Elem().Name(), name)
+                       }
+                       // Check if user specified empty credential name
+                       if name == "" {
+                               return fmt.Errorf("Secret name specified in %s %s cannot be empty", reflect.TypeOf(location).Elem().Name(), name)
+                       }
+               }
+               // Check if the secret key configured in the DPA exists with a secret data
+               secretName, secretKey := getLocationNameAndKeyFunc(location)
+               locationSecret, err := r.getProviderSecret(secretName)
+               if err != nil {
+                       return err
+               }
+               data, foundKey := locationSecret.Data[secretKey]
+               if !foundKey || len(data) == 0 {
+                       return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
+               }
+       }
+       return nil
+}
+
 // setting defaults to avoid emitting update events
 func setContainerDefaults(container *corev1.Container) {
        // setting defaults to avoid emitting update events
diff --git a/controllers/vsl.go b/controllers/vsl.go
index 7e4cdb2..ad7efc6 100644
--- a/controllers/vsl.go
+++ b/controllers/vsl.go
@@ -268,31 +268,9 @@ func containsPlugin(d []oadpv1alpha1.DefaultPlugin, value string) bool {
 }
 
 func (r *DPAReconciler) ensureVslSecretDataExists(dpa *oadpv1alpha1.DataProtectionApplication, vsl *oadpv1alpha1.SnapshotLocation) error {
-       // Check if the Velero feature flag 'no-secret' is not set
-       if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret")) {
-               // Check if the user specified credential under velero
-               if vsl.Velero != nil && vsl.Velero.Credential != nil {
-                       // Check if user specified empty credential key
-                       if vsl.Velero.Credential.Key == "" {
-                               return fmt.Errorf("Secret key specified in SnapshotLocation %s cannot be empty", "SL")
-                       }
-                       // Check if user specified empty credential name
-                       if vsl.Velero.Credential.Name == "" {
-                               return fmt.Errorf("Secret name specified in SnapshotLocation %s cannot be empty", "SL")
-                       }
-
-               }
-               // Check if the VSL secret key configured in the DPA exists with a secret data
-               secretName, secretKey := r.getSecretNameAndKeyforSnapshotLocation(*vsl)
-               vslSecret, err := r.getProviderSecret(secretName)
-               if err != nil {
-                       return err
-               }
-               data, foundKey := vslSecret.Data[secretKey]
-               if !foundKey || len(data) == 0 {
-                       return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
-               }
-
+       getLocationNameAndKeyFunc := func(loc interface{}) (string, string) {
+               vsl := loc.(*oadpv1alpha1.SnapshotLocation)
+               return r.getSecretNameAndKeyforSnapshotLocation(*vsl)
        }
-       return nil
+       return r.ensureSecretDataExistsCommon(dpa, vsl, getLocationNameAndKeyFunc)
 }

```

stillalearner: yes i tried exactly like this typecasting approach, using interface{}, maybe it will work, but dont you think it is more complicated approach?

stillalearner: i couldnt get it working though, using this as well.

stillalearner: AFAIR, one of the problem i faced is comparing "loc.Velero!=nil" inside the common function (which is missing in above code) as bsl.velero and vsl.velero are of different types BackupLocationSpec, and Volumesnapshotlocationspec.

mpryc: It's definitely less readable code. Other option would be to have very simple functs that does just parts of the common logic, e.g.:
```
if !(dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret"))
```
Or the following to write as fucnt getting secretName and secretKey as args
```
		if !foundKey || len(data) == 0 {
			return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
		}
```

stillalearner: @mpryc yes, this would be better and much readable, Updated! Please have a look.

controllers/bsl.go
```diff
@@ -418,14 +418,10 @@ func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionA
 		}
 		// Check if the BSL secret key configured in the DPA exists with a secret data
 		secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
-		bslSecret, err := r.getProviderSecret(secretName)
+		err := r.verifySecretContent(secretName, secretKey)
```
mpryc, line 421: I like it, but we could have even further optimization of this code.

The following function cab be modified to do the validation, right?

```
getSecretNameAndKeyforBackupLocation
```

stillalearner: yes, I can modify this function something like this:
`func (r *DPAReconciler) validateSecretforBackupLocation(bslspec oadpv1alpha1.BackupLocation) (error) {

	var secretName, secretKey string
	if bslspec.CloudStorage != nil {
		if bslspec.CloudStorage.Credential != nil {
			secretName = bslspec.CloudStorage.Credential.Name
			secretKey = bslspec.CloudStorage.Credential.Key
		}
	}
	if bslspec.Velero != nil {
		secretName, secretKey = r.getSecretNameAndKey(bslspec.Velero.Config, bslspec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslspec.Velero.Provider))
	}

	err := r.verifySecretContent(secretName, secretKey)
	if err != nil {
		return err
	}

	return nil
}`


but this function is being used at many places to get secret Name and Key, like in the ReconcileBackupStorageLocation function as well.
Modifying it further will create some more mess. 

@mpryc

mpryc: Getting and validating secret by it's name and key should be the same across all OADP places as it's always be the same result: `secretName` and `secretKey` existing and valid or not. So I don't think it's actually wrong approach to have unique common function to do that.

stillalearner: @mpryc Can you please have a look? I have tried to further simplify it.

controllers/vsl.go
stillalearner: 1 function for getting Secret name and key, and 1 function for validation.

stillalearner:
PS: will add the CloudStorage tests separately.

controllers/bsl.go
```diff
@@ -416,16 +422,23 @@ func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionA
...
 			}
 		}
+
 		// Check if the BSL secret key configured in the DPA exists with a secret data
-		secretName, secretKey := r.getSecretNameAndKeyforBackupLocation(*bsl)
-		bslSecret, err := r.getProviderSecret(secretName)
-		if err != nil {
-			return err
+
+		if bsl.CloudStorage != nil {
```
mateusoliveira43, line 428: I am very confused here: why CloudStorage does not have the same checks for credentials as above? (checking if name and key are not empty)

mateusoliveira43: It should not be something like this?
```go
		// Check if the user specified credential under cloud storage
		if bsl.CloudStorage != nil && bsl.CloudStorage.Credential != nil {
			// Check if user specified empty credential key
			if bsl.CloudStorage.Credential.Key == "" {
				return fmt.Errorf("Secret key specified in CloudStorage of BackupLocation %s cannot be empty", bsl.Name)
			}
			// Check if user specified empty credential name
			if bsl.CloudStorage.Credential.Name == "" {
				return fmt.Errorf("Secret name specified in CloudStorage of BackupLocation %s cannot be empty", bsl.Name)
			}
			err := r.verifySecretContent(bsl.CloudStorage.Credential.Name, bsl.CloudStorage.Credential.Key)
			if err != nil {
				return err
			}
		}
```

controllers/bsl.go
```diff
@@ -416,16 +422,23 @@ func (r *DPAReconciler) ensureSecretDataExists(dpa *oadpv1alpha1.DataProtectionA
...
+			if err != nil {
+				return err
+			}
 		}
-		data, foundKey := bslSecret.Data[secretKey]
-		if !foundKey || len(data) == 0 {
-			return fmt.Errorf("Secret name %s is missing data for key %s", secretName, secretKey)
+
+		if bsl.Velero != nil {
+			_, _, err := r.getSecretNameAndKey(bsl.Velero.Config, bsl.Velero.Credential, oadpv1alpha1.DefaultPlugin(bsl.Velero.Provider))
```
mateusoliveira43, line 436: I think this function should be removed from here (and vsl). The idea is to check for errors, right? It can be done direct in velero `if` with this, not? (which avoid checking again if name and key are not empty)
```go
		if bsl.Velero != nil && bsl.Velero.Credential != nil {
			...
			err := r.verifySecretContent(bsl.Velero.Credential.Name, bsl.Velero.Credential.Key)
			if err != nil {
				return err
			}
		}
```

controllers/registry_test.go
```diff
@@ -209,50 +209,6 @@ func TestDPAReconciler_getSecretNameAndKeyforBackupLocation(t *testing.T) {
 
 			wantProfile: "aws-provider-no-cred",
 		},
-		{
```
mateusoliveira43, line 212: why these tests were removed?

controllers/registry.go
```diff
@@ -243,33 +243,31 @@ func replaceCarriageReturn(data map[string][]byte, logger logr.Logger) map[strin
...
-func (r *DPAReconciler) getSecretNameAndKeyforBackupLocation(bslspec oadpv1alpha1.BackupLocation) (string, string) {
-
-	if bslspec.CloudStorage != nil {
-		if bslspec.CloudStorage.Credential != nil {
-			return bslspec.CloudStorage.Credential.Name, bslspec.CloudStorage.Credential.Key
+func (r *DPAReconciler) getSecretNameAndKeyFromCloudStorage(cloudStorage *oadpv1alpha1.CloudStorageLocation) (string, string, error) {
+	if cloudStorage.Credential != nil {
+		secretName, secretKey := cloudStorage.Credential.Name, cloudStorage.Credential.Key
+		err := r.verifySecretContent(secretName, secretKey)
+		if err != nil {
```
mateusoliveira43, line 250: be careful in these situations: note that the if is not necessary, i.e. this
```go
		if err != nil {
			return secretName, secretKey, err
		}

		return secretName, secretKey, err
```
is the same as 
```go
		return secretName, secretKey, err
```

mateusoliveira43 (changes requested):
There are some simplifications that can be made and some questions

controllers/bsl.go
stillalearner: These Credential checks are also needed at other places, like in the ReconcileBackupStorageLocations , So i have included it in the below function at a common place "getSecretNameandKeyFromCloudStorage"

controllers/bsl.go
stillalearner: same logic will be written 2 times then..

controllers/registry_test.go
stillalearner: Only keeping the velero tests in this function, will write a separate Test function for CloudStorage tests, as both the functions are separate now

controllers/registry.go
stillalearner: ah right! we need to return in both cases. updating this

controllers/bsl.go
stillalearner: this getSecretNameandKey from velero credentials, is also being used at multiple places, so i have made it as a common function

stillalearner: Also @mpryc suggested that, the function getting the secret name and key should also validate the content of the secret, hence the return type of the function is changed to (string, string, error)

mateusoliveira43: But like this, we would be checking the same thing multiple times, not?

here would be the necessary check
https://github.com/openshift/oadp-operator/blob/91e1aacf47ea2b44bafdb9d6d3c14ffee6907baf/controllers/dpa_controller.go#L95

and then here, instead of just using the validated data, I would validate it again...
https://github.com/openshift/oadp-operator/blob/91e1aacf47ea2b44bafdb9d6d3c14ffee6907baf/controllers/dpa_controller.go#L97

controllers/bsl.go
mateusoliveira43: But `getSecretNameandKeyFromCloudStorage` function https://github.com/openshift/oadp-operator/pull/1326/files?diff=split&w=0#diff-14341d832b55f374f353d693c68774d04617876f76c0de474e91763d14594614R246 

only runs `verifySecretContent` function https://github.com/openshift/oadp-operator/pull/1326/files?diff=split&w=0#diff-14341d832b55f374f353d693c68774d04617876f76c0de474e91763d14594614R786 

it does not validate non empty name and key

controllers/registry_test.go
mateusoliveira43: saw the comment that it will be included later, but I believe this should not be a separate PR

controllers/bsl.go
stillalearner: @mateusoliveira43 thanks for the clarification! I understand the ensureSecretDataExists() function is being called in both ValidateBackupStorageLocations & ReconcileBackupStorageLocations.

So, my question is , can we have a scenario where only one function executes out of the 2, like on updating BSL directly? Or no?

@sseago @kaovilai @shubham-pampattiwar anyone can please clarify?

controllers/bsl.go
stillalearner: Agree, Thank you!

controllers/registry_test.go
stillalearner: Added.

controllers/registry.go
```diff
@@ -783,3 +790,15 @@ func (r *DPAReconciler) populateAzureRegistrySecret(bsl *velerov1.BackupStorageL
...
 	return nil
 }
+
+func (r *DPAReconciler) verifySecretContent(secretName string, secretKey string) error {
+	secret, err := r.getProviderSecret(secretName)
+	if err != nil {
+		return err
+	}
+	data, foundKey := secret.Data[secretKey]
+	if !foundKey || len(data) == 0 {
```
mateusoliveira43, line 800: @kaovilai this is the check you wanted for the bug you are working on Azure?

controllers/bsl.go
stillalearner: ok, lets discuss if it is fine to check the content twice.

controllers/registry.go
kaovilai: Not quite.. the azure bug from meeting for imagestream is related to decoded data at secret key, not the length of secret key itself.

controllers/bsl.go
kaovilai: Twice is fine as long as it's not too much duplicate code otherwise make it a function.

stillalearner: not much duplicate i believe

kaovilai (dismissed):

controllers/bsl.go
shubham-pampattiwar: There are 3 options:
- use a `skipVerify` flag in the function
- do not use a function
- make peace with extra verification

stillalearner: I am OK with either 1 or 3

stillalearner: what do you suggest @shubham-pampattiwar

controllers/registry.go
mateusoliveira43: alright, got it now 

so, solution would be adding a next step here, that would differ from azure, aws and gcp, and checking the content of `data` (not only its length)?

this can be a separate PR

kaovilai: > so, solution would be adding a next step here, that would differ from azure, aws and gcp, and checking the content of `data` (not only its length)?
> 
> this can be a separate PR

Tracked in https://issues.redhat.com/browse/OADP-3562

comment there as you see fit.

controllers/bsl.go
shubham-pampattiwar: I am good with either of the approaches I listed. This is not a big concern for me. I am always in favor of simple, readable and maintainable code.

weshayutin:
@shubham-pampattiwar can we take this patch over?

shubham-pampattiwar:
@stillalearner Can you please rebase the PR ? Thanks !

stillalearner:
> @stillalearner Can you please rebase the PR ? Thanks !

Done.

controllers/bsl.go
```diff
@@ -119,10 +119,16 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
 		}
 		// Add the following labels to the bsl secret,
 		//	 1. oadpApi.OadpOperatorLabel: "True"
-		// 	 2. dataprotectionapplication.name: <name>
-		// which in turn will be used in the label handler to trigger the reconciliation loop
+		// 	 2. <namespace>.dataprotectionapplication: <name>
```
mateusoliveira43, line 122: this line change should be reverted, it was changed here https://github.com/openshift/oadp-operator/commit/71ea5b686962954b20874d0e03bc4387a537c09f

stillalearner: done thanks!

controllers/registry_test.go
```diff
@@ -210,6 +210,51 @@ func TestDPAReconciler_getSecretNameAndKeyforBackupLocation(t *testing.T) {
...
+			}
+			if tt.wantProfile == "aws-provider-no-cred" {
+				tt.wantSecretKey = "cloud"
+				tt.wantSecretName = "cloud-credentials"
+			}
+
+			gotName, gotKey, _ := r.getSecretNameAndKey(tt.bsl.Velero.Config, tt.bsl.Velero.Credential, oadpv1alpha1.DefaultPlugin(tt.bsl.Velero.Provider))
+
+			if !reflect.DeepEqual(tt.wantSecretName, gotName) {
+				t.Errorf("expected registry container env var to be %#v, got %#v", tt.wantSecretName, gotName)
```
mateusoliveira43, line 240: message here could be something like `"expected secret name to be %#v, got %#v"`?

would also assert error is nil when calling `getSecretNameAndKey`

mateusoliveira43 (approved):
tested by running unit/integration tests :heavy_check_mark: 

I think we need a follow up to improve error mesages

example, if we create a VSL without secret, the error is `secrets "cloud-credentials" not found`. Some weeks ago I saw the team waste time on an error like this, because user defined a VSL in DPA (but was not using it)

changing error to `SnapshotLocation Secret 'cloud-credentials' not found` can be very good for both debugging and informing the user (also for cloudStorage, BSL, etc)

controllers/registry_test.go
stillalearner: Sure, will add a follow up pr to improve error messages. Thanks !

shubham-pampattiwar (approved):

weshayutin:
@stillalearner @kaovilai doesn't look like we landed this on 1.4?

stillalearner:
> @stillalearner @kaovilai doesn't look like we landed this on 1.4?

We did, right? -> https://github.com/openshift/oadp-operator/pull/1516

kaovilai:
yes. we did

PR #1274, opened 2024-01-17
OADP-3189: do not remove labels from OADP namespace

mateusoliveira43:
## Description

Instead of adding/patching PSA labels to OADP namespace, controller manager would remove all labels, and then add PSA labels to OADP namespace. This PR fixes this behavior.

## How to test

From master branch and this PR branch, create a namespace and check its labels
```sh
oc get namespace <name> --show-labels
```
Add some custom labels, for example
```sh
oc label ns <name> app.kubernetes.io/instance=in-cluster-oadp-operator
oc label ns <name> mateus=hi
oc label ns <name> day=today
```
Run `OADP_TEST_NAMESPACE=<name> make deploy-olm` and check the namespace's labels again. In master branch some labels should have been removed. In this PR branch, no labels should have been removed (and only `pod-security.kubernetes.io/...` labels should have been modified).

PR #1328, opened 2024-02-12
Merge existing namespace labels and privileged labels added by operator

MuneebAijaz:
Problem Statement:
We are using a multi tenancy operator which works by setting namespace labels. When we run OADP Operator inside one of the tenant namespaces, OADP Pods keep crashing with errors coming from tenancy operator that tenant label does not exist anymore, since the update OADP sends on namespaces removes existing labels and adds its own privileged labels. 

Solution:
The fix first gets the existing labels, merges privileged labels onto them. and then apply it onto namespaces.

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,6 +20,8 @@ import (
 	"context"
 	"flag"
 	"fmt"
+	"os"
+
 	"github.com/openshift/oadp-operator/pkg/common"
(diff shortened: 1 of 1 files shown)


MuneebAijaz:
closing in favor of https://github.com/openshift/oadp-operator/pull/1274. apologies

weshayutin:
@shubham-pampattiwar FYI

PR #1333, opened 2024-02-15
CI: Back up and restore a VM.

mrnold:
This pull request adds an actual backup and restore of a running VM. This includes some rework of the existing backup and restore test code, so that different parts of the sequence could be specialized for VM purposes. The result is a little bit awkward, but it makes almost no functional changes to existing tests.

### DIFF
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -171,16 +171,20 @@ func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expecte
 	gomega.Eventually(lib.AreApplicationPodsRunning(kubernetesClientForSuiteRun, brCase.Namespace), timeoutMultiplier*time.Minute*9, time.Second*5).Should(gomega.BeTrue())
 
 	// Run optional custom verification
-	log.Printf("Running post-restore function for case %s", brCase.Name)
-	err = brCase.PostRestoreVerify(dpaCR.Client, brCase.Namespace)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	if brCase.PostRestoreVerify != nil {
+		log.Printf("Running post-restore function for case %s", brCase.Name)
+		err = brCase.PostRestoreVerify(dpaCR.Client, brCase.Namespace)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}
 }
 
 func runBackup(brCase BackupRestoreCase, backupName string) bool {
 	// Run optional custom verification
-	log.Printf("Running pre-backup function for case %s", brCase.Name)
-	err := brCase.PreBackupVerify(dpaCR.Client, brCase.Namespace)
-	gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	if brCase.PreBackupVerify != nil {
+		log.Printf("Running pre-backup function for case %s", brCase.Name)
+		err := brCase.PreBackupVerify(dpaCR.Client, brCase.Namespace)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+	}
 
 	nsRequiresResticDCWorkaround, err := lib.NamespaceRequiresResticDCWorkaround(dpaCR.Client, brCase.Namespace)
 	gomega.Expect(err).ToNot(gomega.HaveOccurred())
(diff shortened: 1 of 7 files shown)


kaovilai:
/hold
> re:Do not merge this yet

mrnold:
Okay, the VM backup and restore test has worked successfully at least once. This is at least worth having @mateusoliveira43 and @mpryc take a look at it.

mrnold:
Looks like failures in "Mongo application DATAMOVER" instead of "MySQL application two Vol CSI" this time, maybe one more retest for entertainment.

mateusoliveira43:
@weshayutin this resolves https://issues.redhat.com/browse/OADP-2982 ?

weshayutin:
@mateusoliveira43 it only partially resolves OADP-2982.    I'll review how to associate a pr w/ a jira again w/ @mrnold .  It's a just a jira task and for CI.

tests/e2e/lib/backup.go
```diff
@@ -24,6 +25,9 @@ func CreateBackupForNamespaces(ocClient client.Client, veleroNamespace, backupNa
 			SnapshotMoveData:         &snapshotMoveData,
 		},
 	}
+	if snapshotVolumes {
+		backup.Spec.SnapshotVolumes = pointer.Bool(true)
```
mateusoliveira43, line 29: maintainers, I never used this flag. This is needed if the PV is not in the same namespace as one of in `IncludedNamespaces`?

tests/e2e/lib/restore.go
```diff
@@ -22,6 +23,9 @@ func CreateRestoreFromBackup(ocClient client.Client, veleroNamespace, backupName
 			BackupName: backupName,
 		},
 	}
+	if restorePVs {
+		restore.Spec.RestorePVs = pointer.Bool(true)
```
mateusoliveira43, line 27: maintainers, what this option does with the restore?

tests/e2e/backup_restore_suite_test.go
mateusoliveira43, line 1: I would suggest breaking this PR in 2

- add backup/restore virt
- remove duplications from backup/restore and must-gather tests

tests/e2e/lib/backup.go
sseago: PVs are cluster-scoped and are not in any namespace. PVCs that are not in an included namespace will be excluded from backup.
This flag tells Velero to snapshot PVCs that are 1) included in the backup, and 2) not backed up using fs-backup. If this is false, then non-fs-backup PVCs will not be snapshotted.

tests/e2e/lib/restore.go
sseago: If I'm remembering correctly, this the restore equivalent to snapshotVolumes -- @shubham-pampattiwar is this correct?
Assuming I'm not misremembering, this means, similar to above, if it's true then we restore the PVC content from snapshot if it is a volume that was backed up via snapshot.

shubham-pampattiwar: RestorePVs specifies whether to restore all included PVs from snapshot

shubham-pampattiwar: +1 @sseago

sseago: @shubham-pampattiwar It's confusing that the restore and backup flags aren't named similarly, but they are essentially equilalent flags, but one for backup, the other for restore.

sseago: > The snapshot can be cloud provider based or csi snapshot. Both are applicable.

Same with SnapshotVolumes. There was a fix a while back to make this flag do the same thing for CSI and native snapshots.

shubham-pampattiwar: yes since velero 1.12 I think, initially these 2 flags were only applicable for cloud provider based snapshots.

tests/e2e/backup_restore_suite_test.go
mrnold: Okay, I have started this process with #1345.

weshayutin (approved):
/LGTM

kaovilai:

> [ya.. the change is for a completely diff test](https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1709662028012249?thread_ts=1709661802.249529&cid=C0144ECKUJ0)

:)
ps. we've seen this flake b4

PR #1343, opened 2024-02-25
fix: clean up code around velero server flags CLI command

mateusoliveira43:
## Description

This PR should be fix duplication, missing fields and wrong descriptions around velero server flags code.

## What was done

I was checking for wrong descriptions in our CRDs and then came across the velero server flags code. I tried to remove duplication and add TODOs around missing fields. As discussed, that code should be automated: each time velero is updated in `go.mod`
- a check should be run to check if all fields from `velero help server` are present in DPA (maybe skip some, but add why in code as comments) 
- and if their descriptions are correct (with some auto corrections).

## How to test

Check if the removal of duplication around velero server flags code did not change the behavior of the args passed to velero container.

## TODO

This PR SHOULD NOT be merged. It does to many things. Break up in smaller ones.


api/v1alpha1/oadp_types.go
```diff
@@ -53,17 +45,214 @@ type CustomPlugin struct {
...
+	// DUPLICATE OF VeleroConfig.DefaultItemOperationTimeout
+
+	// How often (in nanoseconds) 'maintain' is run for backup repositories by default.
+	// +optional
+	RepoMaintenanceFrequency *time.Duration `json:"default-repo-maintain-frequency,omitempty"`
+
+	// --default-snapshot-move-data                          Move data by default for all snapshots supporting data movement.
+	// VeleroConfig.DefaultSnapshotMoveData
+
+	// TODO --default-volume-snapshot-locations mapStringString   List of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)
```
mateusoliveira43, line 109: Are any of this TODO fields important and should be added?

api/v1alpha1/oadp_types.go
```diff
@@ -326,16 +524,15 @@ func (veleroConfig *VeleroConfig) HasFeatureFlag(flag string) bool {
...
 
-func init() {
-	SchemeBuilder.Register(&DataProtectionApplication{}, &DataProtectionApplicationList{}, &CloudStorage{}, &CloudStorageList{})
-}
-
 // AutoCorrect is a collection of auto-correction functions for the DPA CR
 // These auto corrects are in-memory only and do not persist to the CR
 // There should not be another place where these auto-corrects are done
 func (dpa *DataProtectionApplication) AutoCorrect() {
+	// TODO error instead of changing user object?
```
mateusoliveira43, line 531: @shawn-hurley I think you once said we should not change values. Should we change this to be in validator or it is ok?

controllers/velero.go
```diff
@@ -177,25 +170,25 @@ func (r *DPAReconciler) buildVeleroDeployment(veleroDeployment *appsv1.Deploymen
...
 		// last label overrides previous ones
 		install.WithLabels(veleroDeployment.Labels),
 		// use WithSecret false even if we have secret because we use a different VolumeMounts and EnvVars
 		// see: https://github.com/vmware-tanzu/velero/blob/ed5809b7fc22f3661eeef10bdcb63f0d74472b76/pkg/install/deployment.go#L223-L261
 		// our secrets are appended to containers/volumeMounts in credentials.AppendPluginSpecificSpecs function
 		install.WithSecret(false),
 		install.WithServiceAccountName(common.Velero),
+		// TODO remove
+		install.WithFeatures(dpa.Spec.Configuration.Velero.FeatureFlags),
+		install.WithUploaderType(uploaderType),
```
mateusoliveira43, line 187: this flag is never passed to velero container, because we override it afterwards. Should it be added?

controllers/velero.go
```diff
@@ -313,12 +306,14 @@ func (r *DPAReconciler) customizeVeleroDeployment(dpa *oadpv1alpha1.DataProtecti
 	}
 
 	var veleroContainer *corev1.Container
+	// TODO did not understand this
```
mateusoliveira43, line 309: what this part does?

pkg/velero/server/args.go
```diff
@@ -4,156 +4,278 @@ import (
...
-		if err != nil {
-			return []string{}, fmt.Errorf("invalid log level %s, use: %s", logLevel, "trace, debug, info, warning, error, fatal, or panic")
+func getFsBackupTimeout(dpa *oadpv1alpha1.DataProtectionApplication) string {
+	if dpa.Spec.Configuration.Restic != nil && len(dpa.Spec.Configuration.Restic.Timeout) > 0 {
+		fsBackupTimeout, err := time.ParseDuration(dpa.Spec.Configuration.Restic.Timeout)
+		if err == nil {
+			return fsBackupTimeout.String()
 		}
-		args = append(args, fmt.Sprintf("--log-level=%s", logrusLevel.String()))
+		// TODO should not error out?
```
mateusoliveira43, line 65: if user puts wrong string in restic or node agent timeout, code uses default and does not tell nothing to user

pkg/velero/server/args.go
mateusoliveira43, line 1: why do we have duplication of fields around velero server flags? some in VeleroConfig, some in VeleroConfig.Args

pkg/common/common.go
```diff
@@ -18,43 +17,14 @@ const (
...
-	},
-	LowPriorities: []string{
-		"clusterbootstraps.run.tanzu.vmware.com",
-		"clusters.cluster.x-k8s.io",
-		"clusterresourcesets.addons.cluster.x-k8s.io",
-	},
-}
+// Labels
+const (
+	OadpOperatorLabel       = "openshift.io/oadp"
```
kaovilai, line 25: Can you move this outside package common? I think this would make importing github/oadp/pkg/common add a lot of indirect imports for QE if they are using this.

Could be pkg/common/labels

kaovilai: api package via _types.go was nice cause imports there is minimal

pkg/velero/server/args.go
kaovilai: See [discussion for context](https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1709062774155549?thread_ts=1709049938.887259&cid=C0144ECKUJ0)

weshayutin:
@mateusoliveira43 I like the thought, but this will have to wait for 1.4.  
I'm going to place a hold so that we're not confused with regards to what we're working on atm.

pkg/common/common.go
mateusoliveira43: I think it has more imports than this file, but creating something like `Common/consts.go` (and `Common/functions.go`) would make this file have no imports

pkg/velero/server/args.go
```diff
@@ -4,156 +4,278 @@ import (
...
+	if len(dpa.Spec.Configuration.Velero.FeatureFlags) > 0 {
+		args = append(args, fmt.Sprintf("--features=%s", strings.Join(dpa.Spec.Configuration.Velero.FeatureFlags, ",")))
 	}
-	// features set outside Args
-	// args = append(args, "--kubeconfig")        // string
-	// args = append(args, "--kubecontext")       // string
-	if a.TraceLocation != "" {
-		args = append(args, fmt.Sprintf("--log_backtrace_at=%s", a.TraceLocation)) // traceLocation
+	// TODO --kubeconfig
+	// TODO --kubecontext
```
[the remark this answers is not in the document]
mateusoliveira43: comment was it was set elsewhere, where is it set?

pkg/common/common.go
kaovilai: imports are based on package level too.. so it has to be a different subdir

kaovilai: common/consts/labels.go

pkg/velero/server/args.go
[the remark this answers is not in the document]
kaovilai: There's no need to set these as velero will responds to current cluster and current context is the velero namespace. There's no point specifying these

mateusoliveira43:
/remove-lifecycle stale
discuss in scrum

PR #1350, opened 2024-03-04
feat: Add PR template to project

mateusoliveira43:
## Why the changes were made

Help the project review process by:
- adding a why section, so we understand why that PR is important and can help us prioritizing our work
- adding a how section, so we can understand the thought process behind a PR
- adding a testing section, so we do not only read the proposed code changes, but also test the proposed changes prior to approving a PR

Help the project quality by:
- standardizing PR's descriptions
- documenting processes 

## How the changes were made

Searched for some templates in the internet to help create the one proposed in this PR.

Read the proposed template and make sure it is clear and helps achieve this PR's goal.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,11 @@
+## Why the changes were made
+
+<!-- Explain why this PR is important, what problems it fixes, link related issues -->
+
+## How the changes were made
+
+<!-- Explain how was the thought process behind this PR, what commands were run to achieve the changes -->
+
+## How to test the changes made
+
+<!-- Explain how the changes introduced by this PR can be tested and verified, with commands and pictures -->
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -2,10 +2,6 @@
 
 <!-- Explain why this PR is important, what problems it fixes, link related issues -->
 
-## How the changes were made
-
-<!-- Explain how was the thought process behind this PR, what commands were run to achieve the changes -->
-
 ## How to test the changes made
 
 <!-- Explain how the changes introduced by this PR can be tested and verified, with commands and pictures -->


.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
+## Why the changes were made
+
+<!-- Explain why this PR is important, what problems it fixes, link related issues -->
+
+## How the changes were made
```
weshayutin, line 5: Personally I don't think this is a clear question..

The upstream as the second question has: "Does your change fix a particular issue?" which is NOT required if the PR relates to a jira issue as it often does.   Perhaps we can repurpose to say...
"If there is no Jira, what issue is being fixed with the PR" or something like that.

.github/pull_request_template.md
```diff
@@ -0,0 +1,11 @@
...
+
+<!-- Explain how was the thought process behind this PR, what commands were run to achieve the changes -->
+
+## How to test the changes made
```
weshayutin, line 9: ++ I like it :)

.github/pull_request_template.md
kaovilai: "Link to origin of request", slack, jira, github, gitlab, reddit, google meet, google doc.

mateusoliveira43: Here I do not want to know about issues (WHY), added issue part to the first section (we can add more description to it)

Here I want to know HOW, what was done to achieve the goal, which commands were run, what files were modified, what possibilities were considered, etc

kaovilai: I think this feedback would be for the why section :)

weshayutin:
to be most effective @kaovilai and @mateusoliveira43 IMHO we should align our template as closely to https://github.com/vmware-tanzu/velero/blob/main/.github/pull_request_template.md as possible. Over time the two templates should converge to be exactly the same ideally.

sseago:
@weshayutin they won't be exactly the same, but the differences should coincide with actual differences in development model. For example, we don't have the DCO signoff or changelog files per PR.

weshayutin:
@sseago 100% agree :)

.github/pull_request_template.md
rayfordj: **Big** **+1** for 'how to test...'

.github/pull_request_template.md
mateusoliveira43: removed

kaovilai (approved):

weshayutin (approved):
/LGTM

PR #1264, opened 2024-01-08
Fix OADP-3228: Controller Manager lease duration too short for SNO

shubham-pampattiwar:
This PR Updates the controller manager lease based on cluster topology

PR #1354, opened 2024-03-06
fix: E2E subscription test

mateusoliveira43:
## Why the changes were made

To fix CI errors around subscription E2E tests. They were probably introduced by https://github.com/openshift/oadp-operator/pull/1264

## How the changes were made

Run the subscription E2E tests locally to reproduce the error and studied what was happening. It was noticed controller-manager pod would wait around 3 minutes to acquired lease and run normally, which was erroring out subscription tests timeouts.

Tests were updated to wait controller-manager pod to acquired lease.

Run E2E tests locally, focusing on subscription tests. They should not fail. Also, read the new logs and confirm they are informative and clear.

### DIFF
--- a/tests/e2e/subscription_suite_test.go
+++ b/tests/e2e/subscription_suite_test.go
@@ -3,18 +3,20 @@ package e2e_test
 import (
 	"context"
 	"log"
+	"strings"
 	"time"
 
 	ginkgov2 "github.com/onsi/ginkgo/v2"
 	"github.com/onsi/gomega"
 	operators "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
 
 	"github.com/openshift/oadp-operator/tests/e2e/lib"
 )
 
-var _ = ginkgov2.Describe("Subscription Config Suite Test", func() {
+var _ = ginkgov2.FDescribe("Subscription Config Suite Test", func() {
 	type SubscriptionConfigTestCase struct {
 		operators.SubscriptionConfig
 		failureExpected *bool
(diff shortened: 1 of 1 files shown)


tests/e2e/subscription_suite_test.go
```diff
@@ -3,18 +3,20 @@ package e2e_test
...
 	operators "github.com/operator-framework/api/pkg/operators/v1alpha1"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/utils/pointer"
 
 	"github.com/openshift/oadp-operator/tests/e2e/lib"
 )
 
-var _ = ginkgov2.Describe("Subscription Config Suite Test", func() {
+var _ = ginkgov2.FDescribe("Subscription Config Suite Test", func() {
```
mateusoliveira43, line 19: remove prior to merge

tests/e2e/subscription_suite_test.go
```diff
@@ -39,16 +41,57 @@ var _ = ginkgov2.Describe("Subscription Config Suite Test", func() {
...
+						return false
+					}
+					log.Printf("waiting leaderelection")
+					return strings.Contains(podLogs, "leaderelection.go:258] successfully acquired lease")
+				}
+				gomega.Eventually(isLeaseReady, time.Minute*4, time.Second*5).Should(gomega.BeTrue())
 
 				log.Printf("Creating test Velero")
-				err := dpaCR.Build(lib.CSI)
+				err = dpaCR.Build(lib.CSI)
```
kaovilai, line 94: Why are we writing to err outside the else block?
What does this change do?

mateusoliveira43: I think in my early implementations I was creating error before that line, then removed here

it could be reverted, but if harmless, would let it be to not push again

kaovilai: harmless enough, nothing else reading it after the block atm.

kaovilai (approved):

mateusoliveira43:
/unhold
no errors related to subscription tests
/retest

mateusoliveira43:
no errors related to subscription tests

PR #1357, opened 2024-03-07
DO NOT MERGE, Test CI

weshayutin:


mateusoliveira43:
@weshayutin can we close this?
AWS failure was the flake after restore (restore succeeds, but things do not work after, which I tried to attack since last year); AZURE was just infra, failed to create cluster

weshayutin:
@mateusoliveira43 I will close this PR, but not quite yet :) I will use this to access the current state of ci

mateusoliveira43:
@weshayutin is not better to keep track of periodic jobs instead of a PR?

kaovilai:
periodic jobs run once per day. this ci job wesley can trigger whenever as he please.

PR #1359, opened 2024-03-12
OADP-2982: Test VMs from YAML templates

mrnold:
Allow creation of disks and VMs from YAML files in the E2E sample-applications directory, instead of directly from code. This does not change the initial CirrOS download, because of the code that does active retrieval of the latest version. But the test VM and its disk are now created from YAML.

### DIFF
--- a/tests/e2e/lib/virt_helpers.go
+++ b/tests/e2e/lib/virt_helpers.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"log"
+	"path/filepath"
 	"strings"
 	"time"
 
@@ -601,58 +602,6 @@ func (v *VirtOperator) checkVmStatus(namespace, name, expectedStatus string) boo
 	return status == expectedStatus
 }
 
-func (v *VirtOperator) createVm(namespace, name, source string) error {
-	unstructuredVm := unstructured.Unstructured{
-		Object: map[string]interface{}{
-			"apiVersion": "kubevirt.io/v1",
-			"kind":       "VirtualMachine",
-			"metadata": map[string]interface{}{
-				"name":      name,
-				"namespace": namespace,
-			},
-			"spec": map[string]interface{}{
-				"running": true,
-				"template": map[string]interface{}{
-					"spec": map[string]interface{}{
-						"domain": map[string]interface{}{
-							"devices": map[string]interface{}{
-								"disks": []map[string]interface{}{
-									{
-										"disk": map[string]interface{}{
-											"bus": "virtio",
-										},
-										"name": "rootdisk",
-									},
-								},
-							},
-							"resources": map[string]interface{}{
-								"requests": map[string]interface{}{
-									"cpu":    "1",
-									"memory": "256Mi",
-								},
-							},
-						},
-						"volumes": []map[string]interface{}{
-							{
-								"persistentVolumeClaim": map[string]interface{}{
-									"claimName": name,
-								},
-								"name": "rootdisk",
-							},
-						},
-					},
-				},
-			},
-		},
-	}
-
-	if _, err := v.Dynamic.Resource(virtualMachineGvr).Namespace(namespace).Create(context.TODO(), &unstructuredVm, metav1.CreateOptions{}); err != nil {
-		return fmt.Errorf("error creating VM %s/%s: %w", namespace, name, err)
-	}
-
-	return nil
-}
-
 func (v *VirtOperator) removeVm(namespace, name string) error {
 	if err := v.Dynamic.Resource(virtualMachineGvr).Namespace(namespace).Delete(context.TODO(), name, metav1.DeleteOptions{}); err != nil {
 		if !apierrors.IsNotFound(err) {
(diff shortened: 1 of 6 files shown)


weshayutin:
nice! thanks @mrnold

tests/e2e/lib/virt_storage_helpers.go
```diff
@@ -323,3 +324,27 @@ func (v *VirtOperator) CloneDisk(sourceNamespace, sourceName, cloneNamespace, cl
...
 	return nil
 }
+
+// Create a DataVolume from a sample YAML template. The namespace argument
+// should match a subdirectory under sample-applications/virtual-machines, and
+// the name argument should match a .yaml file in that directory, for example:
+//
+//	sample-applications/virtual-machines/cirros-test/cirros-test-disk.yaml
+//
+// This file should create a DataVolume with the following annotations set:
```
mateusoliveira43, line 334: since this specified in yaml, I would remove from doc. Afraid we update yaml, but not the doc in the future

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-disk.yaml
mateusoliveira43, line 1: NIT both yaml just install one thing, right? no need to be list?

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -44,32 +44,49 @@ type VmBackupRestoreCase struct {
...
 	backupName, restoreName := prepareBackupAndRestore(brCase.BackupRestoreCase, func() {})
 
 	err := lib.CreateNamespace(v.Clientset, brCase.Namespace)
 	gomega.Expect(err).To(gomega.BeNil())
 
-	// Create VM from clone of CirrOS image
-	err = v.CloneDisk(brCase.SourceNamespace, brCase.Source, brCase.Namespace, brCase.Name, 5*time.Minute)
+	// Create a standalone VM disk. This defaults to cloning the CirrOS image.
+	// This uses a DataVolume so import and clone progress can be monitored.
+	// Behavioral notes: CDI will garbage collect DataVolumes if they are not
```
mateusoliveira43, line 58: maybe better to add this notes direct to yaml?

tests/e2e/lib/virt_helpers.go
```diff
@@ -601,58 +602,6 @@ func (v *VirtOperator) checkVmStatus(namespace, name, expectedStatus string) boo
 	return status == expectedStatus
 }
 
-func (v *VirtOperator) createVm(namespace, name, source string) error {
```
mateusoliveira43, line 604: the disk in code yaml can also be deleted?

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-disk.yaml
mrnold: That's right, the only reason they are lists is because InstallApplication expects a list.

mateusoliveira43: resolving then !

tests/e2e/lib/virt_helpers.go
mrnold: Do you mean the DataVolume creation code in createDataVolumeFromSource? It is still used for downloading the CirrOS image. I haven't found a good way to combine fixed YAML with the code that gets the latest CirrOS version automatically.

mateusoliveira43: I see, disk yaml is only from pvc part. still missing a from url part, right?

mateusoliveira43: since it is optimization, I would just create an issue and continue this PR (when is ready to merge, I suggest rebasing with master, it gets merged way quicker)

mrnold: Yes, I didn't add a YAML for a disk with a URL source, because the CirrOS URL can change. We could go back to updating it manually if it would be better to have that in YAML too. But right now there's a getLatestCirrosImageURL that checks for the latest URL, and I don't see a nice way to get the resulting URL into YAML without adding a bunch of code for templating or string substitution or something like that.

tests/e2e/lib/virt_storage_helpers.go
mrnold: Okay, I changed the explicit reference to cirros-test.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: If there are multiple tests, it will be awkward to have this documentation in every one. The test code is doing a procedure that makes specific demands on the YAML, I think it makes more sense to document this procedure in that code, or maybe in a separate file under the docs directory.

I added a comment in the cirros-test-disk YAML, but I think the code comment should also still stay.

mateusoliveira43 (approved):
/lgtm
The kubevirt test passed, so everything looking good with the change

Will follow up with adding a template replacer function to latest url

weshayutin (approved):
/LGTM

mateusoliveira43:
/override ci/prow/4.12-e2e-test-azure
restore succeeds, but application work fully working flake

PR #1364, opened 2024-03-17
fix: More YAML templates for E2E virt tests

mateusoliveira43:
## Why the changes were made

Follow up for https://github.com/openshift/oadp-operator/pull/1359 to add more YAML templates.

~Blocked by https://github.com/openshift/oadp-operator/pull/1386~

## How the changes were made

Studied code to find and remove "YAML literals" from it. Added feature to replace strings in code in templates.

If you have 4.14+ cluster, you can run E2E virt tests locally with `make TEST_VIRT=true test-e2e`

Check logs, they should not be less informative and readable as they were before this change.

### DIFF
--- a/tests/e2e/lib/apps.go
+++ b/tests/e2e/lib/apps.go
@@ -10,6 +10,7 @@ import (
 	"os/exec"
 	"reflect"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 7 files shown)


mrnold (commented):
I like having this code simplified into YAML templates, but I know @weshayutin is really in favor of having plain YAML files. I am wondering if a good compromise is to move the infrastructure-related templates (the HCO installation, KVM emulation, and maybe a new one for the latest CirrOS image fetching) out of sample-applications into a separate directory under tests/e2e, and renamed to something like *.yaml.template. Then the actual VMs themselves can be plain .yaml files under sample-applications like they are now, with no templating.

Otherwise, @mpryc had an interesting idea about parsing the sample YAML into actual CRD structs and replacing the relevant fields That way, there could be fixed YAML files in the current directory, and each test could decide to override any fields as needed. Maybe with a command line argument or environment variable to enable or disable this replacement, to force the test to respect any manually-added YAML changes.

Do either of these ideas sound useful?

weshayutin:
I would say the following..  How the templates are constructed is NOT a big deal to me unless the templates are used in documentation in oadp-operator/docs.   I am happy with working templates and we shouldn't bike shed too much about it. 

YAML templates are fine, we should have a little shell script that can populate and  render the templates.  Breaking out files into directories that make more sense is also fine.

Re the sample apps in our tests/e2e 
1. should be good enough for our CI 
2. a user should be able to manually use the sample apps with simple instructions
3. At the first pass of this work does NOT need to be perfect for all users.  We'll get there

tests/e2e/lib/apps.go
```diff
@@ -45,17 +47,40 @@ var (
...
 
-func InstallApplicationWithRetries(ocClient client.Client, file string, retries int) error {
+func InstallApplicationWithRetries(ocClient client.Client, file string, retries int, replace ...string) error {
 	template, err := os.ReadFile(file)
 	if err != nil {
 		return err
 	}
 	obj := &unstructured.UnstructuredList{}
 
+	// YAML templates can have replace directives, in the form of "<<template-replace-N>>", where N
```
kaovilai, line 61: Is this oadp specific directive or we're following some published standard?

mateusoliveira43: just from my head :rofl: 

tried to think what would be good in the future using this

I accept suggestions

mateusoliveira43:
@mrnold did not understand how would be @mpryc idea

we can document yes @weshayutin . This can be very useful for me in the future. Today (in IBM) I need to manually edit all snapclasses (because I do not use ceph), for example. This could be automated with templates (and not change how other people run `make test-e2e`

But seems we have some discussion, so we can focus on working here after 1.3.1 (unless @mrnold wants anything from here now)

mrnold (approved):
This is fine with me as-is, then.

weshayutin:
@mateusoliveira43 thank you for putting up this change, it fell off my radar tbh.   I have strong opinions about this that should be tempered by a group discussion.   I'll raise it in thrs scrum for discussion.   I'd like Matthew to review in depth but he is focused on dataupload/datadown changes, this will have to wait a bit.

mateusoliveira43:
Created issue to discuss solution https://github.com/openshift/oadp-operator/issues/1531

PR #1365, opened 2024-03-18
update mysql/mongo/other for 4.15 error on restartPolicy

weshayutin:
This seems to be a OCP 4.15 issue only.  The restartPolicy is not required on any version afaict.

Error:
`The Deployment "mysql" is invalid:
spec.template.spec.containers[0].restartPolicy:
Forbidden: may not be set for non-init containers`

### DIFF
--- a/tests/e2e/sample-applications/deprecated/parks-app/manifest.yaml
+++ b/tests/e2e/sample-applications/deprecated/parks-app/manifest.yaml
@@ -197,7 +197,6 @@ items:
           - name: mongodb-data
             persistentVolumeClaim:
               claimName: mongodb-data-claim
-          restartPolicy: Always
     
   - kind: Service
     apiVersion: v1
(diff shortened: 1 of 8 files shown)


sseago (dismissed):

sseago:
spec.container.restartPolicy is new as of kube 1.28/OCP4.15 and only valid for init containers. It didn't exist at all prior to 1.28, so removing it from the mysql and mongo yaml files is necessary. The parks-app removal is a different field (spec.restartPolicy) which should be valid across all versions, but it's fine to remove if the test doesn't require it to remove any confusion it might cause.

mateusoliveira43:
@sseago `deprecated` folder is not used by our tests, so tests will not care

trying to remove it in https://github.com/openshift/oadp-operator/pull/1283

mateusoliveira43 (approved):

sseago (approved):

PR #1382, opened 2024-04-05
fix: makefile temp logic

mateusoliveira43:
## Why the changes were made

Makefile would create 4 temporary directories per call.

Makefile would not clean up temporary directory if command failed.

Makefile would not check if temporary directory is empty.

## How the changes were made

Fixed explained problems.

Run Makefile commands. All created temporary directories should be deleted

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -188,17 +188,13 @@ test: vet envtest ## Run unit tests; run Go linters checks; and check if api and
 	@make bundle-isupdated
 
 .PHONY: api-isupdated
-api-isupdated: TEMP:= $(shell mktemp -d)
 api-isupdated:
-	@cp -r ./ $(TEMP) && cd $(TEMP) && make generate && cd - && diff -ruN api/ $(TEMP)/api/ && echo "api is up to date" || (echo "api is out of date, run 'make generate' to update" && exit 1)
-	@chmod -R 777 $(TEMP) && rm -rf $(TEMP)
+	$(call run-in-temp,make generate,compare)
 
 .PHONY: bundle-isupdated
-bundle-isupdated: TEMP:= $(shell mktemp -d)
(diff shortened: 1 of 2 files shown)


mateusoliveira43:
@mpryc review

hack/run_in_temp.go
```diff
@@ -0,0 +1,59 @@
...
+	log.Printf("directory to copy into temporary directory: %v", pwd)
+	copy := exec.Command("cp", "--recursive", pwd, tempDir)
+	err = copy.Run()
+	if err != nil {
+		log.Print(err.Error())
+		return 1
+	}
+
+	// cmd := exec.Command("make", "docker-build", "IMG=ttl.sh/oadp-operator-mateus-123456:1h")
+	cmd := exec.Command("make", "help")
```
mateusoliveira43, line 46: tried to create a go script, but could not call make inside it :exploding_head: 
going to delete this prior to merge

Makefile
```diff
@@ -544,3 +535,19 @@ lint: golangci-lint ## Run Go linters checks against all project's Go files.
 .PHONY: lint-fix
 lint-fix: golangci-lint ## Fix Go linters issues.
 	$(GOLANGCI_LINT) run --fix
+
+define run-in-temp
+@{ \
+set -e ;\
+TEMP=$$(mktemp -d) ;\
+echo $$TEMP ;\
```
mateusoliveira43, line 543: remove echo

mpryc:
I *really* think that using go script to run shell commands is not the right implementation:

Shell -> build go -> spawn new go process -> execute `commands` in Shell

That's a bit too much. We should use Shell that invokes `commands` directly or the built-in shell equivalents.

mpryc:
Here is implementation using shell script (not yet integrated within Makefile).

Please comment which way to proceed (go vs shell):

https://github.com/openshift/oadp-operator/commit/7ed10e5d1e9f3eff8a7021e57906e8027390a78e#diff-8de5118a56011d553059dd47da58cea16ecd90598ef82d0d64bcba3ce9cc0423

mateusoliveira43:
closing in favor of other PR

PR #1386, opened 2024-04-12
CI: Convert CirrOS test to YAML. 

mrnold:
Change CirrOS test to run from a YAML template, and remove any related code that is no longer needed.


tests/e2e/lib/virt_storage_helpers.go
```diff
@@ -21,57 +20,10 @@ var dataVolumeGVK = schema.GroupVersionResource{
...
-	}
-
-	for _, owner := range pvc.OwnerReferences {
-		if owner.Kind == "DataVolume" {
-			return true
-		}
-	}
-
-	return false
+var dataSourceGVK = schema.GroupVersionResource{
```
mateusoliveira43, line 23: here should be GVR? (Resource instead of Kind) (this may also need fixing in `virt_helpers.go` for standardization)

mrnold: Done.

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -189,8 +150,8 @@ var _ = ginkgov2.Describe("VM backup and restore tests", ginkgov2.Ordered, func(
 		},
 
 		ginkgov2.Entry("no-application CSI datamover backup and restore, CirrOS VM", ginkgov2.Label("virt"), VmBackupRestoreCase{
-			Source:          "cirros-dv",
-			SourceNamespace: "openshift-cnv",
+			Template:  "./sample-applications/virtual-machines/cirros-test/cirros-test.yaml",
+			InitDelay: 2 * time.Minute,
```
mateusoliveira43, line 154: why we need delay now with new approach?

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -165,14 +123,17 @@ var _ = ginkgov2.Describe("VM backup and restore tests", ginkgov2.Ordered, func(
 
 		url, err := getLatestCirrosImageURL()
 		gomega.Expect(err).To(gomega.BeNil())
-		err = v.EnsureDataVolumeFromUrl("openshift-cnv", "cirros-dv", url, "128Mi", 5*time.Minute)
+		err = v.EnsureDataVolumeFromUrl("openshift-virtualization-os-images", "cirros", url, "128Mi", 5*time.Minute)
+		gomega.Expect(err).To(gomega.BeNil())
+		err = v.CreateDataSourceFromPvc("openshift-virtualization-os-images", "cirros")
```
mateusoliveira43, line 128: this new way, CirrOS VM appears in cluster like a defualt one? (fedora, centos, etc)

tests/e2e/virt_backup_restore_suite_test.go
mrnold: The VM gets marked running while kernel boot messages are still scrolling on the console, and I would rather start the backup when it's sitting idle like at the login screen. I don't like having to sleep to wait for the VM to get ready, but I don't have a better solution ready yet.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Yes, I figured out how to use the DataSource CRD, and now I get a "cirros" template alongside the others in the openshift-virtualization-os-images namespace. I was previously seeing permissions errors trying to clone the downloaded PVC, but having a real DataSource avoids that problem.

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43: fair enough, a comment about this is enough for me for now

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43: niceeeee

mateusoliveira43 (approved):
code looks good, did not test, will wait CI do its job

shubham-pampattiwar (approved):

PR #1391, opened 2024-05-02
fix: (different approach) makefile temp logic

mpryc:
This is PR similar to https://github.com/openshift/oadp-operator/pull/1382, however reimplemented using shell script.

Note I am unable to create commits to 1382, that's why opening new PR so developers can comment which way to go.

*Note* This is not 100% finished, because it lacks the diff comparison `compare` part, but it can be easily added as part of the shell script and not Makefile if we agree on this implementation.

The logic is much simpler, because Makefile is not the right place to include functions from the scripts.


hack/run_in_temp.sh
```diff
@@ -0,0 +1,71 @@
...
+# Create top level directory where subdirs will be stored
+if [ ! -d "${DEFAULT_TMP_DIR}" ]; then
+    mkdir -p "${DEFAULT_TMP_DIR}" || error_exit "Failed to create temporary top level directory ${DEFAULT_TMP_DIR}"
+fi
+
+# We need to get the git hash from the directory that was used to run this script
+# so we get the hash of the oadp-operator source code
+pushd "$ABS_OADP_GIT_DIR"
+  CURRENT_GIT_HASH=$(git rev-parse HEAD)
+  COMMAND_TEMP_DIR=$(TMPDIR="${DEFAULT_TMP_DIR}" mktemp -d -t "${OADP_TMP_DIR_PREFIX}XXXX-${CURRENT_GIT_HASH}") || exit 1
```
mateusoliveira43, line 59: I think `-t` is deprecated by this command

mpryc: I would not worry about that. It's handy and works on MacOS as well. The deprecation help was way over 17 years ago, so I don't think it will be removed any time.. 

https://github.com/coreutils/coreutils/commit/99b039fead8d72936d0d558198235eee03016904#diff-b034738a99620fcc3b4df8d87d50191f19cb37d848e9ee95e401fd32162b1921R93-R96

hack/run_in_temp.sh
```diff
@@ -0,0 +1,71 @@
...
+          if [[ "${OADP_TMP_DIR}" =~ "${OADP_TMP_DIR_PREFIX}"* ]]; then
+              echo "Cleaning up temporary OADP files"
+              rm -rf "${COMMAND_TEMP_DIR}"
+              rmdir "${DEFAULT_TMP_DIR}" || exit 1
+          fi
+      fi
+    fi
+
+    # Propagate exit value if was provided
+    [ -n "${exit_val}" ] && echo "Exit code: ${exit_val}" && exit "$exit_val"
```
mrnold, line 43: Any reason not to put curly braces around the third instance of exit_val? It looks like they were very carefully used on everything else.

hack/run_in_temp.sh
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash
```
weshayutin, line 1: Lev would like to know why you have a HACK in this code.

hack/run_in_temp.sh
mpryc: No reason, small mistake which does not have any implications. The `exit_val=$1` makes it a single argument anyway so there is no difference between "$exit_val" and "${exit_val}"

hack/run_in_temp.sh
mpryc: The `hack` folder does already exists in the oadp-operator codebase:
   https://github.com/openshift/oadp-operator/tree/master/hack
   
And it seems natural to put scripts within one folder that may have any name.

However in operator-sdk go-based operators it's a common to use hack/ folder for utility scripts as described in the:
https://sdk.operatorframework.io/docs/overview/project-layout/#golang

We use an absolute path of the place where the script resides, but we need to go one level up to be able to copy/rsync folder structure of the main oadp-operator, that is why we have on line number 4 comment about that.

mpryc:
@mateusoliveira43 is this PR still something we want to address (I can work on rebase) or do we want to go in differeht path ?

mateusoliveira43:
yes, we need this to move forward with go mod tidy CI job

but I do not understand why we have a folder in gitignore if we are using temporary folders

mpryc:
> yes, we need this to move forward with go mod tidy CI job
> 
> but I do not understand why we have a folder in gitignore if we are using temporary folders

In cases things goes wrong the temporary folder may be preserved for debug, and the last thing we want to do is to commit that:

https://github.com/openshift/oadp-operator/pull/1391/files#diff-8de5118a56011d553059dd47da58cea16ecd90598ef82d0d64bcba3ce9cc0423R11

hack/run_in_temp.sh
weshayutin: @levhayutin FYI.. thank you @mpryc :)

PR #1392, opened 2024-05-03
Fix OADP-3960: Split labels avoiding label key exceed allowed 63 chars

muellerfabi:
This PR should fix https://issues.redhat.com/browse/OADP-3960 by adding another label containing the namespace, while removing the namespace from the label key. 

Summary: Right now the operator labels the "cloud-credentials"-secret like `<namespace>.dataprotectionapplication=<name>`. If the operator is deployed in a namespace with more than 37 characters, the operation will fail, as the label key exceeds the allowed 63 char length limit.
- Both namespace-name and label-key may contain up to 63 chars [1].
- ".dataprotectionapplication" spends 26 chars.

[1] https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#syntax-and-character-set

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -211,8 +211,12 @@ func (r *DPAReconciler) UpdateCredentialsSecretLabels(secretName string, namespa
 		secret.Labels[oadpv1alpha1.OadpOperatorLabel] = "True"
 		needPatch = true
 	}
-	if secret.Labels[namespace+".dataprotectionapplication"] != dpaName {
-		secret.Labels[namespace+".dataprotectionapplication"] = dpaName
+	if secret.Labels["dataprotectionapplication.name"] != dpaName {
+		secret.Labels["dataprotectionapplication.name"] = dpaName
+		needPatch = true
+	}
+	if secret.Labels["dataprotectionapplication.namespace"] != namespace {
+		secret.Labels["dataprotectionapplication.namespace"] = namespace
 		needPatch = true
 	}
 	if needPatch {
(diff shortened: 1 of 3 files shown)


controllers/bsl.go
```diff
@@ -211,8 +211,12 @@ func (r *DPAReconciler) UpdateCredentialsSecretLabels(secretName string, namespa
 		secret.Labels[oadpv1alpha1.OadpOperatorLabel] = "True"
 		needPatch = true
 	}
-	if secret.Labels[namespace+".dataprotectionapplication"] != dpaName {
-		secret.Labels[namespace+".dataprotectionapplication"] = dpaName
+	if secret.Labels["dataprotectionapplication.name"] != dpaName {
+		secret.Labels["dataprotectionapplication.name"] = dpaName
+		needPatch = true
+	}
+	if secret.Labels["dataprotectionapplication.namespace"] != namespace {
```
mateusoliveira43, line 218: do we need to store this info in the secret :question: is not the same namespace it lives in :question:

mateusoliveira43 (approved):
Thanks @muellerfabi 

I believe the changes are correct (I ran `grep -Inr '\.dataprotectionapplication' .` and did not find any other occurrences)

An improvement could be moving the label `dataprotectionapplication.name` to https://github.com/openshift/oadp-operator/blob/master/pkg/common/common.go, but can a follow up PR

kaovilai (approved):

kaovilai:
```
  Event: Unable to attach or mount volumes: unmounted volumes=[applog], unattached volumes=[kube-api-access-brbkh applog]: timed out waiting for the condition, Type: Warning, Count: 1, Src: {Pod mysql-persistent todolist-1-9hlmh f15e212e-cec2-487c-a2f0-43224044c350 v1 40682 }, Reason: FailedMount

```
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1392/pull-ci-openshift-oadp-operator-master-4.13-e2e-test-aws/1792617560445816832#1:build-log.txt%3A1362

/retest
/lgtm

shubham-pampattiwar (approved):

PR #1393, opened 2024-05-09
Handle NAC feature enablement via OADP for tech preview and GA scenarios

shubham-pampattiwar:
Fixes https://github.com/migtools/oadp-non-admin/issues/47 
Approach discussed: https://github.com/migtools/oadp-non-admin/issues/47#issuecomment-2064362353


controllers/validator.go
```diff
@@ -63,6 +64,17 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 	if _, err := getResticResourceReqs(&dpa); err != nil {
 		return false, err
 	}
+
+	// validate non-admin enable and tech-preview-ack
+	if !r.checkNonAdminEnabled(&dpa) {
+		if dpa.Spec.NonAdmin != nil &&
+			dpa.Spec.NonAdmin.Enable != nil &&
+			boolptr.IsSetToTrue(dpa.Spec.NonAdmin.Enable) &&
```
mateusoliveira43, line 72: is not this unnecessary?

we already check that `dpa.Spec.NonAdmin.Enable` is not `nil`

controllers/nonadmin_controller.go
```diff
@@ -177,10 +178,11 @@ func ensureRequiredSpecs(deploymentObject *appsv1.Deployment, image string) {
 
 func (r *DPAReconciler) checkNonAdminEnabled(dpa *oadpv1alpha1.DataProtectionApplication) bool {
 	// TODO https://github.com/openshift/oadp-operator/pull/1316
-	if dpa.Spec.Features != nil &&
-		dpa.Spec.Features.NonAdmin != nil &&
-		dpa.Spec.Features.NonAdmin.Enable != nil {
-		return *dpa.Spec.Features.NonAdmin.Enable
+	if dpa.Spec.NonAdmin != nil &&
+		dpa.Spec.NonAdmin.Enable != nil &&
+		boolptr.IsSetToTrue(dpa.Spec.NonAdmin.Enable) &&
```
mateusoliveira43, line 183: I would not add validation of unsupported override here, it is already checked by validator https://github.com/openshift/oadp-operator/blob/fc6afbf2920123fb4edb573a20561ed40632bca5/controllers/dpa_controller.go#L94

mateusoliveira43: no, that is just user feedback info

this is what actually checks if it is enabled

would try to simplify to
```go
	if dpa.Spec.Features != nil &&
		dpa.Spec.Features.NonAdmin != nil &&
		dpa.Spec.Features.NonAdmin.Enable != nil {
		return dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal && *dpa.Spec.Features.NonAdmin.Enable
```

controllers/validator.go
mateusoliveira43: would not be simple to do not change `checkNonAdminEnabled` function and just add check here?
```go
if r.checkNonAdminEnabled(&dpa) {
    if !(dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
        return false, errors.New("in order to enable the non-admin feature please set dpa.spec.unsupportedOverrides[tech-preview-ack]: true")
}
```
this way, when it GA, we just need to remove these lines here in validator?

mateusoliveira43: actually, I think is best to add this only in non admin controller

if I enable both unsupported override and enable dpa field, non admin deployment will appear in cluster

if later, I remove the unsupported, validation here will break, right? but this happens before non admin reconcile, which would let no admin deployment still running, right?

api/v1alpha1/oadp_types.go
```diff
@@ -288,6 +288,9 @@ type DataProtectionApplicationSpec struct {
 	// features defines the configuration for the DPA to enable the OADP tech preview features
 	// +optional
 	Features *Features `json:"features"`
+	// nonAdmin defines the configuration for the DPA to enable backup and restore operations for non-admin users
```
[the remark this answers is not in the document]
shubham-pampattiwar: I removed it from one of the places and only kept the description here.

controllers/validator.go
shubham-pampattiwar: @mateusoliveira43 I decoupled non-admin enable and tech preview ack flag in the latest commit, PTAL, hope that resolves everything, Thanks !

controllers/validator.go
```diff
@@ -63,6 +63,14 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 	if _, err := getResticResourceReqs(&dpa); err != nil {
 		return false, err
 	}
+
+	// validate non-admin enable and tech-preview-ack
+	if r.checkNonAdminEnabled(&dpa) {
+		if !(dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
+			return false, errors.New("in order to enable the non-admin feature please set dpa.spec.unsupportedOverrides[tech-preview-ack]: true")
```
mateusoliveira43, line 70: Is valid to change error message to `"in order to enable the non-admin feature please set dpa.spec.unsupportedOverrides[tech-preview-ack]: 'true'"` to avoid users telling us they can not input bool here?

controllers/validator.go
mateusoliveira43: I tested with `make deploy-olm`

semantically, I believe is correct to put unsupportedOverrides validation in validator (it is a DPA validation) 

But the scenario where user deploys non-admin correctly, and later changes unsupportedOverrides, makes reconcile fail for DPA, but does not delete non-admin deployment. Would be safer to put all validation in non-admin controller? It can also confuse users I think, as the DPA message will tell `in order to enable the non-admin feature please ...`, but it will be still enabled

shubham-pampattiwar: Ok maybe we just re-frame the DPA message to `in order to enable/disable the non-admin feature please...` WDYT ?

mateusoliveira43: makes sense to me!

shubham-pampattiwar: Done, updated the PR in f7e2b1200ff8d53359b7585a1f087b87f9d0bccd

controllers/validator.go
shubham-pampattiwar: Done, updated the PR in f7e2b1200ff8d53359b7585a1f087b87f9d0bccd

mateusoliveira43 (approved):

mateusoliveira43 (approved):

sseago (approved):

mpryc (approved):
/lgtm

Thank you @mateusoliveira43

PR #1395, opened 2024-05-22
OADP-4076: Add qps and burst to DPA

mateusoliveira43:
## Why the changes were made

In Velero 1.14, Burst and QPS will change defaults from 30 and 20 to 100 and 100. 
References: 
- https://github.com/vmware-tanzu/velero/commit/d676bfde2233d0d038286df91162c434331e8856
- 1.13 code https://github.com/vmware-tanzu/velero/blob/release-1.13/pkg/cmd/server/server.go#L93
- 1.14 code https://github.com/vmware-tanzu/velero/blob/release-1.14/pkg/cmd/server/server.go#L96

This PR ensures OADP users can change it.

Create DPA with new fields and check if values are passed to Velero container.


api/v1alpha1/oadp_types.go
```diff
@@ -110,6 +110,12 @@ type VeleroConfig struct {
 	// Default is 10m
 	// +optional
 	ResourceTimeout string `json:"resourceTimeout,omitempty"`
+	// Maximum number of requests by the server to the Kubernetes API in a short period of time. (default 30)
+	// +optional
+	ClientBurst *int `json:"client-burst,omitempty"`
+	// Maximum number of requests per second by the server to the Kubernetes API once the burst limit has been reached. (default 20)
+	// +optional
+	ClientQPS *float32 `json:"client-qps,omitempty"`
```
mateusoliveira43, line 118: I think we have 2 options here:
- like this (I had to include `allowDangerousTypes=true` in Makefile CRD_OPTIONS to use float here)
- or in unsupportedOverride

since it is a new field, it needs to be documented, right? So I think either locations are ok

kaovilai: if the only reason its dangerous is because of float, we could just use int and validate?

mateusoliveira43: did not understand the part about validate, but do not think we would lose much by being int here

another option is to be as in args, it is a string and is parsed as float

mrnold: Would resource.Quantity work here?

mateusoliveira43: I saw that in the issue I found the workaround with `allowDangerousTypes=true`, but did not test
going to try out

mateusoliveira43: from my tests, I would not use `resource.Quantity`

It is good for users (it can write `1000` or `1k`) but for us to read the value does not seem to be easy

From what I saw, we should use `String()` or `AsInt64()` method to get value. If I pass `1.5` I can not use `AsInt64()` to pass value to flag, and if I pass `1G` I can not use `String()` to pass value to flag (maybe I did not search enough)

If we do not want to use `allowDangerousTypes=true`, I think best option is to use string as in args https://github.com/openshift/oadp-operator/blob/145a6b82e4bec688294f90eba4a5ce2d96701752/pkg/velero/server/config.go#L58 (`int` we would be changing behavior of flag in Velero, so best to be cautious I think and convert string to float)

mrnold: I also see AsDec and AsApproximateFloat64, which might be more appropriate than AsInt64. But if it's troublesome then probably String is the easiest way to go. Thanks for trying it out.

controllers/velero_test.go
```diff
@@ -4108,6 +4168,161 @@ func TestDPAReconciler_buildVeleroDeployment(t *testing.T) {
...
+							},
+							Volumes:        baseVolumes,
+							InitContainers: []corev1.Container{},
+						},
+					},
+				},
+			},
+		},
+		{
+			name: "Conflicting burst and qps",
```
mateusoliveira43, line 4246: another question: now that we have a new field, should we forbidden the usage of old ones (qps and burst) in args? They would not be removed from DPA, but if any are used, reconcile fails with message telling user to use new ones

kaovilai: .args was always assumed to override everything else.. so .args related code came last and overrides whenever conflict is possible.

kaovilai: Adding validation when args is used and that it's overriding something else is a nice idea.

controllers/velero.go
```diff
@@ -448,6 +451,22 @@ func (r *DPAReconciler) customizeVeleroContainer(dpa *oadpv1alpha1.DataProtectio
 	return nil
 }
 
+func getClientBurst(dpa *oadpv1alpha1.DataProtectionApplication) int {
+	if dpa.Spec.Configuration.Velero != nil && dpa.Spec.Configuration.Velero.ClientBurst != nil {
+		return *dpa.Spec.Configuration.Velero.ClientBurst
+	}
+	// Velero default until 1.13
+	return 30
```
shubham-pampattiwar, line 459: no need set defaults via DPA, velero controller already does this internally

controllers/velero.go
```diff
@@ -448,6 +451,22 @@ func (r *DPAReconciler) customizeVeleroContainer(dpa *oadpv1alpha1.DataProtectio
...
+}
+
+func getClientQPS(dpa *oadpv1alpha1.DataProtectionApplication) float32 {
+	if dpa.Spec.Configuration.Velero != nil && dpa.Spec.Configuration.Velero.ClientQPS != nil {
+		return *dpa.Spec.Configuration.Velero.ClientQPS
+	}
+	// Velero default until 1.13
+	return 20.0
```
shubham-pampattiwar, line 467: same here, no need set defaults via DPA, velero controller already does this internally

Makefile
```diff
@@ -112,7 +112,7 @@ BUNDLE_IMG ?= $(IMAGE_TAG_BASE)-bundle:v$(VERSION)
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/konveyor/oadp-operator:latest
 # Produce CRDs that work back to Kubernetes 1.11 (no version conversion)
-CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false"
+CRD_OPTIONS ?= "crd:trivialVersions=true,preserveUnknownFields=false,allowDangerousTypes=true"
```
shubham-pampattiwar, line 115: I am not so sure that we should add `allowDangerousTypes=true`

controllers/velero_test.go
mateusoliveira43: as discussed in scrum we will not touch in validation now. Can be done in 1.5

controllers/velero.go
mateusoliveira43: done

controllers/velero.go
mateusoliveira43: done

Makefile
mateusoliveira43: to have conversation in one place, will move this comment to here https://github.com/openshift/oadp-operator/pull/1395#discussion_r1610114100

api/v1alpha1/oadp_types.go
mateusoliveira43: @shubham-pampattiwar comment "I am not so sure that we should add `allowDangerousTypes=true`"

mateusoliveira43: tried with controller-gen 0.14.0 but it also needs dangerous type. error message:

> found float, the usage of which is highly discouraged, as support for them varies across languages. Please consider serializing your float as string instead. If you are really sure you want to use them, re-run with crd:allowDangerousTypes=true

sseago: I think using int is fine for OADP. We convert to string to pass to velero -- the fact that velero uses float isn't really relevant. I can't see any  reason to set QPS to 100.5 instead of 100.

shubham-pampattiwar: +1

kaovilai: > it's technically supported, but floats are highly discouraged (and disallowed in official kubernetes APIs) because they won't round-trip properly, can't actually represent certain numbers exactly.

https://github.com/kubernetes-sigs/controller-tools/issues/245#:~:text=it%27s%20technically%20supported%2C%20but%20floats%20are%20highly%20discouraged%20(and%20disallowed%20in%20official%20kubernetes%20APIs)%20because%20they%20won%27t%20round%2Dtrip%20properly%2C%20can%27t%20actually%20represent%20certain%20numbers%20exactly

mateusoliveira43:
/hold
general hold for 1.4

api/v1alpha1/oadp_types.go
mateusoliveira43: as discussed in scrum, changed qps to int and removed dangerous type

mateusoliveira43:
waiting https://github.com/openshift/oadp-operator/pull/1463 merge to undraft this one

mateusoliveira43:
/override ci/prow/4.16-e2e-test-aws
restore succeeds, but deployment/deploymentConfig never goes to ready state flake

kaovilai (dismissed):

mateusoliveira43:
/override ci/prow/4.15-e2e-test-aws
restore succeeds, but deployment/deploymentConfig never goes to ready state flake

shubham-pampattiwar (approved):
Looks good ! 
```
shubham@fedora ~/oadp-operator -  (mateus-qps) $ oc get po velero-5d9b8546f4-trn26 -oyaml | grep args -7
    kind: ReplicaSet
    name: velero-5d9b8546f4
    uid: 8a40bfaa-fde7-4428-8e7b-f91a29330db3
  resourceVersion: "10670511"
  uid: 78de38b8-3e88-4295-9278-388c270dca47
spec:
  containers:
  - args:
    - server
    - --features=EnableCSI
    - --uploader-type=kopia
    - --fs-backup-timeout=4h
    - --restore-resource-priorities=securitycontextconstraints,customresourcedefinitions,klusterletconfigs.config.open-cluster-management.io,managedcluster.cluster.open-cluster-management.io,namespaces,roles,rolebindings,clusterrolebindings,klusterletaddonconfig.agent.open-cluster-management.io,managedclusteraddon.addon.open-cluster-management.io,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io
    - --client-burst=700
    - --client-qps=111
shubham@fedora ~/oadp-operator -  (mateus-qps) $ oc get dpa ts-dpa -oyaml | grep client
      client-burst: 700
      client-qps: 111
shubham@fedora ~/oadp-operator -  (mateus-qps) $ 

```

weshayutin (approved):
/LGTM

PR #1398, opened 2024-05-23
OADP-4137: updated for velero 1.14

sseago:
## Why the changes were made
go.mod changes for Velero 1.14 (including golang 1.22) 
velero CRD changes for 1.14
some oadp operator code changes to cope with controller-runtime upgrade to 0.17.2


go.mod
```diff
@@ -1,165 +1,185 @@
 module github.com/openshift/oadp-operator
 
-go 1.20
+go 1.22.0
```
kaovilai, line 3: I think we said this is no longer necessary if you got the 1.22.3 or above builder. The requirement for patch version was a bug.

sseago: There are 2 options: specify the .0 here *or* add the toolchain line.

In other words, we have 2 choices:
```
go 1.22.0
```
or:
```
go 1.22

toolchain go1.22.3
```

I don't really have much of a preference. Note that if you remove ".0" and run `go mod tidy` the toolchain line is added. We need one or the other.

mateusoliveira43: I can run `go mod tidy` without toolchain being added (go version go1.22.3 linux/amd64)

But we can copy how is done in Velero go.mod

shubham-pampattiwar: I think upstream is gonna be 1.22.2 https://github.com/vmware-tanzu/velero/pull/7821/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6

sseago: @shubham-pampattiwar ok - -we'll get that on the next rebase for velero. so should we set `go 1.22.2` in our go.mod for oadp-operator then?

sseago: @mateusoliveira43 for me, tidy added toolchain if I have `go 1.22` but not if I have `go 1.22.0` -- I'm also on 1.22.3.

rayfordj: FYI, in case it matters, we don't have a downstream/internal golang-builder newer than `v1.22.1` yet

shubham-pampattiwar: @sseago yeah if upstream is gonna release velero 1.14 with Golang 1.22.2 then we should do the same for oadp-1.4 imo

sseago: updated to 1.22.2

kaovilai (dismissed):

Makefile
mateusoliveira43, line 1: Also update here https://github.com/openshift/oadp-operator/blob/oadp-1.4/Makefile#L447 to oadp-1.3

mateusoliveira43:
README links can also be updated
Example https://github.com/openshift/oadp-operator/blob/oadp-1.4/README.md?plain=1#L9

bundle/manifests/oadp.openshift.io_cloudstorages.yaml
```diff
@@ -96,5 +104,5 @@ status:
   acceptedNames:
     kind: ""
     plural: ""
-  conditions: []
-  storedVersions: []
+  conditions: null
+  storedVersions: null
```
mateusoliveira43, line 108: checking the changes done by the update of controller gen
- this empty list to null seems harmless
- description changes are harmless
- `x-kubernetes-map-type: atomic` were introduced by packages we use (example https://github.com/kubernetes/api/blob/v0.29.0/core/v1/types.go#L2272 ), and seem ok as well. Ref about these changes https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy

Makefile
```diff
@@ -314,7 +313,7 @@ operator-sdk:
 	# Download operator-sdk locally if does not exist
 	if [ ! -f $(OPERATOR_SDK) ]; then \
 		mkdir -p bin ;\
-		curl -Lo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v1.23.0/operator-sdk_$(shell go env GOOS)_$(shell go env GOARCH) ; \
+		curl -Lo $(OPERATOR_SDK) https://github.com/operator-framework/operator-sdk/releases/download/v1.34.2/operator-sdk_$(shell go env GOOS)_$(shell go env GOARCH) ; \
```
mateusoliveira43, line 316: I got lost in slack thread, why we are updating here?

if it because of new flag, should be safer to update to lowest release that contains it? I think it is https://github.com/operator-framework/operator-sdk/releases/tag/v1.24.1

and if we do update, I think at least the annotations telling which operator-sdk is used by OADP need to be updated. I have similar changes here https://github.com/openshift/oadp-operator/pull/1250

sseago: So are we  good with this as-is to get this merged ASAP to have a working oadp-1.4 branch? Follow-on PRs to refine it as needed.

weshayutin: Yes, Rayford indicated that the operator-sdk is NOT used in d.s. builds anyway.  This is dev only.  Clear to proceed.

mateusoliveira43: we need operator-sdk to generate `bundle` folder (`make bundle` command). So I think it is not only for development

I will make some tests, but maybe is ok to update version (and if needed changes, can be a follow up PR)

kaovilai: Generating bundle is a dev activity. Unless we specifically ask or they feel adventurous, `make bundle` is only something our team touches.

mateusoliveira43: my fear is the new version of operator-sdk will generate new bundle, changing delivered yamls to users. BUT need to test if that really happens

kaovilai: > changing delivered yamls to users

Not our downstream builds no. Upstream builds are not published in operatorhub so we should be good.

kaovilai: Usually the (upstream) changes are just CSV annotations/labels

kaovilai:
retest when https://github.com/openshift/release/pull/52511 goes in

kaovilai:
```
# github.com/openshift/oadp-operator/pkg/leaderelection
pkg/leaderelection/leaderelection.go:41:4: k8s.io/klog/v2.Warning call has possible Printf formatting directive %v how much
```
[error](https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1398/pull-ci-openshift-oadp-operator-oadp-1.4-unit-test/1795889220414345216)

Dockerfile
```diff
@@ -1,5 +1,5 @@
 # Build the manager binary
-FROM quay.io/konveyor/builder:ubi9-v1.20 AS builder
+FROM quay.io/konveyor/builder:ubi9-latest AS builder
```
mateusoliveira43, line 2: do we need any of Tiger's changes from here? https://github.com/openshift/oadp-operator/pull/1401/files

kaovilai: very minor diff.. 1.22.3 vs 1.22.2. I think we agreed on 1.22.2 in some calls but im good with either.

kaovilai: the ocp/release already pinned to 1.22.2 but should be close enough that it doesn't matter.

mateusoliveira43 (changes requested):
for CI pass, need to run
```
make fmt 
make generate 
```
and a workaround with update of timestamp every time :cry:

kaovilai (approved):

shubham-pampattiwar (approved):
/lgtm

PR #1396, opened 2024-05-22
OADP-4110: Add support for unsupported node-agent args

shubham-pampattiwar:
## Why the changes were made

This PR 
- fixes [OADP-4110](https://issues.redhat.com//browse/OADP-4110) https://issues.redhat.com/browse/OADP-4110
- adds support for unsupported node-agent args via DPA
- this PR unblocks the customers when the DPA does not have server args parity with upstream Velero node-agent server args

## How to test the changes made

In order to test these changes:
- checkout the PR and use `make deploy-olm` command to install the operator and CRDs (make sure you do not have stale CRDs in cluster)
- create a valid sample DPA and add the following to spec (In this example we will try to supply 2 node-agent server args):
```
spec.nodeAgent.unsupportedNodeAgentArgs: "--data-mover-prepare-timeout=45m,--resource-timeout=15m"
``` 
- check the node-agent daemon set container args to verify the values supplied
- To cross verify you can use the image (`veleroImageFqin: quay.io/spampatt/velero:unsupp-node-agent-args`) and view node-agent logs for the arg values we supplied:
```
time="2024-05-22T18:50:48Z" level=info msg="Testing value for data-mover-prepare-timeout: 45m0s" logSource="pkg/cmd/cli/nodeagent/server.go:275"  
time="2024-05-22T18:50:48Z" level=info msg="Testing value for resource-timeout: 15m0s" logSource="pkg/cmd/cli/nodeagent/server.go:276"
```

PR #1400, opened 2024-05-29
Add Design for the unsupported arguments for the OADP

mpryc:
Initial design to address [OADP-4159](https://issues.redhat.com//browse/OADP-4159)

Also may be included as part of [OADP-4110](https://issues.redhat.com//browse/OADP-4110) with PR #1396

## Why the changes were made

To design unsupported arguments.

n/a it's a design.
One test was performed to create ConfigMap with `--` as key name.

### DIFF
--- a/docs/design/unsupported-args-design.md
+++ b/docs/design/unsupported-args-design.md
@@ -1,0 +1,118 @@
+# Unsupported Arguments schema and common design
+Date: 2024-06-29
+
+## Abstract
+OADP allows to configure it's components parameters via number of DataProtectionApplication Schema.
+
+Unsupported arguments within the OADP operator are those that do not conform to the standard DataProtectionApplication Schema or are explicitly identified as unsupported or custom arguments.
+
+There is need to provide a common way for defining and managing unsupported arguments within the OADP operator for its' components such as velero server or node-agent arguments.
+
+## Background
+The OADP allows configuration of several predefined arguments via the DataProtectionApplication Schema. These parameters are carefully selected and tested to meet the highest quality standards. However, in some circumstances, there is a need to pass other parameters that may be under development, not yet fully tested, or specifically overriding the default ones.
+
+Over the past OADP releases, some implementations have provided access to extra arguments. However, there has been no standardization, and unknown to DataProtectionApplication Schema arguments were not allowed.
+
+## Goals
+- Define common way to allow passing extra arguments to various executables used by OADP.
+- Ensure the new common way do not interfere with already existing implementations.
+
+## Non Goals
+- Deprecate current ways of defining extra arguments.
+- Change precedences within current ways of defining extra arguments.
+- Validation of unsupported args, they are passed to the appropriate executables without any prior validation.
+
+## Design
+Unsupported arguments are defined as a `ConfigMap` within the same Namespace as OADP Operator.
+
+There may be multiple `ConfigMaps` for particular section of the OADP configuration.
+
+There are default naming convention for the `ConfigMaps`, however user may override those using `DPA` annotations.
+
+When ConfigMap is found, controller takes those values as the highest priority and replaces **all** of the other configuration options, even if they were defined within `DPA` schema for the section for which the `ConfigMap` was created.
+
+The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
+
+   ```
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: oadp-unsupported-velero-server-args
+     namespace: openshift-adp
+   data:
+     --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
+     --log-level: 'debug'
+   ```
+
+If the name of the ConfigMap is recognized by the OADP controller it is consumed, otherwise user has to create DPA with the annotation pointing to a relevant ConfigMap as in the example:
+
+
+    ```yaml
+    kind: DataProtectionApplication
+    apiVersion: oadp.openshift.io/v1alpha1
+    metadata:
+    name: sample-dpa
+    namespace: openshift-adp
+    annotations:
+        custom.configmap.dpa/velero-server-args: 'oadp-unsupported-velero-server-args'
+    spec:
+        [...]
+    ```
+
+## Alternatives Considered
+ - Resource specific configuration / CRD options
+ 
+   The configuration data is directly integrated into the Custom Resource Definition (CRD). This is how currently some of the extra arguments are defined within OADP.
+
+   There is however schema enforcement, which is problematic, because it requires new release of OADP to include unknown or not available options.
+   
+   This approach also introduces more complexity as it requires logic to handle and validate each of the options. On top of that having CRD options introduces migration and upgrade problems requiring migration strategies to be taken into account for existing resources. 
+
+   Adding the unsupported arguments and options within CRD makes those options highly visible to the users and may encourage to use them.
+
+ - Annotations to include unsupported args directly
+   This mmethod allows easily to add args to existing resources, however they have size limitations, restricting the amount of data that can be stored and may have impact on the scenarios where limit is reached.
+
+   Annotations are also problematic to maintain when used for more complex types of configurations.
+
+   Sample custom unsupported args:
+
+    ```yaml
+    kind: DataProtectionApplication
+    apiVersion: oadp.openshift.io/v1alpha1
+    metadata:
+    name: sample-dpa
+    namespace: openshift-adp
+    annotations:
+        custom.configmap.dpa/velero-server-args: '{"arg_1":"val_1","arg2":"val2"}'
+        custom.configmap.dpa/node-agent-args: '{"arg_1":"val_1","arg_2":"val_2"}'
+    spec:
+        [...]
+    ```
+
+ - Labels to include unsupported args directly
+
+   This approach is similar to annotations, with all the pros/cons of them with additional cons of labels being designed for resource identification, querying, and organizing within clusters and not configurations.
+
+## Security Considerations
+Passing input data without validation may cause malformed configurations to be included.
+
+This design does not allow to store sensitive information within configuration, as it's using ConfigMaps and not Secrets.
+
+Proper RBAC access restriction to the relevant ConfigMaps is required to ensure least privilege is granted.
+
+## Compatibility
+This design is backwards compatible with previous unsupported arguments passed via DPA specification, however it takes precendence over them.
+
+This may lead to the situation where users will have running previous configurations that are not reflected with the OADP instances that are using new way of defining unsupported arguments.
+
+## Implementation
+The implementation will take place within OADP controller only, without changes to the DPA specification.
+
+Once ConfigMap(s) are discovered to be in the same Namespace as OADP their config will take precendence and replace all the other arguments passed to the executable as a simple string without any validations.
+
+## Open Issues
+ - Should we include the unsupported args in all cases when the ConfigMaps are found in the napespace or maybe only when annotations are within DPA?
+ - Naming convention for the annotations - currently it's not well defined
+ - We may consider using one ConfigMap instead of multiple ConfigMaps, but this will lock down it's schema and may lead to future issues with enabling other unsupported args.
+
--- a/docs/design/unsupported-args-design.md
+++ b/docs/design/unsupported-args-design.md
@@ -33,31 +33,48 @@ When ConfigMap is found, controller takes those values as the highest priority a
 
 The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
 
-   ```yaml
-   apiVersion: v1
-   kind: ConfigMap
-   metadata:
-     name: oadp-unsupported-velero-server-args
-     namespace: openshift-adp
-   data:
-     --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
-     --log-level: 'debug'
-   ```
+<a name="notes"></a>
+> ℹ️ **Note 1:** If an argument name is passed without `-` prefix or `--` prefix, then `--` is added to the argument name.
+
+> ℹ️ **Note 2:** If an non-boolean argument value is passed without `'` or `"` sourrounding, then the single quote is added to the argument value.
+
+> ℹ️ **Note 3:** An boolean argument value `true` or `false` is always converted to lower-case and not sourrounded by single quote `'`.
+
+> ℹ️ **Note 4:** An argument value is always combined with the argument name using `=` character in between.
+
+
+  ```yaml
+  apiVersion: v1
+  kind: ConfigMap
+  metadata:
+    name: oadp-unsupported-velero-server-args
+    namespace: openshift-adp
+  data:
+    default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
+    log-level: 'debug'
+    default-snapshot-move-data: 'true'
+  ```
+
+Above example will translate to the `velero` args:
+
+```shell
+velero server --default-volume-snapshot-locations='aws:backups-primary,azure:backups-secondary' --log-level='debug' --default-snapshot-move-data=true
+```
 
 The user has to create DPA with the annotation pointing to a relevant ConfigMap as in the example where both `oadp-unsupported-velero-server-args` and `unsupported-node-agent` ConfigMaps are expected to be present:
 
-    ```yaml
-    kind: DataProtectionApplication
-    apiVersion: oadp.openshift.io/v1alpha1
-    metadata:
-    name: sample-dpa
-    namespace: openshift-adp
-    annotations:
-        oadp.openshift.io/unsupported-velero-server-args: 'oadp-unsupported-velero-server-args'
-        oadp.openshift.io/unsupported-noda-agent-args: 'unsupported-node-agent'
-    spec:
-        [...]
-    ```
+  ```yaml
+  kind: DataProtectionApplication
+  apiVersion: oadp.openshift.io/v1alpha1
+  metadata:
+  name: sample-dpa
+  namespace: openshift-adp
+  annotations:
+      oadp.openshift.io/unsupported-velero-server-args: 'oadp-unsupported-velero-server-args'
+      oadp.openshift.io/unsupported-node-agent-args: 'unsupported-node-agent'
+  spec:
+      [...]
+  ```
 
 ## Alternatives Considered
  - Resource specific configuration / CRD options
@@ -86,7 +103,7 @@ The user has to create DPA with the annotation pointing to a relevant ConfigMap 
     namespace: openshift-adp
     annotations:
         oadp.openshift.io/unsupported-velero-server-args: '{"arg_1":"val_1","arg2":"val2"}'
-        oadp.openshift.io/unsupported-noda-agent-args: '{"arg_1":"val_1","arg_2":"val_2"}'
+        oadp.openshift.io/unsupported-node-agent-args: '{"arg_1":"val_1","arg_2":"val_2"}'
     spec:
         [...]
     ```
@@ -114,6 +131,4 @@ The implementation will take place within OADP controller only, without changes 
 
 Once ConfigMap(s) are discovered to be in the same Namespace as OADP their config will take precendence and replace all the other arguments passed to the executable as a simple string without any validations.
 
-## Open Issues
- - Naming convention for the annotations - currently it's not well defined
-
+There may be slight modification to the argument name or it's value taken from the `ConfigMap` to align with the `Cobra` library as described in the [notes](#notes).
--- a/docs/design/unsupported-args-design.md
+++ b/docs/design/unsupported-args-design.md
@@ -1,5 +1,5 @@
 # Unsupported Arguments schema and common design
-Date: 2024-06-29
+Date: 2024-05-29
 
 ## Abstract
 OADP allows to configure it's components parameters via number of DataProtectionApplication Schema.
@@ -25,15 +25,15 @@ Over the past OADP releases, some implementations have provided access to extra 
 ## Design
 Unsupported arguments are defined as a `ConfigMap` within the same Namespace as OADP Operator.
 
-There may be multiple `ConfigMaps` for particular section of the OADP configuration.
+Multiple `ConfigMaps` may exist to represent specific sections of the OADP configuration, such as dedicated `ConfigMap` instances for Velero server arguments or node-agent arguments.
 
-There are default naming convention for the `ConfigMaps`, however user may override those using `DPA` annotations.
+The user must provide `DPA` annotation(s) with the `name` of the `ConfigMap` where the OADP configuration for the unsupported arguments is present.
 
 When ConfigMap is found, controller takes those values as the highest priority and replaces **all** of the other configuration options, even if they were defined within `DPA` schema for the section for which the `ConfigMap` was created.
 
 The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
 
-   ```
+   ```yaml
    apiVersion: v1
    kind: ConfigMap
    metadata:
@@ -44,8 +44,7 @@ The Schema of an `ConfigMap` includes data that corresponds to the arguments pas
      --log-level: 'debug'
    ```
 
-If the name of the ConfigMap is recognized by the OADP controller it is consumed, otherwise user has to create DPA with the annotation pointing to a relevant ConfigMap as in the example:
-
+The user has to create DPA with the annotation pointing to a relevant ConfigMap as in the example where both `oadp-unsupported-velero-server-args` and `unsupported-node-agent` ConfigMaps are expected to be present:
 
     ```yaml
     kind: DataProtectionApplication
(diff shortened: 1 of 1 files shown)


docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+   kind: ConfigMap
+   metadata:
+     name: oadp-unsupported-velero-server-args
+     namespace: openshift-adp
+   data:
+     --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
+     --log-level: 'debug'
+   ```
+
+If the name of the ConfigMap is recognized by the OADP controller it is consumed, otherwise user has to create DPA with the annotation pointing to a relevant ConfigMap as in the example:
```
shubham-pampattiwar, line 47: Lets skip over the naming convention part and make DPA annotation for config-map name a must, or else we would not get an update event for DPA reconcile to be triggered. DPA annotation patch call would ensure that reconcile is triggered or else maybe we would have to implement a custom watch for configmaps.

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+    ```yaml
+    kind: DataProtectionApplication
+    apiVersion: oadp.openshift.io/v1alpha1
+    metadata:
+    name: sample-dpa
+    namespace: openshift-adp
+    annotations:
+        custom.configmap.dpa/velero-server-args: 'oadp-unsupported-velero-server-args'
```
shubham-pampattiwar, line 57: annotation would be something like `oadp.openshift.io/unsupported-velero-args` ?

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+   The configuration data is directly integrated into the Custom Resource Definition (CRD). This is how currently some of the extra arguments are defined within OADP.
+
+   There is however schema enforcement, which is problematic, because it requires new release of OADP to include unknown or not available options.
+   
+   This approach also introduces more complexity as it requires logic to handle and validate each of the options. On top of that having CRD options introduces migration and upgrade problems requiring migration strategies to be taken into account for existing resources. 
+
+   Adding the unsupported arguments and options within CRD makes those options highly visible to the users and may encourage to use them.
+
+ - Annotations to include unsupported args directly
+   This mmethod allows easily to add args to existing resources, however they have size limitations, restricting the amount of data that can be stored and may have impact on the scenarios where limit is reached.
```
shubham-pampattiwar, line 74: `s/mmethod/method`

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+Passing input data without validation may cause malformed configurations to be included.
+
+This design does not allow to store sensitive information within configuration, as it's using ConfigMaps and not Secrets.
+
+Proper RBAC access restriction to the relevant ConfigMaps is required to ensure least privilege is granted.
+
+## Compatibility
+This design is backwards compatible with previous unsupported arguments passed via DPA specification, however it takes precendence over them.
+
+This may lead to the situation where users will have running previous configurations that are not reflected with the OADP instances that are using new way of defining unsupported arguments.
```
shubham-pampattiwar, line 107: We need to explicitly document this in developer docs and support docs

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Implementation
+The implementation will take place within OADP controller only, without changes to the DPA specification.
+
+Once ConfigMap(s) are discovered to be in the same Namespace as OADP their config will take precendence and replace all the other arguments passed to the executable as a simple string without any validations.
+
+## Open Issues
+ - Should we include the unsupported args in all cases when the ConfigMaps are found in the napespace or maybe only when annotations are within DPA?
```
shubham-pampattiwar, line 115: only when the annotations are present on DPA

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+ - Naming convention for the annotations - currently it's not well defined
+ - We may consider using one ConfigMap instead of multiple ConfigMaps, but this will lock down it's schema and may lead to future issues with enabling other unsupported args.
```
shubham-pampattiwar, line 117: I believe we should use 2 configmaps - one for velero and the other for node-agent, and these should have 2 separate annotation keys too.

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
+# Unsupported Arguments schema and common design
+Date: 2024-06-29
```
mateusoliveira43, line 2: still may :eyes:

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
+
+   ```
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: oadp-unsupported-velero-server-args
+     namespace: openshift-adp
+   data:
+     --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
```
mrnold, line 43: Are there any argument names that won't be acceptable as ConfigMap keys? I tried some different strings with `/` and `:` characters, and got an error message that says `regex used for validation is '[-._a-zA-Z0-9]+'`.

I'm not sure how restrictive this will be in practice, but it might be worth considering in case any of our binaries have any arguments that won't work this way.

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+
+There may be multiple `ConfigMaps` for particular section of the OADP configuration.
+
+There are default naming convention for the `ConfigMaps`, however user may override those using `DPA` annotations.
+
+When ConfigMap is found, controller takes those values as the highest priority and replaces **all** of the other configuration options, even if they were defined within `DPA` schema for the section for which the `ConfigMap` was created.
+
+The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
+
+   ```
```
mateusoliveira43, line 36: please add yaml

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+   Adding the unsupported arguments and options within CRD makes those options highly visible to the users and may encourage to use them.
+
+ - Annotations to include unsupported args directly
+   This mmethod allows easily to add args to existing resources, however they have size limitations, restricting the amount of data that can be stored and may have impact on the scenarios where limit is reached.
+
+   Annotations are also problematic to maintain when used for more complex types of configurations.
+
+   Sample custom unsupported args:
+
+    ```yaml
```
mateusoliveira43, line 80: rendering did not work here 
Screenshot from 2024-05-29 13-19-41

docs/design/unsupported-args-design.md
mateusoliveira43: I think we can poke round how cobra works (velero CLI is built using it), if it also has the same restrictions

docs/design/unsupported-args-design.md
mateusoliveira43: also jump line or lines 73 and 74 appear together when rendered

docs/design/unsupported-args-design.md
mateusoliveira43: tried and I can have a flag name like `name/flag` :grimacing: BUT today, I think there is none like this (would say velero flags name are composed of  lowercase letters, `-` and `_`

Maybe we can enforce this regex in velero flags names using some cobra functionalities

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,118 @@
...
+
+   ```
+   apiVersion: v1
+   kind: ConfigMap
+   metadata:
+     name: oadp-unsupported-velero-server-args
+     namespace: openshift-adp
+   data:
+     --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
+     --log-level: 'debug'
```
mateusoliveira43, line 44: one fear with this approach is if there is any flag that we just activate by passing it. But if that happens, we can just pass an empty string here and in the logic to passing flags, just check value length, right?

weshayutin:
@mpryc can you please get this PR updated for final review and merge.  Thank you

docs/design/unsupported-args-design.md
mpryc: The K8s doesn't restrict special characters in the `value` part, however the key must be within regex as @mrnold wrote:

```
Invalid value: "--default-volume-snapshot_/locations": a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+')
``` 

If we ever want to have such special args, which I don't think it's a valid use case at the moment - the args are within the regex imo, we can not translate them into base64 encodings as that may produce `=` chars, what we could do is to have list of args numerated in configmap and have configmap annotated that it's using this method?, e.g.:

Instead of:
```yaml
   apiVersion: v1
   kind: ConfigMap
   metadata:
     name: oadp-unsupported-velero-server-args
     namespace: openshift-adp
    annotations:
      numerated-args: "true"
data:
   --my-super&strange/arg: 'arg value arg second value'
   --my-second-arg: 'arg value 2'
```
```yaml
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: oadp-unsupported-velero-server-args
    namespace: openshift-adp
    annotations:
      numerated-args: "true"
data:
   arg-1: '--my-super&strange/arg: arg value arg second value'
   arg-2: '--my-second-arg: 'arg value 2'
```

docs/design/unsupported-args-design.md
mpryc: @mateusoliveira43 
You are very much right here, we may want to have empty string to represent boolean value, for example:
```yaml
  --enable-some-option
```

But `ConfigMap` does not allow empty strings, so I wonder if we want to either include some constant string that represents the empty string or boolean flag? Can't think of other option?

docs/design/unsupported-args-design.md
mpryc: `oadp.openshift.io/unsupported-velero-server-args` ?

docs/design/unsupported-args-design.md
mpryc: wired, it's perfectly fine on my rendering:

image

docs/design/unsupported-args-design.md
mpryc: Adding that sentence to cover documentation part.

docs/design/unsupported-args-design.md
mpryc: Removed from Open Issues and added to the examples.

docs/design/unsupported-args-design.md
mateusoliveira43: sorry, put comment in wrong place. Its the second code block (beginning in line 49) that is not rendering. I believe its because it has 2 indents instead of 1 (or 4 spaces instead of 2)

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,119 @@
...
+
+    ```yaml
+    kind: DataProtectionApplication
+    apiVersion: oadp.openshift.io/v1alpha1
+    metadata:
+    name: sample-dpa
+    namespace: openshift-adp
+    annotations:
+        oadp.openshift.io/unsupported-velero-server-args: 'oadp-unsupported-velero-server-args'
+        oadp.openshift.io/unsupported-noda-agent-args: 'unsupported-node-agent'
```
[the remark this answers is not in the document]
mpryc: Thanks, will fix.

docs/design/unsupported-args-design.md
mateusoliveira43: Empty string like this, is accepted, right? i.e. a string of length zero `''`
```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: mateus-test
  namespace: openshift-adp
  uid: e05e4bde-e250-4df6-a90c-8c29bbe7a07d
  resourceVersion: '380824521'
  creationTimestamp: '2024-06-06T13:11:38Z'
  managedFields:
    - manager: Mozilla
      operation: Update
      apiVersion: v1
      time: '2024-06-06T13:11:38Z'
      fieldsType: FieldsV1
      fieldsV1:
        'f:data':
          .: {}
          'f:--another': {}
          'f:--log-level': {}
data:
  '--another': ''
  '--log-level': ''
```

From what I searched, that would not affect us
```
$ go run cmd/velero/velero.go help server | grep empty
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
```

But constant string that represents the empty string seems good idea

sseago: Can we just use "true" here? Bool arg without value is just a short-cut for true. If we require a constant value anyway, lets just have the users explicitly set it to true. We can just document that for bool args, an explicit true or false is needed.

mateusoliveira43: I think that would work, from what I looked, only fields like this are boolean

mpryc: I believe that in some CLI's it may not work, so passing unexpected arg (empty string) may actually cause error ?

sseago: @mpryc My suggestion was just to require args. This is only relevant for bool args where "=true" can be left off with no change in meaning. Via the configmap, lets just require them to specify `true` explicitly -- that way we hve no unexpected empty strings.

docs/design/unsupported-args-design.md
mpryc: ah see it

docs/design/unsupported-args-design.md
mpryc: Removed from the last version.

docs/design/unsupported-args-design.md
mpryc: @sseago that will it work, so for the arguments that are boolean we require `=true` or `=false` to be passed, however we don't validate those.

Below is a full list of server args from `velero server` and in our case for example `--default-snapshot-move-data` would require `--default-snapshot-move-data=true` right?

```shell
velero server --help
Run the velero server

Usage:
  velero server [flags]

Flags:
      --backup-sync-period duration                         How often to ensure all Velero backups in object storage exist as Backup API objects in the cluster. This is the default sync period if none is explicitly specified for a backup storage location. (default 1m0s)
      --client-burst int                                    Maximum number of requests by the server to the Kubernetes API in a short period of time. (default 30)
      --client-page-size int                                Page size of requests by the server to the Kubernetes API when listing objects during a backup. Set to 0 to disable paging. (default 500)
      --client-qps float32                                  Maximum number of requests per second by the server to the Kubernetes API once the burst limit has been reached. (default 20)
      --default-backup-storage-location string              Name of the default backup storage location. DEPRECATED: this flag will be removed in v2.0. Use "velero backup-location set --default" instead. (default "default")
      --default-backup-ttl duration                         How long to wait by default before backups can be garbage collected. (default 720h0m0s)
      --default-item-operation-timeout duration             How long to wait on asynchronous BackupItemActions and RestoreItemActions to complete before timing out. Default is 4 hours (default 4h0m0s)
      --default-repo-maintain-frequency duration            How often 'maintain' is run for backup repositories by default.
      --default-snapshot-move-data                          Move data by default for all snapshots supporting data movement.
      --default-volume-snapshot-locations mapStringString   List of unique volume providers and default volume snapshot location (provider1:location-01,provider2:location-02,...)
      --default-volumes-to-fs-backup                        Backup all volumes with pod volume file system backup by default.
      --disable-controllers strings                         List of controllers to disable on startup. Valid values are backup,backup-operations,backup-deletion,backup-finalizer,backup-sync,download-request,gc,backup-repo,restore,restore-operations,schedule,server-status-request
      --disable-informer-cache                              Disable informer cache for Get calls on restore. WIth this enabled, it will speed up restore in cases where there are backup resources which already exist in the cluster, but for very large clusters this will increase velero memory usage. Default is false (don't disable).
      --fs-backup-timeout duration                          How long pod volume file system backups/restores should be allowed to run before timing out. (default 4h0m0s)
      --garbage-collection-frequency duration               How often garbage collection is run for expired backups.
  -h, --help                                                help for server
      --item-operation-sync-frequency duration              How often to check status on backup/restore operations after backup/restore processing. Default is 10 seconds
      --log-format                                          The format for log output. Valid values are text, json. (default text)
      --log-level                                           The level at which to log. Valid values are trace, debug, info, warning, error, fatal, panic. (default info)
      --max-concurrent-k8s-connections int                  Max concurrent connections number that Velero can create with kube-apiserver. Default is 30. (default 30)
      --metrics-address string                              The address to expose prometheus metrics (default ":8085")
      --plugin-dir string                                   Directory containing Velero plugins (default "/plugins")
      --profiler-address string                             The address to expose the pprof profiler. (default "localhost:6060")
      --resource-timeout duration                           How long to wait for resource processes which are not covered by other specific timeout parameters. Default is 10 minutes. (default 10m0s)
      --restore-only                                        Run in a mode where only restores are allowed; backups, schedules, and garbage-collection are all disabled. DEPRECATED: this flag will be removed in v2.0. Use read-only backup storage locations instead.
      --restore-resource-priorities stringArray             Desired order of resource restores, the priority list contains two parts which are split by "-" element. The resources before "-" element are restored first as high priorities, the resources after "-" element are restored last as low priorities, and any resource not in the list will be restored alphabetically between the high and low priorities. (default customresourcedefinitions,namespaces,storageclasses,volumesnapshotclass.snapshot.storage.k8s.io,volumesnapshotcontents.snapshot.storage.k8s.io,volumesnapshots.snapshot.storage.k8s.io,datauploads.velero.io,persistentvolumes,persistentvolumeclaims,serviceaccounts,secrets,configmaps,limitranges,pods,replicasets.apps,clusterclasses.cluster.x-k8s.io,endpoints,services,-,clusterbootstraps.run.tanzu.vmware.com,clusters.cluster.x-k8s.io,clusterresourcesets.addons.cluster.x-k8s.io)
      --store-validation-frequency 0s                       How often to verify if the storage is valid. Optional. Set this to 0s to disable sync. Default 1 minute. (default 1m0s)
      --terminating-resource-timeout duration               How long to wait on persistent volumes and namespaces to terminate during a restore before timing out. (default 10m0s)
      --uploader-type string                                Type of uploader to handle the transfer of data of pod volumes (default "restic")

Global Flags:
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
      --colorized optionalBool           Show colored output in TTY. Overrides 'colorized' value from $HOME/.config/velero/config.json if present. Enabled by default
      --features stringArray             Comma-separated list of features to enable for this Velero process. Combines with values from $HOME/.config/velero/config.json if present
      --kubeconfig string                Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration
      --kubecontext string               The context to use to talk to the Kubernetes apiserver. If unset defaults to whatever your current-context is (kubectl config current-context)
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint           Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
  -n, --namespace string                 The namespace in which Velero should operate (default "openshift-adp")
      --one_output                       If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
      --s2a_enable_appengine_dialer      If true, opportunistically use AppEngine-specific dialer to call S2A.
      --s2a_timeout duration             Timeout enforced on the connection to the S2A service for handshake. (default 3s)
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity         logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=false) (default 2)
  -v, --v Level                          number for the log level verbosity
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
```

mpryc: @sseago how would you go about this:

```
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: oadp-unsupported-velero-server-args
    namespace: openshift-adp
  data:
    --default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
    --log-level: 'debug'
    --default-snapshot-move-data: '=true'
```

We can see that some arguments must have space between value and for `boolean` there must be no space, which isn't that clear from the above ConfigMap example.

e.g. how to distinguish between:

`velero --my-sample-arg=true --my-sample-arg true`

docs/design/unsupported-args-design.md
mpryc: I think that's addressed now?

docs/design/unsupported-args-design.md
mpryc: I am closing this as it's not relevant anymore.

docs/design/unsupported-args-design.md
mpryc: @sseago please review current change to the design, it uses `=true` or `=false` to indicate boolean args, so they need to start with `=` as a first value character.

mpryc: Because code speak itself, there is a small function that implements ConfigMap to args on how this design describes it:

https://github.com/openshift/oadp-operator/pull/1423

In our case it's going to replace this in the Deployment:

https://github.com/vmware-tanzu/velero/blob/main/pkg/install/deployment.go#L290

mateusoliveira43: I think there is no need for `=true`, we just do logic with string being `"%s=%s", key, value`, for all flags

weshayutin: If the operator-manager ( I think ) is too strict in how it reformats commands and enforces too much we run the risk of making a mistake or limiting the future proofness off this design.  
An Example:
1. If a customer typos a config value in the configmap, and there is an error in velero server args or node-agent the velero pod will not come up and there will be an error.  In this case the customer ( NOTE: along side engineering) will have to fix the configmap.
2.  If the code that transforms the configmap has a mistake or error or didn't account for something engineering will have to ship the customer new code.

Noting that this feature is undocumented and "should" only be employed in partnership with engineering I would much rather have to deal with #1 vs #2

Garbage in -->  Garbage out is a MUCH easier situation to deal with than updating the controller-manager code. In fact we do not have a way to ship a custom controller-manager.  
Being too strict with how arguments are formated has the potential to defeat the whole purpose of the design.

sseago: @mpryc I think we can make a couple simplifications:
1) We can skip the "--" in the string the user is adding. `--foo` -- the name is "foo", the "--" is just how flags are passed
2) `--param-name=param-value` should work in all cases, no need for "=" in some names, and no need to worry about spacing.

Basically, you have k/v pairs in the config map, and then when generating the flags, `"--" + paramName + "=" + paramValue`

So the above configmap would look like this:
```
  apiVersion: v1
  kind: ConfigMap
  metadata:
    name: oadp-unsupported-velero-server-args
    namespace: openshift-adp
  data:
    default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
    log-level: 'debug'
    default-snapshot-move-data: 'true'
```

shubham-pampattiwar: +1 @sseago

docs/design/unsupported-args-design.md
```diff
@@ -0,0 +1,132 @@
...
+The user must provide `DPA` annotation(s) with the `name` of the `ConfigMap` where the OADP configuration for the unsupported arguments is present.
+
+When ConfigMap is found, controller takes those values as the highest priority and replaces **all** of the other configuration options, even if they were defined within `DPA` schema for the section for which the `ConfigMap` was created.
+
+The Schema of an `ConfigMap` includes data that corresponds to the arguments passed to the relevant executable as in the below example:
+
+<a name="notes"></a>
+> ℹ️ **Note 1:** If an argument name is passed without `-` prefix or `--` prefix, then `--` is added to the argument name.
+
+> ℹ️ **Note 2:** If an argument value is passed without sourrounding `'`, then that single quote is added to the argument value.
```
mateusoliveira43, line 39: also check for surrounding `"`?

mpryc: That is an option, but then we need additional logic to:
1. Do the same as single quotes `'` so check if the `"` is at the beginning `and` end of the string
2. If there is we do nothing (not even putting single quote, because that would cause problems).

mpryc: Done.

docs/design/unsupported-args-design.md
mpryc: Ok so I have aligned with all the comments and updated design doc with function that implements it available for review at https://github.com/openshift/oadp-operator/pull/1423

The above function is commented with numerous test scenarios to present how this behaves with different ConfigMap value/key pairs. Please review and let me know if current design / implementation can be final, so I will integrate with oadp.

@sseago @shubham-pampattiwar @mateusoliveira43 ^^

weshayutin (approved):
/LGTM

shubham-pampattiwar (approved):

PR #1402, opened 2024-05-30
remove restartPolicy for OCP 4.16

weshayutin:
* How to test Install Virt and deploy the fedora vm
Alternatively allow the e2e virt ci to hit it.

## Why the changes were made

4.16 deployments will fail if restartPolicy is defined

 Install Virt and deploy the fedora vm
Alternatively allow the e2e virt ci to hit it.

### DIFF
--- a/tests/e2e/sample-applications/virtual-machines/fedora-todolist/fedora-todolist.yaml
+++ b/tests/e2e/sample-applications/virtual-machines/fedora-todolist/fedora-todolist.yaml
@@ -140,7 +140,6 @@ items:
               timeoutSeconds: 2
               successThreshold: 1
               failureThreshold: 40 # 40x30sec before restart pod
-            restartPolicy: Always
           - image: docker.io/curlimages/curl:8.5.0
             name: curl-tool
             command: ["/bin/sleep", "infinity"]


mateusoliveira43 (changes requested):
I think more changes are needed

If you run `grep -Iinr 'restartPolicy' .`, you can see that in `controllers/common.go` we enforce that restartPolicy is set (I believe that is passed to Velero and node-agent deployments)

EDIT
looking again, I think it is fine. The changes were made in `spec.template.spec.containers` and I was worried about doing changes in `spec.template.spec`

mateusoliveira43 (approved):
/lgtm
Tested in 4.16 cluster. This is basically a follow up of https://github.com/openshift/oadp-operator/commit/d06f5f71271c2915d94550a1b696b7d41ca7e2f8 , i.e., the problem is only in `spec.template.spec.containers` and not in `spec.template.spec`

weshayutin:
> /lgtm Tested in 4.16 cluster. This is basically a follow up of [d06f5f7](https://github.com/openshift/oadp-operator/commit/d06f5f71271c2915d94550a1b696b7d41ca7e2f8) , i.e., the problem is only in `spec.template.spec.containers` and not in `spec.template.spec`

ah.. nice catch. I have not yet been able to get virt up on OCP4.16, not working well on my cluster :/ We should hold the patch, perhaps it's not needed as you suggest.

mateusoliveira43:
no need for virt to test this one (just OCP 4.15+), the change was just in the deployment object 

Did not understand last comment. I got confused reviewing the PR, because I though more things needed chnages, but what was changed is enough (and needed to work in 4.15+)

mateusoliveira43:
restore succeeds but application not working flake

PR #1407, opened 2024-05-31
Remove Velero-plugin-for-csi from operator csv

shubham-pampattiwar:

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -856,8 +856,6 @@ spec:
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_GCP
                   value: quay.io/konveyor/velero-plugin-for-gcp:oadp-1.4
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_CSI
-                  value: quay.io/konveyor/velero-plugin-for-csi:oadp-1.3
-                - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
                   value: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
                 - name: RELATED_IMAGE_MUSTGATHER
                   value: registry.redhat.io/oadp/oadp-mustgather-rhel8:v1.2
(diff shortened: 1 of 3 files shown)


mateusoliveira43:
does this file `deploy/disconnected-prep.sh` need update as well?

shubham-pampattiwar:
> does this file `deploy/disconnected-prep.sh` need update as well?

We usually don't use the `deploy/disconnected.sh` script, downstream uses it and it resides downstream. But I still updated the script for reference purposes.

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -856,8 +856,6 @@ spec:
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_GCP
                   value: quay.io/konveyor/velero-plugin-for-gcp:oadp-1.4
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_CSI
-                  value: quay.io/konveyor/velero-plugin-for-csi:oadp-1.3
-                - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
```
sseago, line 860: Looks like the wrong name was removed. 858 above should be removed, this one should stay.

sseago (changes requested):

bundle/manifests/oadp-operator.clusterserviceversion.yaml
rayfordj: Agreed; wrong line removed. :sunglasses: 

FWIW, `make bundle` appears to do the correct thing when I run it locally

```
-                - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_CSI
+                - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
```

rayfordj: Resolved in https://github.com/openshift/oadp-operator/pull/1407/commits/bfc3f5e75954999bad1067af7f165177fd01cdbf

rayfordj (approved):
/lgtm

PR #1413, opened 2024-06-06
fix: update OADP version relation with plugins 

mateusoliveira43:
## Why the changes were made

Fix non admin schedule and csi plugin removed from OADP 1.4.

Check that all necessary places were modified.


sseago (approved):

kaovilai (approved):
/lgtm

PR #1423, opened 2024-06-12
Implementation function for unsupported args.

mpryc:
Draft implementation to calculate args from the provided configmap with test.

Related #1400

## Why the changes were made

To implement #1400 and allow later PR's to implement that functionality.

Please see #1430 for how to use the function. Also the function has many tests and docstring to let know developers how to use it.

Run unit tests.


mpryc:
This is ready to be reviewed and merged, so @shubham-pampattiwar can rework his PR on the node-agent args #1396 and I can work on server args.

pkg/common/common.go
```diff
@@ -239,3 +240,53 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
+//
+// Args:
+//
+//	cliSubCommand: The CLI subcommand as a string, for example 'server'
+//	configMap: A pointer to a corev1.ConfigMap containing key-value pairs.
+//
+// Returns:
+//
+//	A slice of strings representing the CLI arguments.
+func GenerateCliArgsFromConfigMap(cliSubCommand string, configMap *corev1.ConfigMap) []string {
```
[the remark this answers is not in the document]
mpryc: If it's surrounded by only one " or ' then it's ending up as string, so for ConfigMap:

```shell
# ConfigMap:
# myValue: 'myVal

# we will have broken arg:

$ velero server myValue=''myVal'
```

The above will break CLI argument, however for the:

```shell
# ConfigMap:
# myValue: "myVal

# we will have a valid arg:

$ velero server myValue='"myVal'
```

I am not sure if we want to mess parsing strings so much... ?

mateusoliveira43:
will this one be merged prior to https://github.com/openshift/oadp-operator/pull/1430 or just substituted by it and we can close this one?

pkg/common/common.go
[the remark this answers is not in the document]
shubham-pampattiwar: I am thinking, lets strip characters like `", ', `` from the values and add the quotes ourselves, WDYT ? @mpryc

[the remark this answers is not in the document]
mpryc: I think we should not, because we will end up having a lot of validations to make the value `shell` compatible. There are many other gotchas to consider and we would have to take them all what would create possible bugs.

I really do not want to manipulate what user entered as it's unsupported arg and it should go as is.

Here is just a short list of characters to be escaped that comes to my mind and there may be more, so really dunno if we want to go there...

```
'
"
\
$
`
(, )
*
?
[, ]
|
&
;
<
>
\n
space
```

mpryc:
> will this one be merged prior to #1430 or just substituted by it and we can close this one?

I think we should merge #1430 first. I did that in 2 PRs because another PR was to be implemented by @shubham-pampattiwar for the node-agent using common function from #1430.

pkg/common/common.go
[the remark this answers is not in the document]
shubham-pampattiwar: I understand your point but how likely is the user to make mistake regarding other characters in comparison with single quotes/double quotes ? The probability is the highest for the quotes characters as compared to others. Hence the thought.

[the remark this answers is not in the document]
mpryc: Right, but this is unsupported args which is already very custom, so the user will need to be double careful of what args are there, I believe the risk either way is very small. Removing single quite from the beginning if there is no ending and double quote is something we can do, so your call :)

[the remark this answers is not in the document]
mpryc: When I think more about this, what if someone enters:
```
ConfigMap:
  myarg: 'my value' value
```

This would be converted by us to:
```
$ cli server --myarg='my value' value'
```

Remember we removed first char as it didn't have ending ' and wrapped around single quotes to make sure it's string. Really I think we should not do any magic here with edits of what user inputs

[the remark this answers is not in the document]
mpryc: @shubham-pampattiwar updated function, tested with your ConfigMap as well the ConfigMap that consists of malformed data and everything is passed as string so there is no risk of breakage:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: oadp-unsupported-velero-server-args
  namespace: openshift-adp
data:
  default-volume-snapshot-locations: $(echo cat /proc/1/cmdline)
  features: EnableCSI
  uploader-type: kopia

```

pkg/common/common.go
```diff
@@ -239,3 +240,46 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
```
shubham-pampattiwar, line 265: I think you need to modify the function signature here to accept a variable number of string arguments by using variadic parameters. For node-agent daemonset the clisubcommand is `node-agent` and then `server` 
So the function signature should be:
```
func GenerateCliArgsFromConfigMap(cliSubCommand ...string, configMap *corev1.ConfigMap) []string
```

shubham-pampattiwar (changes requested):
change the function signature https://github.com/openshift/oadp-operator/pull/1423/files#r1653179562

pkg/common/common.go
mpryc: Are you sure this is the case?

I do not see another subcommand for the node-agent, just flags.

```shell
$ velero node-agent --help
Work with node-agent

Usage:

Flags:
  -h, --help   help for node-agent

Global Flags:
      --add_dir_header                   If true, adds the file directory to the header of the log messages
      --alsologtostderr                  log to standard error as well as files (no effect when -logtostderr=true)
      --colorized optionalBool           Show colored output in TTY. Overrides 'colorized' value from $HOME/.config/velero/config.json if present. Enabled by default
      --features stringArray             Comma-separated list of features to enable for this Velero process. Combines with values from $HOME/.config/velero/config.json if present
      --kubeconfig string                Path to the kubeconfig file to use to talk to the Kubernetes apiserver. If unset, try the environment variable KUBECONFIG, as well as in-cluster configuration
      --kubecontext string               The context to use to talk to the Kubernetes apiserver. If unset defaults to whatever your current-context is (kubectl config current-context)
      --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
      --log_dir string                   If non-empty, write log files in this directory (no effect when -logtostderr=true)
      --log_file string                  If non-empty, use this log file (no effect when -logtostderr=true)
      --log_file_max_size uint           Defines the maximum size a log file can grow to (no effect when -logtostderr=true). Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
      --logtostderr                      log to standard error instead of files (default true)
  -n, --namespace string                 The namespace in which Velero should operate (default "velero")
      --one_output                       If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
      --skip_headers                     If true, avoid header prefixes in the log messages
      --skip_log_headers                 If true, avoid headers when opening log files (no effect when -logtostderr=true)
      --stderrthreshold severity         logs at or above this threshold go to stderr when writing to files and stderr (no effect when -logtostderr=true or -alsologtostderr=false) (default 2)
  -v, --v Level                          number for the log level verbosity
      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging

```

sseago: @mpryc `velero node-agent server ...` on the node agent vs `velero server ...` on the velero pod

mpryc: Updated the function to allow multiple artguments (here the order in which they are passed to the function matters, so `"node-agent", "server"` will be different from `"server", "node-agent"`) + added sorting of flags so the result is consistent across different function calls.

pkg/common/common.go
```diff
@@ -239,3 +241,57 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
+// Args:
+//
+//		configMap: A pointer to a corev1.ConfigMap containing key-value pairs.
+//		cliSubCommand: The CLI subcommand(s) as a string, for example 'server'
+//	                or 'node-agent', 'server'
+//
+// Returns:
+//
+//	A slice of strings representing the CLI arguments.
+func GenerateCliArgsFromConfigMap(configMap *corev1.ConfigMap, cliSubCommand ...string) []string {
```
mateusoliveira43, line 267: would remove `cliSubCommand` arg from function to make it simpler

pkg/common/common.go
```diff
@@ -239,3 +241,57 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
 	return corev1.PullAlways, nil
 }
+
+// GenerateCliArgsFromConfigMap generates CLI arguments from a ConfigMap.
+//
+// This function takes a ConfigMap and a CLI subcommand(s), and returns a slice of strings representing
+// the CLI arguments from the subcommand(s) followed by the arguments from the ConfigMap.
+// The function processes each key-value pair in the ConfigMap as follows:
+//
+//  1. If the ConfigMaps' value starts and ends with single (') or double (") quotes, it is left unchanged.
```
mateusoliveira43, line 251: this is not valid anymore, right?

pkg/common/common.go
```diff
@@ -239,3 +241,57 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
+			// Convert true/false to lowercase if not surrounded by quotes - boolean
+			value = strings.ToLower(value)
+		}
+
+		keyValueArgs = append(keyValueArgs, fmt.Sprintf("%s=%s", key, value))
+
+	}
+	// We ensure the flags are alphabetically sorted, so they
+	// are always added to the cliSubCommand(s) the same way
+	sort.Strings(keyValueArgs)
```
shubham-pampattiwar, line 291: Why did we have to add sort here ? Unit tests were failing ?

docs/design/unsupported-args-design.md
```diff
@@ -36,9 +36,7 @@ The Schema of an `ConfigMap` includes data that corresponds to the arguments pas
 <a name="notes"></a>
 > ℹ️ **Note 1:** If an argument name is passed without `-` prefix or `--` prefix, then `--` is added to the argument name.
 
-> ℹ️ **Note 2:** If an non-boolean argument value is passed without `'` or `"` sourrounding, then the single quote is added to the argument value.
-
-> ℹ️ **Note 3:** An boolean argument value `true` or `false` is always converted to lower-case and not sourrounded by single quote `'`.
+> ℹ️ **Note 2:** An boolean argument value `true` or `false` is always converted to lower-case.
 
 > ℹ️ **Note 4:** An argument value is always combined with the argument name using `=` character in between.
```
mateusoliveira43, line 41: need to change to 3

docs/design/unsupported-args-design.md
```diff
@@ -50,15 +48,15 @@ The Schema of an `ConfigMap` includes data that corresponds to the arguments pas
     name: oadp-unsupported-velero-server-args
     namespace: openshift-adp
   data:
-    default-volume-snapshot-locations: 'aws:backups-primary,azure:backups-secondary'
-    log-level: 'debug'
-    default-snapshot-move-data: 'true'
+    default-volume-snapshot-locations: aws:backups-primary,azure:backups-secondary'
```
mateusoliveira43, line 51: remove quote in the end?

pkg/common/common.go
mpryc: how would you make it generic then?

We want to pass part of the args from the reconcile and second part from the configmap

pkg/common/common.go
mpryc: it is valid, because we don't remove it.

pkg/common/common.go
mpryc: yes I observed the tests were failing as the values from configmap were random.

Also it's nice because they are passed to the CLI always the same. Of course this sort applies only to the flag parts, so the Cobra is happy and we do have control over first args `node-agent server`

docs/design/unsupported-args-design.md
mpryc: kk

docs/design/unsupported-args-design.md
mpryc: kk

pkg/common/common.go
mpryc: will remove that point because we don't modify so there is no need to document that

pkg/common/common.go
mateusoliveira43: just append return from this function to an array that already contains the sub commands when using it in controllers

mpryc: I prefer this command to contain all args and sub-args, so we don't have mistakenly two implementations of the cliSubCommands within two different reconcile loops.

shubham-pampattiwar (dismissed):

mpryc:
All the tests passed, are we ok with the current implementation? The only outstanding question is from @mateusoliveira43 which I've answered with different opinion.

pkg/common/common.go
```diff
@@ -239,3 +241,56 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
+//	                or 'node-agent', 'server'
+//
+// Returns:
+//
+//	A slice of strings representing the CLI arguments.
+func GenerateCliArgsFromConfigMap(configMap *corev1.ConfigMap, cliSubCommand ...string) []string {
+
+	// cliSubCommand elements to be a separate items in args slice
+	args := make([]string, len(cliSubCommand))
+	copy(args, cliSubCommand)
```
kaovilai, line 270: I think this isn't necessary. cliSubCommand should already be `[]string`. You can append to and return cliSubCommand param directly.

mpryc: @kaovilai do you want me to change this or should we agree that immutability is also important here and leave as is ?

docs/design/unsupported-args-design.md
```diff
@@ -36,11 +36,9 @@ The Schema of an `ConfigMap` includes data that corresponds to the arguments pas
 <a name="notes"></a>
 > ℹ️ **Note 1:** If an argument name is passed without `-` prefix or `--` prefix, then `--` is added to the argument name.
 
-> ℹ️ **Note 2:** If an non-boolean argument value is passed without `'` or `"` sourrounding, then the single quote is added to the argument value.
+> ℹ️ **Note 2:** An boolean argument value `true` or `false` is always converted to lower-case.
```
sseago, line 39: nit: s/An boolean/A boolean/

pkg/common/common.go
```diff
@@ -239,3 +241,56 @@ func GetImagePullPolicy(image string) (corev1.PullPolicy, error) {
...
+
+	var keyValueArgs []string
+	// Iterate through each key-value pair in the ConfigMap
+	for key, value := range configMap.Data {
+		// Ensure the key is prefixed by "--" if it doesn't start with "--" or "-"
+		if !strings.HasPrefix(key, "--") && !strings.HasPrefix(key, "-") {
```
sseago, line 276: Checking for "--" is redundant here. If `!strings.HasPrefix(key, "-")` evaluates to `true`, then `!strings.HasPrefix(key, "--")` always will as well.

pkg/common/common_test.go
```diff
@@ -200,3 +200,127 @@ func TestGetImagePullPolicy(t *testing.T) {
...
+			expectedArgs: []string{
+				"server",
+				"--another-key-mix-letters=true",
+				"--default-snapshot-move=true",
+				"--key-no-prefix=false",
+				"--string-not-bool='False'",
+			},
+		},
+		{
+			name:          "All arguments with spaces, some without sigle quotes",
```
sseago, line 231: nit: s/sigle/single/

pkg/common/common_test.go
```diff
@@ -200,3 +200,127 @@ func TestGetImagePullPolicy(t *testing.T) {
...
+				"--default-volume-snapshot-locations=aws:backups-primary, azure:backups-secondary",
+				"--log-level='debug'",
+			},
+		},
+		{
+			name:          "Preserve single and double '-' as key prefix",
+			cliSubCommand: []string{"server"},
+			configMap: &corev1.ConfigMap{
+				Data: map[string]string{
+					"-default-volume-snapshot-locations": "aws:backups-primary, azure:backups-secondary",
```
sseago, line 250: We should probably test with "-d" or something with one char here, just because single-dash flags are only valid with single-char short names.

pkg/common/common_test.go
```diff
@@ -200,3 +200,127 @@ func TestGetImagePullPolicy(t *testing.T) {
...
+					"--log-level":                        "debug",
+					"-string-bool":                       "False",
```
sseago, line 252: same as above

pkg/common/common_test.go
mpryc: So you are suggesting that if the value name:
 1) starts with one `-` it should always be followed with one character
 2) is exactly one character we should append only one `-`

So valid ConfigMap:

```yaml
"-n": "mynamespace"
"n": "mynamespace"

Translates to:
-n=namespace
```

Invalid ConfigMap:
```yaml
"--n": "mynamespace"
```


?

mpryc: @sseago 

I think this one breaks the rule of one dash / one character:

```
  -v, --v Level                          number for the log level verbosity

```

mateusoliveira43: lets not validate configmap

I think Scott just suggested a more real life test

sseago: @mpryc No, I'm not suggesting that we modify the code. one char flag is also a valid full name, so `-` isn't guaranteed to work. I'm just saying that `-foo` will never be valid to pass to velero, while `-f` might be -- so change the test examples to use single char flag name when the example prepends a single `-`.

pkg/common/common.go
kaovilai: Immutability is important, I'm saying, you shouldn't need to make a new copy here as the cliSubcommand param is already a new copy. **The only exception is if the argument is followed by ...**, then **it is passed unchanged**.

[ref](https://go.dev/ref/spec#Passing_arguments_to_..._parameters)

> Passing arguments to ... parameters
> If f is [variadic](https://go.dev/ref/spec#Function_types) with a final parameter p of type ...T, then within f the type of p is equivalent to type []T. If f is invoked with no actual arguments for p, the value passed to p is nil. Otherwise, the value passed is a new slice of type []T with a new underlying array whose successive elements are the actual arguments, which all must be [assignable](https://go.dev/ref/spec#Assignability) to T. The length and capacity of the slice is therefore the number of arguments bound to p and may differ for each call site.
> 
> Given the function and calls
> 
> func Greeting(prefix string, who ...string)
> Greeting("nobody")
> Greeting("hello:", "Joe", "Anna", "Eileen")
> within Greeting, who will have the value nil in the first call, and []string{"Joe", "Anna", "Eileen"} in the second.
> 
> If the final argument is assignable to a slice type []T and is followed by ..., it is passed unchanged as the value for a ...T parameter. In this case no new slice is created.
> 
> Given the slice s and call
> 
> s := []string{"James", "Jasmine"}
> Greeting("goodbye:", s...)
> within Greeting, who will have the same value as s with the same underlying array.

pkg/common/common_test.go
sseago: `-v` and `--v` doesn't break what I suggested -- which is "-" is followed by a name of only one char. "--" can be followed by one or more -- the practical reason for this is that the "=" is optional with shortcuts. i.e. `-oyaml` is equivalent to `-o=yaml` -- the only way `-oyaml` works when parsed is that the parser knows that only the first char is the flag shortname, and what follows is the value -- but we're not using that shortcut here, as we're always using the `=` for consistency. That's just a shortcut for users to save keystrokes when typing commands manually.

pkg/common/common.go
kaovilai: More reading material
https://stackoverflow.com/a/47296376/4590470

mpryc: @kaovilai updated and removed copy()

docs/design/unsupported-args-design.md
mpryc: thanks, resolved.

pkg/common/common.go
mpryc: Thanks, resolved.

pkg/common/common_test.go
mpryc: @sseago ok here I did modify a little the logic and now single name characters from the ConfigMap by default are prefixed with single `-`. Longer are prefixed by default with `--`.  Of course user have an option to prefix single character with double `--` if the double `--` are part of the key name in ConfigMap, so the valid are:
```
ConfigMap:
 "-v": "3"
 "--v": 3"
 "v": "3"

The first and last will end up being arg (example for velero server):
$ velero server -v=3

The middle one will end up being:
$ velero server --v=3
```

pkg/common/common_test.go
mpryc: fixed and added more options to cover single quotes.

pkg/common/common_test.go
sseago: While this won't really make a difference with the current Velero args, since I think `v` is the only one that's the same for both, I don't think it's a safe assumption that if a user enters a single-name arg that it's a shortcut rather than a name. It's possible for an arg to be one char but without a defined shortcut -- and in that case, we're forcing the user to specify name as "--v". I think it's safer to always prepend "--" and document that if the user intends to use the shortcut name instead, they must prepend the single `-`

sseago: @mpryc ^^

mpryc: @sseago updated, so currently it behaves exactly as you wrote. Thanks.

mpryc:
Note this PR does not cover developer documentation which will be written later stages once #1430 is merged and once @shubham-pampattiwar implements `node-agent server` part, so we have consistent and complete developer doc to cover all unsupported args.

mateusoliveira43 (approved):

sseago (approved):

kaovilai (approved):

PR #1435, opened 2024-06-20
up kubevirt to 0.7

kaovilai:

Why the changes were made
We're using a VERY old kubevirt plugin in prow tests and dev setup

How to test the changes made
make deploy-olm and check the deployment

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -170,7 +170,7 @@ metadata:
       Cloud Provider,Developer Tools,Modernization & Migration,OpenShift Optional,Storage
     certified: "false"
     containerImage: quay.io/konveyor/oadp-operator:latest
-    createdAt: "2024-05-29T17:22:16Z"
+    createdAt: "2024-06-20T18:28:23Z"
     description: OADP (OpenShift API for Data Protection) operator sets up and installs
       Data Protection Applications on the OpenShift platform.
     features.operators.openshift.io/cnf: "false"
@@ -917,7 +917,7 @@ spec:
                 - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_GCP
                   value: quay.io/konveyor/velero-plugin-for-gcp:latest
                 - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
-                  value: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
+                  value: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
                 - name: RELATED_IMAGE_MUSTGATHER
                   value: registry.redhat.io/oadp/oadp-mustgather-rhel8:v1.2
                 - name: RELATED_IMAGE_NON_ADMIN_CONTROLLER
@@ -1067,7 +1067,7 @@ spec:
     name: velero-plugin-for-microsoft-azure
   - image: quay.io/konveyor/velero-plugin-for-gcp:latest
     name: velero-plugin-for-gcp
-  - image: quay.io/konveyor/kubevirt-velero-plugin:v0.2.0
+  - image: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
     name: kubevirt-velero-plugin
   - image: registry.redhat.io/oadp/oadp-mustgather-rhel8:v1.2
     name: mustgather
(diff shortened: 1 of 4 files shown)


weshayutin (approved):
/LGTM

weshayutin:
@mrnold FYI

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -170,7 +170,7 @@ metadata:
       Cloud Provider,Developer Tools,Modernization & Migration,OpenShift Optional,Storage
     certified: "false"
     containerImage: quay.io/konveyor/oadp-operator:latest
-    createdAt: "2024-05-29T17:22:16Z"
+    createdAt: "2020-09-08T12:21:00Z"
```
mateusoliveira43, line 173: NIT: this should not be changed by make bundle (but it is ok, wont break anything)

kaovilai: well.. it is a change.. it is a valid change, it is a new creation of CSV :)

Unless this is the only change.. it is technically a new CSV.

rayfordj: agreed on the adjustment due to valid change, but... createdAt **2020**?? :laughing:

rayfordj: At least have it close to `2024-06-20`, no?

rayfordj, line 173: How about?
```suggestion
    createdAt: "2024-06-20T18:18:07Z"
```

mateusoliveira43: oh I though it was updated from the command, but I think tiger just copied what was our create at (same as in oadp-1.3 branch for example)

would be good practice to update once in a while (every release?), but would discuss a little how/when change the date

kaovilai: good catch.. with the date..

rayfordj: I think how/when to update should happen whenever there is a legitimate change to CSV -- as is the case here with the RELATED_IMAGE update.

Bumping `createdAt` for the sake of updating the timestamp with no other changes creates additional burden downstream reconciling against the CSV because we look at its digest and fail the build if it doesn't match what we expect.

rayfordj: Updating  `createdAt:` value at Z-stream intervals (with no other accompany CSV changes) is not unreasonable.  If that is something desired upstream, I won't fight it, but it would introduce additional effort, however nominal, downstream that does not currently exist. :sunglasses:

rayfordj (approved):
/lgtm

shubham-pampattiwar (approved):

shubham-pampattiwar:
@kaovilai seperate PR for 1.4 or cherry-pick ?

rayfordj:
> @kaovilai seperate PR for 1.4 or cherry-pick ?
@shubham-pampattiwar , cherry-pick comment: https://github.com/openshift/oadp-operator/pull/1435#issuecomment-2181257003

shubham-pampattiwar:
> > @kaovilai seperate PR for 1.4 or cherry-pick ?
> 
> @shubham-pampattiwar , cherry-pick comment: [#1435 (comment)](https://github.com/openshift/oadp-operator/pull/1435#issuecomment-2181257003)

I need a new pair of glasses XD

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mateusoliveira43: will try to think a way to update createdAt always there is change in CSV, and enforce CI to not let just the date change without any other changes 

also enforce date to be higher than previous one heheheh

mateusoliveira43 (changes requested):
I think this image will not work. I tried with DPA containing kubevirt plugin, and velero container never starts running

Screenshot from 2024-06-25 10-22-05

kaovilai:
not working in oadp-1.3 e2e either :/

kaovilai:
hopefully just arch related

kaovilai:
```
❯ docker image inspect quay.io/konveyor/kubevirt-velero-plugin:v0.7.0 | grep Architect
        "Architecture": "arm64",
```
🤦

kaovilai:
fixes pushed. image should be amd64 now.

will figure out multi arch cron in github actions later

/retest

kaovilai:
PRs now passing en-mass so I think we're good here. @mateusoliveira43 do you need a hold?

PR #1439, opened 2024-06-21
CI: add single test for powered-off CirrOS VM

mrnold:
## Why the changes were made

Add a test for backup/restore on powered-off VMs. This is the first step in addressing [OADP-4231](https://issues.redhat.com/browse/OADP-4231).

Run virtualization E2E tests with: make TEST_VIRT=true test-e2e

### DIFF
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -112,7 +112,7 @@ func prepareBackupAndRestore(brCase BackupRestoreCase, updateLastInstallTime fun
 	return backupName, restoreName
 }
 
-func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expectedBackupErr, expectedRestoreErr error, updateLastBRcase func(brCase ApplicationBackupRestoreCase), updateLastInstallTime func()) {
+func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expectedErr error, updateLastBRcase func(brCase ApplicationBackupRestoreCase), updateLastInstallTime func()) {
 	updateLastBRcase(brCase)
 
 	// create DPA
@@ -149,7 +149,7 @@ func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expecte
 	gomega.Eventually(lib.AreApplicationPodsRunning(kubernetesClientForSuiteRun, brCase.Namespace), timeoutMultiplier*time.Minute*9, time.Second*5).Should(gomega.BeTrue())
 
 	// do the backup for real
-	nsRequiredResticDCWorkaround := runBackup(brCase.BackupRestoreCase, backupName, expectedBackupErr)
+	nsRequiredResticDCWorkaround := runBackup(brCase.BackupRestoreCase, backupName)
 
 	// uninstall app
 	log.Printf("Uninstalling app for case %s", brCase.Name)
@@ -162,7 +162,7 @@ func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expecte
 	updateLastInstallTime()
 
 	// run restore
-	runRestore(brCase.BackupRestoreCase, backupName, restoreName, nsRequiredResticDCWorkaround, expectedRestoreErr)
+	runRestore(brCase.BackupRestoreCase, backupName, restoreName, nsRequiredResticDCWorkaround)
 
 	// verify app is running
 	gomega.Eventually(lib.AreAppBuildsReady(dpaCR.Client, brCase.Namespace), timeoutMultiplier*time.Minute*3, time.Second*5).Should(gomega.BeTrue())
@@ -176,7 +176,7 @@ func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expecte
 	}
 }
 
-func runBackup(brCase BackupRestoreCase, backupName string, expectedErr error) bool {
+func runBackup(brCase BackupRestoreCase, backupName string) bool {
 	// Run optional custom verification
 	if brCase.PreBackupVerify != nil {
 		log.Printf("Running pre-backup function for case %s", brCase.Name)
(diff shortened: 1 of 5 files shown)


tests/e2e/backup_restore_suite_test.go
```diff
@@ -227,7 +227,7 @@ func runBackup(brCase BackupRestoreCase, backupName string) bool {
 	return nsRequiresResticDCWorkaround
 }
 
-func runRestore(brCase BackupRestoreCase, backupName, restoreName string, nsRequiresResticDCWorkaround bool) {
+func runRestore(brCase BackupRestoreCase, backupName, restoreName string, nsRequiresResticDCWorkaround bool, expectedErr error) {
```
mateusoliveira43, line 230: to avoid changes in backup_restore_suite_test.go and in must-gather_suite_test.go, I would copy this function to virt_backup_restore_suite_test.go with a TODO to be removed after problem is fixed (this is planed to be fixed in on our sprints, right?)

tests/e2e/backup_restore_suite_test.go
```diff
@@ -176,7 +176,7 @@ func runApplicationBackupAndRestore(brCase ApplicationBackupRestoreCase, expecte
 	}
 }
 
-func runBackup(brCase BackupRestoreCase, backupName string) bool {
+func runBackup(brCase BackupRestoreCase, backupName string, expectedErr error) bool {
```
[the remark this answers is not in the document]
mrnold: I guess technically not, since I haven't added any tests with errors in the backup stage. But it looked weird to have it all plumbed through runRestore and not runBackup.

weshayutin:
let's revive this work asap. Thank you!!

tests/e2e/backup_restore_suite_test.go
mrnold: Okay, I tried this out in the latest push. The changes are definitely more contained now.

mateusoliveira43 (dismissed):
changes seems fine, CI is happy with it

Would just increase a timeout for VM terminating, saw it timeout in my PR as well

mateusoliveira43 (approved):

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -73,18 +92,26 @@ func runVmBackupAndRestore(brCase VmBackupRestoreCase, expectedErr error, update
 		time.Sleep(brCase.InitDelay)
 	}
 
+	// Check if this VM should be running or stopped for this test.
+	// Depend on pre-backup verification function to poll state.
+	if brCase.PowerState == "Stopped" {
+		log.Print("Stopping VM before backup as specified in test case.")
+		err = v.StopVm(brCase.Namespace, brCase.Name)
+		gomega.Expect(err).To(gomega.BeNil())
```
kaovilai, line 100: Do we need any waits or are stops instant?

mrnold: I have a wait as the pre-backup verification function in the test table. I forget why I thought that made sense.

kaovilai: +1 makes sense to me... just didn't connect the dots that there could be waits outside this function.

kaovilai (approved):

shubham-pampattiwar (approved):

mateusoliveira43:
do we need to cherry-pick this oadp-1.4?

mrnold:
> do we need to cherry-pick this oadp-1.4?

Yes, I think this was the plan. It doesn't merge cleanly though, so I will do it without the bot.

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test.yaml
```diff
@@ -33,6 +33,10 @@ items:
               - disk:
                   bus: virtio
                 name: rootdisk
+            firmware:
+              bootloader:
+                efi:
+                  secureBoot: false
```
mpryc, line 39: Why is this needed?

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -188,3 +230,45 @@ var _ = ginkgov2.Describe("VM backup and restore tests", ginkgov2.Ordered, func(
...
+	if brCase.RestoreErr != nil {
+		gomega.Expect(succeeded).To(gomega.Equal(false))
+		gomega.Expect(err).NotTo(gomega.BeNil())
+		gomega.Expect(err.Error()).To(gomega.ContainSubstring(brCase.RestoreErr.Error()))
+	} else {
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		gomega.Expect(succeeded).To(gomega.Equal(true))
+	}
+
+	if nsRequiresResticDCWorkaround {
```
mpryc, line 264: Is this still needed?

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -37,10 +39,27 @@ func getLatestCirrosImageURL() (string, error) {
 	return imageURL, nil
 }
 
+func vmPoweredOff(vmnamespace, vmname string) VerificationFunction {
+	return VerificationFunction(func(ocClient client.Client, namespace string) error {
+		isOff := func() bool {
+			status, err := lib.GetVmStatus(dynamicClientForSuiteRun, vmnamespace, vmname)
+			if err != nil {
+				log.Printf("Error getting VM status: %v", err)
```
mpryc, line 47: should there be a return on error?

tests/e2e/virt_backup_restore_suite_test.go
mrnold: I think possibly not, but I'm not really sure. I just copied all of this from the original runRestore, since it is planned to be temporary.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: I think this could include an error, but I don't know that it would meaningfully improve the Eventually. I guess there might be cases where it would quit faster than the timeout, but then it would also need to check for errors that it should be expected to ride through. I will see if I can think of something.

mpryc: I just see some possible logic problem. Saying first that there is err getting status and then without return logging off vm status and returning "Stopped", for me if there was error getting status the return of the status should not be "Stopped" but something unknown

PR #1440, opened 2024-06-21
CI: Search virtualization package manifest for stable channel.

mrnold:
## Why the changes were made

Fixes virtualization CI failures with operator versions like 4.99.0-0.1713770549. The operator installation code only looked at the first channel in the package manifest, which is sometimes "dev-preview" instead of "stable". This change looks explicitly for "stable".  

Run virtualization E2E tests against a cluster with this condition. It should fail without the fix, and proceed when the fix is applied.

### DIFF
--- a/tests/e2e/lib/virt_helpers.go
+++ b/tests/e2e/lib/virt_helpers.go
@@ -126,12 +126,27 @@ func getCsvFromPackageManifest(dynamicClient dynamic.Interface, name string) (st
 		return "", nil, errors.New("no channels listed in package manifest " + name)
 	}
 
-	firstChannel, ok := channels[0].(map[string]interface{})
-	if !ok {
-		return "", nil, errors.New("failed to read first channel from package manifest " + name)
+	var stableChannel map[string]interface{}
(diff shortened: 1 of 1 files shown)


weshayutin (commented):
NICE! Thank you @mrnold

mateusoliveira43 (approved):
All beforeAll jobs for virt passed

:white_check_mark: 4.14: Current CSV is: kubevirt-hyperconverged-operator.v4.14.6. Current operator version is: 4.14.6

:white_check_mark: 4.15: Current CSV is: kubevirt-hyperconverged-operator.v4.15.2. Current operator version is: 4.15.2

:warning: 4.16: Current CSV is: kubevirt-hyperconverged-operator.v4.14.0. Current operator version is: 4.14.0 (before this change, it was also installing this, so I think it is safe. And once 4.16 is released, it should automatically use a 4.16 version)

mateusoliveira43:
restore succeeds, but application not working flake (and PR does not touch that code)

weshayutin (approved):
/LGTM

PR #1437, opened 2024-06-21
[oadp-1.3] fix: 4.15 CI error with restart policy

mateusoliveira43:
## Why the changes were made

Fix CI for OCP 4.15+ as done in https://github.com/openshift/oadp-operator/pull/1365

## How to test the changes made

Check CI logs, no restartPolicy errors should happen.

PR #1441, opened 2024-06-22
oadp-1.3: Make deploy-olm work on 4.16 and golang 1.22 built-ins

kaovilai:
- **`make deploy-olm` for 4.16** adding ` --security-context-config restricted`
- **Upgrade controller-gen so `make manifests` runs**

## Why the changes were made
https://github.com/kubernetes-sigs/controller-tools/issues/945


mateusoliveira43:
Do we want these changes in oadp-1.3? Since its Go 1.20 (ref https://github.com/openshift/oadp-operator/blob/oadp-1.3/go.mod#L3) this is not needed, but it makes development easier

If there is an easy way to install multiple Go versions in system, would try this way (in Python world, there is a tool that lets you install a specific Python version on a folder only, for example)

What I have been doing to work on old Go versions is to use containers. Example:

Create container image from this Dockerfile 
```dockerfile
FROM golang:1.20

ARG USER_NAME=dev

RUN useradd --create-home "$USER_NAME"

USER $USER_NAME

WORKDIR /home/$USER_NAME
```
```sh
docker image build --tag oadp-development -f dev.Dockerfile .
```
Then, connect to container shell
```sh
docker container run \
  -u $(id -u):$(id -g) -h container \
  -v $PWD:/home/dev/repo \
  -it --rm oadp-development bash
```
and run Go command inside it

kaovilai:
> If there is an easy way to install multiple Go versions in system, would try this way (in Python world, there is a tool that lets you install a specific Python version on a folder only, for example)

Pre golang 1.21: https://github.com/go-simpler/goversion
Post golang 1.21: `GOTOOLCHAIN=go1.18 go version`

However GOTOOLCHAIN still does not download older go .../libexec/ builtins libs, thus it's not a complete replacement of `goversion`

kaovilai (commented):
controller-gen bump can be removed if you use golang 1.20.

operator-sdk bump is mandatory if we want to use 
```Makefile
$(OPERATOR_SDK) run bundle $(THIS_BUNDLE_IMAGE) --namespace $(OADP_TEST_NAMESPACE) --security-context-config restricted
```

kaovilai:
mongo app needs to remove restartPolicy as well from its deploy template spec on non-init containers.
```
error creating or updating resource: apps/v1, Kind=Deployment; name: mysql; error: Deployment.apps "mysql" is invalid: spec.template.spec.containers[0].restartPolicy: Forbidden: may not be set for non-init containers; retrying for 1 more times

```

kaovilai:
restart policy addressed in: https://github.com/openshift/oadp-operator/pull/1437

sseago:
I thought 4.16 wasn't supported on oadp-1.3? Do we need 4.16 changes here?

kaovilai:
Well I don't have a cluster older than 4.16 to test on :/

Makefile
```diff
@@ -328,6 +328,8 @@ bundle: manifests kustomize operator-sdk ## Generate bundle manifests and metada
 	# TODO: update CI to use generated one
 	cp bundle.Dockerfile build/Dockerfile.bundle
 	GOFLAGS="-mod=mod" $(OPERATOR_SDK) bundle validate ./bundle
+	sed -e 's/    createdAt: .*/$(shell grep -I '^    createdAt: ' bundle/manifests/oadp-operator.clusterserviceversion.yaml)/' bundle/manifests/oadp-operator.clusterserviceversion.yaml > bundle/manifests/oadp-operator.clusterserviceversion.yaml.tmp
```
mpryc, line 331: You could use `sed -i[SUFFIX]` or `sed --in-place=[SUFFIX]` which will create backup for you.

mpryc (changes requested):

Makefile
kaovilai: ICYMI sed --in-place [doesn't work well on macos ;P](https://github.com/openshift/oadp-operator/pull/1405#issuecomment-2142425360)

kaovilai: Also per https://stackoverflow.com/questions/7573368/in-place-edits-with-sed-on-os-x

> The Unix Way™ would (IMHO) be to use sed non-destructively, test that it exited cleanly, and only then remove the extraneous file.

kaovilai: oh yeah.. with suffix should work. I guess that's shorter.

Dockerfile
kaovilai, line 1: Contains cherry-pick https://github.com/openshift/oadp-operator/pull/1478/commits/139f4f56281c55cfac6c66b930d6942e66dfb536 from https://github.com/openshift/oadp-operator/pull/1478 to deploy and test on 4.16 ARM courtesy of Mr. Green Helmet.

mateusoliveira43:
can we discuss this in scrum? I would try another approach for this https://github.com/openshift/oadp-operator/pull/1441#issuecomment-2185065402 

Ans also, keeping production requirements https://github.com/openshift/oadp-operator/pull/1441#issuecomment-2191932705

kaovilai:
adding to scrum

kaovilai:
Scrum concluded that this is not going forward.. block 4.16 deploy-olm install on oadp-1.3 via makefile.

PR #1430, opened 2024-06-19
Unsupported server args

mpryc:
## Why the changes were made

Add unsupported args for the velero server via ConfigMap within the DPA namespace.

## How to test the changes made

1. Run the tests
2. Deploy oadp with this change, then

- Create ConfigMap with some name
- Inside ConfigMap create key/value pairs, where key will correspond to the velero server argument name and value is it's value
- Create DPA with annotation (replace <ConfigMap-Name> to adopt to your ConfigMap:
  `oadp.openshift.io/unsupported-velero-server-args=<ConfigMap-Name>`
- Verify the Velero server pod runs velero with the expected flags

Notes:
 - NonExisting ConfigMap referenced by the annotation will report error
 - Empty ConfigMap name will not cause override

PR #1450, opened 2024-07-02
Fix OADP-4110:Unsupported Node-Agent Server args

shubham-pampattiwar:
## Why the changes were made

Add unsupported server args for the node-agent daemon set via ConfigMap within the DPA namespace.

Fixes https://issues.redhat.com/browse/OADP-4110

- Deploy OADP Operator using the PR
- Create ConfigMap with key/value pairs, where key will correspond to the node-agent server argument name and value is it's value
- Create DPA with annotation (replace <ConfigMap-Name> to adopt to your ConfigMap:
  `oadp.openshift.io/unsupported-node-agent-server-args=<ConfigMap-Name>`
- Verify the node-agent server pod runs with the expected server args

Notes:
 - NonExisting ConfigMap referenced by the annotation will report error
 - Empty ConfigMap name will not cause override

### DIFF
--- a/controllers/nodeagent.go
+++ b/controllers/nodeagent.go
@@ -335,8 +335,8 @@ func (r *DPAReconciler) customizeNodeAgentDaemonset(dpa *oadpv1alpha1.DataProtec
 			setContainerDefaults(nodeAgentContainer)
 
 			if configMapName, ok := dpa.Annotations[common.UnsupportedNodeAgentServerArgsAnnotation]; ok {
-				unsupportedServerArgsCM := corev1.ConfigMap{}
 				if configMapName != "" {
+					unsupportedServerArgsCM := corev1.ConfigMap{}
 					if err := r.Get(r.Context, types.NamespacedName{Namespace: dpa.Namespace, Name: configMapName}, &unsupportedServerArgsCM); err != nil {
 						return nil, err
 					}
--- a/controllers/nodeagent.go
+++ b/controllers/nodeagent.go
@@ -29,8 +29,6 @@ const (
 	FsRestoreHelperCM     = "fs-restore-action-config"
 	HostPods              = "host-pods"
 	HostPlugins           = "host-plugins"
-
-	UnsupportedNodeAgentServerArgsAnnotation = "oadp.openshift.io/unsupported-node-agent-server-args"
 )
 
 var (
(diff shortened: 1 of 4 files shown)


controllers/nodeagent.go
```diff
@@ -369,9 +371,35 @@ func (r *DPAReconciler) customizeNodeAgentDaemonset(dpa *oadpv1alpha1.DataProtec
 	if ds.Spec.RevisionHistoryLimit == nil {
 		ds.Spec.RevisionHistoryLimit = pointer.Int32(10)
 	}
+
+	if err := r.applyUnsupportedNodeAgentServerArgsOverride(dpa, nodeAgentContainer); err != nil {
```
mateusoliveira43, line 375: I would this function along container changes 

would refactor line 305 to be like this
```go

	// fetch nodeAgent container in order to customize it
	for _, container := range ds.Spec.Template.Spec.Containers {
		if container.Name == common.NodeAgent {
			// append certs volume mount
			container.VolumeMounts = append(container.VolumeMounts, corev1.VolumeMount{
				Name:      "certs",
				MountPath: "/etc/ssl/certs",
			})
			// append PodConfig envs to nodeAgent container
			if useResticConf {
				if dpa.Spec.Configuration.Restic.PodConfig != nil && dpa.Spec.Configuration.Restic.PodConfig.Env != nil {
					container.Env = common.AppendUniqueEnvVars(container.Env, dpa.Spec.Configuration.Restic.PodConfig.Env)
				}
			} else if dpa.Spec.Configuration.NodeAgent.PodConfig != nil && dpa.Spec.Configuration.NodeAgent.PodConfig.Env != nil {
				container.Env = common.AppendUniqueEnvVars(container.Env, dpa.Spec.Configuration.NodeAgent.PodConfig.Env)
			}

			// append env vars to the nodeAgent container
			container.Env = common.AppendUniqueEnvVars(container.Env, proxy.ReadProxyVarsFromEnv())

			container.SecurityContext = &corev1.SecurityContext{
				Privileged: pointer.Bool(true),
			}

			imagePullPolicy, err := common.GetImagePullPolicy(getVeleroImage(dpa))
			if err != nil {
				r.Log.Error(err, "imagePullPolicy regex failed")
			}

			container.ImagePullPolicy = imagePullPolicy
			setContainerDefaults(&container)

			if err := r.applyUnsupportedNodeAgentServerArgsOverride(dpa, &container); err != nil {
				return nil, err
			}
			break
		}
	}
```

controllers/nodeagent.go
```diff
@@ -369,9 +371,35 @@ func (r *DPAReconciler) customizeNodeAgentDaemonset(dpa *oadpv1alpha1.DataProtec
...
+		return nil, err
+	}
+
 	return ds, nil
 }
 
+// Apply Override unsupported Node agent Server Args
+func (r *DPAReconciler) applyUnsupportedNodeAgentServerArgsOverride(dpa *oadpv1alpha1.DataProtectionApplication, nodeAgentContainer *corev1.Container) error {
```
mateusoliveira43, line 383: would refactor function created in https://github.com/openshift/oadp-operator/pull/1430 to be used here

It is extremely similar to this one. Adding annotation and sub command as parameters in it, it can be used in both places

controllers/nodeagent.go
mateusoliveira43: actually, I think my suggestion does not work "The analyzer reports instances of writes to struct fields and arrays that are never read. Specifically, when a struct object or an array is copied, its elements are copied implicitly by the compiler, and any element write to this copy does nothing with the original object." Ref https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/unusedwrite

would need to be something like this
```go
	// fetch nodeAgent container in order to customize it
	for index, container := range ds.Spec.Template.Spec.Containers {
		if container.Name == common.NodeAgent {
			nodeAgentContainer := &ds.Spec.Template.Spec.Containers[index]
			// append certs volume mount
			nodeAgentContainer.VolumeMounts = append(nodeAgentContainer.VolumeMounts, corev1.VolumeMount{
				Name:      "certs",
				MountPath: "/etc/ssl/certs",
			})
			// append PodConfig envs to nodeAgent container
			if useResticConf {
				if dpa.Spec.Configuration.Restic.PodConfig != nil && dpa.Spec.Configuration.Restic.PodConfig.Env != nil {
					nodeAgentContainer.Env = common.AppendUniqueEnvVars(nodeAgentContainer.Env, dpa.Spec.Configuration.Restic.PodConfig.Env)
				}
			} else if dpa.Spec.Configuration.NodeAgent.PodConfig != nil && dpa.Spec.Configuration.NodeAgent.PodConfig.Env != nil {
				nodeAgentContainer.Env = common.AppendUniqueEnvVars(nodeAgentContainer.Env, dpa.Spec.Configuration.NodeAgent.PodConfig.Env)
			}

			// append env vars to the nodeAgent container
			nodeAgentContainer.Env = common.AppendUniqueEnvVars(nodeAgentContainer.Env, proxy.ReadProxyVarsFromEnv())

			nodeAgentContainer.SecurityContext = &corev1.SecurityContext{
				Privileged: pointer.Bool(true),
			}

			imagePullPolicy, err := common.GetImagePullPolicy(getVeleroImage(dpa))
			if err != nil {
				r.Log.Error(err, "imagePullPolicy regex failed")
			}

			nodeAgentContainer.ImagePullPolicy = imagePullPolicy
			setContainerDefaults(nodeAgentContainer)

			if err := r.applyUnsupportedNodeAgentServerArgsOverride(dpa, nodeAgentContainer); err != nil {
				return nil, err
			}
			break
		}
	}
```

controllers/nodeagent_test.go
```diff
@@ -95,11 +95,12 @@ func TestDPAReconciler_buildNodeAgentDaemonset(t *testing.T) {
 		},
 	}
 	tests := []struct {
-		name    string
-		fields  fields
-		args    args
-		want    *appsv1.DaemonSet
-		wantErr bool
+		name          string
+		fields        fields
```
mateusoliveira43, line 99: lets remove `fields`, it not necessary. Only one test sets it, but removing it, did not break it. On test run, we just need client for dpaReconciler, which now use fakeClient

controllers/nodeagent.go
mpryc: The proposed implementation is nicer, but it does change one condition where the `nodeAgentContainer` may be nil based on the reference from `ds.Spec.Template.Spec.Containers` I am unsure if it's valid case, but does change a bit how the crash may occur in the application.

mateusoliveira43: I think that will never happen

if container is nil, calling container.Name would break

controllers/nodeagent.go
mateusoliveira43: not GenerateCliArgsFromConfigMap, but [applyUnsupportedArgsOverride](https://github.com/openshift/oadp-operator/blob/master/controllers/velero.go#L150)

mpryc: and make the applyUnsupportedArgsOverride a public function living somewhere in common? Either way works for me.

controllers/nodeagent_test.go
shubham-pampattiwar: Done, updated

controllers/nodeagent.go
shubham-pampattiwar: Done, updated the PR

controllers/nodeagent.go
shubham-pampattiwar: Done

mpryc (approved):
One nit which isn't required to be fixed.

Tested and it correctly applied unsupported args (fake ones for testing purposes) to all node-agents pods created by node-agent DaemonSet:

```
      args:
        - node-agent
        - server
        - '--myarg=test'
        - '--mysecondarg=true'
```

controllers/velero.go
```diff
@@ -513,7 +490,20 @@ func (r *DPAReconciler) appendPluginSpecificSpecs(dpa *oadpv1alpha1.DataProtecti
 				})
 		}
 	}
-	return r.applyUnsupportedArgsOverride(dpa, veleroDeployment, veleroContainer)
+
+	if configMapName, ok := dpa.Annotations[common.UnsupportedVeleroServerArgsAnnotation]; ok {
+		unsupportedServerArgsCM := corev1.ConfigMap{}
```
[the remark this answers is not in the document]
shubham-pampattiwar: Done, updated !

mateusoliveira43 (approved):
Ran
```sh
git clone --depth=1 git@github.com:shubham-pampattiwar/oadp-operator.git -b unsupported-node-agent-args test-pr
make deploy-olm
```
Then created DPA with kopia, checked NodeAgent containers
Screenshot from 2024-07-10 14-51-21

Got `velero node-agent server` options running `go run cmd/velero/velero.go help node-agent server` in velero repo

Created configMap and added annotation in DPA

Checked NodeAgent containers again
Screenshot from 2024-07-10 14-58-51
Since pods were still running, args were accepted successfully 

STILL MISSING DEVELOPER DOCUMENTATION!

shubham-pampattiwar:
@mateusoliveira43 Thank you for your review ! Dev Docs will be in a separate PR.
Note: Might be worth to test the Unsupported Velero server args as well, as this PR refactors that code flow too

mpryc:
@shubham-pampattiwar 

I've tested with `node-agent` and `server`:
 1. created CMs and added annotations for both, they got reflected on the pods
 2. removed annotation for node-agent, reconcile properly restarted those pods reflecting annotation
 3. added back annotation, everything was updated properly
 4. modified values in annotations, everything was updated properly
 5. modified values in CMs, they were reflected properly in the pods

/lgtm

mpryc (approved):

PR #1446, opened 2024-06-28
fix: automate upgrade tests

mateusoliveira43:
## Why the changes were made

Automate upgrade tests process.

## How to test the changes made

Follow E2E testing documentation and run
```sh
make catalog-test-upgrade
TEST_UPGRADE=true make test-e2e
```

PR #1451, opened 2024-07-03
[oadp-1.4] fix: automate upgrade tests

mateusoliveira43:
## Why the changes were made

Automate upgrade tests process, with focus in 1.4.0. (master branch PR https://github.com/openshift/oadp-operator/pull/1446)

Also, cherry-pick of https://github.com/openshift/oadp-operator/pull/1463

Follow [E2E testing documentation](https://github.com/mateusoliveira43/oadp-operator/blob/fix/automate-test-upgrade/docs/developer/testing/TESTING.md) and run
```sh
make catalog-test-upgrade
TEST_UPGRADE=true make test-e2e
```

To run tests manually, check [upgrade testing documentation](https://github.com/mateusoliveira43/oadp-operator/blob/fix/automate-test-upgrade/docs/developer/testing/test_oadp_version_upgrade.md).


mateusoliveira43:
/hold
Want to run CI, not ready yet

Makefile
```diff
@@ -18,7 +18,7 @@ BSL_AWS_PROFILE ?= default
 # BSL_AWS_PROFILE ?= migration-engineering
 
 # vsl secret
-CREDS_SECRET_REF ?= cloud-credentials
+CREDS_SECRET_REF ?= cloud-credentials-aws
```
mateusoliveira43, line 21: just to confirm VSL test is broken, undo change afterwards

kaovilai: in case you need more contexts, VSL requires cred that can trigger snapshot of disks on the cluster.. in this case, prow CI. Cluster is provisioned in prow CI account, and therefor needs to use their secret to trigger snapshots of those volumes.

This VSL secret is linked to the prow ci aws account.. whereas the normal cred we use elsewhere is for our department's account which is only responsible for buckets, not cluster/disks/native-snapshots in prow CI.

mateusoliveira43: do not if I understand

here what I want to confirm is that we miss adding credentials struct to VSL, as we do for BSLs. Which means, DPA will not reconcile true, because it will try to use default secret

kaovilai: Default secret is what is used currently in e2e for VSL. Custom secret name only applies to BSL.

kaovilai: It would be helpful to read https://github.com/openshift/oadp-operator/blob/master/docs/developer/testing/%20MULTI_CLOUD_TESTING_UPDATES.md

kaovilai: this was authored when VSL did not have the option to specify custom credential name. thus only default credentials were usable in VSL at the time.

mateusoliveira43: Hmm I did not know VSL had 3 defaults

But I still believe makes sense to remove option, as BSL do not have this flag in E2E

mateusoliveira43:
I am happy with CI results. Will break PR in 2 now: one for refactoring and other just for upgrade test

kaovilai:
fixed bucket region, tests passing.

shubham-pampattiwar (approved):

PR #1472, opened 2024-07-17
OADP-4290: Add flake for snapshot creation to ignore error list 

weshayutin:
## Why the changes were made

This will add ci tests that hit:
"VolumeSnapshot has a temporary error Failed to create snapshot: error updating status for volume snapshot content snapcontent-"  
Which has been determined to be temporary and a flake and the error can be safely ignored

Added error description to json list.

### DIFF
--- a/tests/e2e/lib/flakes.go
+++ b/tests/e2e/lib/flakes.go
@@ -19,6 +19,7 @@ var errorIgnorePatterns = []string{
 	// Ignore managed fields errors per https://github.com/vmware-tanzu/velero/pull/6110 and avoid e2e failure.
 	// https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1126/pull-ci-openshift-oadp-operator-master-4.10-operator-e2e-aws/1690109468546699264#1:build-log.txt%3A686
 	"level=error msg=\"error patch for managed fields ",
+        "VolumeSnapshot has a temporary error Failed to create snapshot: error updating status for volume snapshot content snapcontent-",
 }
 
 type FlakePattern struct {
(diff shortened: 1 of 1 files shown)


sseago (approved):

shubham-pampattiwar (approved):

kaovilai (dismissed):
/lgtm

mateusoliveira43:
/hold
dont we want this in backup/restore ignore logs instead of flakes

tests/e2e/lib/flakes.go
```diff
@@ -46,6 +46,11 @@ func CheckIfFlakeOccured(input string, knownFlake *bool) {
 			Description:         "Transient S3 bucket errors and limits",
 			StringSearchPattern: "Error copying image: writing blob: uploading layer chunked: received unexpected HTTP status: 500 Internal Server Error",
 		},
+		{
```
mateusoliveira43, line 49: more information:

adding this to flake, when the problem is hit, that test will fail and then will be retried

adding to ignore backup/restore errors/warnings, if the backup/restore completes successfully (which I saw happening in the past), test will just continue, which is faster, and then, better in my opinion

so we at least, we should also add this to logs ignore list. If we know this can also break backup/restore, then we should also add it here in flakes

kaovilai: > test will just continue

and it will still fail if backup/restore weren't actually in status completed. So win/win.

tests/e2e/lib/flakes.go
```diff
@@ -19,6 +19,7 @@ var errorIgnorePatterns = []string{
 	// Ignore managed fields errors per https://github.com/vmware-tanzu/velero/pull/6110 and avoid e2e failure.
 	// https://prow.ci.openshift.org/view/gs/origin-ci-test/pr-logs/pull/openshift_oadp-operator/1126/pull-ci-openshift-oadp-operator-master-4.10-operator-e2e-aws/1690109468546699264#1:build-log.txt%3A686
 	"level=error msg=\"error patch for managed fields ",
+        "VolumeSnapshot has a temporary error Failed to create snapshot: error updating status for volume snapshot content snapcontent-",
```
mateusoliveira43, line 22: I think this will fail code formatter (extra space), you can run `make test` to fix/see fix

kaovilai (approved):

kaovilai (approved):

shubham-pampattiwar (approved):

mateusoliveira43 (approved):
/lgtm

PR #1479, opened 2024-07-26
OADP-4291 Add no-pod/immediate bind VM tests.

mrnold:
## Why the changes were made

For [OADP-4291](https://issues.redhat.com//browse/OADP-4291), adds requested tests for no-pod VM backup/restore and immediate volume binding mode. Still need to see how some of these new tests run in CI before merging.

make TEST_VIRT=true test-e2e


tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -186,6 +188,10 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
 		gomega.Expect(err).To(gomega.BeNil())
 
 		dpaCR.VeleroDefaultPlugins = append(dpaCR.VeleroDefaultPlugins, v1alpha1.DefaultPluginKubeVirt)
+
+		storageClassYaml := fmt.Sprintf("./sample-applications/storageclass-immediate/%s-immediate.yaml", provider)
```
kaovilai, line 192: Where are these files coming from?
https://github.com/search?q=repo%3Aopenshift%2Foadp-operator%20storageclass-immediate&type=code

mrnold: Thanks, I just forgot to add them.

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43, line 1: Can we remove `Ordered` from here https://github.com/openshift/oadp-operator/blob/master/tests/e2e/virt_backup_restore_suite_test.go#L154 ?

I think this causes test to exit on failure, instead of running all of them and then reporting which tests failed (if any)

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -195,6 +201,9 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
 		if v != nil && wasInstalledFromTest {
 			v.EnsureVirtRemoval()
 		}
+
+		err := v.Clientset.StorageV1().StorageClasses().Delete(context.TODO(), "test-gp3-csi-immediate", metav1.DeleteOptions{})
```
mateusoliveira43, line 205: can we use `UninstallApplication` function here instead?

tests/e2e/sample-applications/storageclass-immediate/aws-immediate.yaml
```diff
@@ -0,0 +1,14 @@
+apiVersion: v1
+kind: List
+items:
+  - kind: StorageClass
+    apiVersion: storage.k8s.io/v1
+    metadata:
+      name: test-gp3-csi-immediate
+    provisioner: ebs.csi.aws.com
```
mateusoliveira43, line 8: we need to create one file for each cloud (aws, ibm, gcp and azure) right?

otherwise, virt tests can only be run in AWS

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -245,6 +254,73 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
...
+				Namespace:         "cirros-test",
+				Name:              "cirros-test",
+				SkipVerifyLogs:    true,
+				BackupRestoreType: lib.CSIDataMover,
+				BackupTimeout:     20 * time.Minute,
+				PreBackupVerify:   vmPoweredOff("cirros-test", "cirros-test"),
+			},
+		}, nil),
+
+		ginkgo.Entry("immediate binding no-application CSI+datamover backup and restore, powered-off CirrOS VM", ginkgo.Label("virt"), VmBackupRestoreCase{
```
mateusoliveira43, line 310: this one is a duplication of above test, no?

mateusoliveira43:
Saw dynamic PV errors in virt tests logs. Should we wait a rebase in velero fork default branch to merge this one?

tests/e2e/sample-applications/storageclass-immediate/aws-immediate.yaml
mrnold: Yes, this should just be a copy of the existing storage classes with the volumeBindingMode set to immediate. I didn't add them because I don't have all the different cluster types set up to test.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Yes, thanks, fixed.

mrnold:
> Saw dynamic PV errors in virt tests logs. Should we wait a rebase in velero fork default branch to merge this one?

I think it is not actually choosing my immediate-mode storage class copy:
`failed to provision volume with StorageClass "test-gp3-csi-immediate": claim Selector is not supported`

So probably the test is just using WaitForFirstConsumer, and failing in the expected way. I haven't figured out the right way to use a particular storage class per-VM yet, I guess. We should definitely wait to merge until all the VM tests pass though.

tests/e2e/sample-applications/storageclass-immediate/aws-immediate.yaml
mateusoliveira43: Maybe we can automate logic here instead of creating files

The StorageClass I usually use in IBM has this annotation
```
metadata:
  annotations:
    storageclass.kubernetes.io/is-default-class: "true"
```
Does this one you are using also have it in AWS?

if yes, we could create StorageClass for tests by deepCopying result and changing its name and volumeBindingMode

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-immediate.yaml
```diff
@@ -0,0 +1,48 @@
...
+          spec:
+            sourceRef:
+              kind: DataSource
+              name: cirros
+              namespace: openshift-virtualization-os-images
+            storage:
+              resources:
+                requests:
+                  storage: 128Mi
+              storageClassName: test-gp3-csi-immediate
```
mateusoliveira43, line 24: only difference between `tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test.yaml` and this file is this line, right?

would be better to add logic in code to add this to installed VM?

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Yes, so far this seems to be working fine for me locally.

tests/e2e/sample-applications/storageclass-immediate/aws-immediate.yaml
mrnold: Yes, that would be nicer. It's definitely a good way to add the default class annotation too, I just didn't add it yet because the YAML templates got me to test results faster.

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-immediate.yaml
mrnold: Yes, that would be fine to add at VM installation time. There would just need to be a new VmBackupRestoreCase field to specify the volumeMode.

mrnold: On second thought, I don't think this is configurable after the VM is created. It could be another use case for templating, or maybe some kind of hook inside InstallApplication? I will probably keep this file for this pull request, and try some other options out later.

tests/e2e/virt_backup_restore_suite_test.go
mrnold: Okay, trying this out.

mrnold: Looks like the answer to this is no: 

`[BeforeAll] setup nodes must appear inside an Ordered container.`

mateusoliveira43: ahh :cry: thanks for trying!

tests/e2e/sample-applications/storageclass-immediate/aws-immediate.yaml
mrnold: Okay, I have moved this out of the YAML.

tests/e2e/lib/virt_storage_helpers.go
```diff
@@ -203,3 +205,42 @@ func (v *VirtOperator) CreateDataSourceFromPvc(namespace, name string) error {
...
+	immediateStorageClass.Name = name
+	immediateStorageClass.ResourceVersion = ""
+	immediateStorageClass.Annotations["storageclass.kubernetes.io/is-default-class"] = "false"
+
+	_, err = v.Clientset.StorageV1().StorageClasses().Create(context.Background(), immediateStorageClass, metav1.CreateOptions{})
+	return err
+}
+
+func (v *VirtOperator) RemoveStorageClass(name string) error {
+	return v.Clientset.StorageV1().StorageClasses().Delete(context.TODO(), name, metav1.DeleteOptions{})
```
mateusoliveira43, line 245: should we always use `context.Background()`?

tests/e2e/virt_backup_restore_suite_test.go
mateusoliveira43, line 1: with new tests, virt job almost takes 2h to finish (which breaks PROW)

the one current failed hit the timeout, and the other 2 took 1h55min and 1h57min

should we create a new virt job to avoid timeouts?

mateusoliveira43: or maybe wait rebase, with fix to dynamic PV, maybe tests run faster

mrnold: The easiest thing is rebasing, I know the expected failures sit spinning on FinalizingPartiallyFailed for ten minutes each. Hopefully getting rid of that will keep things inside the timeout.

mpryc:
Added comments on the https://github.com/openshift/oadp-operator/pull/1488

tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-immediate.yaml
mateusoliveira43, line 1: https://github.com/openshift/oadp-operator/pull/1500#issuecomment-2302991667

mateusoliveira43: rebase was done, I think this PR can be updated to get merged now

mateusoliveira43 (approved):

weshayutin:
need an override, our tests passed on 4.16-e2e-test-aws
SUCCESS! -- 24 Passed | 0 Failed | 0 Pending | 9 Skipped
https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1479/pull-ci-openshift-oadp-operator-master-4.16-e2e-test-aws/1828491099992231936/artifacts/e2e-test-aws/e2e/build-log.txt

shubham-pampattiwar (approved):
/lgtm

PR #1493, opened 2024-08-14
use the multiarch build of mariadb-todolist

weshayutin:
Current source for this image is:
https://github.com/weshayutin/todolist-mariadb-go.git
quay.io/migtools/oadp-ci-todolist-mariadb-go:multiarch

This is an updated version that is cleaned up a bit and also works with VM's or with containers.  Will push to migtools when this is passing.

### DIFF
--- a/tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi.yaml
+++ b/tests/e2e/sample-applications/mysql-persistent/mysql-persistent-csi.yaml
@@ -176,7 +176,7 @@ items:
         spec:
           containers:
           - name: todolist
-            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:latest
+            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:multiarch
             env:
               - name: foo
                 value: bar
--- a/tests/e2e/sample-applications/mysql-persistent/mysql-persistent-twovol-csi.yaml
+++ b/tests/e2e/sample-applications/mysql-persistent/mysql-persistent-twovol-csi.yaml
@@ -167,7 +167,7 @@ items:
         spec:
           containers:
           - name: todolist
-            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:latest
+            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:multiarch
             env:
               - name: foo
                 value: bar
--- a/tests/e2e/sample-applications/mysql-persistent/mysql-persistent.yaml
+++ b/tests/e2e/sample-applications/mysql-persistent/mysql-persistent.yaml
@@ -189,7 +189,7 @@ items:
         spec:
           containers:
           - name: todolist
-            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:latest
+            image: quay.io/migtools/oadp-ci-todolist-mariadb-go:multiarch
             env:
               - name: foo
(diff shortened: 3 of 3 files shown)


mateusoliveira43 (changes requested):
all tests failed with same error, which I never saw before, in two vol test

weshayutin:
> all tests failed with same error, which I never saw before, in two vol test

@mateusoliveira43 looks like this exposed an error in our testing.  The old mysql-todolist containers always created some sample data in the db, 2 rows.  This version leaves the DB empty when we start up.

AFICT it looks like we're checking what data is in the db prior to adding any data.  Not sure sure yet.  That's my guess
```
2024/08/15 14:14:25 Data before the curl request: 
 []

2024/08/15 14:14:25 Payload encoded parameters: description=2024-08-15+14%3A14%3A25.114899618+%2B0000+UTC+m%3D%2B898.006557829
2024/08/15 14:14:25 Payload encoded parameters: description=Thursday
2024/08/15 14:14:25 Writing data to backupFile (backup-data.txt): 
 [{"Id":1,"Description":"2024-08-15 14:14:25.114899618 +0000 UTC m=+898.006557829","Completed":false},{"Id":2,"Description":"Thursday","Completed":false}]
```

weshayutin:
updated image
/retest

weshayutin:
well this is interesting.. 
Screenshot from 2024-08-15 15-54-18

kaovilai:
Once the image are identical, I would push for latest to be multiarch.. and not separate tag.

mpryc:
> Once the image are identical, I would push for latest to be multiarch.. and not separate tag.

I will second that. With a single multi-arch image, users and developers don't need to worry about which specific tag corresponds to which architecture. Secondly having separate multi-arch tag rather then latest is less clear on what is the tag about? Is the `multiarch` build out of stable/branch or latest? That's not really clear.

I can quickly check what arch is supported by the image by calling `podman manifest inspect <image>`, but it's much harder to tell what the image corresponds to e.g.
```shell
podman manifest inspect quay.io/migtools/oadp-ci-todolist-mariadb-go:multiarch

[...]

{
    "schemaVersion": 2,
    "mediaType": "application/vnd.docker.distribution.manifest.list.v2+json",
    "manifests": [
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 814,
            "digest": "sha256:4623b177778d720f3707ef9f9e652e7bb671afea05896c03be4f1611121ec025",
            "platform": {
                "architecture": "arm64",
                "os": "linux"
            }
        },
        {
            "mediaType": "application/vnd.oci.image.manifest.v1+json",
            "size": 814,
            "digest": "sha256:c05264eda64207e0251dc4b69a71e399ac6e08c2836bff5d5561921de2ab87ad",
            "platform": {
                "architecture": "amd64",
                "os": "linux"
            }
        }
    ]
}
```

weshayutin:
> Once the image are identical, I would push for latest to be multiarch.. and not separate tag.

100% just consider that multiarch tag as a test :)

weshayutin:
Updated tags https://quay.io/repository/migtools/oadp-ci-todolist-mariadb-go?tab=tags
multiarch is now tagged as latest :)   PR to konveyor/mig-demo-apps
https://github.com/konveyor/mig-demo-apps/pull/105

Issue #1487, opened 2024-08-09
E2E uses old must-gather image


mateusoliveira43:
E2E tests run uses must-gather image not updated for a long time 

References
- https://quay.io/repository/konveyor/oadp-must-gather?tab=tags
- https://github.com/openshift/oadp-operator/blob/54fed94af3a26484b56bc8f92608bcfd04786bad/tests/e2e/lib/apps.go#L355

To fix the problem, must-gather image must be built constantly like other project images

PR #1494, opened 2024-08-15
fix: multi-arch must-gather image

mateusoliveira43:
## Why the changes were made

Add support for multi-arch builds of OADP must-gather image

Related to https://github.com/openshift/oadp-operator/issues/1487

Go to must-gather folder and change `PLATFORM` in Makefile and run `make docker-build`. Build should succeed 

Check that tools removed from Dockerfile are not needed

### DIFF
--- a/must-gather/Dockerfile
+++ b/must-gather/Dockerfile
@@ -13,13 +13,21 @@ RUN curl --location --output velero.tgz https://github.com/openshift/velero/arch
     curl --location --output restic.tgz https://github.com/openshift/restic/archive/refs/heads/${RESTIC_BRANCH}.tar.gz && \
     tar -xzvf restic.tgz && cd restic-${RESTIC_BRANCH} && \
     CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -mod=mod -ldflags '-extldflags "-static"' -o /restic github.com/restic/restic/cmd/restic && \
-    cd .. && rm -rf restic.tgz restic-${RESTIC_BRANCH}
+    cd .. && rm -rf restic.tgz restic-${RESTIC_BRANCH} && \
+    curl --location --output oc.tgz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.16/openshift-client-linux-${TARGETARCH}-rhel9.tar.gz && \
+    tar -xzvf oc.tgz && rm -rf oc.tgz README.md kubectl
 
 FROM registry.access.redhat.com/ubi9/go-toolset:latest AS gobuilder
 
 RUN go install -v github.com/google/pprof@latest
 
-FROM quay.io/openshift/origin-must-gather:latest AS builder
+# FROM registry.access.redhat.com/openshift4/ose-cli-rhel9:v4.16 AS builder
+
+# RUN which oc
+# # /usr/bin/oc
+# RUN oc version
+# # Client Version: v4.2.0-alpha.0-2197-g91ba77a
+# # Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
 
 FROM registry.access.redhat.com/ubi9-minimal:latest
 
@@ -27,7 +35,10 @@ RUN echo -ne "[centos-9-appstream]\nname = CentOS 9 (RPMs) - AppStream\nbaseurl 
 RUN microdnf -y install rsync tar gzip graphviz findutils
 
 COPY --from=gobuilder /opt/app-root/src/go/bin/pprof /usr/bin/pprof
-COPY --from=builder /usr/bin/oc /usr/bin/oc
+# COPY --from=builder /usr/bin/oc /usr/bin/oc
+COPY --from=konveyor-builder /build/oc /usr/bin/oc
+# Client Version: 4.16.6
+# Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
 COPY --from=konveyor-builder /velero /usr/bin/velero
 COPY --from=konveyor-builder /restic /usr/bin/restic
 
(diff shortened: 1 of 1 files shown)


must-gather/Dockerfile
```diff
@@ -13,21 +13,32 @@ RUN curl --location --output velero.tgz https://github.com/openshift/velero/arch
...
 RUN go install -v github.com/google/pprof@latest
 
-FROM quay.io/openshift/origin-must-gather:latest AS builder
+# FROM registry.access.redhat.com/openshift4/ose-cli-rhel9:v4.16 AS builder
+
+# RUN which oc
+# # /usr/bin/oc
+# RUN oc version
+# # Client Version: v4.2.0-alpha.0-2197-g91ba77a
+# # Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
```
mateusoliveira43, line 30: will remove comments prior to merge, but want to check if it ok to bump `oc` version. It was 4.2, now will be 4.16

must-gather/Dockerfile
mateusoliveira43, line 1: @rayfordj these changes do not impact you, right?

rayfordj: @mateusoliveira43 , not in any substantial way, but I will need to update the Dockerfile's digest once this is merged.  Thanks for the heads-up!  :sunglasses:

must-gather/Dockerfile
rayfordj: @mateusoliveira43 , I see no reason not to _**especially**_ as old as you've identified it still is.

FWIW, 
```
$ podman run -it --rm --entrypoint /usr/bin/oc registry.redhat.io/oadp/oadp-mustgather-rhel9:1.4.0 version
Client Version: 4.16.0-202405281506.p0.g12a2e98.assembly.stream.el9-12a2e98
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
```

mateusoliveira43: thanks for the info, will update then!

must-gather/Dockerfile
```diff
@@ -13,21 +13,32 @@ RUN curl --location --output velero.tgz https://github.com/openshift/velero/arch
     curl --location --output restic.tgz https://github.com/openshift/restic/archive/refs/heads/${RESTIC_BRANCH}.tar.gz && \
     tar -xzvf restic.tgz && cd restic-${RESTIC_BRANCH} && \
     CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -mod=mod -ldflags '-extldflags "-static"' -o /restic github.com/restic/restic/cmd/restic && \
-    cd .. && rm -rf restic.tgz restic-${RESTIC_BRANCH}
+    cd .. && rm -rf restic.tgz restic-${RESTIC_BRANCH} && \
+    curl --location --output oc.tgz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable-4.16/openshift-client-linux-${TARGETARCH}-rhel9.tar.gz && \
```
mpryc, line 17: Here you could use stable (which is currently 4.16), this way we won't need to update once 4.17 is there:

```
curl --location --output oc.tgz https://mirror.openshift.com/pub/openshift-v4/clients/ocp/stable/openshift-client-linux-${TARGETARCH}-rhel9.tar.gz
```

must-gather/Dockerfile
```diff
@@ -13,21 +13,32 @@ RUN curl --location --output velero.tgz https://github.com/openshift/velero/arch
...
+    tar -xzvf oc.tgz && rm -rf oc.tgz README.md kubectl
```
mpryc, line 18: you can use:
```
tar -xzvf oc.tgz -C . oc
```

Above will extract `oc` only to the current folder path `.`, so no need to rm -rf README.md and kubectl.

must-gather/Dockerfile
mateusoliveira43: I do not like automatically updates

how does prod handles updates for `oc` @rayfordj ?

must-gather/Dockerfile
mateusoliveira43: will test

mateusoliveira43: applied

must-gather/Dockerfile
rayfordj: At the moment, manually, until we have an `oc` available from an el9/ubi9 base image at which time we'll then resume updating via `follow_tag` as the base image is updated.  _I'm guessing this likely won't happen until >=v4.17._

rayfordj: While I lean towards `stable` to ensure it isn't later forgotten and doesn't rot as has been observed to occur over the years across our repos.  I've no objection for upstream to use either a) latest `stable` or b) latest versioned (such as, `stable-v4.16`).  However, the former will continue to be updated over releases, while the latter will eventually stop being maintained.  Regardless, it isn't something that will be carried downstream. 
:grimacing:

kaovilai: could copy oc from container image

mateusoliveira43: changed to `stable`

there is a handy image for that, Tiger? ones I found, where not in `registry.access`, so needed auth

must-gather/Dockerfile
```diff
@@ -16,19 +16,13 @@ RUN curl --location --output velero.tgz https://github.com/openshift/velero/arch
...
 
-FROM registry.access.redhat.com/ubi9/go-toolset:latest AS gobuilder
-
-RUN go install -v github.com/google/pprof@latest
-
 FROM registry.access.redhat.com/ubi9-minimal:latest
 
-RUN echo -ne "[centos-9-appstream]\nname = CentOS 9 (RPMs) - AppStream\nbaseurl = https://mirror.stream.centos.org/9-stream/AppStream/x86_64/os/\nenabled = 1\ngpgcheck = 0" > /etc/yum.repos.d/centos-9-appstream.repo
-RUN microdnf -y install rsync tar gzip graphviz findutils
+RUN microdnf -y install tar
```
kaovilai, line 24:
```suggestion
RUN microdnf -y install tar && microdnf clean all
```
https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/9/html/building_running_and_managing_containers/assembly_adding-software-to-a-ubi-container_building-running-and-managing-containers#proc_building-ubi-based-images_assembly_adding-software-to-a-ubi-container

kaovilai (changes requested):

PR #1496, opened 2024-08-20
CI: Remove VM test that's expected to fail.

mrnold:
## Why the changes were made

This removes a test that fails consistently. It is currently marked as expected to fail, but it takes so much time to fail that it's better off not being run at all.

make TEST_VIRT=true test-e2e

Specific test is WFFC+datamover+no-pod backup/restore.

### DIFF
--- a/tests/e2e/virt_backup_restore_suite_test.go
+++ b/tests/e2e/virt_backup_restore_suite_test.go
@@ -1,7 +1,6 @@
 package e2e_test
 
 import (
-	"errors"
 	"fmt"
 	"io"
 	"log"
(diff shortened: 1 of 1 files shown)


tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -231,21 +229,6 @@ var _ = Describe("VM backup and restore tests", Ordered, func() {
 			},
 		}, nil),
 
-		Entry("no-application CSI+datamover backup and restore, powered-off CirrOS VM", Label("virt"), VmBackupRestoreCase{
```
mateusoliveira43, line 234: just to make sure (PR description got me confused), as discussed in Scrum, this is a case that currently does not work and need Velero update to work in the future?

mrnold: Yes, this is a case that does not work right now. I added this test in the first place because I thought Velero had a fix, but it does not. I think I heard there is a Jira ticket or some other plan to fix it in the future though.

mateusoliveira43 (approved):

shubham-pampattiwar (approved):

kaovilai (approved):
/lgtm

PR #1508, opened 2024-09-05
add openstack settings file for e2e

weshayutin:
## Why the changes were made

We need to be able run virt tests on an internal openshift hosted on openstack

### DIFF
--- a/tests/e2e/scripts/openstack_settings.sh
+++ b/tests/e2e/scripts/openstack_settings.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+cat > $TMP_DIR/oadpcreds <<EOF
+{
+  "spec": {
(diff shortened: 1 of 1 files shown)


tests/e2e/scripts/openstack_settings.sh
mateusoliveira43, line 1: how/when to run it?

I think good place to add its doc is here https://github.com/openshift/oadp-operator/blob/master/docs/developer/testing/TESTING.md

weshayutin: agree, we'll doc it once we for sure know what to doc :)

weshayutin:
https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1725631699796829
will reopen if needed

PR #1519, opened 2024-09-13
add a support matrix for backup, volumetypes, incremental

weshayutin:
## Why the changes were made

We need document and distinguish our support across backup types, volume types and incremental backup support.

READ!

### DIFF
--- a/docs/supported_backups.md
+++ b/docs/supported_backups.md
@@ -1,0 +1,29 @@
+# OADP Supported Backups
+
+OADP supports backup and restore disaster recovery operations for both containerized and OpenShift Virtualization workloads. It is important to distinguish the nuances and support for each workload type.
+
+## Containerized Workloads
+
+OADP supports backup and restore operations for containerized workloads. 
+
+| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover ** |
+|------------|--------------|-------------|-----|------------------|
+| Filesystem | S, I         | S, I        | S   | S, I             |
+| Block      | N            | N           | S   | S, I             |
+
+
+## OpenShift Virtualization Workloads
+
+OADP supports backup and restore operations for OpenShift Virtualization workloads.
+
+| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover ** |
+|------------|--------------|-------------|-----|------------------|
+| Filesystem | Supported    | Supported   | Supported | Supported  |
+| Block      | Not Supported| Not Supported| Supported| Supported  |
+
+* Legend:
+  * S - Supported
+  * I - Incremental Backup supported
+  * N - Not Supported
+  * FSB - File System Backup
+  * **Note: DataMover upload and download operations use kopia regardless of uploader type.
--- a/docs/supported_backups.md
+++ b/docs/supported_backups.md
@@ -6,24 +6,25 @@ OADP supports backup and restore disaster recovery operations for both container
 
 OADP supports backup and restore operations for containerized workloads. 
 
-| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover ** |
-|------------|--------------|-------------|-----|------------------|
-| Filesystem | S, I         | S, I        | S   | S, I             |
-| Block      | N            | N           | S   | S, I             |
+| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover (ii) | Volume Snapshot |
+|------------|--------------|-------------|-----|------------------| ----------------|
+| Filesystem | S, I         | S, I        | S   | S, I             | S               |
+| Block      | N            | N           | S   | S, I             | S               |
 
 
 ## OpenShift Virtualization Workloads
 
 OADP supports backup and restore operations for OpenShift Virtualization workloads.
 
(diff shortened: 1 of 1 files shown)


docs/supported_backups.md
```diff
@@ -0,0 +1,29 @@
+# OADP Supported Backups
+
+OADP supports backup and restore disaster recovery operations for both containerized and OpenShift Virtualization workloads. It is important to distinguish the nuances and support for each workload type.
+
+## Containerized Workloads
+
+OADP supports backup and restore operations for containerized workloads. 
+
+| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover ** |
```
shubham-pampattiwar, line 9: Maybe also add another column of cloud provider snapshots ? (via VSL )

weshayutin: ya.. I will before we merge. I just want to keep this clear atm for the doc team.
/hold

sseago: @shubham-pampattiwar We'll need another dimension in the chart for that. Based on VolumeMode, CSI and native snapshots are supported in the same cases (i.e. all). We'll need to add columns for "Volume uses CSI driver" and "Cloud Volume (aws/gcp/azure)" -- so if the first new column is false, then CSI/DM are `N`, and if the second new column is false, then native snapshots are `N`

docs/supported_backups.md
```diff
@@ -0,0 +1,29 @@
...
+|------------|--------------|-------------|-----|------------------|
+| Filesystem | N            | N           | S   | S, I             |
+| Block      | N            | N           | S   | S, I             |
+
+* Legend:
+  * S - Supported
+  * I - Incremental Backup supported
+  * N - Not Supported
+  * FSB - File System Backup
+  * **Note: DataMover upload and download operations use kopia regardless of uploader type.
```
shubham-pampattiwar, line 29: `s/**Note/**Note**`

docs/supported_backups.md
weshayutin: added a section for volume snapshots

docs/supported_backups.md
```diff
@@ -0,0 +1,30 @@
+# OADP Supported Backups
+
+OADP supports backup and restore disaster recovery operations for both containerized and OpenShift Virtualization workloads. It is important to distinguish the nuances and support for each workload type.
+
+## Containerized Workloads
+
+OADP supports backup and restore operations for containerized workloads. 
+
+| VolumeMode | FSB - restic | FSB - kopia | CSI | CSI DataMover (ii) | Volume Snapshot |
```
sseago, line 9: Calling it "Volume Snapshot" is confusing, since VolumeSnapshot is the CR created for CSI snapshots. Maybe "native snapshots (VolumeSnapshotter plugin)? If that's too long, then just "Native snapshots" and a note below the chart saying that native snapshots are only supported if there's an included VolumeSnapshotter plugin.

sseago: Oh, there's already a note below for the scope of this column. In that case, either "Native snapshots" or "Native snapshots (VolumeSnapshotter plugin)" for the heading here, and use the same wording beginning the note below.

shubham-pampattiwar: Cloud Provider Native Snapshots

weshayutin: nice, thank you both!

docs/supported_backups.md
weshayutin: updated to:
  * ii -  DataMover upload and download operations use kopia regardless of uploader type.

kaovilai:
Plugin Native Snapshots sounds clearer..

Both **Plugin Native Snapshots** and **CSI Snapshot** could trigger snapshot at the Cloud Provider natively.

docs/supported_backups.md
```diff
@@ -0,0 +1,31 @@
...
+|------------|--------------|-------------|-----|------------------| ----------------|
+| Filesystem | N            | N           | S   | S, I             | N               |
+| Block      | N            | N           | S   | S, I             | N               |
+
+* Legend:
+  * **S:** Supported
+  * **I:** Incremental Backup supported
+  * **N:** Not Supported
+  * **FSB:** File System Backup
+  * **Cloud Provider Native Snapshots:** Non CSI Volume support for AWS, GCP, Azure via plugins.
```
shubham-pampattiwar, line 29: Maybe just update this to: Non CSI Volume support for AWS, GCP, Azure via VolumeSnapshotter plugins.
WDYT @kaovilai ?

kaovilai: Sure. Perhaps also have to change description to say

```diff
+OADP supports volume snapshot operations for OpenShift Virtualization workloads.
-OADP supports backup and restore operations for OpenShift Virtualization workloads.
```

weshayutin: I don't think `volume snapshot operations` is clear enough for customers.  They are using OADP to backup and resore or disaster recovery.  Not to snapshot volumes.

sseago, line 29: non-CSI is inaccurate. gp3-csi, for example, also supports native snapshots.

weshayutin: Thank you @sseago, updated and added links to plugins

kaovilai (approved):

sseago (approved):

PR #1525, opened 2024-09-19
OADP-3050: Remove unwanted BSL/VSLs on create/edit DPA

### LINKED ISSUE #1469: Remove BSL/VSL when DPA updates
BSL/VSL may be left out in Cluster depending on DPA updates. For example:
- if an action updates BSL name in DPA, 2 BSL will be in the cluster: new one with name, and old one with old name
- if an action removes BSL from DPA (by using `NoDefaultBackupLocation` for example), old BSL is not removed from cluster

> **Note**: when DPA is deleted, all BSL/VSL are deleted

stillalearner:
Fixes #1469 of unwanted BSL or VSLs in the cases:

1. BSLs/VSLs removed from DPA spec.
2. BSL/VSLs names modified in-place from DPA spec.
3. Default-named BSL/VSLs removed from DPA spec.

hope that covers all cases.

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -3,6 +3,7 @@ package controllers
 import (
 	"errors"
 	"fmt"
+	"slices"
 
 	"github.com/go-logr/logr"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
@@ -100,6 +101,8 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool, 
 	if err := r.Get(r.Context, r.NamespacedName, &dpa); err != nil {
 		return false, err
 	}
+
+	dpaBSLNames := []string{}
 	// Loop through all configured BSLs
 	for i, bslSpec := range dpa.Spec.BackupLocations {
 		// Create BSL as is, we can safely assume they are valid from
@@ -110,6 +113,7 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool, 
 		if bslSpec.Name != "" {
 			bslName = bslSpec.Name
 		}
+		dpaBSLNames = append(dpaBSLNames, bslName)
 
 		bsl := velerov1.BackupStorageLocation{
 			ObjectMeta: metav1.ObjectMeta{
@@ -196,6 +200,33 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool, 
 			)
 		}
 	}
+
+	dpaBSLs := velerov1.BackupStorageLocationList{}
+	dpaBSLLabels := map[string]string{
+		"app.kubernetes.io/name":       common.OADPOperatorVelero,
+		"app.kubernetes.io/managed-by": common.OADPOperator,
+		"app.kubernetes.io/component":  "bsl",
+	}
+	err := r.List(r.Context, &dpaBSLs, client.InNamespace(r.NamespacedName.Namespace), client.MatchingLabels(dpaBSLLabels))
+	if err != nil {
+		return false, err
+	}
+	// If current BSLs do not match the spec, delete extra BSLs
+	if len(dpaBSLNames) != len(dpaBSLs.Items) {
+		for _, bsl := range dpaBSLs.Items {
+			if !slices.Contains(dpaBSLNames, bsl.Name) {
+				if err := r.Delete(r.Context, &bsl); err != nil {
+					return false, err
+				}
+				// Record event for BSL deletion
+				r.EventRecorder.Event(&bsl,
+					corev1.EventTypeNormal,
+					"BackupStorageLocationDeleted",
+					fmt.Sprintf("BackupStorageLocation %s created by OADP in namespace %s was deleted as it was not in DPA spec.", bsl.Name, bsl.Namespace))
+			}
+		}
+	}
+
 	return true, nil
 }
 
--- a/controllers/vsl.go
+++ b/controllers/vsl.go
@@ -3,12 +3,14 @@ package controllers
 import (
 	"errors"
 	"fmt"
+	"slices"
 	"strings"
 
 	"github.com/go-logr/logr"
 	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
@@ -216,6 +218,7 @@ func (r *DPAReconciler) ReconcileVolumeSnapshotLocations(log logr.Logger) (bool,
 		return false, err
 	}
 
+	dpaVSLNames := []string{}
 	// Loop through all configured VSLs
 	for i, vslSpec := range dpa.Spec.SnapshotLocations {
 		// Create VSL as is, we can safely assume they are valid from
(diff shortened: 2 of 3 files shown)


controllers/bsl.go
```diff
@@ -196,9 +197,76 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+			if err := r.Delete(r.Context, &bsl); err != nil {
+				return false, err
+			}
+		}
+	}
+
 	return true, nil
 }
 
+func countDefaultNamedBSLs(backupLocations []oadpv1alpha1.BackupLocation) int {
```
sseago, line 241: I don't think "default" is the right word here. Default has a specific meaning for a BSL -- the field `spec.default` is a bool, and if a backup doesn't specify a BSL, then the default bsl is used. I think what you're calling default are BSLs without user-defined named. Maybe call this func `countUnnamedConfiguredBSLs`?

stillalearner: ya i thought of it for a moment, and added "named" after default. But if you also observed that, that means it is something which can create confusion, Let me change that @sseago ! Thanks.

controllers/bsl.go
```diff
@@ -196,9 +197,76 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+
+	if err != nil {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func (r *DPAReconciler) syncBSLs(dpa *oadpv1alpha1.DataProtectionApplication) (bool, error) {
+	bslLabels := map[string]string{
```
kaovilai, line 211: dpaBslLabels?

labels for BSLs managed via dpa.

controllers/bsl.go
kaovilai: countUnnamedDpaBSLs?

kaovilai (commented):
Have to double check but hopefully this doesn't [break the workaround](https://issues.redhat.com/browse/OADP-3630#:~:text=As%20a%20workaround,populated%20into%20it) in
 https://issues.redhat.com/browse/OADP-3630

controllers/bsl.go
stillalearner: Done.

controllers/bsl.go
stillalearner: Done

stillalearner:
> Have to double check but hopefully this doesn't [break the workaround](https://issues.redhat.com/browse/OADP-3630#:~:text=As%20a%20workaround,populated%20into%20it) in https://issues.redhat.com/browse/OADP-3630

I think it should not, as these changes only deal with the BSLs having all of the 3 labels.

While the workaround is only adding one of those. What do you think @kaovilai ?

kaovilai (dismissed):
> While the workaround is only adding one of those. What do you think @kaovilai ?

works for me

controllers/bsl.go
```diff
@@ -196,9 +197,76 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
 			)
 		}
 	}
+
+	_, err := r.syncBSLs(&dpa)
```
mateusoliveira43, line 201: I studied the difference between BSLs created by DPA and by users, and I think only difference is these metadata that DPA includes to the objects
```yaml
metadata:
  labels:
    app.kubernetes.io/component: bsl
    app.kubernetes.io/instance: <bsl-name>
    app.kubernetes.io/name: oadp-operator-velero
    openshift.io/oadp: "True"
    openshift.io/oadp-registry: "True"
  ownerReferences:
    blockOwnerDeletion: true
    controller: true
    name: <dpa-name>
    uid: <uid>
``` 

so looking labels is a good idea (BUT ALSO NEED TO CHECK NAMESPACE, not?)

Did not understand what the current solution is doing, I suggest simplifying

What we need is to do is:
- store the BSLs names created by DPA , in OADP namespace
- list BSLs created by DPA, in OADP namespace
- if count differs, delete BSLs not defined in DPA

right?

Something like this would solve the problem?
```go
	dpaBSLNames := []string{}
	// Loop through all configured BSLs
	for i, bslSpec := range dpa.Spec.BackupLocations {
		// Create BSL as is, we can safely assume they are valid from
		// ValidateBackupStorageLocations

		// check if BSL name is specified in DPA spec
		bslName := fmt.Sprintf("%s-%d", r.NamespacedName.Name, i+1)
		if bslSpec.Name != "" {
			bslName = bslSpec.Name
		}
		dpaBSLNames = append(dpaBSLNames, bslName)

		...
	}

	dpaBSLs := velerov1.BackupStorageLocationList{}
	dpaBSLLabels := map[string]string{
		"app.kubernetes.io/name":       common.OADPOperatorVelero,
		"app.kubernetes.io/managed-by": common.OADPOperator,
		"app.kubernetes.io/component":  "bsl",
	}
	err := r.List(r.Context, &dpaBSLs, client.InNamespace(r.NamespacedName.Namespace), client.MatchingLabels(dpaBSLLabels))
	if err != nil {
		return false, err
	}
	if len(dpaBSLNames) != len(dpaBSLs.Items) {
		for _, bsl := range dpaBSLs.Items {
			if !slices.Contains(dpaBSLNames, bsl.Name) {
				if err := r.Delete(r.Context, &bsl); err != nil {
					return false, err
				}
				// event recorder
			}
		}
	}

	return true, nil
}
```

stillalearner: @mateusoliveira43 This solution should also work, basically it's the same thing,

You are getting the current BSLs in env. using the labels, and then looping over Dpa spec BSLs, and matching the names with slice variable stored.

I am also getting the current BSLs in env. using the labels, just that I am looping the other way around on the BSLs present in the environment, and then checking the name format using the same (length of dpaBSL.Items)

It really looks like the other way of doing the same thing to me..
Do you want me to change and test again?

stillalearner: But yes, it is definitely easier to understand, and lighter. So....changes pushed. 

Thanks @mateusoliveira43 !!

kaovilai (dismissed):

controllers/bsl.go
```diff
@@ -196,6 +200,33 @@ func (r *DPAReconciler) ReconcileBackupStorageLocations(log logr.Logger) (bool,
...
+		for _, bsl := range dpaBSLs.Items {
+			if !slices.Contains(dpaBSLNames, bsl.Name) {
+				if err := r.Delete(r.Context, &bsl); err != nil {
+					return false, err
+				}
+				// Record event for BSL deletion
+				r.EventRecorder.Event(&bsl,
+					corev1.EventTypeNormal,
+					"BackupStorageLocationDeleted",
+					fmt.Sprintf("BackupStorageLocation %s in namespace %s was deleted as it was not in spec.", bsl.Name, bsl.Namespace))
```
mateusoliveira43, line 225: Would be better to be more specific in message? Telling user that the deleted BSL/VSL was created by OADP (not by the user) and telling the object of the spec

A suggestion
> BackupStorageLocation %s **created by OADP** in namespace %s was deleted, because it was not in **DPA** spec.

mateusoliveira43: also need to change VSL message

mateusoliveira43 (dismissed):
Did some manual tests with BSL

> also good to do some tests with VSL

add new BSL to DPA **expect:** one creation/no deletion

    master ✅ PR ✅

change existing BSL spec from DPA **expect:** no creation/no deletion

    master ✅ PR ✅

change existing BSL name from DPA **expect:** one creation/one deletion

    master ❌ PR ✅

remove existing BSL from DPA **expect:** no creation/one deletion

    master ❌ PR ✅

delete DPA **expect:** no creation/all deletion

    master ✅ PR ✅

stillalearner:
> Did some manual tests with BSL
> 
> > also good to do some tests with VSL
> 
> add new BSL to DPA **expect:** one creation/no deletion
> 
> ```
> master ✅ PR ✅
> ```
> 
> change existing BSL spec from DPA **expect:** no creation/no deletion
> 
> ```
> master ✅ PR ✅
> ```
> 
> change existing BSL name from DPA **expect:** one creation/one deletion
> 
> ```
> master ❌ PR ✅
> ```
> 
> remove existing BSL from DPA **expect:** no creation/one deletion
> 
> ```
> master ❌ PR ✅
> ```
> 
> delete DPA **expect:** no creation/all deletion
> 
> ```
> master ✅ PR ✅
> ```

Thanks @mateusoliveira43 !! I also performed most of these test cases with BSL and VSL both and some other corner cases as well.

mateusoliveira43 (approved):

sseago (approved):

shubham-pampattiwar (approved):

kaovilai (approved):
/lgtm

weshayutin:
@stillalearner fyi ^

PR #1541, opened 2024-10-03
Set unsupported images via shell in our e2e tests

weshayutin:
## Why the changes were made

Generally speaking this will be handy for dev's to be able to easily use custom images in our e2e tests.
This especially is handy as we build out and use openshift/oadp-operator e2e tests across our various repos.

Each cloud provider has a json template for the dpa under tests/e2e/scripts.  All the unsupported Image Overrides were added to each provider.   If the variable is empty oadp will use the default images [1].  The map of strings is patched to the created dpa.

[1]
```
      podConfig:
        resourceAllocations: {}
  podDnsConfig: {}
  unsupportedOverrides:
    openshiftPluginImageFqin: ''
    gcpPluginImageFqin: ''
    veleroImageFqin: 'quay.io/rhn_engineering_whayutin/velero:testlatest'
    azurePluginImageFqin: ''
    resticRestoreImageFqin: ''
    nonAdminControllerImageFqin: ''
    kubevirtPluginImageFqin: ''
    awsPluginImageFqin: ''
    csiPluginImageFqin: ''
```

### DIFF
--- a/docs/developer/testing/TESTING.md
+++ b/docs/developer/testing/TESTING.md
@@ -94,6 +94,22 @@ You can also execute make test-e2e with a $GINKGO_ARGS variable set. Example:
 make test-e2e GINKGO_ARGS="--ginkgo.focus='MySQL application DATAMOVER'"
 ```
 
+### Run tests with custom images
+
+You can run tests with custom images by setting the following environment variables:
+```bash
+export VELERO_IMAGE=<velero_image>
+export AWS_PLUGIN_IMAGE=<aws_plugin_image>
+export OPENSHIFT_PLUGIN_IMAGE=<openshift_plugin_image>
+export AZURE_PLUGIN_IMAGE=<azure_plugin_image>
+export GCP_PLUGIN_IMAGE=<gcp_plugin_image>
+export CSI_PLUGIN_IMAGE=<csi_plugin_image>
+export RESTORE_IMAGE=<restore_image>
+export KUBEVIRT_PLUGIN_IMAGE=<kubevirt_plugin_image>
+export NON_ADMIN_IMAGE=<non_admin_image>
+```
+For further details, see tests/e2e/scripts/
+
 ## Clean up
 
 To clean environment after running E2E tests, run
--- a/docs/developer/testing/TESTING.md
+++ b/docs/developer/testing/TESTING.md
@@ -108,7 +108,7 @@ export RESTORE_IMAGE=<restore_image>
 export KUBEVIRT_PLUGIN_IMAGE=<kubevirt_plugin_image>
 export NON_ADMIN_IMAGE=<non_admin_image>
 ```
-For further details, see tests/e2e/scripts/
+For further details, see [tests/e2e/scripts/](../../../tests/e2e/scripts/)
 
 ## Clean up
 
--- a/tests/e2e/dpa_deployment_suite_test.go
+++ b/tests/e2e/dpa_deployment_suite_test.go
@@ -32,7 +32,6 @@ type TestDPASpec struct {
 	NoS3ForcePathStyle      bool
 	NoRegion                bool
 	DoNotBackupImages       bool
-	UnsupportedOverrides    map[oadpv1alpha1.UnsupportedImageKey]string
 }
 
 func createTestDPASpec(testSpec TestDPASpec) *oadpv1alpha1.DataProtectionApplicationSpec {
@@ -66,7 +65,7 @@ func createTestDPASpec(testSpec TestDPASpec) *oadpv1alpha1.DataProtectionApplica
 			},
 		},
 		SnapshotLocations:    testSpec.SnapshotLocations,
-		UnsupportedOverrides: testSpec.UnsupportedOverrides,
+		UnsupportedOverrides: dpaCR.UnsupportedOverrides,
 	}
 	if testSpec.EnableNodeAgent {
 		dpaSpec.Configuration.NodeAgent = &oadpv1alpha1.NodeAgentConfig{
@@ -357,14 +356,6 @@ var _ = ginkgo.Describe("Configuration testing for DPA Custom Resource", func() 
 				NoRegion:           true,
 				NoS3ForcePathStyle: true,
 				DoNotBackupImages:  true,
-			}),
-		}),
-		ginkgo.Entry("DPA CR with unsupportedOverrides", ginkgo.Label("aws", "ibmcloud"), InstallCase{
-			DpaSpec: createTestDPASpec(TestDPASpec{
-				BSLSecretName: bslSecretName,
-				UnsupportedOverrides: map[oadpv1alpha1.UnsupportedImageKey]string{
-					"awsPluginImageFqin": "quay.io/konveyor/velero-plugin-for-aws:oadp-1.4",
-				},
 			}),
 		}),
 	)
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -49,8 +49,7 @@ var (
 	knownFlake          bool
 	accumulatedTestLogs []string
 
-	kvmEmulation         bool
-	unsupportedOverrides map[string]string
+	kvmEmulation bool
 )
 
 func init() {
--- a/tests/e2e/e2e_suite_test.go
+++ b/tests/e2e/e2e_suite_test.go
@@ -49,7 +49,8 @@ var (
 	knownFlake          bool
 	accumulatedTestLogs []string
 
-	kvmEmulation bool
+	kvmEmulation         bool
+	unsupportedOverrides map[string]string
 )
 
 func init() {
@@ -169,6 +170,7 @@ func TestOADPE2E(t *testing.T) {
 		BSLBucketPrefix:      veleroPrefix,
 		VeleroDefaultPlugins: dpa.DeepCopy().Spec.Configuration.Velero.DefaultPlugins,
 		SnapshotLocations:    dpa.DeepCopy().Spec.SnapshotLocations,
+		UnsupportedOverrides: dpa.DeepCopy().Spec.UnsupportedOverrides,
 	}
 
 	ginkgo.RunSpecs(t, "OADP E2E using velero prefix: "+veleroPrefix)
--- a/tests/e2e/lib/dpa_helpers.go
+++ b/tests/e2e/lib/dpa_helpers.go
@@ -94,9 +94,7 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) *oadpv1al
 				},
 			},
 		},
-		UnsupportedOverrides: map[oadpv1alpha1.UnsupportedImageKey]string{
-			oadpv1alpha1.VeleroImageKey: v.UnsupportedOverrides[oadpv1alpha1.VeleroImageKey],
-		},
+		UnsupportedOverrides: v.UnsupportedOverrides,
 	}
 	switch backupRestoreType {
 	case RESTIC, KOPIA:
@@ -120,7 +118,7 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) *oadpv1al
 	// dpaSpec.UnsupportedOverrides = map[oadpv1alpha1.UnsupportedImageKey]string{
 	// 	oadpv1alpha1.VeleroImageKey: "quay.io/rhn_engineering_whayutin/velero:testlatest",
 	// }
-	// log.Printf("WES3 - DPA: %v", dpaSpec)
+	log.Printf("WES-Michal - DPA: %v", dpaSpec)
 	return &dpaSpec
 }
 
@@ -148,10 +146,6 @@ func (v *DpaCustomResource) Get() (*oadpv1alpha1.DataProtectionApplication, erro
 }
 
 func (v *DpaCustomResource) CreateOrUpdate(c client.Client, spec *oadpv1alpha1.DataProtectionApplicationSpec) error {
-	// for debugging
-	// prettyPrint, _ := json.MarshalIndent(spec, "", "  ")
-	// log.Printf("DPA with spec\n%s\n", prettyPrint)
-
 	dpa, err := v.Get()
 	log.Printf("WES2 - DPA: %v", dpa)
 	if err != nil {
--- a/tests/e2e/lib/dpa_helpers.go
+++ b/tests/e2e/lib/dpa_helpers.go
@@ -153,9 +153,8 @@ func (v *DpaCustomResource) CreateOrUpdate(c client.Client, spec *oadpv1alpha1.D
 				},
 				Spec: *spec.DeepCopy(),
 			}
-			dpaPatch := dpa.DeepCopy()
-			dpaPatch.Spec.UnsupportedOverrides = v.UnsupportedOverrides
-			return v.Create(dpaPatch)
+			dpa.Spec.UnsupportedOverrides = v.UnsupportedOverrides
+			return v.Create(dpa)
 		}
 		return err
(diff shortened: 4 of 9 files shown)


kaovilai (dismissed):
Wow. Just Amazing.

docs/developer/testing/TESTING.md
```diff
@@ -94,6 +94,22 @@ You can also execute make test-e2e with a $GINKGO_ARGS variable set. Example:
...
+export AWS_PLUGIN_IMAGE=<aws_plugin_image>
+export OPENSHIFT_PLUGIN_IMAGE=<openshift_plugin_image>
+export AZURE_PLUGIN_IMAGE=<azure_plugin_image>
+export GCP_PLUGIN_IMAGE=<gcp_plugin_image>
+export CSI_PLUGIN_IMAGE=<csi_plugin_image>
+export RESTORE_IMAGE=<restore_image>
+export KUBEVIRT_PLUGIN_IMAGE=<kubevirt_plugin_image>
+export NON_ADMIN_IMAGE=<non_admin_image>
+```
+For further details, see tests/e2e/scripts/
```
mateusoliveira43, line 111: would add link here (changing to `[tests/e2e/scripts/](../../../tests/e2e/scripts/)`)

tests/e2e/e2e_suite_test.go
```diff
@@ -49,7 +49,8 @@ var (
 	knownFlake          bool
 	accumulatedTestLogs []string
 
-	kvmEmulation bool
+	kvmEmulation         bool
+	unsupportedOverrides map[string]string
```
mateusoliveira43, line 53: unused, remove (a LINTER could have catch this...)

tests/e2e/lib/dpa_helpers.go
```diff
@@ -155,13 +150,16 @@ func (v *DpaCustomResource) CreateOrUpdate(c client.Client, spec *oadpv1alpha1.D
 				},
 				Spec: *spec.DeepCopy(),
 			}
-			return v.Create(dpa)
```
mateusoliveira43, line 158: All changes from this line on, I think are unnecessary

the workflow is: UnsupportedOverrides will be in json, `tests/e2e/e2e_suite_test.go` will load the values, line 96 of this file will add values to DPAs created by test run

changes being done on this line on are duplication of was already done in line 96

tests/e2e/lib/dpa_helpers.go
```diff
@@ -141,10 +140,6 @@ func (v *DpaCustomResource) Get() (*oadpv1alpha1.DataProtectionApplication, erro
 }
 
 func (v *DpaCustomResource) CreateOrUpdate(c client.Client, spec *oadpv1alpha1.DataProtectionApplicationSpec) error {
-	// for debugging
-	// prettyPrint, _ := json.MarshalIndent(spec, "", "  ")
-	// log.Printf("DPA with spec\n%s\n", prettyPrint)
```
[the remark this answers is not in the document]
weshayutin: fixed :) thanks

docs/developer/testing/TESTING.md
weshayutin: good suggestion

tests/e2e/e2e_suite_test.go
weshayutin: sure, but that is also useless imho as none of the vars are truly aligned. 
It's up for debate I guess, but I don't find that meaningful, other than I actually didn't change that code, so I should not have let my editor realign it.

tests/e2e/lib/dpa_helpers.go
weshayutin: You are incorrect.  There is duplication in how the tests create the dpa.  My addition requires the above.  I was also confused about that.   DPA creation is handled in two different ways.
Follow the dpa creation from the two points:
* make test-e2e GINKGO_ARGS="--ginkgo.focus='MySQL application DATAMOVER'"
* make test-e2e GINKGO_ARGS="--ginkgo.focus='DPA reconciled to true'"

Should we perhaps fix that yes, but it's certainly NOT in scope for this change :)  1 scope per PR.

docs/developer/testing/TESTING.md
```diff
@@ -94,6 +94,22 @@ You can also execute make test-e2e with a $GINKGO_ARGS variable set. Example:
...
+
+You can run tests with custom images by setting the following environment variables:
+```bash
+export VELERO_IMAGE=<velero_image>
+export AWS_PLUGIN_IMAGE=<aws_plugin_image>
+export OPENSHIFT_PLUGIN_IMAGE=<openshift_plugin_image>
+export AZURE_PLUGIN_IMAGE=<azure_plugin_image>
+export GCP_PLUGIN_IMAGE=<gcp_plugin_image>
+export CSI_PLUGIN_IMAGE=<csi_plugin_image>
+export RESTORE_IMAGE=<restore_image>
```
mateusoliveira43, line 107: These are the available values
https://github.com/openshift/oadp-operator/blob/d5adc852931f1e23a24c7f0be7f86a27d107ce2d/api/v1alpha1/oadp_types.go#L57-L64

CSI is not there anymore (it is not an image anymore)

What is `RESTORE_IMAGE`?

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: Understood what you want to do, but I do not think this is the correct approach

Since tests here https://github.com/openshift/oadp-operator/blob/d5adc852931f1e23a24c7f0be7f86a27d107ce2d/tests/e2e/dpa_deployment_suite_test.go do not use `DpaCustomResource.Build` function, you were not seeing `UnsupportedOverrides` there

But this change breaks this test
https://github.com/openshift/oadp-operator/blob/d5adc852931f1e23a24c7f0be7f86a27d107ce2d/tests/e2e/dpa_deployment_suite_test.go#L362-L369

If I export `AWS_PLUGIN_IMAGE`, or not, the value defined in json will overwrite value defined in test

mateusoliveira43: The way I would go with this:
- remove the test mentioned above (since the idea is to run these tests with values of `UnsupportedOverrides` in the near future, so test would not be necessary anymore)
- remove this
  https://github.com/openshift/oadp-operator/blob/d5adc852931f1e23a24c7f0be7f86a27d107ce2d/tests/e2e/dpa_deployment_suite_test.go#L35
- change `testSpec.UnsupportedOverrides` here to `dpaCR.UnsupportedOverrides`
  https://github.com/openshift/oadp-operator/blob/d5adc852931f1e23a24c7f0be7f86a27d107ce2d/tests/e2e/dpa_deployment_suite_test.go#L69

tests/e2e/e2e_suite_test.go
weshayutin: thanks for clarifying this bit, I misread your comment. appologies.

tests/e2e/lib/dpa_helpers.go
weshayutin: ok.. thank you for the in depth review!  I think I have made the changes you requested, I'm going to push an update and take some time to double check the changes and test.

docs/developer/testing/TESTING.md
weshayutin: CSI is removed.  We probably won't cp this back to 1.3 :)
no idea what ResticRestoreImage is, however it's in the spec. ?? ¯\_(ツ)_/¯

kaovilai: restic restore image is an init container added to pods to copy files to pvc before finishing so workload can start.

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: with the current changes, I think you can safely undo changes in lines 156-158 and line 165

the same changes are being applied to line 68 of `tests/e2e/dpa_deployment_suite_test.go`

docs/developer/testing/TESTING.md
mateusoliveira43: do not forget to also update documentation

weshayutin: thanks @kaovilai I had no clue, will update the doc

tests/e2e/lib/dpa_helpers.go
weshayutin: yes, I understand, however afaict you are still incorrect.
Here is what I see, and what happens when I test.
1.  line 68 of tests/e2e/dpa_deployment_suite_test.go will configure the dpa w/ overrides ONLY for the deployment test suite
2.  line 157/8 in dpa_helpers adds the overrides if a DPA exists at the time the backup_restore test suite runs
3.  line 165 in dpa_helpers adds the overrides if a DPA does NOT exist at the time the backup_restore test suite runs

Either follow the code paths, or test.. 
make test-e2e GINKGO_ARGS="--ginkgo.focus='DPA reconciled to true'"
w/ DPA make test-e2e GINKGO_ARGS="--ginkgo.focus='MySQL application CSI'"
w/o DPA make test-e2e GINKGO_ARGS="--ginkgo.focus='MySQL application CSI'"

weshayutin:
going to look at improving the mongo block mode test :/
/test 4.15-e2e-test-aws

tests/e2e/lib/dpa_helpers.go
mpryc: Reviewed the code again and the workflow that @mateusoliveira43 is raising it's because not all of the DPA CR's are generated using `Build()` function that set's the `UnsupportedOverrides`. The function itself is in:
   https://github.com/openshift/oadp-operator/blob/a5b0c0f9ed1c0cb41d745ab84e869635e87b5b07/tests/e2e/lib/dpa_helpers.go#L60-L97


We have two set's of tests which are taking advantage of the above function:
1. https://github.com/openshift/oadp-operator/blob/a5b0c0f9ed1c0cb41d745ab84e869635e87b5b07/tests/e2e/backup_restore_suite_test.go#L57
2. https://github.com/openshift/oadp-operator/blob/a5b0c0f9ed1c0cb41d745ab84e869635e87b5b07/tests/e2e/dpa_deployment_suite_test.go#L384

There is also `CreateOrUpdate` function where the DPA CR is being created outside of `Build` function, which also needs to be adjusted to take advantage of `UnsupportedOverrides`.
https://github.com/openshift/oadp-operator/blob/a5b0c0f9ed1c0cb41d745ab84e869635e87b5b07/tests/e2e/lib/dpa_helpers.go#L156-L164

So I do understand that the workflow may be simplified to have only one place to update `UnsupportedOverrides`, but I think it's outside of this PR and it's ok if it's being set even multiple times in favor of making smaller changes to the overall test functions.

We can have work to simplify this and ensure DPA CR is update via one generic function for all methods such as `Build` or `CreateOrUpdate`, but please correct me if I am wrong that it's outside of what it's being addressed by this PR?

mpryc: Yes, I wrote my comment to double that :)

mpryc (approved):
/lgtm

mpryc:
The build from images jobs is very much infrastructure one, I think it can be overriden:
```
Build src-arm64 previously failed from an infrastructure error (PullBuilderImageFailed)
```

tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -66,7 +65,7 @@ func createTestDPASpec(testSpec TestDPASpec) *oadpv1alpha1.DataProtectionApplica
 			},
 		},
 		SnapshotLocations:    testSpec.SnapshotLocations,
-		UnsupportedOverrides: testSpec.UnsupportedOverrides,
+		UnsupportedOverrides: dpaCR.UnsupportedOverrides,
```
kaovilai, line 68: You should be merging the two, or check that testSpec does not contain unsupportedOverrides since we are now no longer reading testSpec.UnsupportedOverrides.

kaovilai (changes requested):

tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -359,14 +358,6 @@ var _ = ginkgo.Describe("Configuration testing for DPA Custom Resource", func()
...
 			}),
 		}),
-		ginkgo.Entry("DPA CR with unsupportedOverrides", ginkgo.Label("aws", "ibmcloud"), InstallCase{
-			DpaSpec: createTestDPASpec(TestDPASpec{
-				BSLSecretName: bslSecretName,
-				UnsupportedOverrides: map[oadpv1alpha1.UnsupportedImageKey]string{
-					"awsPluginImageFqin": "quay.io/konveyor/velero-plugin-for-aws:oadp-1.4",
-				},
-			}),
-		}),
```
kaovilai, line 369: Why is this case removed? Is there a replacement?

tests/e2e/dpa_deployment_suite_test.go
kaovilai: And by check I mean, a code in this func that errors out on testSpec.UnsupportedOverrides. but still provide some way of testing unsupportedOverrides in CI, not just manual Makefile.

tests/e2e/dpa_deployment_suite_test.go
weshayutin: ya.. we'll be consistently running w/ overridden images in other repos and this test conflicts now w/ a users setting or ci setting to override the image.

kaovilai: ok cool. A comment in code to say we will have this coverage via other repos would be neat :).

kaovilai:
/hold for [checks](https://github.com/openshift/oadp-operator/pull/1541#pullrequestreview-2357135200)

feel free to unhold if ask is unreasonable.

tests/e2e/dpa_deployment_suite_test.go
weshayutin: hrm.. I could merge, but which one wins if they both set the same image?  IMHO since we're going to use this feature in openshift/velero ASAP and mig-tools/oadp-non-admin ASAP it's ok to leave out.

kaovilai:
Mr. ASAP wins..
/unhold

tests/e2e/dpa_deployment_suite_test.go
kaovilai: If we do ever get back to this, the testSpec wins since the cases with testSpec.unsupportedOverrides relies on those specific images. The rest will honor the read in environment for the rest of the test suite.. what we wanted for plugins/other repos tests.

weshayutin: I can put it back and retest.  My old brain *thinks* it wasn't overwritten by the test spec.  I can look again

kaovilai: the testSpec *should* win if we implement a merge functionality which isn't currently implemented.

kaovilai: im not saying anything is overwritten by testspec.. just saying testspec should win if we ever do a merge as testSpec is higher priority/required by a single case whereas env is for everywhere else.

mpryc: then we need to ensure none of testSpec's are currently overriding what's in the env, I think. Either way it's not a problem imo now

kaovilai: testSpec should override. testSpec is defined per case. We want ideally at least one case in oadp-operator main repo that check that override still works irregardless of what else is going on in other repos.

kaovilai: There should not be more than one or two testSpec so the majority should be using env.

PR #1552, opened 2024-10-10
fix issue #1464: Added VSL Backup/Restore Tests

### LINKED ISSUE #1464: VSL backup/restore test
Create a backup/restore test that uses a VSL.

stillalearner:
## Why the changes were made

This PR adds VSL backup / restore tests to the e2e tests, for both mysql and mongo apps.

Fixes #1464 

Use "make test-e2e" and run the new test cases with proper ginkgo variables.

### DIFF
--- a/tests/e2e/backup_restore_suite_test.go
+++ b/tests/e2e/backup_restore_suite_test.go
@@ -68,6 +68,8 @@ func prepareBackupAndRestore(brCase BackupRestoreCase, updateLastInstallTime fun
 		gomega.Eventually(lib.AreNodeAgentPodsRunning(kubernetesClientForSuiteRun, namespace), time.Minute*3, time.Second*5).Should(gomega.BeTrue())
 	}
 
+	// Velero does not change status of VSL objects. Users can only confirm if VSLs are correct configured when running a native snapshot backup/restore
+
 	log.Print("Checking if BSL is available")
 	gomega.Eventually(dpaCR.BSLsAreAvailable(), time.Minute*3, time.Second*5).Should(gomega.BeTrue())
 
@@ -404,6 +406,28 @@ var _ = ginkgo.Describe("Backup and restore tests", func() {
 				Namespace:         "mongo-persistent",
 				Name:              "mongo-blockdevice-e2e",
 				BackupRestoreType: lib.CSIDataMover,
+				PreBackupVerify:   todoListReady(true, false, "mongo"),
+				PostRestoreVerify: todoListReady(false, false, "mongo"),
+				BackupTimeout:     20 * time.Minute,
+			},
+		}, nil),
+		ginkgo.Entry("MySQL application Native-Snapshots", ginkgo.FlakeAttempts(flakeAttempts), ginkgo.Label("aws", "azure", "gcp"), ApplicationBackupRestoreCase{
+			ApplicationTemplate: "./sample-applications/mysql-persistent/mysql-persistent.yaml",
+			BackupRestoreCase: BackupRestoreCase{
+				Namespace:         "mysql-persistent",
+				Name:              "mysql-native-snapshots-e2e",
+				BackupRestoreType: lib.NativeSnapshots,
+				PreBackupVerify:   todoListReady(true, false, "mysql"),
+				PostRestoreVerify: todoListReady(false, false, "mysql"),
+				BackupTimeout:     20 * time.Minute,
+			},
+		}, nil),
+		ginkgo.Entry("Mongo application Native-Snapshots", ginkgo.FlakeAttempts(flakeAttempts), ginkgo.Label("aws", "azure", "gcp"), ApplicationBackupRestoreCase{
+			ApplicationTemplate: "./sample-applications/mongo-persistent/mongo-persistent.yaml",
+			BackupRestoreCase: BackupRestoreCase{
+				Namespace:         "mongo-persistent",
+				Name:              "mongo-native-snapshots-e2e",
+				BackupRestoreType: lib.NativeSnapshots,
 				PreBackupVerify:   todoListReady(true, false, "mongo"),
 				PostRestoreVerify: todoListReady(false, false, "mongo"),
 				BackupTimeout:     20 * time.Minute,
--- a/tests/e2e/lib/dpa_helpers.go
+++ b/tests/e2e/lib/dpa_helpers.go
@@ -24,16 +24,18 @@ import (
 type BackupRestoreType string
 
 const (
-	CSI          BackupRestoreType = "csi"
-	CSIDataMover BackupRestoreType = "csi-datamover"
-	RESTIC       BackupRestoreType = "restic"
-	KOPIA        BackupRestoreType = "kopia"
+	CSI             BackupRestoreType = "csi"
+	CSIDataMover    BackupRestoreType = "csi-datamover"
+	RESTIC          BackupRestoreType = "restic"
+	KOPIA           BackupRestoreType = "kopia"
+	NativeSnapshots BackupRestoreType = "native-snapshots"
 )
 
 type DpaCustomResource struct {
 	Name                 string
 	Namespace            string
 	Client               client.Client
+	VSLSecretName        string
 	BSLSecretName        string
 	BSLConfig            map[string]string
 	BSLProvider          string
@@ -112,6 +114,13 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) *oadpv1al
 		dpaSpec.Configuration.Velero.DefaultPlugins = append(dpaSpec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPluginCSI)
 		dpaSpec.Configuration.Velero.FeatureFlags = append(dpaSpec.Configuration.Velero.FeatureFlags, velero.CSIFeatureFlag)
 		dpaSpec.SnapshotLocations = nil
+	case NativeSnapshots:
+		dpaSpec.SnapshotLocations[0].Velero.Credential = &corev1.SecretKeySelector{
+			LocalObjectReference: corev1.LocalObjectReference{
+				Name: v.VSLSecretName,
+			},
+			Key: "cloud",
+		}
 	}
 
 	return &dpaSpec
--- a/tests/e2e/dpa_deployment_suite_test.go
+++ b/tests/e2e/dpa_deployment_suite_test.go
@@ -198,7 +198,7 @@ var _ = ginkgo.Describe("Configuration testing for DPA Custom Resource", func() 
 			}
 
 			if len(installCase.DpaSpec.SnapshotLocations) > 0 {
-				// TODO Check if VSLs are available creating new backup/restore test with VSL
+				// Velero does not change status of VSL objects. Users can only confirm if VSLs are correct configured when running a native snapshot backup/restore
 				for _, vsl := range installCase.DpaSpec.SnapshotLocations {
 					log.Printf("Checking for VSL spec")
 					gomega.Expect(dpaCR.DoesVSLSpecMatchesDpa(namespace, *vsl.Velero)).To(gomega.BeTrue())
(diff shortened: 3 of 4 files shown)


mateusoliveira43:
@stillalearner please link the issue to the PR. Here is some reference https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue

tests/e2e/lib/dpa_helpers.go
```diff
@@ -112,6 +114,13 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) *oadpv1al
 		dpaSpec.Configuration.Velero.DefaultPlugins = append(dpaSpec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPluginCSI)
 		dpaSpec.Configuration.Velero.FeatureFlags = append(dpaSpec.Configuration.Velero.FeatureFlags, velero.CSIFeatureFlag)
 		dpaSpec.SnapshotLocations = nil
+	case VSL:
+		dpaSpec.SnapshotLocations[0].Velero.Credential = &corev1.SecretKeySelector{
+			LocalObjectReference: corev1.LocalObjectReference{
+				Name: v.VSLSecretName,
+			},
+			Key: "cloud",
+		}
```
mateusoliveira43, line 123: also need to remove BSLs, right?

```go
dpaSpec.BackupLocations = nil
```

stillalearner: no, why? 

BSL + VSL is fine right.

tests/e2e/lib/dpa_helpers.go
```diff
@@ -28,12 +28,14 @@ const (
 	CSIDataMover BackupRestoreType = "csi-datamover"
 	RESTIC       BackupRestoreType = "restic"
 	KOPIA        BackupRestoreType = "kopia"
+	VSL          BackupRestoreType = "vsl"
```
mateusoliveira43, line 31: following  these changes https://github.com/openshift/oadp-operator/commit/68a1c47edd28a9acfd5f93bb76dd08989a6667ac I think is better to rename `vsl` here

Maybe `native-snapshots`?

tests/e2e/backup_restore_suite_test.go
```diff
@@ -409,5 +409,27 @@ var _ = ginkgo.Describe("Backup and restore tests", func() {
 				BackupTimeout:     20 * time.Minute,
 			},
 		}, nil),
+		ginkgo.Entry("MySQL application VSL", ginkgo.FlakeAttempts(flakeAttempts), ApplicationBackupRestoreCase{
```
mateusoliveira43, line 412: need to label these tests, as they will not work everywhere

```go
ginkgo.Label("aws", "azure", "gcp")
```

tests/e2e/lib/dpa_helpers.go
stillalearner: actually its VSL everywhere in the product documentation as well, whenever snapshotLocations are being configured. And its smaller and easy as well, maybe in this https://github.com/openshift/oadp-operator/commit/68a1c47edd28a9acfd5f93bb76dd08989a6667ac we can add  "(VSL)" in brackets?

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: just to be sure VSL is used? (for the other tests, VSLs are removed from DPA) I do not have experience with VSLs, some other maintainers input would be good

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: @weshayutin is this going to change in the docs?

tests/e2e/backup_restore_suite_test.go
stillalearner: its a VSL backup, i expect it to work everywhere. Not specific to AWS.

tests/e2e/lib/dpa_helpers.go
stillalearner: For other tests, VSLs are removed from the DPA. Because, the volumes are being backed up using other ways like csi, file system backup.

In the VSL case, Other resources are still backed up using BSL.
Just the volumes are being backed up using native snapshots workflow.

I am quite sure about that.

stillalearner: @shubham-pampattiwar @kaovilai can you please confirm.

kaovilai: BSLs are independent from VSL. To use VSL do not require removal of BSL.

kaovilai: If you use CSI, or file system backup, you do not need VSL. VSL is to allow use of velero cloud provider plugin "native snapshots" which do not require csi driver or a running pod on a pvc.

tests/e2e/backup_restore_suite_test.go
kaovilai: VSL only works where there is a velero cloud provider plugin with VSL support. Which is aws, azure, gcp, but not IBM cloud.

stillalearner: ah for IBM..sorry i misunderstood.

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: So, I am still confused. How Velero will know to use BSL or VSL to perform the backup?

stillalearner: It is not BSL or VSL

BSL always need to be there, to backup other K8s resources.

It is just about the volume backup part

If we specify, defaultVolumestoFsBackup -> it will do file system backup using restic/kopia

If we specify, snapshotMoveData -> it will perform csi snapshot with native datamover

If both are false, and VSL is there in spec, it will use Native Snapshot (VSL)

stillalearner: spec.VolumeSnapshotLocations is by default added, and used in case other methods of volume backups are false. So should not be needed to change this function.

tests/e2e/backup_restore_suite_test.go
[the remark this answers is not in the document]
stillalearner: No, same creds are fine. I have tested it already.

The issue with checking this availability of VSL might be because

VSL does not report any status.

Example: Below is a VSL created and ready to use. No status is reported in VSL CR.

`apiVersion: v1
items:
- apiVersion: velero.io/v1
  kind: VolumeSnapshotLocation
  metadata:
    creationTimestamp: "2024-10-11T17:05:48Z"
    generation: 1
    name: ts-dpa-1
    namespace: openshift-adp
    ownerReferences:
      blockOwnerDeletion: true
      controller: true
      name: ts-dpa
      uid: 4aa4e26d-a4aa-4643-b3ce-d3471b63308e
    resourceVersion: "218344"
    uid: a70a7894-be17-4616-8845-255b89d257bb
  spec:
    config:
      profile: default
      region: us-east-2
    provider: aws
  status: {}
kind: List
metadata:
  resourceVersion: ""
`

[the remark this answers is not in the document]
kaovilai: https://github.com/search?q=repo%3Avmware-tanzu%2Fvelero+VolumeSnapshotLocationPhaseAvailable&type=code

VSL velero has no code referencing phases so probably no status updates.

tests/e2e/lib/dpa_helpers.go
sseago: So VSL is not what we call the backup type -- that would be "native snapshots". More general, the terminology is as follows:
1) "native snapshots" -- Backup via cloud provider snapshots rather than CSI snapshots or filesystem backup.
2) "Volume Snapshotter Plugin" -- plugin that performs the native snapshot backup.
3) "VolumeSnapshotLocation (VSL)" -- the CR that tells Velero how/where the snapshots are saved. Usually this just means configuring the region and credentials for a particular provider.

So VSL is not the functionality as a whole, it's a configuration CR.

stillalearner: ack.

tests/e2e/backup_restore_suite_test.go
```diff
@@ -409,5 +409,27 @@ var _ = ginkgo.Describe("Backup and restore tests", func() {
 				BackupTimeout:     20 * time.Minute,
 			},
 		}, nil),
+		ginkgo.Entry("MySQL application VSL", ginkgo.FlakeAttempts(flakeAttempts), ginkgo.Label("aws", "azure", "gcp"), ApplicationBackupRestoreCase{
```
mateusoliveira43, line 412: Also change tests names from VSL to native snapshots, for consistency

tests/e2e/lib/dpa_helpers.go
mateusoliveira43: So, after discussion with the team, I was wrong about removal of BSL, it is needed for native snapshots backups

And yes, velero automatically adds bsl/vsl ref to backup spec

tests/e2e/backup_restore_suite_test.go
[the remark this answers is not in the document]
mateusoliveira43: We looked into velero code, and unfortunately it does not populate VSL status :( 

Can you please update the comment here 
https://github.com/openshift/oadp-operator/blob/fb2411e2948874a6abe61ca31b1cd9957621ebb6/tests/e2e/dpa_deployment_suite_test.go#L201

to something like
> Velero does not change status of VSL objects. Users can only confirm if VSLs are correct configured when running a native snapshot backup/restore

And also add the same comment here
https://github.com/openshift/oadp-operator/blob/fb2411e2948874a6abe61ca31b1cd9957621ebb6/tests/e2e/backup_restore_suite_test.go#L71-L72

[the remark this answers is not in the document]
stillalearner: done.

mateusoliveira43 (approved):

tests/e2e/backup_restore_suite_test.go
```diff
@@ -68,6 +68,8 @@ func prepareBackupAndRestore(brCase BackupRestoreCase, updateLastInstallTime fun
 		gomega.Eventually(lib.AreNodeAgentPodsRunning(kubernetesClientForSuiteRun, namespace), time.Minute*3, time.Second*5).Should(gomega.BeTrue())
 	}
 
+	// Velero does not change status of VSL objects. Users can only confirm if VSLs are correct configured when running a native snapshot backup/restore
```
kaovilai, line 71: I think it was said in some scrum in 2021/2022 that we don't need VSL tests when implementing e2e due to "CSI is the future, we will not suggests VSL to anyone new onboarding". Just an info.

tests/e2e/lib/dpa_helpers.go
```diff
@@ -112,6 +114,13 @@ func (v *DpaCustomResource) Build(backupRestoreType BackupRestoreType) *oadpv1al
 		dpaSpec.Configuration.Velero.DefaultPlugins = append(dpaSpec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPluginCSI)
 		dpaSpec.Configuration.Velero.FeatureFlags = append(dpaSpec.Configuration.Velero.FeatureFlags, velero.CSIFeatureFlag)
 		dpaSpec.SnapshotLocations = nil
+	case NativeSnapshots:
+		dpaSpec.SnapshotLocations[0].Velero.Credential = &corev1.SecretKeySelector{
```
shubham-pampattiwar, line 118: Shouldn't we just load the VSL spec only for the `NativeSnapshots` spec ? Here we are only adding the credential reference, IMO we should configure VSL only for this case and not by default.

tests/e2e/backup_restore_suite_test.go
stillalearner: your call. I guess keeping tests if functionality is there / not deprecated yet is fine..

kaovilai: adding tests is fine yes. Pretty sure we had it at some point [here](https://github.com/openshift/oadp-operator/pull/568/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R8)

tests/e2e/lib/dpa_helpers.go
shubham-pampattiwar: Cool, I am fine with it.

shubham-pampattiwar (approved):

PR #1558, opened 2024-10-15
Add Design for Upload Speed Test with BSL Config for Object Storage

shubham-pampattiwar:
## Why the changes were made

Design for Upload Speed Test CRD and Controller

Review the Design


docs/design/speed-test.md
```diff
@@ -0,0 +1,118 @@
...
+
+- UploadSpeedTest (UST) CRD:
+  - Captures the BSL configuration used for testing directly in the CRD.
+  - Holds the test results, including success/failure and upload speed.
+- UploadSpeedTest (UST) Controller:
+  - Watches for changes to the UploadSpeedTest CRs and executes uploads based on the provided BSL configuration. 
+  - Uses appropriate cloud SDKs (AWS, Azure, or GCP) to perform uploads.
+  - Updates the status of UST CR with upload speed, upload success/failure and error message if any.
+- OADP/DPA-UST Integration:
+  - Fetches BackupStorageLocation (BSL) details from the OADP operator’s DPA CR.
```
weshayutin, line 43: if a customer has multiple bsl's configured in the dpa will it go through each?

docs/design/speed-test.md
```diff
@@ -0,0 +1,118 @@
...
+metadata:
+  name: sample-dpa
+spec:
+  backupStorageLocations:
+    - name: aws-backup-location
+      provider: aws
+      config:
+        region: us-east-1
+        bucket: my-backup-bucket
+        uploadSpeedTest:
```
weshayutin, line 87: If we're not going to support automatic speed checks, I'm confused why the DPA needs to change initially.

docs/design/speed-test.md
```diff
@@ -0,0 +1,118 @@
...
+- OADP/DPA-UST Integration:
+  - Fetches BackupStorageLocation (BSL) details from the OADP operator’s DPA CR. 
+  - Populates the UploadSpeedTest CR with relevant configurations from the BSL.
+  - Creates the UST CR for the BSL configured in DPA instance.
+## Detailed Design
+
+Proposed Upload Speed Test CRD:
+
+```yaml
+apiVersion: uploadspeedtest.io/v1alpha1
```
kaovilai, line 51: why not oadp.openshift.io? or velero.io?
Make this not dependent on dpa?

docs/design/speed-test.md
shubham-pampattiwar: Auto-Check or not depends on the user configuration here. For each BSL configured we have a `uploadSpeedTest` struct that has `enabled` flag and `uploadSpeedTest` config. If the user sets these then UST CR would be automatically created by OADP Operator for that particular BSL, else the user could create the UST CR themselves.

docs/design/speed-test.md
shubham-pampattiwar: yes its per BSL configured in the DPA, we are adding a uploadSpeedTest struct, please refer the proposed CRD section and DPA spec changes section.

docs/design/speed-test.md
shubham-pampattiwar: yeah need to update that to `oadp.openshift.io`

docs/design/speed-test.md
```diff
@@ -0,0 +1,210 @@
+# Upload Speed Test for Backup Storage Location (Object Storage)
+
+## Abstract
+
+This document presents the design of a Custom Resource Definition (CRD) and its controller to test the upload speed from an OpenShift cluster to cloud object storage.
+The design leverages the BackupStorageLocation configuration from the OADP operator’s DPA CRD to specify object storage locations for testing. The controller will use
```
mateusoliveira43, line 6: we intend to only measure BSLs created by DPA and not all BSLs in OADP namespace?

docs/design/speed-test.md
```diff
@@ -0,0 +1,210 @@
...
+      credential:
+        name: cloud-credentials  # Secret containing cloud credentials
+        key: cloud
+  uploadSpeedTestConfig:
+    fileSize: "100MB"    # File size for the test upload
+    testTimeout: "60s"   # Maximum duration allowed for the test
+  cloudProviderSecretRef:
+    name: cloud-credentials
+    namespace: openshift-adp
+```
```
mateusoliveira43, line 78: Could we change here to only have 2 fields in spec (`uploadSpeedTestConfig` and `backupLocationName`)?

`cloudProviderSecretRef` can be fetched from BSL, right? So does not seem needed

Having BSL spec copied here makes it harder for user to manually create one (even thinking about DPA integration, if BSL spec changes, UST must change. Having just name reference would not simplify workflow?)

docs/design/speed-test.md
```diff
@@ -0,0 +1,210 @@
...
+## Non-Goals
+
+- This design will not create or modify BackupStorageLocation entries in OADP.
+- It will not implement download or latency tests, focusing solely on upload speed.
+- Scheduling of recurring tests is not supported in the initial version.
+
+## High-Level Design
+Components involved and their responsibilities:
+
+- UploadSpeedTest (UST) CRD:
```
mateusoliveira43, line 35: Reading whole design, would make sense to change CRD name to something more generic (NetworkSpeedTest?). If more features, like latency and download speed, are added, the CRD name would not be telling users all things it can do?

docs/design/speed-test.md
shubham-pampattiwar: So these are 2 separate things:
- You can independently create UST CR for any BSL config (so this would work for OADP BSL and Non-OADP BSLs)
- We have a trigger for OADP BSLs (in this case DPA controller will create a UST CR)
So all the scenarios are covered I believe.

docs/design/speed-test.md
shubham-pampattiwar: Yeah I had thought about that, Any Network speed would consist of upload as well as download speed, the risk here with `NetworkSpeedTest` is that it might give a false impression that we are catering/calculating download speed as well, but thats not our goal here. Hence `UploadSpeedTest`.

docs/design/speed-test.md
shubham-pampattiwar: Done, removed `cloudProviderSecretRef` from UST CRD spec, tested AWS POC as well.

docs/design/speed-test.md
```diff
@@ -0,0 +1,203 @@
...
+      config:
+        region: us-east-1
+        profile: "default"
+        insecureSkipTLSVerify: "true"
+      credential:
+        name: cloud-credentials  # Secret containing cloud credentials
+        key: cloud
+  uploadSpeedTestConfig:
+    fileSize: "100MB"    # File size for the test upload
+    testTimeout: "60s"   # Maximum duration allowed for the test
```
kaovilai, line 74: I would add number of files WDYT? but could be follow up enhancements.
I know that some server are able to speed up upload via multiple connections vs single connection.
```suggestion
  uploadSpeedTestConfig:
    numBigFiles: 2 # number of files to divide file size after small files are uploaded
    numSmallFiles: 100 # number of 1KB files (simulating file system backup of dirs with small files) 
    fileSize: "100MB"    # File size for the test upload
    testTimeout: "60s"   # Maximum duration allowed for the test
```

docs/design/speed-test.md
```diff
@@ -0,0 +1,203 @@
...
+
+The UploadSpeedTest CRD structure includes fields to capture backup storage location configurations, upload test parameters, and the location of cloud provider secrets.
+
+```yaml
+apiVersion: oadp.openshift.io/v1alpha1
+kind: UploadSpeedTest
+metadata:
+  name: my-upload-speed-test
+spec:
+  backupLocation:
```
kaovilai, line 58:
```suggestion
    backupStorageLocationName: <name of bsl to test>
    backupLocation:
```

kaovilai (changes requested):
we said we [want to](https://github.com/openshift/oadp-operator/pull/1558/files#r1824480005) ref by name so make `.spec.backupLocation.<>` optional if `.spec.backupLocationName` is specified.

docs/design/speed-test.md
shubham-pampattiwar: I think we want UST to be able to be used independently (in addition to via DPA) by providing the BSLConfig. Thinking of the the use-case:
- User has a BSL Config,
- they test the BSLConfig via UST
- now use the BSLConfig to create BSL

kaovilai: So we do not want to reference by name to an already existing BSL?

shubham-pampattiwar: Yeah because we already have the functionality to test a particular BSL via DPA, [proposed here](https://github.com/openshift/oadp-operator/blob/7250bac87d17dcc25417ee87b8644906bda24523/docs/design/speed-test.md?plain=1#L91) . Why keep it bi-directional ! I believe this keeps the CRD simpler.

kaovilai: So an upstream velero user would not be able to use this without copying the whole bsl.spec?

kaovilai: Perhaps that was not the plan.

kaovilai (commented):
Can you clarify which path it would be writing to?
How will it handle <insert local provider> "bucket full" related errors?
Let say I wanted to test upload of 2TB, we hit bucket full at 1TB. Do we throw away current results and error? or will it preserve info gathered so far ie. how long it took to upload 1TB even tho user specified 2TB?

kaovilai (commented):
Do we want to have continuous test as an option?
ie. don't stop uploading until user change spec from running to false.

Will user see live progress of the current speed (ie. speed over short time interval like 1s)? or only after full test will they get results?

docs/design/speed-test.md
weshayutin: not an upstream feature :)

weshayutin (approved):
This looks good to me.. Can you please clarify that when an admin is configuring the DPA w/ usl settings that nothing ( velero ) is restarted etc..   Just to make that point obvious

docs/design/speed-test.md
kaovilai: Sounds like an easy add. If it does not complicate our designs, I would propose we do things in an upstream usable way.

It'd be much cooler to say "you can try our team's tool to debug your bsl speed"
vs
"If you were using OADP, you'd have ability to debug your bsl speed"

weshayutin:
@shubham-pampattiwar and I were in a mtg where it was pointed out the storage system may also be a bottle neck. 
We could potentially as part of this test.
1. create a pod/pv/pvc 
2.  Execute a csi snapshot ( NOT w/ OADP/velero)
3.  Time and post the results. 

This would provide a baseline for the time it takes OpenShift to snapshot $x data.

shubham-pampattiwar:
Closed PR because GitHub was stuck at processing for 2 hours after a force push. 😓

shubham-pampattiwar:
Created a new PR: https://github.com/openshift/oadp-operator/pull/1706 :)

Issue #1536, opened 2024-09-30
Improve DPA reconcile error messages


mateusoliveira43:
## Problem

It is difficult for users (and developers) to debug OADP errors without more context (studying DPA, for example).

## Example

Some months ago, a new user was not able to set custom credential name to BSL (managed by DPA), as DPA would reconcile false with the generic error
```log
Secret "cloud-credentials" not found
```
After some hours, OADP team asked for user DPA, and the problem was found. There was also VSL defined in DPA (probably copy/paste from example DPA)

https://github.com/openshift/oadp-operator/blob/68a1c47edd28a9acfd5f93bb76dd08989a6667ac/config/samples/oadp_v1alpha1_dpa.yaml#L28-L33

This problem could have been resolved immediately if the error message was more informative. Example
 ```log
VolumeSnapshotLocation <VSL-name> defined in DPA spec.snapshotLocations[<index>] is invalid: 
    create Secret "cloud-credentials" in namespace <namespace-name> or change Secret name in DPA spec.snapshotLocations[<index>].velero.credential.name
```

## Solution

Solution would be to rewrite OADP error messages to be more informative.

A first approach suggestion is to check error messages in tests that are already written, BUT do NOT check error messages.

### Example

This test

https://github.com/openshift/oadp-operator/blob/68a1c47edd28a9acfd5f93bb76dd08989a6667ac/controllers/bsl_test.go#L57

only checks if error occurred, without testing the error message.

> This is very WRONG. Some of these tests are broken, because their error messages do not relate to what is being tested. But they are are not failing in CI, as we check only if an error occurred, and not what error occurred.

Maybe this command can help found other tests like this
```sh
grep -Iinr 'wantErr:' . --include=\*_test.go
```

PR #1559, opened 2024-10-15
fix OADP-5044 and other VSL validation errors

stillalearner:
## Why the changes were made

Fixed some VSL validation errors like:

-> throwing error on incorrect provider in VSL
-> throwing error on missing region in AWS VSL
-> throwing error if missing plugins for AWS, GCP, Azure VSLs

Use "make deploy-olm" and configure VSL in DPA spec.

1. Provide incorrect provider in VSL
2. Configure AWS VSL, and dont provide region in VSL spec.
3. Add AWS/GCP/Azure VSL and dont add relevant velero plugin.

### DIFF
--- a/controllers/validator_test.go
+++ b/controllers/validator_test.go
@@ -1038,13 +1038,16 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
 					SnapshotLocations: []oadpv1alpha1.SnapshotLocation{
 						{
 							Velero: &v1.VolumeSnapshotLocationSpec{
-								Provider: "velero.io/aws",
+								Provider: "aws",
 								Credential: &corev1.SecretKeySelector{
 									LocalObjectReference: corev1.LocalObjectReference{
 										Name: "custom-vsl-credentials",
 									},
 									Key:      "cloud",
 									Optional: new(bool),
+								},
+								Config: map[string]string{
+									"region": "us-east-1",
 								},
 							},
 						},
--- a/controllers/vsl.go
+++ b/controllers/vsl.go
@@ -110,53 +110,37 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
 
 func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataProtectionApplication) (bool, error) {
 	for i, vslSpec := range dpa.Spec.SnapshotLocations {
+		vslYAMLPath := fmt.Sprintf("spec.snapshotLocations[%v]", i)
+		veleroVSLYAMLPath := vslYAMLPath + ".velero"
+		veleroConfigYAMLPath := "spec.configuration.velero"
+
 		if vslSpec.Velero == nil {
 			return false, errors.New("snapshotLocation velero configuration cannot be nil")
-		}
-		vsl := velerov1.VolumeSnapshotLocation{
-			ObjectMeta: metav1.ObjectMeta{
-				// TODO: Use a hash instead of i
-				Name:      fmt.Sprintf("%s-%d", r.NamespacedName.Name, i+1),
-				Namespace: r.NamespacedName.Namespace,
-			},
-			Spec: *vslSpec.Velero,
 		}
 
 		// check for valid provider
 		if vslSpec.Velero.Provider != AWSProvider && vslSpec.Velero.Provider != GCPProvider &&
-			vslSpec.Velero.Provider != Azure {
-			r.Log.Info("Non-supported provider specified, might be a misconfiguration")
-
-			r.EventRecorder.Event(&vsl,
-				corev1.EventTypeWarning,
-				"VSL provider is invalid",
-				fmt.Sprintf("VSL provider %s is invalid, might be a misconfiguration", vslSpec.Velero.Provider),
-			)
+			vslSpec.Velero.Provider != AzureProvider {
+			return false, fmt.Errorf("DPA %s.provider %s is invalid: only %s, %s and %s are supported", veleroVSLYAMLPath, vslSpec.Velero.Provider, AWSProvider, GCPProvider, AzureProvider)
 		}
 
 		//AWS
 		if vslSpec.Velero.Provider == AWSProvider {
 			//in AWS, region is a required field
 			if len(vslSpec.Velero.Config[AWSRegion]) == 0 {
-				return false, errors.New("region for AWS VSL is not configured, please ensure a region is configured")
+				return false, fmt.Errorf("region for %s VSL in DPA %s.config is not configured, please ensure a region is configured", AWSProvider, veleroVSLYAMLPath)
 			}
 
 			// check for invalid config key
 			for key := range vslSpec.Velero.Config {
 				valid := validAWSKeys[key]
 				if !valid {
-					return false, fmt.Errorf("%s is not a valid AWS config value", key)
+					return false, fmt.Errorf("DPA %s.config key %s is not a valid %s config key", veleroVSLYAMLPath, key, AWSProvider)
 				}
 			}
 			//checking the aws plugin, if not present, throw warning message
 			if !containsPlugin(dpa.Spec.Configuration.Velero.DefaultPlugins, AWSProvider) {
-				r.Log.Info("VSL for AWS specified, but AWS plugin not present, might be a misconfiguration")
-
-				r.EventRecorder.Event(&vsl,
-					corev1.EventTypeWarning,
-					"VolumeSnapshotLocation is invalid",
-					fmt.Sprintf("could not validate vsl for AWS plugin on: %s/%s", vsl.Namespace, vsl.Name),
-				)
+				return false, fmt.Errorf("to use VSL for %s specified in DPA %s, %s plugin must be present in %s.defaultPlugins", AWSProvider, vslYAMLPath, AWSProvider, veleroConfigYAMLPath)
 			}
 		}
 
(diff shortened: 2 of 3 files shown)


controllers/vsl.go
```diff
@@ -122,41 +122,38 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
 			Spec: *vslSpec.Velero,
 		}
 
+		// check if VSL name is specified in DPA spec
```
mateusoliveira43, line 125: Here, lets delete `vsl` variable (it was used only by `EventRecorder`) and use same approach as the other function in this file does to get VSL name

https://github.com/openshift/oadp-operator/blob/fb2411e2948874a6abe61ca31b1cd9957621ebb6/controllers/vsl.go#L227-L231

controllers/vsl.go
```diff
@@ -122,41 +122,38 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
...
+		if vslSpec.Name != "" {
+			vsl.Name = vslSpec.Name
+		}
+
 		// check for valid provider
 		if vslSpec.Velero.Provider != AWSProvider && vslSpec.Velero.Provider != GCPProvider &&
 			vslSpec.Velero.Provider != Azure {
 			r.Log.Info("Non-supported provider specified, might be a misconfiguration")
```
mateusoliveira43, line 133: I think we should remove these logs, OR, change them to be level error, since we are erroring it out now

controllers/vsl.go
```diff
@@ -122,41 +122,38 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
...
 
-			r.EventRecorder.Event(&vsl,
-				corev1.EventTypeWarning,
-				"VSL provider is invalid",
-				fmt.Sprintf("VSL provider %s is invalid, might be a misconfiguration", vslSpec.Velero.Provider),
-			)
+			return false, fmt.Errorf("invalid provider %s for VSL %s", vslSpec.Velero.Provider, vsl.Name)
```
mateusoliveira43, line 135: I really think we should start to give more verbose error messages to users, as I explain here https://github.com/openshift/oadp-operator/issues/1536

My suggestion for here, for example, would be
> DPA spec.snapshotLocations[**index**].velero.provider is invalid: **wrong-one** is not a valid value. Use of of **options**

controllers/validator_test.go
```diff
@@ -1038,14 +1038,17 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
 					SnapshotLocations: []oadpv1alpha1.SnapshotLocation{
 						{
 							Velero: &v1.VolumeSnapshotLocationSpec{
-								Provider: "velero.io/aws",
```
mateusoliveira43, line 1041: Looking this code, I think this is not a valid change
https://github.com/openshift/oadp-operator/blob/fb2411e2948874a6abe61ca31b1cd9957621ebb6/controllers/bsl.go#L50-L68
validation should accept both `aws` and `velero.io/aws` (but this should be fixed in the future, we can open an issue to discuss only allowing `aws`), right maintainers?

controllers/vsl.go
```diff
@@ -122,41 +122,38 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
...
-				return false, errors.New("region for AWS VSL is not configured, please ensure a region is configured")
+				return false, fmt.Errorf("region for AWS VSL %s is not configured, please ensure a region is configured", vsl.Name)
 			}
 
 			// check for invalid config key
 			for key := range vslSpec.Velero.Config {
 				valid := validAWSKeys[key]
 				if !valid {
-					return false, fmt.Errorf("%s is not a valid AWS config value", key)
+					return false, fmt.Errorf("%s is not a valid AWS config value for VSL %s", key, vsl.Name)
```
mateusoliveira43, line 149: here message is wrong, no?

It is referencing the `key` not the `value`, in formatting

controllers/vsl.go
```diff
@@ -122,41 +122,38 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
...
 			//checking the aws plugin, if not present, throw warning message
 			if !containsPlugin(dpa.Spec.Configuration.Velero.DefaultPlugins, AWSProvider) {
 				r.Log.Info("VSL for AWS specified, but AWS plugin not present, might be a misconfiguration")
 
-				r.EventRecorder.Event(&vsl,
-					corev1.EventTypeWarning,
-					"VolumeSnapshotLocation is invalid",
-					fmt.Sprintf("could not validate vsl for AWS plugin on: %s/%s", vsl.Namespace, vsl.Name),
-				)
+				return false, fmt.Errorf("VSL %s for AWS specified, but AWS plugin not present, might be a misconfiguration", vsl.Name)
```
mateusoliveira43, line 156: we do not have this kind of validation for BSL, right? Should we add it (if yes, another PR)

mateusoliveira43 (changes requested):
Added some comments, but not much changes needed for approval :smile:

controllers/validator_test.go
stillalearner: @mateusoliveira43 I have fixed this intentionally, for now , for this test to pass.

Because this is a bug here actually, there are multiple issues regarding this, which i plan to open soon.

Then, we can separately fix this issue.

controllers/vsl.go
stillalearner: ack

controllers/vsl.go
stillalearner: ack

controllers/vsl.go
stillalearner: Is the name of VSL needed then?

controllers/vsl.go
stillalearner: hmm, probably not, will confirm and create a PR if not. Noted.

controllers/vsl.go
mateusoliveira43: I think just yaml path is fine (and most of the time users do not even put name on those objects, right? so it will not be a valuable information for them) 

With yaml path I would say error message is good enough, because if user does not understand it, we will understand it immediately and can easily help the user fix the problem

stillalearner: ya sounds good. Thanks

controllers/vsl.go
stillalearner: ack

controllers/validator_test.go
mateusoliveira43: Since this changes behavior (prior to this PR, users could specify `aws` or `velero.io/aws` in VSL, and after it, users will be able to only specify `aws`) I think is important to discuss this prior to merge

(easy solution could be that if user specifies `velero.io/aws`, the error message tell it to use `aws` instead, but we need to verity our deprecation policy as well, maybe first we need to deprecate it)

stillalearner: no, no, this PR does not change any behaviour, i have only modified the test case to use "aws" for sometime.
Earlier, velero.io/aws was also creating a Warning event, and was getting ignored.

The reason it is failing now with "velero.io" prefix is a different bug, which is causing issue now. I will raise a thread in slack. That needs to be fixed first, and after that, we can have any value here.

controllers/vsl.go
```diff
@@ -113,50 +113,33 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations(dpa oadpv1alpha1.DataPro
...
 		if vslSpec.Velero.Provider != AWSProvider && vslSpec.Velero.Provider != GCPProvider &&
 			vslSpec.Velero.Provider != Azure {
 			r.Log.Info("Non-supported provider specified, might be a misconfiguration")
 
-			r.EventRecorder.Event(&vsl,
-				corev1.EventTypeWarning,
-				"VSL provider is invalid",
-				fmt.Sprintf("VSL provider %s is invalid, might be a misconfiguration", vslSpec.Velero.Provider),
-			)
+			return false, fmt.Errorf("DPA spec.snapshotLocations[%v].velero.provider %s is invalid: only AWS, GCP and Azure are supported", i, vslSpec.Velero.Provider)
```
mateusoliveira43, line 122: I would use variables inside error messages as much as possible

For example, creating these variables inside for loop (after line 112)
```go
vslYAMLPath := fmt.Sprintf("spec.snapshotLocations[%v]", i)
veleroVSLYAMLPath := vslYAMLPath + ".velero"
```
So if these paths update in the future, it is easy to update error messages (only one place, instead of every message individually)

And then changing this message to
```go
return false, fmt.Errorf("DPA %v.provider %s is invalid: only %v are supported", veleroVSLYAMLPath, vslSpec.Velero.Provider, getVSLPluglins())
```
(`getVSLPluglins` does not exist, created for the example)

What you think?

stillalearner: ya, variable part is good.

stillalearner: changes pushed. Thanks you

controllers/vsl.go
mateusoliveira43: only this log was left inside `ValidateVolumeSnapshotLocations` function, do not forget to remove it :)

stillalearner: oops. done.

controllers/vsl.go
```diff
@@ -110,53 +110,37 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
...
 		if vslSpec.Velero.Provider != AWSProvider && vslSpec.Velero.Provider != GCPProvider &&
 			vslSpec.Velero.Provider != Azure {
-			r.Log.Info("Non-supported provider specified, might be a misconfiguration")
-
-			r.EventRecorder.Event(&vsl,
-				corev1.EventTypeWarning,
-				"VSL provider is invalid",
-				fmt.Sprintf("VSL provider %s is invalid, might be a misconfiguration", vslSpec.Velero.Provider),
-			)
+			return false, fmt.Errorf("DPA %s.provider %s is invalid: only AWS, GCP and Azure are supported", veleroVSLYAMLPath, vslSpec.Velero.Provider)
```
mateusoliveira43, line 123: lets use `AWSProvider`, `GCPProvider` and `AzureProvider` variables here in the message

controllers/vsl.go
```diff
@@ -110,53 +110,37 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
...
 		}
 
 		//AWS
 		if vslSpec.Velero.Provider == AWSProvider {
 			//in AWS, region is a required field
 			if len(vslSpec.Velero.Config[AWSRegion]) == 0 {
-				return false, errors.New("region for AWS VSL is not configured, please ensure a region is configured")
+				return false, fmt.Errorf("region for AWS VSL in DPA %s.Config is not configured, please ensure a region is configured", veleroVSLYAMLPath)
```
mateusoliveira43, line 130: Typo in `config` (better to be lowercase, matching json tag, so commands like `oc get dpa velero-sample -n openshift-adp -o=jsonpath="{.spec.snapshotLocations[0].velero.config}"` work)

controllers/vsl.go
```diff
@@ -110,53 +110,37 @@ func (r *DPAReconciler) LabelVSLSecrets(log logr.Logger) (bool, error) {
...
 			//checking the aws plugin, if not present, throw warning message
 			if !containsPlugin(dpa.Spec.Configuration.Velero.DefaultPlugins, AWSProvider) {
-				r.Log.Info("VSL for AWS specified, but AWS plugin not present, might be a misconfiguration")
 
-				r.EventRecorder.Event(&vsl,
-					corev1.EventTypeWarning,
-					"VolumeSnapshotLocation is invalid",
-					fmt.Sprintf("could not validate vsl for AWS plugin on: %s/%s", vsl.Namespace, vsl.Name),
-				)
+				return false, fmt.Errorf("VSL for AWS specified in DPA %s, but AWS plugin not present, might be a misconfiguration", vslYAMLPath)
```
mateusoliveira43, line 143: better to use imperative form here, no?

would also add path to aws plugin config in DPA

Suggestion 

> to use VSL for AWS specified in DPA spec.snapshotLocations[index], AWS plugin must be present in spec.configuration.velero.defaultPlugins

shubham-pampattiwar (changes requested):
@stillalearner Please add specific unit tests for the validation error case scenarios related to this PR:
-> throwing error on incorrect provider in VSL
-> throwing error on missing region in AWS VSL
-> throwing error if missing plugins for AWS, GCP, Azure VSLs

stillalearner:
> @stillalearner Please add specific unit tests for the validation error case scenarios related to this PR: -> throwing error on incorrect provider in VSL -> throwing error on missing region in AWS VSL -> throwing error if missing plugins for AWS, GCP, Azure VSLs

Sure, Thanks! All scenarios covered now.

controllers/vsl_test.go
mateusoliveira43, line 1: For follow up (you already did an amazing job in this PR @stillalearner!!!)

As I explain in https://github.com/openshift/oadp-operator/issues/1536 these tests do not gives us confidence, because we do not check error messages (only if function errors out). So some tests here may be currently wrong

mateusoliveira43 (approved):

controllers/vsl_test.go
stillalearner: Thanks @mateusoliveira43 !

Yes, for the error messages, dont worry, Its getting ready in another PR as part of 1536 fixes.

kaovilai (approved):
/lgtm

PR #1564, opened 2024-10-17
fix: Remove duplication around plugins' names

mateusoliveira43:
## Why the changes were made

Noted that we use one variable to define what plugins' names DPA accepts, and another in validators. I was afraid this could haunt us in the future.

I ran commands like `grep -Inr '"aws"' .` and `grep -Inr "'aws'" .`  (ignoring `tests` folder) to make sure all occurrences were changed.

### DIFF
--- a/controllers/bsl.go
+++ b/controllers/bsl.go
@@ -47,19 +47,18 @@ func (r *DPAReconciler) ValidateBackupStorageLocations() (bool, error) {
 				return false, fmt.Errorf("no provider specified for one of the backupstoragelocations configured")
 			}
 
-			// TODO: cases might need some updates for IBM/Minio/noobaa
 			switch provider {
-			case AWSProvider, "velero.io/aws":
+			case string(oadpv1alpha1.DefaultPluginAWS), "velero.io/aws":
 				err := r.validateAWSBackupStorageLocation(*bslSpec.Velero)
 				if err != nil {
 					return false, err
 				}
-			case AzureProvider, "velero.io/azure":
+			case string(oadpv1alpha1.DefaultPluginMicrosoftAzure), "velero.io/azure":
 				err := r.validateAzureBackupStorageLocation(*bslSpec.Velero)
 				if err != nil {
 					return false, err
 				}
-			case GCPProvider, "velero.io/gcp":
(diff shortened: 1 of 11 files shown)


controllers/bsl.go
```diff
@@ -47,19 +47,18 @@ func (r *DPAReconciler) ValidateBackupStorageLocations() (bool, error) {
 				return false, fmt.Errorf("no provider specified for one of the backupstoragelocations configured")
 			}
 
-			// TODO: cases might need some updates for IBM/Minio/noobaa
 			switch provider {
-			case AWSProvider, "velero.io/aws":
```
mateusoliveira43, line 52: I you run `grep -Inr '"velero.io/aws"' .`, you will note only place this is used (aside from test files) is here. Should we remove this and only allow `aws`?

controllers/vsl.go
```diff
@@ -116,62 +116,63 @@ func (r *DPAReconciler) ValidateVolumeSnapshotLocations() (bool, error) {
...
 
 			// check for invalid config key
 			for key := range vslSpec.Velero.Config {
 				valid := validAWSKeys[key]
 				if !valid {
-					return false, fmt.Errorf("DPA %s.config key %s is not a valid %s config key", veleroVSLYAMLPath, key, AWSProvider)
+					return false, fmt.Errorf("DPA %s.config key %s is not a valid %s config key", veleroVSLYAMLPath, key, string(oadpv1alpha1.DefaultPluginAWS))
 				}
 			}
 			//checking the aws plugin, if not present, throw warning message
```
mateusoliveira43, line 139: comments are outdated, remove them

kaovilai:
```
2024/11/13 02:35:58 VM cirros-test/cirros-test status is: WaitingForVolumeBinding
```
common failure

/override ci/prow/4.17-e2e-test-kubevirt-aws

weshayutin (approved):

controllers/bsl.go
kaovilai: > only allow aws

so `velero.io/aws` would not work? cause that's what people who move from upstream velero would expect to use.

PR #1565, opened 2024-10-17
OADP-5079 Add support for legacy aws plugin

sseago:
## Why the changes were made

Certain s3 providers don't support the AWS SDK v2 updates made in velero-plugin-for-aws v1.9+ (OADP 1.4+). This PR provides the ability to use the v1.8 plugin in a newer OADP environment.

In the DPA `DefaultPlugins` section, use `legacy-aws` instead of `aws`. S3 providers that work with the regular aws plugin should still work here. S3 providers that error out in various ways with the regular plugin should also work.


docs/developer/PROW_CI.md
mateusoliveira43, line 1: Thank you for adding this. Let's not forget about adding them to PROW

pkg/common/common.go
```diff
@@ -75,6 +76,7 @@ const (
 // Plugin names
```
mateusoliveira43, line 76: This is confusing for me. Plugin names are aws, gcp, etc, for me. This is only used as the container image name of each plugin, right?

pkg/credentials/credentials.go
```diff
@@ -39,6 +40,16 @@ var (
 			MountPath:          "/credentials",
 			EnvCredentialsFile: common.AWSSharedCredentialsFileEnvKey,
 			PluginName:         common.VeleroPluginForAWS,
+			ProviderName:       "aws",
```
mateusoliveira43, line 43: lets not use literal strings, I am trying to change this in this PR https://github.com/openshift/oadp-operator/pull/1564

pkg/credentials/credentials.go
```diff
@@ -142,6 +165,9 @@ func GetPluginImage(pluginName string, dpa *oadpv1alpha1.DataProtectionApplicati
 	case common.VeleroPluginForAWS:
 		return getAWSPluginImage(dpa)
 
+	case common.VeleroPluginForLegacyAWS:
```
mateusoliveira43, line 168: I would refactor this function to receive `plugin oadpv1alpha1.DefaultPlugin` instead of plugin name, as I described my confusion in another comment

pkg/credentials/credentials.go
```diff
@@ -25,6 +25,7 @@ type DefaultPluginFields struct {
 	PluginImage        string
 	PluginSecretKey    string
 	PluginName         string
```
mateusoliveira43, line 27: would change this to `ContainerName` to avoid confusion

controllers/bsl.go
```diff
@@ -398,9 +399,9 @@ func (r *DPAReconciler) validateGCPBackupStorageLocation(bslSpec velerov1.Backup
 	return nil
 }
 
-func pluginExistsInVeleroCR(configuredPlugins []oadpv1alpha1.DefaultPlugin, expectedPlugin oadpv1alpha1.DefaultPlugin) bool {
+func pluginExistsInVeleroCR(configuredPlugins []oadpv1alpha1.DefaultPlugin, expectedProvider string) bool {
```
mateusoliveira43, line 402: This is a duplication of the function changed in `controllers/vsl.go`. I would delete function in that file and use this one

This function name is wrong, no? From what I see in code, DPA was called velero at some point in time. It should be `pluginExistsInDPA` or something like this, right?

controllers/validator.go
```diff
@@ -155,5 +165,10 @@ func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {
 			}
 		}
 	}
+
+	if foundAWSPlugin && foundLegacyAWSPlugin {
```
mateusoliveira43, line 169: I would change code here to something like
```go
if slices.Contains(dpa.Spec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPluginAWS) &&
slices.Contains(dpa.Spec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPluginLegacyAWS) {
```
I think it is easier to read

controllers/validator.go
```diff
@@ -155,5 +165,10 @@ func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {
...
+		return false, errors.New("Only one of aws and legacy-aws DefaultPlugins may be installed")
```
mateusoliveira43, line 170: As I explain here https://github.com/openshift/oadp-operator/issues/1536, I think we need to improve our error messages

I suggest something like

```go
fmt.Errorf("%s and %s can not be both specified in DPA spec.configuration.velero.defaultPlugins", oadpv1alpha1.DefaultPluginAWS, oadpv1alpha1.DefaultPluginLegacyAWS)
```

docs/developer/PROW_CI.md
sseago: @mateusoliveira43 yes -- @weshayutin is working on the prow side on this I think.

controllers/bsl.go
sseago: I was trying to avoid making major functionality changes since we want to backport to 1.4. The only change here is we're passing a regular string rather than a string that's typecast to `expectedPlugin` -- until this change, the plugin name and the provider were always the same. Now they're not, since both `legacy-aws` and `aws` have the provider `aws`.

controllers/bsl.go
```diff
@@ -412,7 +413,7 @@ func (r *DPAReconciler) validateProviderPluginAndSecret(bslSpec velerov1.BackupS
 		return nil
 	}
 	// check for existence of provider plugin and warn if the plugin is absent
-	if !pluginExistsInVeleroCR(r.dpa.Spec.Configuration.Velero.DefaultPlugins, oadpv1alpha1.DefaultPlugin(bslSpec.Provider)) {
+	if !pluginExistsInVeleroCR(r.dpa.Spec.Configuration.Velero.DefaultPlugins, bslSpec.Provider) {
 		r.Log.Info(fmt.Sprintf("%s backupstoragelocation is configured but velero plugin for %s is not present", bslSpec.Provider, bslSpec.Provider))
```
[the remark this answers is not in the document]
sseago: The goal of this PR is not to change how BSLs work but to react to adding a new plugin where plugin name and provider don't match anymore. If we want to change validation rules, that should be a separate PR.

controllers/validator.go
sseago: Readability is subjective here. I would find that change harder to read -- but we're already iterating over the slice once above. Adding this would mean 2 more iterations.

controllers/bsl.go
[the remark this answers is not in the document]
stillalearner: @mateusoliveira43 @sseago Sure, Thanks ! Will create a separate PR for this validation . Noted.

pkg/common/common.go
sseago: The container name is the only actual "name" string that's used in the velero side of the configuration -- as the container name. aws/gcp/etc. only exist as values for the velero default plugin list. So from a velero point of view, this is the plugin name.

config/manager/manager.yaml
```diff
@@ -50,6 +50,8 @@ spec:
             value: quay.io/konveyor/openshift-velero-plugin:latest
           - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_AWS
             value: quay.io/konveyor/velero-plugin-for-aws:latest
+          - name: RELATED_IMAGE_VELERO_PLUGIN_FOR_LEGACY_AWS
```
rayfordj, line 53: :+1:

README.md
```diff
@@ -25,9 +25,9 @@ Periodic AWS E2E Virtualization Tests in OpenShift 4.17
 [![VM tests](https://prow.ci.openshift.org/badge.svg?jobs=periodic-ci-openshift-oadp-operator-master-4.17-e2e-test-kubevirt-aws-periodic)](https://prow.ci.openshift.org/job-history/gs/origin-ci-test/logs/periodic-ci-openshift-oadp-operator-master-4.17-e2e-test-kubevirt-aws-periodic)
 
 OADP repositories images job
-| OADP | OpenShift Velero plugin | Velero | Velero plugin for AWS | Velero plugin for GCP | Velero plugin for Microsoft Azure | Non Admin |
+| OADP | OpenShift Velero plugin | Velero | Velero plugin for AWS | Velero plugin for Legacy AWS | Velero plugin for GCP | Velero plugin for Microsoft Azure | Non Admin |
```
mateusoliveira43, line 28: formatting broke here 
Screenshot from 2024-10-22 11-37-38

probably need one more column in line 29 (adding `---------- |` at the end)

sseago: @mateusoliveira43 Updated.

kaovilai:
TODO: Add legacy-aws to e2e

kaovilai (approved):

controllers/validator.go
```diff
@@ -104,10 +104,20 @@ func (r *DPAReconciler) ValidateVeleroPlugins(log logr.Logger) (bool, error) {
 		}
 	}
 
+	foundAWSPlugin := false
+	foundLegacyAWSPlugin := false
 	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
 		pluginSpecificMap, ok := credentials.PluginSpecificFields[plugin]
 		pluginNeedsCheck, foundInBSLorVSL := providerNeedsDefaultCreds[string(plugin)]
```
mateusoliveira43, line 111: Do we also need to change here because of line 94?

mateusoliveira43 (commented):
I ran E2E tests in my IBM cluster, by changing `$PROVIDER` here
https://github.com/openshift/oadp-operator/blob/20d8c61d1eef343a6463d54d3df0806368e7b263/tests/e2e/scripts/ibmcloud_settings.sh#L19
to `legacy-aws` and they all passed ✅ (did not run virt tests)

I also ran some commands like `grep -Inr "DefaultPlugins" . --exclude-dir=tests --exclude=\*_test.go`, and I think all necessary code changes were made, except for this question https://github.com/openshift/oadp-operator/pull/1565#discussion_r1812653225

controllers/validator.go
sseago: @mateusoliveira43 You are correct again. I'll fix this.

shubham-pampattiwar (approved):

weshayutin:
> I ran E2E tests in my IBM cluster, by changing `$PROVIDER` here
> 
> https://github.com/openshift/oadp-operator/blob/20d8c61d1eef343a6463d54d3df0806368e7b263/tests/e2e/scripts/ibmcloud_settings.sh#L19
> 
> 
> to `legacy-aws` and they all passed ✅ (did not run virt tests)
> I also ran some commands like `grep -Inr "DefaultPlugins" . --exclude-dir=tests --exclude=\*_test.go`, and I think all necessary code changes were made, except for this question [#1565 (comment)](https://github.com/openshift/oadp-operator/pull/1565#discussion_r1812653225)

FYI @mateusoliveira43  If this worked for you and have completed the code review I would expect to see you vote as well sir.  Thank you

weshayutin:
Testing will be covered here: https://issues.redhat.com/browse/OADP-5106

mateusoliveira43 (approved):

PR #1543, opened 2024-10-07
convert the troubleshotting.md to proper md

weshayutin:
* mostly just converting to markdown
* moved some links around, imho better organization.
* not perfect but better

## Why the changes were made

* the old page was too ugly

PR #1574, opened 2024-10-24
OADP-5024 Virtualization troubleshooting document.

mrnold:
## Why the changes were made

It would be good to have a convenient place to keep troubleshooting tips for VM backup and restore. This first pass adds some ways to do backup and restore for individual VMs in multi-VM namespaces/backups.

Open [markdown preview](https://github.com/mrnold/oadp-operator/blob/virt_troubleshooting/docs/virtualization_troubleshooting.md).

### DIFF
--- a/docs/virtualization_troubleshooting.md
+++ b/docs/virtualization_troubleshooting.md
@@ -1,0 +1,55 @@
+# Troubleshooting virtualization backup and restore
+
+1. [Create a backup for a single VM in a namespace with multiple VMs.]( #create-a-backup-for-a-single-vm-in-a-namespace-with-multiple-vms)
+1. [Create a restore for a single VM in a backup with multiple VMs.](#create-a-restore-for-a-single-vm-in-a-backup-with-multiple-vms)
+
+## Create a backup for a single VM in a namespace with multiple VMs.
+
+If you have a namespace with many VMs, and only want to back up one of them, you can use a label selector with the `app: vmname` label to indicate which VM should be included in the backup. For example:
+```
+apiVersion: velero.io/v1
+kind: Backup
+metadata:
+  name: vmbackupsingle
+  namespace: openshift-adp
+spec:
+  snapshotMoveData: true
+  includedNamespaces:
+  - bkp1
+  labelSelector:
+    matchLabels:
+      app: vm1
+  storageLocation: aws-dpa-1
+```
+Creating a restore pointing to `vmbkpsingle` will then create only one VM, `vm1`:
+```
+apiVersion: velero.io/v1
+kind: Restore
+metadata:
+  name: vmrestoresingle
+  namespace: openshift-adp
+spec:
+  backupName: vmbackupsingle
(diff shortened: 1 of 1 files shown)


weshayutin:
Thanks for writing this up.. Looks good to land IMHO.  I do want to add this NEW doc to the troubleshooting index here: https://github.com/openshift/oadp-operator/blob/master/docs/TROUBLESHOOTING.md

Need to land https://github.com/openshift/oadp-operator/pull/1543 first to avoid a conflict.

docs/virtualization_troubleshooting.md
```diff
@@ -0,0 +1,55 @@
+# Troubleshooting virtualization backup and restore
+
+1. [Create a backup for a single VM in a namespace with multiple VMs.]( #create-a-backup-for-a-single-vm-in-a-namespace-with-multiple-vms)
+1. [Create a restore for a single VM in a backup with multiple VMs.](#create-a-restore-for-a-single-vm-in-a-backup-with-multiple-vms)
+
+## Create a backup for a single VM in a namespace with multiple VMs.
+
+If you have a namespace with many VMs, and only want to back up one of them, you can use a label selector with the `app: vmname` label to indicate which VM should be included in the backup. For example:
+```
```
mateusoliveira43, line 9: lets add syntax highlighting for yaml in the 3 code blocks?

mateusoliveira43 (dismissed):

docs/virtualization_troubleshooting.md
mrnold: Sure, done.

mateusoliveira43 (approved):

kaovilai (commented):
This feels more like an example than troubleshooting?

There is no errors to workaround that I can see.

docs/virtualization_troubleshooting.md
```diff
@@ -0,0 +1,55 @@
+# Troubleshooting virtualization backup and restore
+
```
kaovilai, line 2: Perhaps a subheading that signify this is a typical usecase that should pass without errors.

mrnold:
> This feels more like an example than troubleshooting?
> 
> There is no errors to workaround that I can see.

It's really just this one, corresponding to a customer case: `However, this label may be occupied by the containerized-data-importer for DataVolume-created PVCs, so a Data Mover restore will result in a VM stuck in provisioning due to the mismatched `app` label.`

I think this document is expected to get filled out more over time.

kaovilai:
So to make more of a troubleshooting document, I think we need to have table of content of errors, that then explain why, and solution.

Right now the error while highlighted as the potential issue, is not the main topic or a heading in the doc. WDYT?

weshayutin:
@kaovilai @mrnold is transitioning to a new role.  We will have to pick up longer term direction in a virt troubleshooting guide.  I like the idea but this PR isn't the best place to discuss it.

kaovilai:
aight

/lgtm

PR #1577, opened 2024-10-25
fix: NAC install validation

mateusoliveira43:
## Why the changes were made

Only allow one NAC install per cluster.

Related to https://github.com/migtools/oadp-non-admin/pull/107

Also, add validation for only allowing one DPA per OADP installation namespace.

Deploy 2 (or more) OADP operators (`make deploy-olm`) in different namespaces. Create 2 (or more) DPAs in those namespaces with NAC enabled. Only one NAC should be deployed and DPAs should have error in their statuses.

### DIFF
--- a/controllers/nonadmin_controller.go
+++ b/controllers/nonadmin_controller.go
@@ -10,7 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -83,21 +82,6 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 			fmt.Sprintf("Non admin controller deployment %s/%s deleted", nonAdminDeployment.Namespace, nonAdminDeployment.Name),
 		)
 		return true, nil
-	}
-
-	selector, err := fields.ParseSelector(fmt.Sprintf("metadata.namespace!=%s", r.NamespacedName.Namespace))
-	if err != nil {
-		return false, err
-	}
-	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
-	err = r.ClusterWideClient.List(r.Context, dpaList, &client.ListOptions{FieldSelector: selector})
-	if err != nil {
-		return false, err
-	}
-	for _, dpa := range dpaList.Items {
-		if (&DPAReconciler{dpa: &dpa}).checkNonAdminEnabled() {
-			return false, fmt.Errorf("only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is also configured to be installed in %s namespace", dpa.Namespace)
-		}
 	}
 
 	operation, err := controllerutil.CreateOrUpdate(
--- a/controllers/nonadmin_controller.go
+++ b/controllers/nonadmin_controller.go
@@ -10,6 +10,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
@@ -39,8 +40,6 @@ var (
 )
 
 func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, error) {
-	dpa := r.dpa
-
 	nonAdminDeployment := &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      nonAdminObjectName,
@@ -49,7 +48,7 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 	}
 
 	// Delete (possible) previously deployment
-	if !(r.checkNonAdminEnabled() && dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
+	if !r.checkNonAdminEnabled() {
 		if err := r.Get(
 			r.Context,
 			types.NamespacedName{
@@ -86,6 +85,21 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 		return true, nil
 	}
 
+	selector, err := fields.ParseSelector(fmt.Sprintf("metadata.namespace!=%s", r.NamespacedName.Namespace))
+	if err != nil {
+		return false, err
+	}
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err = r.List(r.Context, dpaList, &client.ListOptions{FieldSelector: selector})
+	if err != nil {
+		return false, err
+	}
+	for _, dpa := range dpaList.Items {
+		if (&DPAReconciler{dpa: &dpa}).checkNonAdminEnabled() {
+			return false, fmt.Errorf("only one NAC can be installed in the whole cluster")
+		}
+	}
+
 	operation, err := controllerutil.CreateOrUpdate(
 		r.Context,
 		r.Client,
@@ -97,7 +111,7 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 			}
 
 			// Setting controller owner reference on the non admin controller deployment
-			return controllerutil.SetControllerReference(dpa, nonAdminDeployment, r.Scheme)
+			return controllerutil.SetControllerReference(r.dpa, nonAdminDeployment, r.Scheme)
 		},
 	)
 	if err != nil {
@@ -190,10 +204,8 @@ func ensureRequiredSpecs(deploymentObject *appsv1.Deployment, image string, imag
 }
 
 func (r *DPAReconciler) checkNonAdminEnabled() bool {
-	dpa := r.dpa
-	if dpa.Spec.NonAdmin != nil &&
-		dpa.Spec.NonAdmin.Enable != nil {
-		return *dpa.Spec.NonAdmin.Enable
+	if r.dpa.Spec.NonAdmin != nil && r.dpa.Spec.NonAdmin.Enable != nil {
+		return *r.dpa.Spec.NonAdmin.Enable && r.dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal
 	}
 	return false
 }
--- a/controllers/nonadmin_controller.go
+++ b/controllers/nonadmin_controller.go
@@ -10,7 +10,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	k8serror "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/utils/ptr"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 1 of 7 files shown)


controllers/nonadmin_controller.go
```diff
@@ -86,6 +85,21 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 		return true, nil
 	}
 
+	selector, err := fields.ParseSelector(fmt.Sprintf("metadata.namespace!=%s", r.NamespacedName.Namespace))
+	if err != nil {
+		return false, err
+	}
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err = r.List(r.Context, dpaList, &client.ListOptions{FieldSelector: selector})
```
mateusoliveira43, line 93: A corner case here should be if 2 (or more) DPAs are created at the same time. But since DPA reconciles every minute, that's not to bad? (DPAs would have error message in their statuses, but multiple NAC would also exist in the cluster. To really avoid this, before erroring out, could delete NAC deployment)

Another question here: since DPA reconciles every minute, its too much to do a cluster wide get call every minute?

controllers/nonadmin_controller.go
mateusoliveira43, line 1: TODO write tests for multiple DPAs in same namespace and multiple NACs in cluster

controllers/nonadmin_controller.go
kaovilai: Or a more drastic change would be to move this controller out of the current ReconcileBatch into its own thing.

controllers/validator.go
```diff
@@ -15,26 +16,32 @@ import (
 // ValidateDataProtectionCR function validates the DPA CR, returns true if valid, false otherwise
 // it calls other validation functions to validate the DPA CR
 func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error) {
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err := r.List(r.Context, dpaList, &client.ListOptions{Namespace: r.NamespacedName.Namespace})
+	if err != nil {
+		return false, err
+	}
+	if len(dpaList.Items) > 1 {
+		return false, errors.New("only one DPA CR can exist per OADP namespace")
```
shubham-pampattiwar, line 25: `only one DPA CR can exist per OADP installation namespace`

controllers/validator.go
```diff
@@ -15,26 +16,32 @@ import (
 // ValidateDataProtectionCR function validates the DPA CR, returns true if valid, false otherwise
 // it calls other validation functions to validate the DPA CR
 func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error) {
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err := r.List(r.Context, dpaList, &client.ListOptions{Namespace: r.NamespacedName.Namespace})
```
shubham-pampattiwar, line 20: Lets cover this via a unit test.

controllers/validator.go
shubham-pampattiwar, line 25: Update PR description to highlight this validation.

controllers/nonadmin_controller.go
```diff
@@ -86,6 +85,21 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
...
+	if err != nil {
+		return false, err
+	}
+	for _, dpa := range dpaList.Items {
+		if (&DPAReconciler{dpa: &dpa}).checkNonAdminEnabled() {
+			return false, fmt.Errorf("only one NAC can be installed in the whole cluster")
```
shubham-pampattiwar, line 99: `Only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is already deployed in <NAC-Enabled-DPA.NS>`

controllers/nonadmin_controller.go
mateusoliveira43: thinking again, I think we do not need to worry about having multiple NAC deployments after validation 

- If 2 DPAs with NAC enabled are created at the same time, when they reach this part, both of them will not deploy NAC

- If one DPA is created and after some time another is created, only first DPA NAC will be deployed

Right?

controllers/validator.go
mateusoliveira43: done

controllers/nonadmin_controller.go
```diff
@@ -86,6 +85,21 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 		return true, nil
 	}
 
+	selector, err := fields.ParseSelector(fmt.Sprintf("metadata.namespace!=%s", r.NamespacedName.Namespace))
```
mateusoliveira43, line 88: got `non-exact field matches are not supported by the cache` error with this approach

will change

controllers/nonadmin_controller.go
```diff
@@ -86,6 +84,19 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
 		return true, nil
 	}
 
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err := r.List(r.Context, dpaList)
+	if err != nil {
+		return false, err
+	}
+	r.Log.Info("number of DPAs fetched: ", "number of DPAs", len(dpaList.Items))
```
mateusoliveira43, line 92: TODO remove prior to merge

weshayutin:
@shubham-pampattiwar it's not clear to me we've come to a conclusion on 1 NAC per cluster vs. a NAC married to a particular OADP deployment.  If this is a temporary measure until we can put in the code to pair an oadp deployment with a NAC than I think this is fine.

shubham-pampattiwar:
This is not a temporary measure. The current PR implements the behavior that:
- Only one NAC exists per cluster
- you can install multiple OADP Operators in the cluster but only of them can deploy/enable NAC feature

weshayutin:
Upon installing OADP in another namespace w/ NAC enabled.. A user gets the following error message.
```
 lastTransitionTime: '2024-11-13T01:54:18Z'
      message: only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is also configured to be installed in openshift-adp namespace
      reason: Error
```

controllers/nonadmin_controller.go
```diff
@@ -86,6 +85,21 @@ func (r *DPAReconciler) ReconcileNonAdminController(log logr.Logger) (bool, erro
...
+		return false, err
+	}
+	dpaList := &oadpv1alpha1.DataProtectionApplicationList{}
+	err = r.ClusterWideClient.List(r.Context, dpaList, &client.ListOptions{FieldSelector: selector})
+	if err != nil {
+		return false, err
+	}
+	for _, dpa := range dpaList.Items {
+		if (&DPAReconciler{dpa: &dpa}).checkNonAdminEnabled() {
+			return false, fmt.Errorf("only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is also configured to be installed in %s namespace", dpa.Namespace)
```
shubham-pampattiwar, line 99: Lets re-frame the error text a little bit: `only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is already configured and installed in %s namespace", dpa.Namespace`

controllers/nonadmin_controller.go
```diff
@@ -190,10 +204,8 @@ func ensureRequiredSpecs(deploymentObject *appsv1.Deployment, image string, imag
 }
 
 func (r *DPAReconciler) checkNonAdminEnabled() bool {
-	dpa := r.dpa
-	if dpa.Spec.NonAdmin != nil &&
-		dpa.Spec.NonAdmin.Enable != nil {
-		return *dpa.Spec.NonAdmin.Enable
+	if r.dpa.Spec.NonAdmin != nil && r.dpa.Spec.NonAdmin.Enable != nil {
+		return *r.dpa.Spec.NonAdmin.Enable && r.dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal
```
mpryc, line 208: Do we need to ensure the comparison is against "True", "TRUE" or "true", so regardless of capital/small letters?

```go
import (
	"strings"
[...]

		return *r.dpa.Spec.NonAdmin.Enable && strings.EqualFold(r.dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck], TrueVal)
```

controllers/nonadmin_controller.go
mateusoliveira43: I would like to keep as it is

from what I tested, NAC will never be deployed 2 or more times. Either one is installed (one was installed and now a second is trying to be installed) or none (2 or more are trying to be installed at the same time)

So, your suggestion would make sense to the case of the DPA that tried to install NAC and it already exists in cluster, but not for the DPA that already installed it, or other cases (error messages will appear in all DPAs, because DPA reconciles every minute). Do you agree?

shubham-pampattiwar: Users does the following:
- Creates DAP 1
- Enables NAC for DPA 1
- Now Another NS, OADP is installed and DPA 2 is created there 
- User enabled NAC for DPA 2
I want error message only on DPA 2.

mateusoliveira43: To have error only on "wrong" DPAs, I would need to know if NAC is already deployed (do a GET call for deployments in OADP namespace) prior to erroring out. Would this not complicate validation too much? (and every minute, DPA would do GET call)

shubham-pampattiwar: DPA 1 did nothing wrong, why should DPA 1 be errored out ?

mateusoliveira43: Just because we can not be sure that error is not in DPA 1

examples

- DPA 1 is created first with NAC enabled -> DPA 2 created after with NAC enabled -> by doing a deployment get call we can add error only to DPA 2 ✅

- DPA 1 and DPA 2 are created (or edited) at the same time with NAC enabled -> in this case, which DPA we error out? ❌

shubham-pampattiwar: For case 2, we don't decide that, let the DAP controller do its work. Either of the DPAs will get the error for sure, right ?

mateusoliveira43: @shubham-pampattiwar that is my fear. I think if your suggest approach, both will deploy NAC

@mpryc NAC has leader election, I will test what happens

controllers/validator.go
```diff
@@ -65,26 +75,50 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
+		}
+		for _, dpa := range dpaList.Items {
+			if dpa.Namespace != r.NamespacedName.Namespace && (&DPAReconciler{dpa: &dpa}).checkNonAdminEnabled() {
+				nonAdminDeployment := &appsv1.Deployment{
+					ObjectMeta: metav1.ObjectMeta{
+						Name:      nonAdminObjectName,
+						Namespace: dpa.Namespace,
+					},
+				}
+				if err := r.ClusterWideClient.Get(
```
shubham-pampattiwar, line 104: Shouldn't we be using the older ns-scoped client here ?

shubham-pampattiwar:
Changes worked for: 
```
  snapshotLocations:
  - velero:
      config:
        profile: default
        region: us-east-1
      provider: aws
  unsupportedOverrides:
    tech-preview-ack: "true"
status:
  conditions:
  - lastTransitionTime: "2024-11-19T06:41:53Z"
    message: only a single instance of Non-Admin Controller can be installed across
      the entire cluster. Non-Admin controller is already configured and installed
      in openshift-adp namespace
    reason: Error
    status: "False"
    type: Reconciled

```

Have one question tho: https://github.com/openshift/oadp-operator/pull/1577#discussion_r1847725411

controllers/validator.go
mateusoliveira43: no

the check here is to check if other namespace, different than the namespace this OADP is installed, has NAC installed already. We check that by getting DPAs in other namespaces, and on those namespaces, if nonAdmin is enabled, check if nonAdmin Deployment exist. So both calls need cluster wide privileges

controllers/nonadmin_controller.go
mpryc: Ok thanks, comment can be closed.

controllers/validator.go
```diff
@@ -65,26 +72,23 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
...
-	if _, err := getNodeAgentResourceReqs(dpa); err != nil {
+	if _, err := getNodeAgentResourceReqs(r.dpa); err != nil {
 		return false, err
 	}
 
 	// validate non-admin enable and tech-preview-ack
-	if r.checkNonAdminEnabled() {
-		if !(dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
+	if r.dpa.Spec.NonAdmin != nil && r.dpa.Spec.NonAdmin.Enable != nil && *r.dpa.Spec.NonAdmin.Enable {
+		if !(r.dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
```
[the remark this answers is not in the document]
mpryc: Thanks same as above, comment can be closed.

mpryc (approved):

shubham-pampattiwar (approved):

Issue #1540, opened 2024-10-02
Create `operator-sdk` update guidelines


mateusoliveira43:
## Problem

Currently, we do not have any strategy to update `operator-sdk` version used by the project. Last time we updated it, for example, was to fix problem fixed in newer releases. Reference https://github.com/openshift/oadp-operator/pull/1398

But we only updated the executable version, without any updates to project structure, which can cause `operator-sdk` commands and functionalities not work properly.

Currently, some of  `operator-sdk` functionalities are broken in the project.

### Example

These lines are wrong, pointing to an object that does not exist (`buckets.oadp.openshift.io`)
https://github.com/openshift/oadp-operator/blob/5b0b9fcbc0d86c09158f9a6bcf9ec47fd61ae50e/controllers/bucket_controller.go#L44-L47

If we update it and run `make bundle`, `config/rbac/role.yaml` is not updated.

https://github.com/openshift/oadp-operator/blob/5b0b9fcbc0d86c09158f9a6bcf9ec47fd61ae50e/config/rbac/role.yaml#L51
> `operator-sdk` uses `kubebuilder` under the hood

## Solution

As proposed previously in https://github.com/openshift/oadp-operator/pull/1250 and already being applied in NAC project (references https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#upgrade-kubebuilder-version https://github.com/migtools/oadp-non-admin/blob/master/docs/architecture.md#kubebuilde), we should update all necessary places when updating `operator-sdk` version. 

To avoid human mistakes of forgetting to apply a change describe in `operator-sdk` documentation, we should do this update following these instructions:

Download both current and new version of `operator-sdk` from https://github.com/operator-framework/operator-sdk/releases and run
```sh
mkdir current
mkdir new
cd current
# Run operator-sdk commands pointing to operator-sdk executable with the current version
cd ..
cd new
# Run operator-sdk commands pointing to operator-sdk executable with the new version
cd ..
diff -ruN current new > operator-sdk-upgrade.diff
patch -p1 --verbose -d ./ -i operator-sdk-upgrade.diff
# Resolve possible conflicts
```
> Because last step can fail, needing human intervention, do not know if it is necessary to create a script

### Pros

- Ensures each update does not break any `operator-sdk` commands or functionalities
- Get updates from newer versions (Example: if we already followed these steps, we could change between `docker` and `podman` in Makefile)
- Avoid things getting (very) outdated 
[...]

PR #1582, opened 2024-11-05
Fix for #1581 Allow to use podman as container tool for building images

### LINKED ISSUE #1581: Allow to use podman as container tool for building images
**Is your feature request related to a problem? Please describe.**
It's currently not possible to build and deploy operator with the system where only `podman` command is available.

**Describe the solution you'd like**
Rework `Makefile` to allow auto-discovery of container build CLI available on the system from both `docker` or `podman` with an override opption for the user to select tool.

**Describe alternatives you've considered**
Installing podman-docker packages or symlinking podman to docker CLI.

**Additional context**
Operating systems such as Fedora does not have `docker` CLI as the main build, it's easier to select `podman` or make it transparent for the user, so Makefile selects what's available keeping in mind the order that `docker` should be first, `podman` second unless decided by the user to use something specified in the env vars.

mpryc:
The Makefile auto-discovers the build tool, defaulting to docker or podman if available. Users can override this by setting the CONTAINER_TOOL environment variable to specify their preferred tool. This approach is aligned with other operators, such as non-admin, while also adding automatic tool discovery for improved usability and transparency.

## Why the changes were made

Fix #1581 

Reworked Makefile tested on the system with podman and docker.
Selecting CONTAINER_TOOL which is not available on the system properly gave error before entering Makefile parts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,6 +24,24 @@ AZURE_RESOURCE_FILE ?= /var/run/secrets/ci.openshift.io/multi-stage/metadata.jso
 AZURE_CI_JSON_CRED_FILE ?= ${CLUSTER_PROFILE_DIR}/osServicePrincipal.json
 AZURE_OADP_JSON_CRED_FILE ?= ${OADP_CRED_DIR}/azure-credentials
 
+# CONTAINER_TOOL defines the container tool to be used for building images.
+# By default, this Makefile uses docker, as the target commands have been tested primarily with it.
+# However, if docker is not available, the Makefile will attempt to use podman if it's installed.
+# You may also set CONTAINER_TOOL directly as an environment variable to specify a different tool.
+# If neither docker nor podman is found, or if the specified tool is unavailable, the Makefile will exit with an error.
+
+# Set CONTAINER_TOOL to Docker or Podman if not already defined by the user
+CONTAINER_TOOL ?= $(shell \
+  if command -v docker >/dev/null 2>&1; then echo docker; \
+  elif command -v podman >/dev/null 2>&1; then echo podman; \
+  else echo ""; \
+  fi \
+)
+ifeq ($(shell command -v $(CONTAINER_TOOL) >/dev/null 2>&1 && echo found),)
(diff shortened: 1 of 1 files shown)


mateusoliveira43 (approved):
checked if all places where modified with `grep -Inr "docker" ./Makefile`

Tested with `export CONTAINER_TOOL=batman`

mateusoliveira43:
side note: if we have agreed on https://github.com/openshift/oadp-operator/issues/1540, this change would have already been in the code, as https://github.com/kubernetes-sigs/kubebuilder newer versions already have ability to select `CONTAINER_TOOL` in Makefile

(but would not have the automation to select between docker or podman)

weshayutin (approved):
ha.. I love seeing the code from the past in new places :)

Makefile
```diff
@@ -24,6 +24,24 @@ AZURE_RESOURCE_FILE ?= /var/run/secrets/ci.openshift.io/multi-stage/metadata.jso
...
+# If neither docker nor podman is found, or if the specified tool is unavailable, the Makefile will exit with an error.
+
+# Set CONTAINER_TOOL to Docker or Podman if not already defined by the user
+CONTAINER_TOOL ?= $(shell \
+  if command -v docker >/dev/null 2>&1; then echo docker; \
+  elif command -v podman >/dev/null 2>&1; then echo podman; \
+  else echo ""; \
+  fi \
+)
+ifeq ($(shell command -v $(CONTAINER_TOOL) >/dev/null 2>&1 && echo found),)
```
kaovilai, line 40:
```suggestion
ifeq ($(shell command -v $(CONTAINER_TOOL) >/dev/null 2>&1 && echo $(CONTAINER_TOOL) found),)
```

weshayutin:
https://prow.ci.openshift.org/job-history/gs/test-platform-results/pr-logs/directory/pull-ci-openshift-oadp-operator-master-4.17-e2e-test-kubevirt-aws is broken and needs to be overridden. 

I am not an owner

PR #1584, opened 2024-11-07
fix: Enforce NonAdminBackup spec field values

mateusoliveira43:
## Why the changes were made

Related to https://github.com/migtools/oadp-non-admin/issues/37

This PR adds `EnforceBackupSpec` to DPA `spec.nonAdmin`. With it, admin user can enforce that all NonAdminBackups created have the enforced spec fields with the values admin user configured.

Follow [install-from-source](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source) NAC testing documentation, pointing my NAC fork branch [`fix/enforce-backup-spec-field-values`](https://github.com/mateusoliveira43/oadp-non-admin/tree/fix/enforce-backup-spec-field-values) and this branch.

Test that:
- when changing DPA `spec.nonAdmin.enforceBackupSpec`, NAC Pod recreates (for new enforcement to take effect)
- when not changing DPA `spec.nonAdmin.enforceBackupSpec`, NAC Pod does not get recreated
- when setting  DPA `spec.nonAdmin.enforceBackupSpec`, NonAdminBackups fail validation if not respecting enforcement


controllers/nonadmin_controller.go
```diff
@@ -166,7 +174,7 @@ func ensureRequiredSpecs(deploymentObject *appsv1.Deployment, image string, imag
 			Name:            nonAdminObjectName,
 			Image:           image,
 			ImagePullPolicy: imagePullPolicy,
-			Env:             []corev1.EnvVar{namespaceEnvVar},
+			Env:             []corev1.EnvVar{namespaceEnvVar, enforceBackupSpecEnvVar},
```
mateusoliveira43, line 177: So, goal here is admin user will enforce some NonAdminBackup spec field values through DPA (for example, `SnapshotVolumes` must be `false` to every NonAdminBackup created)

During NAC startup, it reads these values admin user has set. 

we could read from DPA directly, doing a DPA list call in OADP namespace before here https://github.com/migtools/oadp-non-admin/blob/8d8de636b68c0241b937403a98ab4384616458ba/cmd/main.go#L135 and passing result to `NonAdminBackupReconciler`. Problem with this approach are:
1. need to change NAC RBAC (but easy to fix, just add to NAC role list DPA)
2. how to tell NAC deployment to restart when that field changes? 

Current implementation idea is passing DPA field value as env var to deployment. On NAC controller, it would Unmarshal the value and use it. Problem with this approach is that we can reach a char limit on env var value.

Any ideas on a third solution? Or how to fix problem `2` of first idea?

kaovilai: > how to tell NAC deployment to restart when that field changes?

if NAC is in a deployment, simply os.Exit() should result in another pod coming up.

controllers/validator.go
```diff
@@ -75,7 +76,12 @@ func (r *DPAReconciler) ValidateDataProtectionCR(log logr.Logger) (bool, error)
 	// validate non-admin enable and tech-preview-ack
 	if r.checkNonAdminEnabled() {
 		if !(dpa.Spec.UnsupportedOverrides[oadpv1alpha1.TechPreviewAck] == TrueVal) {
-			return false, errors.New("in order to enable/disable the non-admin feature please set dpa.spec.unsupportedOverrides[tech-preview-ack]: 'true'")
+			return false, errors.New("in order to enable the non-admin feature please set dpa.spec.unsupportedOverrides[tech-preview-ack]: 'true'")
+		}
+		if r.dpa.Spec.NonAdmin.EnforceBackupSpec != nil {
```
mateusoliveira43, line 81: added this to avoid headaches with users creating bad configuration

I think this is NAC responsibility, not user. If user can set this, then all NonAdminBackups will fail (or only of a specific namespace) 

do you agree with this safety check?

(if yes, other fields should be added here?)

controllers/nonadmin_controller.go
mateusoliveira43: changed logic here

now I save DPA `spec.nonAdmin.enforceBackupSpec` that created Deployment and compare with current DPA `spec.nonAdmin.enforceBackupSpec`, only update Deployment when those are different

mateusoliveira43:
TODO `make update-non-admin-manifests`

weshayutin:
ok.. checking this out.  
* Created a deployment from your branch
* Created a dpa.
* Added oadp-non-admin spec and overrides

* added enforceBackupSpec
```
    enforceBackupSpec:
      storageLocation: dpa-sample-1
```
* The controller properly restarted \0/

```
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Stopping and waiting for non leader election runnables
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Stopping and waiting for leader election runnables
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Shutdown signal received, waiting for all workers to finish	{"controller": "nonadminbackup", "controllerGroup": "nac.oadp.openshift.io", "controllerKind": "NonAdminBackup"}
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	All workers finished	{"controller": "nonadminbackup", "controllerGroup": "nac.oadp.openshift.io", "controllerKind": "NonAdminBackup"}
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Stopping and waiting for caches
non-admin-controller-5bcd79f674-dtqls non-admin-controller W1113 01:18:50.319171       1 reflector.go:462] pkg/mod/k8s.io/client-go@v0.29.0/tools/cache/reflector.go:229: watch of *v1.Backup ended with: an error on the server ("unable to decode an event from the watch stream: context canceled") has prevented the request from succeeding
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Stopping and waiting for webhooks
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Stopping and waiting for HTTP servers
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	controller-runtime.metrics	Shutting down metrics server with timeout of 1 minute
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	shutting down server	{"kind": "health probe", "addr": "[::]:8081"}
non-admin-controller-5bcd79f674-dtqls non-admin-controller 2024-11-13T01:18:50Z	INFO	Wait completed, proceeding to shutdown the manager
non-admin-controller-6d5b9769dd-jpblw non-admin-controller 2024-11-13T01:18:49Z	INFO	Starting workers	{"controller": "nonadminbackup", "controllerGroup": "nac.oadp.openshift.io", "controllerKind": "NonAdminBackup", "worker count": 1}
- non-admin-controller-5bcd79f674-dtqls › non-admin-controller
```

* Adding invalid backupspec key/values fails appropriate - see screenshot

Screenshot from 2024-11-12 18-22-12

It's not clear to me how to test: 
`when not changing DPA spec.nonAdmin.enforceBackupSpec, NAC Pod does not get recreated`

If you can help me understand that, I'll poke at this a bit more.

kaovilai:
> If you can help me understand that, I'll poke at this a bit more.

See design:
https://github.com/migtools/oadp-non-admin/pull/110/files#diff-788dfa1caf04072e867eaa6c490201f32d2e8bdf4cda035ba397d4d974737850R38

kaovilai:
> [If admin user changes any enforced field value, NAC Pod is recreated (and only NAC Pod) to always be up to date with admin user enforcements.](https://github.com/migtools/oadp-non-admin/pull/110/files#diff-788dfa1caf04072e867eaa6c490201f32d2e8bdf4cda035ba397d4d974737850R38:~:text=If%20admin%20user%20changes%20any%20enforced%20field%20value%2C%20NAC%20Pod%20is%20recreated%20(and%20only%20NAC%20Pod)%20to%20always%20be%20up%20to%20date%20with%20admin%20user%20enforcements.)

Therefore following should happen

> when not changing DPA spec.nonAdmin.enforceBackupSpec, NAC Pod does not get recreated

mateusoliveira43:
@weshayutin change for example velero log level. That should restart Velero Pod, but not NAC Pod

controllers/validator.go
shubham-pampattiwar: Hmm..I think maybe we should keep this optional and give a free reign to the admins.

controllers/nonadmin_controller_test.go
```diff
@@ -296,6 +313,68 @@ func TestEnsureRequiredSpecs(t *testing.T) {
 	if deployment.Spec.Template.Spec.Containers[0].Image != defaultNonAdminImage {
 		t.Errorf("Deployment has wrong Image: %v", deployment.Spec.Template.Spec.Containers[0].Image)
 	}
+	if len(deployment.Spec.Template.Annotations[enforcedBackupSpecKey]) == 0 {
+		t.Errorf("Deployment does not have Annotation")
+	}
+	previousEnforcedBackupSpec := deployment.DeepCopy().Spec.Template.Annotations[enforcedBackupSpecKey]
```
shubham-pampattiwar, line 319: lets use a different variable name here for `previousEnforcedBackupSpec`, something more clearer - maybe `previousEnforcedBackupSpecAnnotation` or just `previousAnnotationValue`
This also shadows declaration of 'previousEnforcedBackupSpec' in nonadmin_controller. go

controllers/nonadmin_controller.go
```diff
@@ -143,23 +149,42 @@ func ensureRequiredLabels(deploymentObject *appsv1.Deployment) {
...
+
 	templateObjectLabels := deploymentObject.Spec.Template.GetLabels()
 	if templateObjectLabels == nil {
 		deploymentObject.Spec.Template.SetLabels(controlPlaneLabel)
 	} else {
 		templateObjectLabels[controlPlaneKey] = controlPlaneLabel[controlPlaneKey]
 		deploymentObject.Spec.Template.SetLabels(templateObjectLabels)
 	}
+
+	templateObjectAnnotations := deploymentObject.Spec.Template.GetAnnotations()
```
shubham-pampattiwar, line 179: ++

controllers/validator.go
mateusoliveira43: ok

controllers/nonadmin_controller_test.go
mateusoliveira43: done

shubham-pampattiwar (approved):
/lgtm

PR #1593, opened 2024-11-20
Fix deploy-olm failing on docker build without `--load`

kaovilai:
## Why the changes were made
This is because docker build has become aliased to docker buildx build whose default no longer load build result to docker images which mean subsequent call to `docker-push` target fails, such as when `make deploy-olm`

We are getting rid of error highlighted in red like below.
```diff
WARNING: No output specified with docker-container driver. Build result will only remain in the build cache. To push result image into registry use --push or to load image into docker use --load
/Applications/Xcode.app/Contents/Developer/usr/bin/make docker-push IMG=ttl.sh/oadp-operator-bundle-a5cd6de2:1h
docker push ttl.sh/oadp-operator-bundle-a5cd6de2:1h
-Error response from daemon: failed to find image ttl.sh/oadp-operator-bundle-a5cd6de2:1h: ttl.sh/oadp-operator-bundle-a5cd6de2:1h: image not known
-make[2]: *** [docker-push] Error 1
-make[1]: *** [bundle-push] Error 2
```

https://docs.docker.com/reference/cli/docker/buildx/build/#load

checkout and see that podman build can also accept this param and not fail. (I did, it worked.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -240,7 +240,7 @@ ifneq ($(CLUSTER_TYPE),)
 	DOCKER_BUILD_ARGS = --platform=$(CLUSTER_OS)/$(CLUSTER_ARCH)
 endif
 docker-build: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t $(IMG) . $(DOCKER_BUILD_ARGS)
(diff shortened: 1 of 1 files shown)


weshayutin (approved):
thanks @kaovilai

PR #1600, opened 2024-11-28
fix: Enforce NonAdminRestore spec field values

mateusoliveira43:
## Why the changes were made

Related to https://github.com/migtools/oadp-non-admin/issues/37

This PR adds `EnforceRestoreSpec` to DPA `spec.nonAdmin`. With it, admin user can enforce that all NonAdminRestores created have the enforced spec fields with the values admin user configured.

Follow [install-from-source](https://github.com/migtools/oadp-non-admin/blob/master/docs/CONTRIBUTING.md#install-from-source) NAC testing documentation, pointing [`fix/enforce-restore-spec-field-values`](https://github.com/mateusoliveira43/oadp-non-admin/tree/fix/enforce-restore-spec-field-values) and this branch.

Test that:
- when changing DPA `spec.nonAdmin.enforceRestoreSpec`, NAC Pod recreates (for new enforcement to take effect)
- when not changing DPA `spec.nonAdmin.enforceRestoreSpec`, NAC Pod does not get recreated
- when setting  DPA `spec.nonAdmin.enforceRestoreSpec`, NonAdminRestores fail validation if not respecting enforcement


mpryc:
@mateusoliveira43 please add documentation about expected enforce if the cluster admin will add in the following enforcement (example):

```yaml
excludedResources: {}
```

Currently this will have *no* effect and the user will be able to specify the value.

Otherwise it's good to go in my opinion.

mateusoliveira43:
@mpryc where?

mpryc:
> @mpryc where?

Actually not within this PR but the other one which depends on it. Somewhere around the design:

https://github.com/migtools/oadp-non-admin/blob/0168f6c5998740fdd59a330e4b14000237c9a7cb/docs/design/admin_control_over_spec.md?plain=1#L112

Possibly creating Issue so we won't loose this.

mpryc (approved):

mpryc (commented):
/lgtm

shubham-pampattiwar (approved):

kaovilai (approved):

PR #1603, opened 2024-12-04
fix: Update NAC CRDs

mateusoliveira43:
## Why the changes were made

Update NAB CRD for https://github.com/migtools/oadp-non-admin/pull/120

Follow NAC development documentation.

### DIFF
--- a/bundle/manifests/oadp.openshift.io_nonadminbackups.yaml
+++ b/bundle/manifests/oadp.openshift.io_nonadminbackups.yaml
@@ -525,18 +525,8 @@ spec:
                   ForceDeleteBackup removes the NonAdminBackup and its associated VeleroBackup from the cluster,
                   regardless of whether deletion from object storage succeeds or fails
                 type: boolean
-              logLevel:
-                description: NonAdminBackup log level (use debug for the most logging,
-                  leave unset for default)
-                enum:
-                - trace
-                - debug
-                - info
-                - warning
-                - error
-                - fatal
-                - panic
-                type: string
+            required:
+            - backupSpec
             type: object
           status:
             description: NonAdminBackupStatus defines the observed state of NonAdminBackup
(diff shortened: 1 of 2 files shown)


shubham-pampattiwar (approved):

kaovilai (approved):
/lgtm

bundle/manifests/oadp.openshift.io_nonadminbackups.yaml
```diff
@@ -525,18 +525,8 @@ spec:
...
-                enum:
-                - trace
-                - debug
-                - info
-                - warning
-                - error
-                - fatal
-                - panic
-                type: string
+            required:
```
mpryc, line 528: As we discussed Today, do we want to remove most of the required ?

sseago: @mpryc I don't think anything we talked about today would necessarily indicate having no required fields -- the concern with required fields is adding them to existing APIs. Required fields in new APIs are fine, as long as they're truly required.

mpryc: @sseago correct, we are adding this to a new API for the NAC and in this case we are embedding entire backupSpec at the top level of the NAB object `spec`. If we agree that we don't really want to change this in the future, for example by having our own subset of `backupSpec` fields rather then embedding entire velero spec then we can make it required, otherwise we could make it optional ?

We have in the controller verification if the backupSpec is valid or not.

Here is an example of current spec. As you can see user still can use `includedNamespaces` as it's taken from the velero spec, however it's not allowed by the controller to use anything else then current object namespace. I see this as possible place to improvement in the future, however if we decide on required we may be limiting such improvement? 

```yaml
kind: NonAdminBackup
metadata:
  name: example
  namespace: default
spec:
  backupSpec:
    includedNamespaces: something
```

PR #1613, opened 2025-01-05
OADP-5245: Operator SDK update guidelines

mateusoliveira43:
## Why the changes were made

Fix https://github.com/openshift/oadp-operator/issues/1540

Changes were made following [documentation about how to upgrade Operator SDK version](https://github.com/mateusoliveira43/oadp-operator/blob/fix/operator-sdk-update-guidelines/docs/developer/update_operator_sdk.md), from `1.23.0` to `1.34.2`.

Changes are necessary to allow creation of webhooks for NAC CRDs.

Run commands to confirm that after the changes they still work. For example
```sh
make help
make generate
make bundle
make lint
make test
make deploy-olm
make test-e2e
```

Read [documentation about how to upgrade Operator SDK version](https://github.com/mateusoliveira43/oadp-operator/blob/fix/operator-sdk-update-guidelines/docs/developer/update_operator_sdk.md) and check if it is clear.


mateusoliveira43:
/hold 
more testing needed

api/v1alpha1/cloudstorage_types.go
```diff
@@ -1,23 +1,26 @@
+/*
+Copyright 2021.
```
kaovilai, line 2: do you know where 2021 is coming from?

mateusoliveira43: https://github.com/openshift/oadp-operator/blob/master/hack/boilerplate.go.txt

kaovilai: should we update that here?

mateusoliveira43: since it is not necessary for it to work, no. Can be a separate PR

internal/controller/predicate.go
```diff
@@ -41,7 +41,7 @@ func isObjectOurs(scheme *runtime.Scheme, object client.Object) bool {
 		return false
 	}
 	gvk := objGVKs[0]
-	if gvk.Group == oadpv1alpha1.GroupVersion.Group && gvk.Version == oadpv1alpha1.GroupVersion.Version && gvk.Kind == oadpv1alpha1.Kind {
+	if gvk.Group == oadpv1alpha1.GroupVersion.Group && gvk.Version == oadpv1alpha1.GroupVersion.Version && gvk.Kind == "DataProtectionApplication" {
```
kaovilai, line 44: is this string not available somewhere as const/var already?

mateusoliveira43: it was in `api/` folder

since it was only referenced here, I removed it

kaovilai (dismissed):

internal/controller/bsl_test.go
```diff
@@ -1,4 +1,4 @@
-package controllers
```
shubham-pampattiwar, line 1: IMO `controllers` (plural) is more apt that  `controller` (singular) as we have multiple controllers in our project. 
Ref: https://sdk.operatorframework.io/docs/overview/project-layout/#golang

cmd/main.go
```diff
@@ -18,6 +18,7 @@ package main
```
shubham-pampattiwar, line 18: Similarly IMO `main.go` at root level seems more apt, I am unsure about moving it to `cmd`
Ref: https://sdk.operatorframework.io/docs/overview/project-layout/#golang

internal/controller/cloudstorage_controller.go
```diff
@@ -31,31 +47,21 @@ const (
 	oadpCloudStorageDeleteAnnotation = "oadp.openshift.io/cloudstorage-delete"
 )
 
-// VeleroReconciler reconciles a Velero object
-type BucketReconciler struct {
+// CloudStorageReconciler reconciles a CloudStorage object
```
shubham-pampattiwar, line 50: ++ this always bugged me :)

shubham-pampattiwar (commented):
Overall changes look sane to me, except the dir changes for `controllers` and `main.go` . Also, just because we have some csv updates and refactoring I would test the PR on various OCP favors (ROSA, OCP-AWS etc), just to be cautious. Thank you @mateusoliveira43

internal/controller/bsl_test.go
mateusoliveira43: this documentation is outdated 

here is how a project scaffold with operator sdk version 1.34 looks like https://github.com/operator-framework/operator-sdk/tree/v1.34.x/testdata/go/v4/memcached-operator

it uses singular https://github.com/operator-framework/operator-sdk/blob/v1.34.x/testdata/go/v4/memcached-operator/internal/controller/memcached_controller.go#L17 

changing this can break things

cmd/main.go
mateusoliveira43: this is how it is generated in 1.34 https://github.com/operator-framework/operator-sdk/tree/v1.34.x/testdata/go/v4/memcached-operator/cmd

weshayutin:
@weshayutin please test install on rosa x86 and aws arm

weshayutin:
ROSA 4.17 x86_64 install and test: https://termbin.com/uq5x
looks good

weshayutin:
hrm.. @mateusoliveira43 arm install is hitting an issue. Not clear to me yet if it's the cluster or code changes.  Issue is on `make deploy-olm` 
```
INFO[0038] Approved InstallPlan install-vwm7k for the Subscription: oadp-operator-v99-0-0-sub 
INFO[0038] Waiting for ClusterServiceVersion "openshift-adp/oadp-operator.v99.0.0" to reach 'Succeeded' phase 
INFO[0039]   Found ClusterServiceVersion "openshift-adp/oadp-operator.v99.0.0" phase: Pending 
INFO[0043]   Found ClusterServiceVersion "openshift-adp/oadp-operator.v99.0.0" phase: InstallReady 
INFO[0044]   Found ClusterServiceVersion "openshift-adp/oadp-operator.v99.0.0" phase: Installing 
FATA[0120] Failed to run bundle: error waiting for CSV to install: deployment openshift-adp-controller-manager has error: client rate limiter Wait returned an error: context deadline exceeded

```

mateusoliveira43:
@weshayutin this error seems platform error (it was not built for ARM architecture)

do you have the whole `make deploy-olm` output or the architecture of the container images?

weshayutin:
> @weshayutin this error seems platform error (it was not built for ARM architecture)
> 
> do you have the whole `make deploy-olm` output or the architecture of the container images?

ya... so I mentioned that ARM is not built yet in my comment.   

> @weshayutin this error seems platform error (it was not built for ARM architecture)
> 
> do you have the whole `make deploy-olm` output or the architecture of the container images?

ya.. I mentioned in my comment that ARM does not appear to be built.  The Makefile looks like it should get the cluster os and arch via $CLUSTER_OS/$CLUSTER_ARCH.  At least on my system that does not work.    Setting DOCKER_BUILD_ARGS does work because we have ?= on line 238.   

```
IMG=ttl.sh/oadp-operator-d1f6ee85:1h  BUNDLE_IMG=ttl.sh/oadp-operator-bundle-d1f6ee85:1h  \
	make docker-build docker-push bundle bundle-build bundle-push; \
chmod -R 777 /tmp/tmp.tJcyAUPAvP/  && rm -rf /tmp/tmp.tJcyAUPAvP/ 
make[1]: Entering directory '/tmp/tmp.tJcyAUPAvP'
Using Container Tool: docker
docker build --load -t ttl.sh/oadp-operator-d1f6ee85:1h .  --platform=linux/arm64
[+] Building 58.8s (18/18) FINISHED                                                                                                                                                                                                            
 => [internal] load build definition from Dockerfile                                                                                                                                                                                      0.1s
 => => transferring dockerfile: 1.51kB                                                                                                                                                                                                    0.0s
 => [internal] load .dockerignore          
```

This one looks like it will build correctly :)

weshayutin:
OK... ARM works https://termbin.com/54zs after setting:
```
echo $DOCKER_BUILD_ARGS 
--platform=linux/arm64
```

weshayutin:
@shubham-pampattiwar ok.. rosa x86_64 and arm were able to deploy.  Please review when you have time.  Thank you!

mpryc:
@kaovilai @shubham-pampattiwar @weshayutin @sseago @stillalearner 

Let's make this PR a priority and blocker for all other PRs in the queue as it's a big one and requires some testing from our side. I am gong to review it asap as well.

**Note:** once this is merged it will make cherry picks to other branches a bit more demanding in some of the situations, due to number of changes across multiple files.

PROJECT
```diff
@@ -1,9 +1,32 @@
...
+- api:
+    crdVersion: v1
+    namespaced: true
+  controller: true
+  domain: openshift.io
+  group: oadp
+  kind: CloudStorage
+  path: github.com/openshift/oadp-operator/api/v1alpha1
+  version: v1alpha1
 version: "3"
```
mpryc, line 32: Should this version on last line of PROJECT be bumped as well?

config/manager/kustomization.yaml
```diff
@@ -1,13 +1,5 @@
...
 - manager.yaml
-
-generatorOptions:
-  disableNameSuffixHash: true
-
-configMapGenerator:
-- files:
-  - controller_manager_config.yaml
-  name: manager-config
 apiVersion: kustomize.config.k8s.io/v1beta1
```
mpryc, line 3: Reading about changes in the following:
 https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.33.0/

This points to https://book.kubebuilder.io/plugins/available/kustomize-v2

Should the kustomize be moved to v2 ?

Makefile
```diff
@@ -239,50 +173,186 @@ DOCKER_BUILD_ARGS ?= --platform=linux/amd64
...
 
+# PLATFORMS defines the target platforms for the manager image be built to provide support to multiple
+# architectures. (i.e. make docker-buildx IMG=myregistry/mypoperator:0.0.1). To use this option you need to:
+# - be able to use docker buildx. More info: https://docs.docker.com/build/buildx/
+# - have enabled BuildKit. More info: https://docs.docker.com/develop/develop-images/build_enhancements/
+# - be able to push the image to your registry (i.e. if you do not set a valid value via IMG=<myregistry/image:<tag>> then the export will fail)
+# To adequately provide solutions that are compatible with multiple platforms, you should consider using this option.
+PLATFORMS ?= linux/arm64,linux/amd64,linux/s390x,linux/ppc64le
+.PHONY: docker-buildx
+docker-buildx: ## Build and push docker image for the manager for cross-platform support
```
mpryc, line 192: Bump instructions to 1.34.0 mensions replacing this:

https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.34.0/

mpryc (changes requested):
I've reviewed this by following upgrading sdk versions starting from 1.23.0 up to the 1.34.0:

   https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.23.0/

@mateusoliveira43 one question, does this requires as well changes to the go.mod?

For example this particular one references go.mod to be modified:
 https://sdk.operatorframework.io/docs/upgrading-sdk-version/v1.28.0/#upgrade-k8s-versions-to-use-126

I see from above other changes such as `internal/controller/ internal/controller/` in the Dockerfile, just asking if we should expect some changes to the go.mod here. Possibly no as we are using newer versions but want to be double sure for that.

Another question, why 1.34 was chosen if there are newer ones available up to 1.39 ?

config/manager/kustomization.yaml
mateusoliveira43: Kustomize plugin is used under the hood, its in the kubebuilder code https://github.com/kubernetes-sigs/kubebuilder/tree/master/pkg/plugins/common/kustomize/v2 (that Operator SDK uses)

this yaml has the correct version https://github.com/operator-framework/operator-sdk/blob/v1.34.x/testdata/go/v4/memcached-operator/config/manager/kustomization.yaml#L3

Makefile
mateusoliveira43: I think we can safely ignore this

we do not use this command (CI builds multi arch images for us and we use `make docker-build` and `make bundle-build` when we want different arch images)

we could remove the command. I prefer to let it be so future Operator SDK upgrades are easier

mateusoliveira43:
@mpryc I did not choose Opeartor SDK version to upgrade to. I am just fixing current project structure, which uses 1.34.2 https://github.com/openshift/oadp-operator/blob/9d8b0d52b97c0b1e81f99b15d3481cb8d607d3fd/Makefile#L329

Updating current Operator SDK version is another PR

My suggestion would be to always use the Operator SDK version that has the same kubebuilder version as NAC

---
It may require changes to `go.mod`. Since this Operator SDK version was released on may last year, all our dependencies have greater version then the ones it uses. Following these instructions https://github.com/mateusoliveira43/oadp-operator/blob/fix/operator-sdk-update-guidelines/docs/developer/update_operator_sdk.md to update version, will show `go.mod` changes. If there are older dependencies, it is good to update them

Makefile
mpryc: I think it's still worth updating it to say we are fully migrated to 1.34.x.
One could use the buildx to work on multi-arch, so either remove that or modify to ensure it's aligned with the upgrading release instructions and then I will approve.

Thanks.

PROJECT
mpryc: ok, can close this comment.

config/manager/kustomization.yaml
mpryc: Ok, thanks. Makes sense.

Makefile
mateusoliveira43: removed

mpryc:
/retest

Last one as there may be some change to the images for 4.18 which are not tagged with -dev or it was infra issue.

mpryc:
I think 4.18 CI jobs may need updating.

mpryc:
This should fix 4.18 jobs: https://github.com/openshift/release/pull/60797

PR #1618, opened 2025-01-23
Update manifests for the OADP master to sync with oadp non-admin

mpryc:
Updates the manifests and rbac to align with current master branch of the oadp-non-admin project.

## Why the changes were made

To allow oadp-non-admin CI jobs passing.

Ran make manifests pointing to the latest version of non-admin with use of go 1.22.7


mateusoliveira43:
correct fix https://github.com/migtools/oadp-non-admin/pull/150

PR #1621, opened 2025-01-30
Fix Makefile to allow building images aligned with cluster architecture

mpryc:
Setting the CLUSTER_TYPE, CLUSTER_OS and CLUSTER_ARCH in the Makefile happened after it was used in the Makefile making them effectively unset.

Moving the part to discover those variables above their usage fixes the problem of images being built for improper arch.

## Why the changes were made

To fix `$make deploy-olm` on the cluster which has different architecture then builder platform.

You can build the images using arm cluster from the x64 box or you could add debug of those fields just before their usage:
```
# Debugging:
$(info CLUSTER_OS: $(CLUSTER_OS))
$(info CLUSTER_ARCH: $(CLUSTER_ARCH))
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -165,6 +165,14 @@ build: manifests generate fmt vet ## Build manager binary.
 run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/main.go
 
+OC_CLI ?= $(shell which oc)
+
+# makes CLUSTER_TYPE quieter when unauthenticated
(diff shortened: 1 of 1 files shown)


mpryc:
@sseago @weshayutin FYI

kaovilai (approved):

shubham-pampattiwar (approved):

weshayutin:
thanks @mpryc !

PR #1623, opened 2025-01-31
feat: Garbage collector input

mateusoliveira43:
## Why the changes were made

Allow admin users to configure NAC garbage collection frequency https://github.com/migtools/oadp-non-admin/issues/145

Follow instructions from NAC PR https://github.com/migtools/oadp-non-admin/pull/152


api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -447,6 +447,11 @@ type NonAdmin struct {
 	// which restore spec field values to enforce
 	// +optional
 	EnforceRestoreSpec *velero.RestoreSpec `json:"enforceRestoreSpec,omitempty"`
+
+	// GarbageCollectionPeriod defines how frequently to look for possible leftover non admin objects in OADP namespace.
+	// By default 5m
```
mateusoliveira43, line 452: the default will be applied in NAC side, but documentation will live here

please tell me what should be the default, I just added a random value

internal/controller/validator.go
```diff
@@ -83,37 +83,45 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
...
+					); err == nil {
+						return false, fmt.Errorf("only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is already configured and installed in %s namespace", dpa.Namespace)
+					}
 				}
 			}
 		}
 
+		garbageCollectionPeriod := r.dpa.Spec.NonAdmin.GarbageCollectionPeriod
+		if garbageCollectionPeriod != nil {
+			if garbageCollectionPeriod.Duration < 0 {
```
mateusoliveira43, line 121: should we not allow zero as well?

should we do as velero and if user pass zero, disable controller?

api/v1alpha1/dataprotectionapplication_types.go
shubham-pampattiwar: The scenario where residual Velero objects get leftover won't be very frequent, right ? 
I think daily cleanup as a default (24 hr) would be reasonable here.

internal/controller/validator.go
shubham-pampattiwar: yeah replicate velero behavior here.

api/v1alpha1/dataprotectionapplication_types.go
mateusoliveira43: will change doc here and value in code in NAC side

internal/controller/validator.go
mateusoliveira43: alright, will make change in NAC code side

api/v1alpha1/dataprotectionapplication_types.go
kaovilai: Probably matching one of these times
https://github.com/search?q=repo%3Avmware-tanzu%2Fvelero+time.Minute+*+%3D+NOT+path%3Atest.go&type=code

kaovilai: 24 hrs works

kaovilai: That should discourage frequent nonAdminSchedules that we sometimes see admins do.. hourly backup of entire clusters.. 😮‍💨

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -447,6 +447,11 @@ type NonAdmin struct {
 	// which restore spec field values to enforce
 	// +optional
 	EnforceRestoreSpec *velero.RestoreSpec `json:"enforceRestoreSpec,omitempty"`
+
```
shubham-pampattiwar, line 450: `s/non admin objects/non admin related objects`
We are cleaning up residual Velero Objects and secrets, right ? (Not the actual Non-Admin prefixed objects)

mateusoliveira43: yes, objects NAC controller created in OADP namespace and for some reason they were not properly deleted, GC will do this job

will make the change

kaovilai: Please revise PR title/description of the same detail.

mateusoliveira43: do not see other places that needs changing. Can you point it to me?

sseago (approved):

api/v1alpha1/dataprotectionapplication_types.go
kaovilai: > feat: Garbage collector input

do not convey that this GC is limited to NAC created resources in oadp-operator namespace.

sseago (approved):

kaovilai (approved):
nit: PR title should convey the scope of functionality.

feat: DPA NAC Garbage collector input

PR #1629, opened 2025-02-07
Adding --ginkgo.junit-report flag

mail2nadeem92:
## Why the changes were made

Adding `--ginkgo.junit-report` flag to generate the junit XML file after test execution such that those results can be uploaded to the report portal.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -596,6 +596,7 @@ test-e2e: test-e2e-setup install-ginkgo ## Run E2E tests against OADP operator i
 	--ginkgo.vv \
 	--ginkgo.no-color=$(OPENSHIFT_CI) \
 	--ginkgo.label-filter="$(TEST_FILTER)" \
+	--ginkgo.junit-report="junit_report.xml" \
 	--ginkgo.timeout=2h \
 	$(GINKGO_ARGS)
 
--- a/.gitignore
+++ b/.gitignore
@@ -31,5 +31,6 @@ tests/e2e/e2e.test
 tests/e2e/templates/*.yaml
 .DS_Store
 test-upgrade/
+junit_report.xml
 # `operator-sdk run bundle` caches files in this directory
 cache/


Makefile
```diff
@@ -596,6 +596,7 @@ test-e2e: test-e2e-setup install-ginkgo ## Run E2E tests against OADP operator i
 	--ginkgo.vv \
 	--ginkgo.no-color=$(OPENSHIFT_CI) \
 	--ginkgo.label-filter="$(TEST_FILTER)" \
+	--ginkgo.junit-report="junit_report.xml" \
```
mateusoliveira43, line 599: what will this do? write to this file after test run? if so, we need to add file to gitignore

And we also need to check if file is stored in PROW to validate PR? or not needed?

weshayutin: I agree, but atm all we would need in this pr is a gitignore.  I have asked for this work to be done so that the QE infra team can get some experience with importing test results into the report portal.

weshayutin (dismissed):
please add the gitignore and we'll merge this up.

mpryc (approved):
This is ok. Another option to consider is using path for the `junit_report.xml` inside `$(ARTIFACT_DIR)` which is already defined  on line 565 of this Makefile. 

This can be done in either defining gingko artifact dir for all outputs by:
```shell
        --ginkgo.output-dir=$(ARTIFACT_DIR) \
	--ginkgo.junit-report="junit_report.xml" \
```

or just for `junit_report.xml`:

```shell
	--ginkgo.junit-report="$(ARTIFACT_DIR)\junit_report.xml" \
```

Of course this is just an improvement and depends if reportportal can easily consume junit_report.xml from such different to current working dir locations.

Approving, because current implementation is good enough and leaving decision to improve or merge to @weshayutin.

weshayutin:
hrm.. odd.  something else in the .gitignore must be handling the junit xml file

```
whayutin@thinkdoe:~/OPENSHIFT/git/OADP/oadp-operator$ ls -ltra tests/e2e/junit_report.xml 
-rw-r--r--. 1 whayutin whayutin 119614 Feb 10 10:28 tests/e2e/junit_report.xml
whayutin@thinkdoe:~/OPENSHIFT/git/OADP/oadp-operator$ git status
On branch add_junitflag
nothing to commit, working tree clean
```

weshayutin (approved):
/LGTM

Issue #1595, opened 2024-11-26
controllers/velero.go: `veleroDeployment *appsv1.Deployment` unused in customizeVeleroContainer


kaovilai:
`veleroDeployment *appsv1.Deployment` is unused
https://github.com/openshift/oadp-operator/blob/d55863cc4150afae779715219e5583521c1049f3/controllers/velero.go#L509

mateusoliveira43:
This the kind of thing linters should take care

This one should do the job https://github.com/openshift/oadp-operator/blob/c0f2825bdba4fb4064cba5609c4c09fe45737b68/.golangci.yaml#L109

Joeavaikath:
I can take a look at this if it involves enabling that flag in the linter and it's a good first issue
I suspect simply removing this unused var isn't the best for the long term

Joeavaikath:
Could close this, since unparam flag has been enabled and unused params have been removed

PR #1630, opened 2025-02-08
Enable unparam flag in linter: Resolve unused func params

Joeavaikath:
## Why the changes were made

Attempting a fix for #1595.
- Enabling unparam flag for golang-ci to catch unused params
- Removing unused params instances

ci/prow/unit-test should now check for unused params.

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -132,7 +132,7 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 		if bslSpec.Velero != nil {
 			secretName, _, _ = r.getSecretNameAndKey(bslSpec.Velero.Config, bslSpec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Velero.Provider))
 		}
-		err := r.UpdateCredentialsSecretLabels(secretName, dpa.Namespace, dpa.Name)
+		err := r.UpdateCredentialsSecretLabels(secretName, dpa.Name)
 		if err != nil {
 			return false, err
 		}
@@ -229,9 +229,8 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 	return true, nil
 }
 
-func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secretName string, namespace string, dpaName string) error {
-	// todo: lot of calls supply this, needs combing
-	_ = namespace
+func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secretName string, dpaName string) error {
+
 	var secret corev1.Secret
 	secret, err := r.getProviderSecret(secretName)
 	if err != nil {
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -132,7 +132,7 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 		if bslSpec.Velero != nil {
 			secretName, _, _ = r.getSecretNameAndKey(bslSpec.Velero.Config, bslSpec.Velero.Credential, oadpv1alpha1.DefaultPlugin(bslSpec.Velero.Provider))
 		}
-		_, err := r.UpdateCredentialsSecretLabels(secretName, dpa.Namespace, dpa.Name)
+		err := r.UpdateCredentialsSecretLabels(secretName, dpa.Namespace, dpa.Name)
 		if err != nil {
 			return false, err
 		}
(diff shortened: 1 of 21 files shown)


kaovilai:
whoa..

internal/controller/bsl_test.go
```diff
@@ -1518,6 +1518,8 @@ func TestDPAReconciler_ValidateBackupStorageLocations(t *testing.T) {
 }
 
 func newContextForTest(name string) context.Context {
+	// todo: used a lot, combing needed
```
kaovilai, line 1521: combing = combining?

kaovilai:
These CI tests take about 2 hours and can be flakey fyi and may require some retrying.

internal/controller/bsl_test.go
Joeavaikath: Oh, I meant combing through the places where this func is called
all the func calls pass the name param, so was not sure if I should remove this param from all those calls. 

there's one other commit like this where I set the unused param to the blank identifier, would like inputs on if it's better to remove it altogether

kaovilai:
As far as lint goes, ci/prow/unit-test should be the one to look at and should take less than 15 minutes.. 5 minutes on a good day.

internal/controller/bsl_test.go
kaovilai: based on the return type of context.Context where I don't think it's possible to add names here, I think this one can just remove the unused param.

kaovilai:
https://prow.ci.openshift.org/view/gs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1630/pull-ci-openshift-oadp-operator-master-unit-test/1889912343991685120
unit test passing.

[lint successful](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1630/pull-ci-openshift-oadp-operator-master-unit-test/1889912343991685120/artifacts/unit-test/unit/build-log.txt#:~:text=/go/src/github.com/openshift/oadp%2Doperator/bin/golangci%2Dlint%20run).

internal/controller/validator_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"testing"
 
 	"github.com/go-logr/logr"
-	"github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```
mpryc, line 7: This change caused small issue in CI:
```
internal/controller/validator_test.go:7:2: redundant-import-alias: Import alias "v1" is redundant (revive)
	v1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
```

Joeavaikath: golang forces an alias if I try removing it, so set it to velerov1

kaovilai (dismissed):

kaovilai:
will wait for more approvals and this should go in.

internal/controller/validator_test.go
mateusoliveira43: what you mean by "golang forces an alias"?

tests/e2e/lib/k8s_common_helpers.go
```diff
@@ -209,7 +209,7 @@ func GetFirstPodByLabel(clientset *kubernetes.Clientset, namespace string, label
 func SavePodLogs(clientset *kubernetes.Clientset, namespace, dir string) error {
 	podList, err := clientset.CoreV1().Pods(namespace).List(context.Background(), metav1.ListOptions{})
 	if err != nil {
-		return nil
+		return err
```
mateusoliveira43, line 212: I would further refactor here

there are other 3 `if err != nil` statements, that do not return an error

(This function only runs if a test fails, so changing its behavior seems harmless to me)

internal/controller/bsl.go
```diff
@@ -229,14 +229,16 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 	return true, nil
 }
 
-func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secretName string, namespace string, dpaName string) (bool, error) {
+func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secretName string, namespace string, dpaName string) error {
```
mateusoliveira43, line 232: lets also remove `namespace` param

tests/e2e/lib/k8s_common_helpers.go
Joeavaikath: Added returns

mateusoliveira43 (approved):

kaovilai (approved):
/lgtm

internal/controller/validator_test.go
Joeavaikath: Removing the explicit v1 package import alias didn't work because it kept adding it back when the file was saved. Not sure what does that

mateusoliveira43: that is not Go, probably your IDE

in VSCode, for example, this setting is required to avoid this
```json
"[go]": {
        "editor.codeActionsOnSave": {
            "source.organizeImports": "never"
        }
    },
```

PR #1631, opened 2025-02-14
Make common function for updating BSL spec

mpryc:
## Why the changes were made

With the self service project (Non Admin) it is required to update BSL spec based on the same conditions as OADP BSL spec would be updated. It makes sense to move out the logic from the bsl specific function to a common publicly available from which NABSL controller can also benefit.

Run tests. New tests have been written to cover public function, old tests are also working proving that the function is correct.

This is required to fix https://github.com/migtools/oadp-non-admin/issues/188 on the Non-Admin side.

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -265,50 +265,22 @@ func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secr
 
 func (r *DataProtectionApplicationReconciler) updateBSLFromSpec(bsl *velerov1.BackupStorageLocation, bslSpec velerov1.BackupStorageLocationSpec) error {
 	// Set controller reference to Velero controller
-	err := controllerutil.SetControllerReference(r.dpa, bsl, r.Scheme)
-	if err != nil {
+	if err := controllerutil.SetControllerReference(r.dpa, bsl, r.Scheme); err != nil {
 		return err
 	}
-	// While using Service Principal as Azure credentials, `storageAccountKeyEnvVar` value is not required to be set.
-	// However, the registry deployment fails without a valid storage account key.
-	// This logic prevents the registry pods from being deployed if Azure SP is used as an auth mechanism.
-	registryDeployment := "True"
-	if bslSpec.Provider == "azure" && bslSpec.Config != nil {
-		if len(bslSpec.Config["storageAccountKeyEnvVar"]) == 0 {
-			registryDeployment = "False"
-		}
-	}
-	// The AWS SDK expects the server providing S3 blobs to remove default ports
-	// (80 for HTTP and 443 for HTTPS) before calculating a signature, and not
-	// all S3-compatible services do this. Remove the ports here to avoid 403
-	// errors from mismatched signatures.
-	if bslSpec.Provider == "aws" && bslSpec.Config != nil {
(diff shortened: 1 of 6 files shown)


pkg/common/common.go
```diff
@@ -299,3 +303,48 @@ func ApplyUnsupportedServerArgsOverride(container *corev1.Container, unsupported
 	}
 	return nil
 }
+
+// UpdateBackupStorageLocation updates the BackupStorageLocation spec and config.
+func UpdateBackupStorageLocation(bsl *velerov1.BackupStorageLocation, bslSpec velerov1.BackupStorageLocationSpec) error {
```
mpryc, line 308: I am aware that currently this function does not really return error, however it's a design to ensure if it does, both NABSL controller and bsl functions from the OADP project does not need to update calls to it.

I can imagine some validation could be done here which would return error.

mpryc:
/hold

Note: If we decide in the oadp-non-admin project that we want to update NABSL spec for the user and as a result the Velero BSL will get updated, then this function will need to be slightly modified. If current implementation is agreed then it's as is.

internal/controller/bsl.go
```diff
@@ -264,50 +264,22 @@ func (r *DataProtectionApplicationReconciler) UpdateCredentialsSecretLabels(secr
...
-		oadpv1alpha1.RegistryDeploymentLabel: registryDeployment,
+	// Update BSL spec and registry-deployment label
+	if err := common.UpdateBackupStorageLocation(bsl, bslSpec); err != nil {
+		return err
 	}
-	bsl.Spec = bslSpec
+
+	// Assign labels individually
+	// bsl.Labels was initialized in the common.UpdateBackupStorageLocation function
+	bsl.Labels["app.kubernetes.io/name"] = common.OADPOperatorVelero
```
kaovilai, line 278: you might need to first make this := map[string]string

kaovilai: ok done by common func. either comment or do a lil nil check.

kaovilai: nvm.. h*k..

pkg/common/common.go
kaovilai: ok so expect nolint here in the future from some linter

kaovilai (dismissed):

shubham-pampattiwar (dismissed):
/lgtm

kaovilai (commented):
regarding modifying nabsl.. I think we can do it behind-the-scenes like in oadp, where we don't apply what we modified to dpa cr in cluster.

mpryc:
> regarding modifying nabsl.. I think we can do it behind-the-scenes like in oadp, where we don't apply what we modified to dpa cr in cluster.

@kaovilai exactly, but I wanted to confirm first before going full ahead with such implementation. Doing it differently would require moving label thing separate as we would not mark label on the BSL object but on the generic one which would be NABSL or BSL depending on who is calling this new function.

pkg/common/common.go
mpryc: yes (maybe, or we add err validation first) :)

mateusoliveira43 (approved):
tested on IBM COS

kaovilai (approved):

shubham-pampattiwar (approved):
/lgtm

PR #1635, opened 2025-02-21
Pass LogLevel to NAC deployment

kaovilai:

## Why the changes were made
To help solve https://github.com/migtools/oadp-non-admin/issues/51

See that NAC deployment would have the LOG_LEVEL env with appropriate logrus loglevel int value passed as string.

### DIFF
--- a/internal/controller/nonadmin_controller.go
+++ b/internal/controller/nonadmin_controller.go
@@ -160,7 +160,10 @@ func ensureRequiredSpecs(deploymentObject *appsv1.Deployment, dpa *oadpv1alpha1.
 			Name: common.LogLevelEnvVar,
 			Value: func() string {
 				// these levels are already validated in another controller.
-				level, _ := logrus.ParseLevel(dpa.Spec.Configuration.Velero.LogLevel)
+				level, err := logrus.ParseLevel(dpa.Spec.Configuration.Velero.LogLevel)
+				if err != nil {
+					return ""
+				}
 				return strconv.FormatUint(uint64(level), 10)
 			}(),
 		})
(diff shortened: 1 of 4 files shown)


internal/controller/nonadmin_controller.go
```diff
@@ -147,9 +149,21 @@ func ensureRequiredLabels(deploymentObject *appsv1.Deployment) {
...
-	namespaceEnvVar := corev1.EnvVar{
-		Name:  "WATCH_NAMESPACE",
-		Value: deploymentObject.Namespace,
+	envVars := []corev1.EnvVar{
+		{
+			Name:  "WATCH_NAMESPACE", // TODO: fix, this is only used to indicate oadp ns, and is not actually used for watching ns.
+			Value: deploymentObject.Namespace,
+		},
+	}
+	if dpa.Spec.Configuration != nil && dpa.Spec.Configuration.Velero != nil {
```
kaovilai, line 158: Perhaps I should have an else that if not specified then default to info

mateusoliveira43: lets have same default as velero, since we using same option for both

mpryc: Current default level is from the:
   https://github.com/migtools/oadp-non-admin/commit/5204eb4fe7a0b72d3eacdfe4db6f2258e80391f1#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R85
   
   Is this answering your question?

mateusoliveira43: yes

mpryc (approved):

mpryc (approved):

sseago (approved):

PR #1638, opened 2025-02-24
fix: BSL sync period validation with NAC

mateusoliveira43:
## Why the changes were made

https://github.com/migtools/oadp-non-admin/issues/192

Create a default BSL that has backup sync period greater or equal to NAC backup sync period, DPA should have error message in status.

### DIFF
--- a/internal/controller/validator.go
+++ b/internal/controller/validator.go
@@ -4,9 +4,11 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"time"
 
 	mapset "github.com/deckarep/golang-set/v2"
 	"github.com/go-logr/logr"
+	velerov1 "github.com/vmware-tanzu/velero/pkg/apis/velero/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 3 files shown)


internal/controller/validator.go
```diff
@@ -139,8 +141,45 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
...
-		// but BSL can not exist yet when we validate the value
+
+		defaultBSLIndex := -1
+		bslsInOADPNamespace := &velerov1.BackupStorageLocationList{}
+		r.List(r.Context, bslsInOADPNamespace, client.InNamespace(r.NamespacedName.Namespace))
+		for index, bsl := range bslsInOADPNamespace.Items {
+			if bsl.Spec.Default {
+				defaultBSLIndex = index
+				break
+			}
```
kaovilai, line 152: Is it safe to assume velero would be running by this time?

If not, we might want to either wait, or use the same logic as velero function, which would [remove older defaults](https://github.com/vmware-tanzu/velero/blob/a45c9f27e890380e889e073496132b53e89cb10c/pkg/controller/backup_storage_location_controller.go#L250-L255) if more than one.

mateusoliveira43: From my tests, If I just install OADP and create DPA with non admin enabled, NAC Pod goes live prior to Velero Pod, so we can not assume that :( 

Will take a look

mateusoliveira43: ah, but DPA reconcile will run every minute, so we do not need to wait, the next reconcile will have velero running, right?

kaovilai: Sure. if image is correct.

internal/controller/validator.go
```diff
@@ -139,8 +141,45 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
...
+			defaultBSLSyncPeriodErrorMessage := "default BSL spec.backupSyncPeriod (%v) can not be greater or equal spec.nonAdmin.backupSyncPeriod (%v)"
+			if defaultBSL.Spec.BackupSyncPeriod != nil {
+				if appliedBackupSyncPeriod <= defaultBSL.Spec.BackupSyncPeriod.Duration {
+					return false, fmt.Errorf(
+						defaultBSLSyncPeriodErrorMessage,
+						defaultBSL.Spec.BackupSyncPeriod.Duration, appliedBackupSyncPeriod,
+					)
+				}
+			} else {
+				if r.dpa.Spec.Configuration.Velero.Args != nil && r.dpa.Spec.Configuration.Velero.Args.BackupSyncPeriod != nil {
```
mateusoliveira43, line 165: changing this value in DPA does not recreate NAC Pod
 
is this a problem?

internal/controller/validator.go
shubham-pampattiwar: Why not add a check re: velero pod running ?

internal/controller/validator.go
shubham-pampattiwar: That means that NAC won't be pulling in the latest updates, right ?

internal/controller/validator.go
mateusoliveira43: I think this is separate problem

we could add validation to check if velero Pod is running, but I think this PR is not the place

internal/controller/validator.go
mateusoliveira43: uhum

mateusoliveira43: alright, I think I fixed, can you test

IBM is currently down, could not test in my cluster

weshayutin:
running this through a test now

weshayutin:
first test worked, but I could not fix the dpa by setting the default bsl to a lower value than the non-admin setting.
```
metadata:
  creationTimestamp: "2025-02-26T18:56:23Z"
  generation: 6
  name: dpa-sample
  namespace: openshift-adp
  resourceVersion: "13296637"
  uid: bcea7ffd-e24b-4899-83f9-0ea232d400b2
spec:
  backupLocations:
  - velero:
      backupSyncPeriod: 8m
      config:
        profile: default
        region: us-west-2
      credential:
        key: cloud
        name: cloud-credentials
      default: true
      objectStorage:
        bucket: cvpbucketuswest2
        prefix: velero
      provider: aws
  configuration:
    nodeAgent:
      enable: true
      uploaderType: kopia
    velero:
      defaultPlugins:
      - kubevirt
      - csi
      - openshift
      - aws
  nonAdmin:
    backupSyncPeriod: 10m
    enable: true
  snapshotLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      provider: aws
status:
  conditions:
  - lastTransitionTime: "2025-02-26T19:00:59Z"
    message: default BSL spec.backupSyncPeriod (11m0s) can not be greater or equal
      spec.nonAdmin.backupSyncPeriod (10m0s)
    reason: Error
    status: "False"
    type: Reconciled
```
please take a look if I screwed something up

weshayutin:
ok. messed around with a bunch of different variations and it all worked!  Thank you @mateusoliveira43 
Here is my latest
```
metadata:
  creationTimestamp: "2025-02-26T19:53:33Z"
  generation: 7
  name: dpa-sample
  namespace: openshift-adp
  resourceVersion: "13320136"
  uid: ac3e7873-fd3e-4398-804d-0705982966f0
spec:
  backupLocations:
  - velero:
      backupSyncPeriod: 9m59s
      config:
        profile: default
        region: us-west-2
      credential:
        key: cloud
        name: cloud-credentials
      default: true
      objectStorage:
        bucket: cvpbucketuswest2
        prefix: velero
      provider: aws
  configuration:
    nodeAgent:
      enable: true
      uploaderType: kopia
    velero:
      defaultPlugins:
      - kubevirt
      - csi
      - openshift
      - aws
  nonAdmin:
    backupSyncPeriod: 10m
    enable: true
  snapshotLocations:
  - velero:
      config:
        profile: default
        region: us-west-2
      provider: aws
status:
  conditions:
  - lastTransitionTime: "2025-02-26T19:56:35Z"
    message: Reconcile complete
    reason: Complete
    status: "True"
    type: Reconciled
```

weshayutin (approved):
/LGTM

sseago (approved):

kaovilai (approved):

shubham-pampattiwar (approved):

PR #1639, opened 2025-02-25
Restrict enforcing NAR spec that are restricted for non-admin users

shubham-pampattiwar:
## Why the changes were made

Fix part of https://github.com/migtools/oadp-non-admin/issues/151
Related to https://github.com/migtools/oadp-non-admin/pull/227

- Try enforcing NAR restore spec values for scheduleName, includedNamespaces, excludedNamespaces and namespacMapping
- DPA should error out with a relevant err message

### DIFF
--- a/internal/controller/validator.go
+++ b/internal/controller/validator.go
@@ -173,6 +173,26 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
 
 		}
 
+		enforcedRestoreSpec := r.dpa.Spec.NonAdmin.EnforceRestoreSpec
+
+		if enforcedRestoreSpec != nil {
+			if len(enforcedRestoreSpec.ScheduleName) > 0 {
+				return false, fmt.Errorf(NACNonEnforceableErr, "spec.nonAdmin.enforcedRestoreSpec.scheduleName")
+			}
+
+			if enforcedRestoreSpec.IncludedNamespaces != nil {
+				return false, fmt.Errorf(NACNonEnforceableErr, "spec.nonAdmin.enforcedRestoreSpec.includedNamespaces")
+			}
+
+			if enforcedRestoreSpec.ExcludedNamespaces != nil {
+				return false, fmt.Errorf(NACNonEnforceableErr, "spec.nonAdmin.enforcedRestoreSpec.excludedNamespaces")
+			}
+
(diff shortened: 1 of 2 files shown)


internal/controller/validator.go
```diff
@@ -173,6 +173,26 @@ func (r *DataProtectionApplicationReconciler) ValidateDataProtectionCR(log logr.
...
+		if enforcedRestoreSpec != nil {
+			if len(enforcedRestoreSpec.ScheduleName) > 0 {
+				return false, fmt.Errorf(NACNonEnforceableErr, "spec.nonAdmin.enforcedRestoreSpec.scheduleName")
+			}
+
+			if enforcedRestoreSpec.IncludedNamespaces != nil {
+				return false, fmt.Errorf(NACNonEnforceableErr, "spec.nonAdmin.enforcedRestoreSpec.includedNamespaces")
+			}
+
+			if enforcedRestoreSpec.ExcludedNamespaces != nil {
```
mpryc, line 187: Can't this become a feature, where an admin want some namespaces to be restricted and not allow users from those namespaces to backup or restore workload? I know it's not part of this PR but as a future request, which will need to change also backup enforcedBackupSpec.ExcludedNamespaces part ?

shubham-pampattiwar: I think you talking along the lines where or which NS we want NAC B/R to work ?

mpryc: Yes

shubham-pampattiwar: Yeah so thats a separate feature where we want to allow admins to configure NAC for a set of namespaces based on some list or criteria. I can track it in a issue.

sseago: Yeah -- restricting namespace would mean "we shouldn't have NAB/NAR at all", not "namespace field in NAB/NAR adds restrictions".

sseago: @shubham-pampattiwar Hmm. So couldn't we handle the "set of namespaces to allow NAC for" via rbac? i.e. don't grant create NAB/NAR role for users on namespaces where we don't want to enable NAC?

sseago (approved):

internal/controller/validator.go
mpryc: Agree, no need to change anything, just thinking laud here.

PR #1641, opened 2025-02-25
update go version: 1.22.8

tareqalayan:


kaovilai (commented):
fyi this is not needed (not blockers) for the CI PRs. The versions can mismatch as long as the CLI version is higher. Also I would not bump go directive unless one of our dependencies requires the bump.

tareqalayan:
@kaovilai ok thanks. I will close this PR.

PR #1662, opened 2025-03-12
OADP-5635: dpa support for parallel item backup

sseago:
## Why the changes were made
Velero 1.16 introduces parallel item backup, allowing multiple items within a backup to be processed in parallel. By default, only one worker thread processes item blocks, so even though item processing is in a separate goroutine, only one item at a time gets backed up. The optional velero server parameter `--item-block-worker-count` allows velero to run additional worker threads to process items in parallel.

To enable this in OADP, set `dpa.Spec.Configuration.Velero.ItemBlockWorkerCount` to an integer value greater than zero.

Set `dpa.Spec.Configuration.Velero.ItemBlockWorkerCount` to an integer value -- velero should be redeployed with a server arg added: `--item-block-worker-count=<dpa-value>`

### DIFF
--- a/api/v1alpha1/dataprotectionapplication_types.go
+++ b/api/v1alpha1/dataprotectionapplication_types.go
@@ -301,6 +301,11 @@ type VeleroConfig struct {
 	// Disable informer cache for Get calls on restore. With this enabled, it will speed up restore in cases where there are backup resources which already exist in the cluster, but for very large clusters this will increase velero memory usage. Default is false.
 	// +optional
 	DisableInformerCache *bool `json:"disableInformerCache,omitempty"`
+	// Number of workers in worker pool for processing item backup. This will allow multiple items within
+	// a Velero backup to be backed up at the same time which may improve performance for backups with
+	// a large number of items. Default is 1.
+	// +optional
+	ItemBlockWorkerCount int `json:"itemBlockWorkerCount,omitempty"`
 	// resourceTimeout defines how long to wait for several Velero resources before timeout occurs,
 	// such as Velero CRD availability, volumeSnapshot deletion, and repo availability.
 	// Default is 10m
--- a/internal/controller/velero.go
+++ b/internal/controller/velero.go
@@ -364,6 +364,10 @@ func (r *DataProtectionApplicationReconciler) customizeVeleroDeployment(veleroDe
 	disableInformerCache := disableInformerCacheValue(dpa)
 	veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--disable-informer-cache=%s", disableInformerCache))
 
+	if dpa.Spec.Configuration.Velero.ItemBlockWorkerCount > 0 {
+		veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--item-block-worker-count=%v", dpa.Spec.Configuration.Velero.ItemBlockWorkerCount))
+	}
+
 	// Set defaults to avoid update events
 	if veleroDeployment.Spec.Strategy.Type == "" {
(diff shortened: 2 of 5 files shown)


api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -301,6 +301,11 @@ type VeleroConfig struct {
 	// Disable informer cache for Get calls on restore. With this enabled, it will speed up restore in cases where there are backup resources which already exist in the cluster, but for very large clusters this will increase velero memory usage. Default is false.
 	// +optional
 	DisableInformerCache *bool `json:"disableInformerCache,omitempty"`
+	// Number of workers in worker pool for processing item backup. This will allow multiple items within
+	// a Velero backup to be backed up at the same time which may improve performance for backups with
+	// a large number of items. Default is 1.
+	// +optional
```
mateusoliveira43, line 307: Can we add `// +kubebuilder:validation:Minimum=1` to validate here?

sseago: @mateusoliveira43 actually, on further thought, we don't want this since it will break upgrades, as it turns this into a required field. The current PR already checks value and ignores the field if it's <1.

internal/controller/velero.go
```diff
@@ -364,6 +364,10 @@ func (r *DataProtectionApplicationReconciler) customizeVeleroDeployment(veleroDe
 	disableInformerCache := disableInformerCacheValue(dpa)
 	veleroContainer.Args = append(veleroContainer.Args, fmt.Sprintf("--disable-informer-cache=%s", disableInformerCache))
 
+	if dpa.Spec.Configuration.Velero.ItemBlockWorkerCount > 0 {
```
shubham-pampattiwar, line 367: Should this check `dpa.Spec.Configuration.Velero.ItemBlockWorkerCount > 1` ?

sseago: @shubham-pampattiwar I think `> 0` is what we want -- if the user wants to explicitly set it to 1, that's fine. Also, this doesn't require code changes if velero changes the default value in the future.

shubham-pampattiwar (approved):

api/v1alpha1/dataprotectionapplication_types.go
mateusoliveira43: should we fail DPA reconcile if invalid value is passed then?

mpryc: I think if you made this an pointer to *int and then the `omitempty` would make it optional and only validate when it's present ?

Another option would be I think to:
```
// +kubebuilder:default=1
// +kubebuilder:validation:Minimum=1
```

This will make default backwards-compatible, but needs some testing if I am right here.

sseago: not sure how "Minimum=1" interacts with pointer vals -- in particular, is that compatible with a pointer val of nil?
I think leaving the default as 0 without any requirements may be easier. I don't see any huge advantage of adding kubebuilder-level validation here  vs. just only passing values on to velero if they're >0.

mpryc: +1 on no real benefit, so without it will be fine

sseago: @mateusoliveira43 Currently we're not failing reconcile, we're just ignoring it if they put something less than 1. If it's zero, that just means the user probably didn't specify anything, so ignoring is completely appropriate. I suppose we could check for negative values and fail reconcile if they are provided, but that's a bit of a silly edge case. If a user wants negative work done, then ignoring the request seems fine to me. But if people really want to check for negative values and error out, I can add that.

mpryc (commented):
/lgtm

PR #1666, opened 2025-03-13
velero rebase Mar 13 2025

sseago:


weshayutin (dismissed):
/LGTM

sseago:
OADP code changes needed. looks like some controller changes due to controller-runtime bump and some unit test fixes due to velero code changes. Will update the PR in a bit.

internal/controller/nodeagent.go
```diff
@@ -290,7 +290,9 @@ func (r *DataProtectionApplicationReconciler) customizeNodeAgentDaemonset(ds *ap
 	// Update with any pod config values
 	if dpa.Spec.Configuration.NodeAgent.PodConfig != nil {
 		ds.Spec.Template.Spec.Tolerations = dpa.Spec.Configuration.NodeAgent.PodConfig.Tolerations
-		ds.Spec.Template.Spec.NodeSelector = dpa.Spec.Configuration.NodeAgent.PodConfig.NodeSelector
+		if len(dpa.Spec.Configuration.NodeAgent.PodConfig.NodeSelector) != 0 {
```
mateusoliveira43, line 293: why change here?

mateusoliveira43: alright, saw that now velero deployment and node-agent daemonSet have nodeSelector `"kubernetes.io/os": "linux"`

but shouldn't we change logic here then? instead of changing nodeSelector, append to it, to do not overwrite what velero set?

sseago: @mateusoliveira43 That's a good question. Maybe we should append here instead? The intent here is to avoid windows nodes. Is that a concern here? @shubham-pampattiwar @kaovilai thoughts?

shubham-pampattiwar: ok how about we drop the windows related selectors (if they exist) and keep everything else and then append to it ?

sseago: @shubham-pampattiwar Currently I'm thinking that we just overwrite -- that way we have a workaround if the velero-generated selectors don't work in a particular customer environment. If the user really needs the velero selector components on top of their own, they can include it in their DPA.

shubham-pampattiwar: yeah that works too, DPA to the rescue

weshayutin (approved):
tested locally w/ mysql-twovol and virt just passed.. GREEN LIGHT :)
/LGTM

shubham-pampattiwar (approved):

PR #1667, opened 2025-03-14
Add option to use ConfigMap in the node agent

mpryc:
Fixes [OADP-5042](https://issues.redhat.com//browse/OADP-5042), add ConfigMap to the NodeAgent

## Why the changes were made

Fix [OADP-5042](https://issues.redhat.com//browse/OADP-5042)

1. Create DPA
2. Use one of the new options in the DPA
3. Check if the CM was created for the DPA (same namespace)
4. Check the CM data
5. Check if the node-agent has the config map arg passed to the command


api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -352,6 +354,52 @@ type NodeAgentCommonFields struct {
...
+	PerNodeConfig []RuledConfigs `json:"perNodeConfig,omitempty"`
+}
+
+// LoadAffinity is the config for data path load affinity.
+type LoadAffinity struct {
+	// NodeSelector specifies the label selector to match nodes
+	NodeSelector metav1.LabelSelector `json:"nodeSelector"`
+}
+
+type NodeAgentConfigMapSettings struct {
```
mateusoliveira43, line 385: why rewrite this instead of importing? https://github.com/vmware-tanzu/velero/blob/6337c52cfb8482cb62d1c0a3dd298f51d5f36c07/pkg/nodeagent/node_agent.go#L90-L105

internal/controller/nodeagent.go
mateusoliveira43, line 1: what about the handling of `NodeAgentCommonFields.PodConfig.NodeSelector`? that and `NodeAgentConfigMapSettings.LoadAffinity.NodeSelector` are supposed to do the same thing

mateusoliveira43, line 1: what about 1.4 configMap, should we tell user it will not be used? should we delete it?

api/v1alpha1/dataprotectionapplication_types.go
mpryc: We can't import directly otherwise we will have issues in the `zz_generated.deepcopy.go` around `DeepCopyInto`.

internal/controller/nodeagent.go
mpryc: I don't understand ?

mpryc: Yes, but my understanding was we want the CM to be same as upstream and we are moving it to the new fields? Otherwise we could have PodConfig with Affinity as well:
   https://github.com/kubernetes/kubernetes/blob/611abd3bcdeb8ee513ab7814c4ac251575e48cbd/pkg/apis/core/types.go#L3487

mpryc: The 1.4 configmap is applied later as unsupported args (no configmap), so I don't think it's issue here?

mpryc:
rebased this PR after oadp rebase

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -357,6 +359,52 @@ type NodeAgentCommonFields struct {
...
+	// LoadConcurrency is the config for data path load concurrency per node.
+	// +optional
+	LoadConcurrency *LoadConcurrency `json:"loadConcurrency,omitempty"`
+	// LoadAffinity is the config for data path load affinity.
+	// +optional
+	LoadAffinity []*LoadAffinity `json:"loadAffinity,omitempty"`
```
sseago, line 396: I don't think we need a separate field here for LoadAffinity. We already have NodeAgentConfig.PodConfig.NodeSelector, and this value (used to schedule datamover pods) needs to match the node selector for the DaemonSet. It's more error-prone if users must manage two separate fields, so it's better to just use the existing node selector value when creating the ConfigMap

mpryc: @sseago we don't need to put anything to the ConfigMap around NodeSelector right? It will apply to node via https://github.com/openshift/oadp-operator/blob/e8b6efd937f3bc60fae55b21ccc7acd9b3205743/internal/controller/nodeagent.go#L293 or am I missing something and we need to include that value in the ConfigMap as well from that setting?

mpryc: https://github.com/openshift/oadp-operator/blob/51cf09b9b8405c66edd7c176afd1f49a8fb9e73c/internal/controller/nodeagent_test.go#L1395-L1418

The above test makes it simpler to review the resulting CM for the above PodConfig, please also see one test below to see full CM with all other config options.

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -357,6 +359,45 @@ type NodeAgentCommonFields struct {
...
+	// BackupPVCConfig is the config for backupPVC (intermediate PVC) of snapshot data movement
+	// +optional
+	BackupPVCConfig map[string]nodeagent.BackupPVC `json:"backupPVC,omitempty"`
+	// RestoreVCConfig is the config for restorePVC (intermediate PVC) of generic restore
+	// +optional
+	RestorePVCConfig *RestorePVC `json:"restorePVC,omitempty"`
+	// PodResources is the resource config for various types of pods launched by node-agent, i.e., data mover pods.
+	// +optional
+	PodResources *kube.PodResources `json:"podResources,omitempty"`
+	// LoadAffinity is not required within NodeAgentConfigMapSettings,
```
mpryc, line 397: @sseago added comment here as well.

weshayutin:
I suspect the e2e test failures are legit
```
Summarizing 2 Failures:
  [FAIL] Configuration testing for DPA Custom Resource DPA reconciled to true [It] DPA CR with NodeAgent enabled with restic and node selector
  /go/src/github.com/openshift/oadp-operator/tests/e2e/dpa_deployment_suite_test.go:153
  [FAIL] Configuration testing for DPA Custom Resource DPA reconciled to true [It] DPA CR with NodeAgent enabled with kopia and node selector
  /go/src/github.com/openshift/oadp-operator/tests/e2e/dpa_deployment_suite_test.go:151
```

mpryc:
yep something br8ken, will fix it.

mpryc:
To be fixed: Restart node-agent pods if the configmap content changed.

kaovilai:
> To be fixed: Restart node-agent pods if the configmap content changed.

Is this something to be done in upstream velero? ie. velero should add configmap to watch and react.

Or do we just wanna handle in OADP?

api/v1alpha1/dataprotectionapplication_types.go
msfrucht: Is there going to be support for LoadAffinity? 

https://github.com/vmware-tanzu/velero/blob/main/pkg/nodeagent/node_agent.go#L60

That is part of normal node-config.

mpryc:
> > To be fixed: Restart node-agent pods if the configmap content changed.
> 
> Is this something to be done in upstream velero? ie. velero should add configmap to watch and react.
> 
> Or do we just wanna handle in OADP?

Currently I've handled this in the OADP as we build the DaemonSet. Don't know if it's something Velero should be messing with directly.

api/v1alpha1/dataprotectionapplication_types.go
kaovilai: `make generate`?

kaovilai: Just tried.. looks like some amount of code..
https://github.com/kaovilai/oadp-operator/commit/f5bdbb5534d9982fde7c13e6fbafd9b11d28a984

kaovilai (changes requested):

kaovilai (dismissed):

kaovilai:
nit: There's missing test cases around CM deletion upon dpa disabling of node agent. probably not required.

internal/controller/nodeagent.go
```diff
@@ -93,19 +95,124 @@ func getNodeAgentObjectMeta(r *DataProtectionApplicationReconciler) metav1.Objec
...
+		err := r.Get(r.Context, cmName, &configMap)
+		if err != nil && !errors.IsNotFound(err) {
+			return false, err
+		}
+		if errors.IsNotFound(err) {
+			return true, nil
+		}
+		deleteContext := context.Background()
+		if err := r.Delete(deleteContext, &configMap); err != nil {
+			return false, err
```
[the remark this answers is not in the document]
mpryc: @kaovilai the CM is managed by the DPA (screenshot below) and we do not have any finalizer on that ConfigMap.

I agree with your use case, it is however very corner case and I can think of other use case which will break when we switch to your logic:
 1. User creates DPA with fields that requires CM
 2. User updates DPA fields that do require CM, so now CM is not required
 3. Controller fails to delete CM due to API transient error (they do happen)
 4. Reconcile checks there was no error during Delete action, we returned `true, nil`
 5. Node-Agents are not restarted, because CM was not deleted nor updated
 6. DPA is out of sync with node-agents, CM is still in use.


image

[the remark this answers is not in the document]
mpryc: Done, misread your comment. @kaovilai you are absolutely right here.

api/v1alpha1/dataprotectionapplication_types.go
shubham-pampattiwar: Hmm, thinking about this, shouldn't keep all the same options as upstream ?

weshayutin:
ok.. toying around w/ this
DPA:
```
  configuration:
    nodeAgent:
      backupPVC:
        pvc-1:
          readOnly: true
          spcNoRelabeling: true
          storageClass: gp3-csi
      enable: true
      loadConcurrency:
        globalConfig: 3
      uploaderType: kopia
<snip>
status:
  conditions:
    - lastTransitionTime: '2025-03-21T17:34:16Z'
      message: Reconcile complete
      reason: Complete
      status: 'True'
      type: Reconciled
```

ConfigMap:
```
  labels:
    app.kubernetes.io/component: node-agent-config
    app.kubernetes.io/instance: dpa-sample
    openshift.io/oadp: 'True'
  ownerReferences:
      name: dpa-sample
      uid: a4004ab0-2d1c-40c6-8806-ac5232980408
      controller: true
      blockOwnerDeletion: true
data:
  node-agent-config: '{"backupPVC":{"pvc-1":{"readOnly":true,"spcNoRelabeling":true,"storageClass":"gp3-csi"}},"loadConcurrency":{"globalConfig":3}}'

```

internal/controller/nodeagent_test.go
```diff
@@ -1352,6 +1444,219 @@ func TestDPAReconciler_buildNodeAgentDaemonset(t *testing.T) {
...
+										{
+											NodeSelector: metav1.LabelSelector{
+												MatchLabels: map[string]string{"app": "velero-2"},
+											},
+											Number: 5,
+										},
+									},
+								},
+								BackupPVCConfig: map[string]nodeagent.BackupPVC{
+									"pvc-1": {
```
shubham-pampattiwar, line 1567: @mpryc This map key isn't the PVC name, its a storageClass name. 
I see that this is what got @weshayutin confused.

weshayutin: well to be fair, I'm easily confused.  I DO think we'll have to be careful w/ this one when handing it off to QE :)

weshayutin:
created pvc's for dataupload as readonly :)
```
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    volume.beta.kubernetes.io/storage-provisioner: ebs.csi.aws.com
    volume.kubernetes.io/selected-node: ip-10-0-48-143.us-west-2.compute.internal
    volume.kubernetes.io/storage-provisioner: ebs.csi.aws.com
  creationTimestamp: "2025-03-21T18:21:01Z"
  finalizers:
  - kubernetes.io/pvc-protection
  name: mysql-persistent-1-mrjl6
  namespace: openshift-adp
  ownerReferences:
  - apiVersion: velero.io/v2alpha1
    controller: true
    kind: DataUpload
    name: mysql-persistent-1-mrjl6
    uid: 705748b1-017e-4962-b05c-6498ecd2dac2
  resourceVersion: "1523789"
  uid: adb689f4-37ba-4424-aa67-d06fcc9a5155
spec:
  accessModes:
  - ReadOnlyMany
  dataSource:
    apiGroup: snapshot.storage.k8s.io
    kind: VolumeSnapshot
    name: mysql-persistent-1-mrjl6
  dataSourceRef:
    apiGroup: snapshot.storage.k8s.io
```

weshayutin:
```
  configuration:
    nodeAgent:
      backupPVC:
        gp3-csi:
          readOnly: false
          spcNoRelabeling: true
          storageClass: gp3-csi
```
now set to readonly: false
and dm works :) 
```
NAME                       STATUS      STARTED   BYTES DONE   TOTAL BYTES   STORAGE LOCATION   AGE   NODE
mysql-persistent-1-mrjl6   Canceled    6m43s                                dpa-sample-1       10m   
mysql-persistent-1-zdvf2   Canceled    6m43s                                dpa-sample-1       10m   
mysql-persistent-2-nmmpt   Completed   44s       1832         1832          dpa-sample-1       90s   ip-10-0-27-13.us-west-2.compute.internal
mysql-persistent-2-zwz5h   Completed   14s       108267980    108267980     dpa-sample-1       95s   ip-10-0-27-13.us-west-2.compute.internal
```

api/v1alpha1/dataprotectionapplication_types.go
kaovilai: I think @mpryc is trying to avoid exposing more than one place to set.

We can keep as a dummy one or perhaps as a higher priority one?

weshayutin:
restores are working w/ the config
```
whayutin@thinkdoe:~/OPENSHIFT/git/OADP/oadp-operator$ oc get pvc mysql-persistent-3-2xfn9 -o yaml
apiVersion: v1
kind: PersistentVolumeClaim
metadata:
  annotations:
    pv.kubernetes.io/bind-completed: "yes"
    pv.kubernetes.io/bound-by-controller: "yes"
    volume.beta.kubernetes.io/storage-provisioner: ebs.csi.aws.com
    volume.kubernetes.io/selected-node: ip-10-0-94-251.us-west-2.compute.internal
    volume.kubernetes.io/storage-provisioner: ebs.csi.aws.com
  creationTimestamp: "2025-03-21T18:41:04Z"
  finalizers:
  - kubernetes.io/pvc-protection
  labels:
    app: mysql
    velero.io/backup-name: mysql-persistent-3
    velero.io/restore-name: mysql-persistent-3
    velero.io/volume-snapshot-name: velero-mysql-7rsv6
  name: mysql-persistent-3-2xfn9
  namespace: openshift-adp
  ownerReferences:
  - apiVersion: velero.io/v2alpha1
    controller: true
    kind: DataDownload
    name: mysql-persistent-3-2xfn9
    uid: bbb6be7c-6f08-451e-93cc-ca57d0269872
  resourceVersion: "1530187"
  uid: 8f134684-5e80-4e23-b607-d5e91c5ad80f
spec:
  accessModes:
  - ReadWriteOnce
  resources:
    requests:
      storage: 1Gi
  storageClassName: gp3-csi
  volumeMode: Filesystem
  volumeName: pvc-8f134684-5e80-4e23-b607-d5e91c5ad80f
status:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 1Gi
  phase: Bound

```

weshayutin:
now in a negative test, I'm still using gp3-csi for app and backup but configured gp2-csi.. no effect should happen on pvc's

```
  configuration:
    nodeAgent:
      backupPVC:
        gp2-csi:
          readOnly: true
          spcNoRelabeling: false
          storageClass: gp2-csi
```
```
mysql-persistent-5-86x5n   Completed   15s       2290         2290          dpa-sample-1       96s     ip-10-0-94-251.u
s-west-2.compute.internal
mysql-persistent-5-nbhnc   Completed   55s       108267980    108267980     dpa-sample-1       101s    ip-10-0-94-251.u
s-west-2.compute.interna
```
woot :)

mpryc:
@weshayutin https://github.com/openshift/oadp-operator/pull/1667/commits/d1cddb421eb3415164aabc38f111c15cac4b486a

weshayutin:
> @weshayutin https://github.com/openshift/oadp-operator/pull/1667/commits/d1cddb421eb3415164aabc38f111c15cac4b486a

@mpryc yes, the key / value for sc is correct, however the top level arbitrary name for the map is more clear with the sc name then pvc1 or something more descriptive as compared to the upstream doc. Not a big deal either way

kaovilai (approved):

mpryc:
> > @weshayutin [d1cddb4](https://github.com/openshift/oadp-operator/commit/d1cddb421eb3415164aabc38f111c15cac4b486a)
> 
> @mpryc yes, the key / value for sc is correct, however the top level arbitrary name for the map is more clear with the sc name then pvc1 or something more descriptive as compared to the upstream doc. Not a big deal either way

Ok, however this is upstream name, so we can't really change it here: https://github.com/vmware-tanzu/velero/blob/d086cb2fc303d00f4e182fc2a8f3e99f7064d4b9/pkg/nodeagent/node_agent.go#L98

It uses names directly from that structure:
https://github.com/vmware-tanzu/velero/blob/d086cb2fc303d00f4e182fc2a8f3e99f7064d4b9/pkg/nodeagent/node_agent.go#L198-L204

weshayutin (approved):
/LGTM

sseago:
> > @weshayutin [d1cddb4](https://github.com/openshift/oadp-operator/commit/d1cddb421eb3415164aabc38f111c15cac4b486a)
> 
> @mpryc yes, the key / value for sc is correct, however the top level arbitrary name for the map is more clear with the sc name then pvc1 or something more descriptive as compared to the upstream doc. Not a big deal either way

BackupPVC is the configmap entry that upstream code is looking for. That distinguishes it from the other fields that are unrelated to backupPVC. Next level is storageclass name, since we could have different configurations by storageclass. Either way. changing BackupPVC to the storageclass name wouldn't make sense, even if we could modify upstream. It's a specific config key telling velero "this is configuration for creating backup PVCs which is specified per-storage-class"

PR #1676, opened 2025-03-20
OADP-5782: Add hypershift-oadp-plugin as a default one in oadp-operator

jparrill:
## Why the changes were made

This allows the [hypershift-oadp-plugin](https://github.com/openshift/hypershift-oadp-plugin) to be part of the default ones delivered into oadp-operator. This also give supportability to the DR scenarios to HCP using OADP as a provider.

Bumping some dependencies too:
- github.com/openshift/api v0.0.0-20240524162738-d899f8877d22
- github.com/google/pprof v0.0.0-20241210010833-40e02aabc2ad // indirect
- github.com/spf13/pflag v1.0.6-0.20210604193023-d5e0c0615ace // indirect

1. Modify the DPA adding the `hypershift` default plugin
2. Perform a backup/restore using the samples from the [hypershift-oadp-plugin](https://github.com/openshift/hypershift-oadp-plugin/tree/main/examples) repository

### DIFF
--- a/api/v1alpha1/dataprotectionapplication_types.go
+++ b/api/v1alpha1/dataprotectionapplication_types.go
@@ -36,7 +36,7 @@ const ReconcileCompleteMessage = "Reconcile complete"
 
 const OadpOperatorLabel = "openshift.io/oadp"
 
-// +kubebuilder:validation:Enum=aws;legacy-aws;gcp;azure;csi;vsm;openshift;kubevirt
+// +kubebuilder:validation:Enum=aws;legacy-aws;gcp;azure;csi;vsm;openshift;kubevirt;hypershift
 type DefaultPlugin string
 
 const DefaultPluginAWS DefaultPlugin = "aws"
@@ -47,6 +47,7 @@ const DefaultPluginCSI DefaultPlugin = "csi"
 const DefaultPluginVSM DefaultPlugin = "vsm"
 const DefaultPluginOpenShift DefaultPlugin = "openshift"
 const DefaultPluginKubeVirt DefaultPlugin = "kubevirt"
+const DefaultPluginHypershift DefaultPlugin = "hypershift"
 
 type CustomPlugin struct {
 	Name  string `json:"name"`
@@ -71,6 +72,7 @@ const AzurePluginImageKey UnsupportedImageKey = "azurePluginImageFqin"
 const GCPPluginImageKey UnsupportedImageKey = "gcpPluginImageFqin"
 const ResticRestoreImageKey UnsupportedImageKey = "resticRestoreImageFqin"
 const KubeVirtPluginImageKey UnsupportedImageKey = "kubevirtPluginImageFqin"
+const HypershiftPluginImageKey UnsupportedImageKey = "hypershiftPluginImageFqin"
 const NonAdminControllerImageKey UnsupportedImageKey = "nonAdminControllerImageFqin"
 const OperatorTypeKey UnsupportedImageKey = "operator-type"
 
@@ -713,6 +715,7 @@ type DataProtectionApplicationSpec struct {
 	//   - gcpPluginImageFqin
 	//   - resticRestoreImageFqin
 	//   - kubevirtPluginImageFqin
+	//   - hypershiftPluginImageFqin
 	//   - nonAdminControllerImageFqin
 	//   - operator-type
 	//   - tech-preview-ack
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -1078,6 +1078,8 @@ spec:
                   value: quay.io/konveyor/velero-plugin-for-gcp:latest
                 - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
                   value: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
+                - name: RELATED_IMAGE_HYPERSHIFT_VELERO_PLUGIN
+                  value: quay.io/hypershift/hypershift-oadp-plugin:latest
                 - name: RELATED_IMAGE_MUSTGATHER
                   value: registry.redhat.io/oadp/oadp-mustgather-rhel8:v1.2
                 - name: RELATED_IMAGE_NON_ADMIN_CONTROLLER
@@ -1236,6 +1238,8 @@ spec:
     name: velero-plugin-for-gcp
   - image: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
     name: kubevirt-velero-plugin
+  - image: quay.io/hypershift/hypershift-oadp-plugin:latest
+    name: hypershift-velero-plugin
   - image: registry.redhat.io/oadp/oadp-mustgather-rhel8:v1.2
     name: mustgather
   - image: quay.io/konveyor/oadp-non-admin:latest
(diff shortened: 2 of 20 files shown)


rayfordj:
Should add the default RELATED_IMAGE like [this](https://github.com/openshift/oadp-operator/blob/master/config/manager/manager.yaml#L77-L78) 
```
            - name: RELATED_IMAGE_HYPERSHIFT_VELERO_PLUGIN
              value: quay.io/konveyor/hypershift-velero-plugin:v0.1.0
```
which should then get added to the [CSV](https://github.com/openshift/oadp-operator/blob/master/bundle/manifests/oadp-operator.clusterserviceversion.yaml#L1217-L1236) after executing _`make bundle`_

config/manager/manager.yaml
mateusoliveira43, line 1: after changing a file in `config/` folder, you need to run `make bundle` to update `bundle/` folder

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -45,6 +45,7 @@ const DefaultPluginCSI DefaultPlugin = "csi"
 const DefaultPluginVSM DefaultPlugin = "vsm"
 const DefaultPluginOpenShift DefaultPlugin = "openshift"
 const DefaultPluginKubeVirt DefaultPlugin = "kubevirt"
+const DefaultPluginHypershift DefaultPlugin = "hypershift"
```
mateusoliveira43, line 48: I think this is still not testable

CRD validation does not include `hypershift` as an option (line 37 in this file. After changing it, you need to run `make bundle` to update YAML files)

shubham-pampattiwar:
@jparrill Is this PR ready for review ?

jparrill:
@shubham-pampattiwar Nope, I need to work on the E2E yet. I just want to bring the PR up to have some reviews and keep enhancing the code accordingly. Thanks for asking :)!

config/manager/manager.yaml
```diff
@@ -76,6 +76,8 @@ spec:
               value: quay.io/konveyor/velero-plugin-for-gcp:latest
             - name: RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN
               value: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
+            - name: RELATED_IMAGE_HYPERSHIFT_VELERO_PLUGIN
+              value: quay.io/hypershift/hypershift-oadp-plugin:latest
```
mateusoliveira43, line 80: should we used fixed tag as kubevirt?

jparrill: not for now I think, let's roll this out and once I have all the issues fixed in the rest of the providers I will create the first release

rayfordj:
From my perspective, for midstream/downstream pieces, 'lgtm'; leaving the overall acks/approvals to the broader team.  :sunglasses:

rayfordj:
[Slack #forum-oadp Thread](https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1741625011483529) for bulk of conversation outside of this PR.

/lgtm

go.mod
```diff
@@ -11,7 +11,7 @@ require (
 	github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0
 	github.com/onsi/ginkgo/v2 v2.19.0
 	github.com/onsi/gomega v1.33.1
-	github.com/openshift/api v0.0.0-20230213134911-7ba313770556 // release-4.12
+	github.com/openshift/api v0.0.0-20240524162738-d899f8877d22 // release-4.12
```
kaovilai, line 14: if you had to bump this, please update comments accordingly if this commit is from a certain release-4.x branch.

kaovilai (changes requested):

go.mod
kaovilai: https://github.com/openshift/api/commit/d899f8877d22
This is coming from master..

kaovilai: matches release-4.19 for the time being.. if that differ in the future, or when we do oadp-1.5 branching, we should use the stable branches.

kaovilai (approved):

pkg/credentials/credentials.go
```diff
@@ -159,6 +163,16 @@ func getKubeVirtPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string
 	return os.Getenv("RELATED_IMAGE_KUBEVIRT_VELERO_PLUGIN")
 }
 
+func getHypershiftPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string {
+	if dpa.Spec.UnsupportedOverrides[oadpv1alpha1.HypershiftPluginImageKey] != "" {
+		return dpa.Spec.UnsupportedOverrides[oadpv1alpha1.HypershiftPluginImageKey]
+	}
```
kaovilai, line 169: There is no unit test for this override. Can follow up.

jparrill: Noted for the follow up PR

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -1236,6 +1238,8 @@ spec:
     name: velero-plugin-for-gcp
   - image: quay.io/konveyor/kubevirt-velero-plugin:v0.7.0
     name: kubevirt-velero-plugin
+  - image: quay.io/hypershift/hypershift-oadp-plugin:latest
```
shubham-pampattiwar, line 1241: Your plugin repo name is `hypershift-oadp-plugin` , so as to be consistent shouldn't the name be `hypershift-oadp-plugin` too ?

pkg/credentials/credentials.go
```diff
@@ -159,6 +163,16 @@ func getKubeVirtPluginImage(dpa *oadpv1alpha1.DataProtectionApplication) string
...
+	if os.Getenv("RELATED_IMAGE_HYPERSHIFT_VELERO_PLUGIN") == "" {
+		return common.HypershiftPluginImage
+	}
+	return os.Getenv("RELATED_IMAGE_HYPERSHIFT_VELERO_PLUGIN")
```
shubham-pampattiwar, line 173: same comment for all the places.

rayfordj: @shubham-pampattiwar , I would rather instead advocate to change the plugin repo name. :pleading_face: 

A lot of downstream work has been completed based on [`oadp-hypershift-velero-plugin-container`](https://redhat-internal.slack.com/archives/C0144ECKUJ0/p1741773839662159?thread_ts=1741625011.483529&cid=C0144ECKUJ0) proposed in Slack (Mar 12th) without objections and subsequently discussed in Scrum (Mar 13th) with consensus for the name which was then fed back ([for example](https://github.com/openshift/oadp-operator/pull/1676#issuecomment-2741460086)) to ongoing efforts in github -- primarily, this PR.  :sweat_smile:

jparrill: I was thinking that both (dowstream images and the repo name with upstream images) were fully separated, is that the case? should I rename all the things based on `oadp-hypershift-velero-plugin-container`?

rayfordj: They are indeed distinct & separate items.  The only real concern is getting the plumbing correct, so we override/inject/replace appropriately as things make their way from upstream through midstream and land in downstream to then produce official RH image build(s).  

As it stands, all the plumbing should be correct (as expected) with this PR merged.
I wouldn't worry too much about renaming anything at the moment while other, arguably more important, work is outstanding.  :sunglasses:

weshayutin:
@jparrill @kaovilai so will this cherry pick to oadp-1.5??  Or is that a fresh PR?

kaovilai:
Once working cherry pick one by one imo.. but either works

PR #1679, opened 2025-03-24
Ensure root fs in the controller manager is readonly

mpryc:
Fixes [OADP-5275](https://issues.redhat.com//browse/OADP-5275) which ensures the controller manager has readOnlyRootFilesystem set to true.

## Why the changes were made

Fix [OADP-5275](https://issues.redhat.com//browse/OADP-5275)

1. Deploy from the source
2. Ensure root fs in the `openshift-adp-controller-manager-*` pod is read only:
```shell
sh-5.1$ touch /testreadonly
touch: cannot touch '/testreadonly': Read-only file system

sh-5.1$ whoami
1000650000
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -87,7 +87,11 @@ spec:
           name: manager
           terminationMessagePolicy: FallbackToLogsOnError
           securityContext:
+            capabilities:
+              drop:
+              - ALL
             allowPrivilegeEscalation: false
+            readOnlyRootFilesystem: true
           startupProbe:
             httpGet:
               path: /healthz
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -1106,6 +1106,10 @@ spec:
                     memory: 128Mi
                 securityContext:
                   allowPrivilegeEscalation: false
+                  capabilities:
+                    drop:
+                    - ALL
+                  readOnlyRootFilesystem: true
                 startupProbe:
                   failureThreshold: 12
                   httpGet:


sseago (approved):

shubham-pampattiwar (approved):

config/manager/manager.yaml
```diff
@@ -87,7 +87,11 @@ spec:
           name: manager
           terminationMessagePolicy: FallbackToLogsOnError
           securityContext:
+            capabilities:
```
mateusoliveira43, line 90: from JIRA, adding only `readOnlyRootFilesystem` should be sufficient, no? why also add this capabilities?

shubham-pampattiwar: I think @mpryc is trying to enforce the _least privilege principle._

mpryc: Yes. Adding `readOnlyRootFilesystem` is enough to fix the issue, but removing `capabilities` is a bit more into the direction on having minimal privilege required without harm on the pod functionality.

mateusoliveira43: ok

mateusoliveira43 (approved):

config/manager/manager.yaml
mpryc: With the current settings our container runs as `restricted-v2`:

https://docs.redhat.com/en/documentation/openshift_container_platform/4.18/html/authentication_and_authorization/managing-pod-security-policies#default-sccs_configuring-internal-oauth

```text
Note
The restricted-v2 SCC is the most restrictive of the SCCs that is included by default with the system. 
```

kaovilai (approved):

PR #1683, opened 2025-03-27
Node Selector Configuration for Data Protection Components - NodeAgent and DataMover

mpryc:
This proposal introduces a standardized approach for configuring node selectors in the DataProtectionApplication (DPA) custom resource for various use cases that are common across different Velero settings.

It implements the changes to fix [OADP-5832](https://issues.redhat.com//browse/OADP-5832)

## Why the changes were made

To propose design to fix the [OADP-5832](https://issues.redhat.com//browse/OADP-5832).

Current configuration options are limited and do not allow precisely schedule pods using Affinity and Anti-Affinity patterns.

### NodeAgent and DataMover pod placement

1. Prepare cluster with the nodes that are labelled, examples below are based on the IPs from the first command, so adjust to your cluster, but the idea is to have combinatorial covering of different scenarios for the nodes:

```shell
$ oc get nodes
NAME          STATUS   ROLES           AGE   VERSION
10.243.64.4   Ready    master,worker   10d   v1.30.10
10.243.64.5   Ready    master,worker   10d   v1.30.10
10.243.64.6   Ready    master,worker   10d   v1.30.10

$ ND1=10.243.64.4 
$ ND2=10.243.64.5
$ ND3=10.243.64.6

# Let's add labels which allows to mix different settings making this a matrix of simple labels
$ oc label node "$ND1" test.io/single=A
$ oc label node "$ND2" test.io/single=B
$ oc label node "$ND3" test.io/single=C

$ oc label node "$ND1" test.io/twinsD=D
$ oc label node "$ND2" test.io/twinsD=D

$ oc label node "$ND1" test.io/twinsE=E
$ oc label node "$ND3" test.io/twinsE=E

$ oc label node "$ND2" test.io/twinsF=F
$ oc label node "$ND3" test.io/twinsF=F


$ oc label node "$ND1" test.io/triplets=G
$ oc label node "$ND2" test.io/triplets=G
$ oc label node "$ND3" test.io/triplets=G
```

2. With the two DPA settings that controls the node-agent scheduling cover different scenarios with the simple label selector or with the `matchExpressions`. The scenarios should cover only `podConfig`, only `loadAffinity` with different selectors and both fields specified. For the ideas on what scenarios should be allowed and disallowed please refer to the validation tests from within this PR and the 3rd point below.

```yaml

spec:
  configuration:
    nodeAgent:
      enable: true
      podConfig:
        nodeSelector:
           [...]

spec:
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchLabels:
              [...]
```
4. Scenarios to consider
After each scenario the following parts should be validated:
 - Is ConfigMap in the openshift-adp namespace properly created and does it contain expected fields from the loadAffinity. The ConfigMap name correspods to the BSL or DPA name with `node-agent-` prefix, example on how to nicely get the CM:
```shell
$ oc get cm -n openshift-adp node-agent-velero-sample -o json | jq -r '.data["node-agent-config"]' | jq .
{
  "loadConcurrency": {
    "globalConfig": 1
  },
  "loadAffinity": [
    {
      "nodeSelector": {
        "matchLabels": {
          "test.io/twinsD": "D"
        }
      }
    }
  ]
}
```
 - Check if actually CM is passed to all the node-agents running in the cluster the below command comes handy:
```shell
$ oc get pods -n openshift-adp -l role=node-agent -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{.spec.containers[?(@.name=="node-agent")].command}{" "}{.spec.containers[?(@.name=="node-agent")].args}{"\n\n"}{end}'
node-agent-9mtvm
["/velero"] ["node-agent","server","--node-agent-configmap=node-agent-velero-sample","--log-format=text"]

node-agent-vv754
["/velero"] ["node-agent","server","--node-agent-configmap=node-agent-velero-sample","--log-format=text"]

``` 
 - One podConfig.nodeSelector label specified with matching loadAffinity.nodeSelector.matchLabels, check if node-agent pods are scheduled there:
```shell
$ oc get pods -n openshift-adp -l role=node-agent -o jsonpath='{range .items[*]}{.metadata.name}{" on node: "}{.spec.nodeName}{"\n"}{end}'
node-agent-9mtvm on node: 10.243.64.4
node-agent-vv754 on node: 10.243.64.5
```
As you can see the above were scheduled using CM with matchLabels `test.io/twinsA: D` so they got scheduled on those two nodes which we labelled them before.

 - Confirm the node-agent DaemonSet has proper nodeSelector and Affinity settings:
 ```shell
$ oc get ds node-agent -n openshift-adp -o jsonpath='{.spec.template.spec.nodeSelector}' | jq
{
  "test.io/twinsD": "D"
}

$ oc get ds node-agent -n openshift-adp -o jsonpath='{.spec.template.spec.affinity}' | jq
{
  "nodeAffinity": {
    "requiredDuringSchedulingIgnoredDuringExecution": {
      "nodeSelectorTerms": [
        {
          "matchExpressions": [
            {
              "key": "test.io/twinsD",
              "operator": "In",
              "values": [
                "D"
              ]
            }
          ]
        }
      ]
    }
  }
}
```

Note: The above is just one simple example of how a single label can be configured, the matrix of different settings and scheduling is much wider and should consists of mixture of multiple labels in one and another field. Also the `matchExpressions` can be used in the `loadAffinity` settings to create Affinity and Anti-Affinity rules. You can use the `matchExpressions` field a logical operator when interpreting the rules. You can use In, NotIn, Exists, DoesNotExist, Gt and Lt. Example of such Anti-Affinity rule where we don't want to schedule node-agent nor the DataMover workload in the nodes where the `test.io/twinsE=E` label exists. To confirm please use steps from the previous point **After each scenario the following parts should be validated**:

```yaml
spec
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchExpressions:
              - key: test.io/twinsE
                operator: 'NotIn'
                values:
                  - "E"
```


sseago (approved):

docs/design/DPA-and-node-selectors.md
```diff
@@ -0,0 +1,253 @@
...
+## High-Level Design
+The proposed change introduces a unified nodeSelector structure for all relevant components within the DPA custom resource.
+This will allow users to specify node selection using either matchLabels or matchExpressions, ensuring flexibility while maintaining consistency across configurations.
+
+The modified structure will be applied to the new DPA sections:
+
+- `configuration.nodeAgent.loadConcurrency`
+- `configuration.nodeAgent.loadAffinity`
+- `configuration.velero.loadAffinity`
+- `configuration.repositoryMaintenance`
```
mateusoliveira43, line 36: repositoryMaintenance only happens if using nodeagent, right? should it be inside `configuration.nodeAgent`?

mpryc: https://velero.io/docs/v1.15/repository-maintenance/#resource-limitation-and-node-affinity

mpryc: Uses similar to node-agent, but it's very different.

mpryc: Here is doc for the loadAffinity node-agent:
   https://velero.io/docs/main/data-movement-backup-node-selection/#affinity

And here is doc for loadConcurrency:
  https://velero.io/docs/main/node-agent-concurrency/

mpryc: velero nodeAffinity - I may not implement now but I do prefer to have it covered in the design doc for complete overview.

mpryc: hmm but maybe @mateusoliveira43 you are right when I thing about it, but would it be very confusing to have:

```
- `configuration.nodeAgent.loadConcurrency`
- `configuration.nodeAgent.loadAffinity`
- `configuration.repositoryMaintenance.global.loadAffinity`
- `configuration.repositoryMaintenance.<backup repository  key>.loadAffinity`
```

I mean very different loadAffinity fields - it needs to be separated from the nodeAgent for the repository maintenance to ensure users can have different nodes for those.

Of course we are talking here only about kopia as restic is deprecated, right?

mpryc:
Please hold off with reviewing this Design. I am currently looking for additional option to actually use Velero scheduler for the datamover and re-use this in the node-agent deployment, which would limit the nodes on which node-agent is running as well. For that I need to run couple of tests and see how repo maintenance would behave outside of node-agent scheduler.

With the above we would have node-agent node selector aligned with the datamover scheduler and separately scheduler for the repo maintenance.

docs/design/DPA-and-node-selectors.md
mpryc: I've updated the design doc and explain why I think repositoryMaintenance should NOT be part of nodeAgent (in short it can be scheduled on very different node then NodeAgent which is not required to run those jobs.).

mpryc:
I am marking this as not in draft anymore. Please note this implementation is only to fix OADP-5832 without repository nor velero affinity. If we want to add those (and possibly we do) it should imo be as separate PR to not make things more complex then they are now.

mpryc:
@shubham-pampattiwar @sseago @mateusoliveira43 @weshayutin FYI ^

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -344,6 +344,8 @@ type PodConfig struct {
 	// labels to add to pods
 	// +optional
 	Labels map[string]string `json:"labels,omitempty"`
+	// (do not use warning) nodeSelector field is for backwards compatibility and
```
mateusoliveira43, line 347: this is also used for velero deployment/pod configuration. Do we want to not allow use there as well?

mateusoliveira43: reading design seems a new field will be added to velero config, but it is still not there, right?

I think we should add both this "deprecation" message and velero change in same PR

mpryc: Repository Maintenance and Velero will be as separate PRs as this one is already too complex imo, so don't want to add even more to it. Repo Maintenance PR https://github.com/openshift/oadp-operator/pull/1688 based on this one, so this one needs to be merged first.

mpryc: Actually I will remove this (do not use warning) because it only applies to the node-agent and we may not want to even remove it. It's explained in the design.

docs/design/DPA-and-node-selectors.md
```diff
@@ -0,0 +1,265 @@
+# Node Selector Configuration for Data Protection Components
+
+## Abstract
+This proposal introduces a standardized approach for configuring node selectors in the DataProtectionApplication (DPA) custom resource for various use cases that are common across different Velero settings.
+
+## Background
+Currently, node selection for Velero and NodeAgent components is inconsistently defined within the DPA specification.
+
+- `nodeAgent.podConfig.nodeSelector` specifies nodes using matchLabels. This setting is used to define in which nodes the node-agent pod will be scheduled.
+- `loadConcurrency.perNodeConfig` supports both matchLabels and matchExpressions. Node-agent concurrency configurations allows to configure the concurrent number of node-agent loads per node, but only for the nodes where the node-agent pod is scheduled.
```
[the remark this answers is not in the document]
mpryc: Done

api/v1alpha1/dataprotectionapplication_types.go
mpryc: I have also updated the design document to make the Velero component optional, noting that we may implement it in the future if needed. This decision is based on the fact that PodConfig.NodeSelector should generally be sufficient. Since the Velero pod itself doesn’t perform actual work but primarily manages operations, and only one pod exists per deployment, simple label matching should be adequate for scheduling.

docs/design/DPA-and-node-selectors.md
```diff
@@ -0,0 +1,270 @@
...
+- `spec.configuration.nodeAgent.loadConcurrency.perNodeConfig` supports both matchLabels and matchExpressions. Node-agent concurrency configurations allows to configure the concurrent number of node-agent loads per node, but only for the nodes where the node-agent pod is scheduled.
+- `spec.configuration.velero.podConfig.nodeSelector` specifies node selector for the velero pod.
+
+The NodeAgent pod is either scheduled on every node in the cluster or limited to specific nodes where the NodeAgent pod is scheduled via `spec.configuration.nodeAgent.podConfig.nodeSelector` setting.
+
+This design enhances flexibility by allowing Affinity and Anti-Affinity settings for scheduling NodeAgent pods with their DataMover workloads or repository maintenance pods.
+
+## Goals
+- Provide a clear and structured approach for users to define node selection criteria for the NodeAgent, DataMover and repository maintenance pods using the DPA specification.
+- Support both `matchLabels` and `matchExpressions` for flexible node selection for the NodeAgent, DataMover and repository maintenance pods.
```
shubham-pampattiwar, line 19: Any specific reason due to which we are not targeting support of `matchLabels` and `matchExpressions` for the Velero Pod ? I think we definitely want this. We have requests for Velero Pod too. For e.g: Avoid scheduling to GPU nodes.

mpryc: Discussed this offline with Wes, but we may change this. There is only one velero pod per deployment and multiple other pods that do the work. The second is no brainer - we should have Affinity and Anti-Affinity. In the Velero pod itself - one per deployment my take is that it's sufficient to have 1-1 mapping and mark node with a specific label. Pls correct me if I am wrong.

shubham-pampattiwar: yeah I get what you are saying. I just think this is the right moment to bring uniform support for all the pods that we create via OADP.

mpryc: Ok I will implement that as separate PR similarly to separate PR for the repository maintenance. In that PR I will update design to remove it's possible in the future. ok?

mpryc: @shubham-pampattiwar without tests and manual testing, but here is something that should work for the velero pod: https://github.com/openshift/oadp-operator/pull/1690

mpryc: @shubham-pampattiwar sorry this is full on the few other commits, here is the separate commit just for the velero pod affinity: https://github.com/openshift/oadp-operator/pull/1690/commits/4b58767ae38bf18ed60084d47dcc620470c4099c

weshayutin:
TWINS
```
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
      - nodeSelector:
          matchLabels:
            test.io/twinsD: D
```
```
 oc get daemonsets.apps node-agent 
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   2         2         2       2            2           kubernetes.io/os=linux   12m
```
TRIPLETS
```
 configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchLabels:
              test.io/triplets: G
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   3         3         2       3            2           kubernetes.io/os=linux   15m
```
SINGLE
```
 configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchLabels:
              test.io/single: A
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   1         1         1       1            1           kubernetes.io/os=linux   17m
```
NONE
```
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchLabels:
              test.io/single: Z
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   0         0         0       0            0           kubernetes.io/os=linux   18m
```

weshayutin:
```
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchExpressions:
              - key: test.io/twinsE
                operator: NotIn
                values:
                  - E
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   1         1         1       1            1           kubernetes.io/os=linux   26m
```   
```
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchExpressions:
              - key: test.io/triplets
                operator: NotIn
                values:
                  - G
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   0         0         0       0            0           kubernetes.io/os=linux   27m
```

```
  configuration:
    nodeAgent:
      enable: true
      loadAffinity:
        - nodeSelector:
            matchExpressions:
              - key: test.io/triplets
                operator: NotIn
                values:
                  - Z
```
```
oc get daemonsets.apps node-agent
NAME         DESIRED   CURRENT   READY   UP-TO-DATE   AVAILABLE   NODE SELECTOR            AGE
node-agent   3         3         3       3            3           kubernetes.io/os=linux   28m
```

weshayutin:
positive and negative cases check out.. I'm able to allow or disallow by label.. very nice work here @mpryc

weshayutin (approved):

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -387,30 +406,25 @@ type RuledConfigs struct {
...
-type LoadConcurrency struct {
-	// GlobalConfig specifies the concurrency number to all nodes for which per-node config is not specified
-	GlobalConfig int `json:"globalConfig,omitempty"`
-
-	// PerNodeConfig specifies the concurrency number to nodes matched by rules
-	PerNodeConfig []RuledConfigs `json:"perNodeConfig,omitempty"`
-}
-
+// NodeAgentConfigMapSettings is the config for node-agent
+// Struct should be same as:
```
mateusoliveira43, line 410: would move this comment to another place, to avoid it being propagated to yaml description

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -374,11 +374,30 @@ type NodeAgentCommonFields struct {
...
+	PerNodeConfig []RuledConfigs `json:"perNodeConfig,omitempty"`
 }
 
+// LoadAffinity is the config for data path load affinity.
+// Used by the Node-Agent, that needs to match the DataMover and the RepositoryMaintenance pods.
+// Struct should be same as:
+// https://github.com/openshift/velero/blob/584cf1148a746838ee67aa27e3e4e0ded1f5c069/pkg/nodeagent/node_agent.go#L60-L63
+type LoadAffinity struct {
+	// NodeSelector specifies the label selector to match nodes
+	// +optional
```
mateusoliveira43, line 394: this is not optional in velero side. Can this be a problem?

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -374,11 +374,30 @@ type NodeAgentCommonFields struct {
...
+	GlobalConfig int `json:"globalConfig,omitempty"`
+
+	// PerNodeConfig specifies the concurrency number to nodes matched by rules
+	PerNodeConfig []RuledConfigs `json:"perNodeConfig,omitempty"`
 }
 
+// LoadAffinity is the config for data path load affinity.
+// Used by the Node-Agent, that needs to match the DataMover and the RepositoryMaintenance pods.
+// Struct should be same as:
+// https://github.com/openshift/velero/blob/584cf1148a746838ee67aa27e3e4e0ded1f5c069/pkg/nodeagent/node_agent.go#L60-L63
```
mateusoliveira43, line 391: it should actually points to this https://github.com/openshift/velero/blob/584cf1148a746838ee67aa27e3e4e0ded1f5c069/pkg/nodeagent/node_agent.go#L95 (which is the same object, do not know why they duplicated it)

api/v1alpha1/dataprotectionapplication_types.go
mpryc: In which Velero CRD are you seeing this as exposed non-optional filed?

api/v1alpha1/dataprotectionapplication_types.go
mpryc: This is not a problem for the NodeAgentConfigMapSettings, because it's included inline, so this comment is not propagated to yaml, but you are right other similar comments that are not included inline should be moved out. Will change this.

api/v1alpha1/dataprotectionapplication_types.go
mateusoliveira43: not CRD (this is used only to read configMap, right?), but on velero struct it does not have optional marker or omitempty json tag

api/v1alpha1/dataprotectionapplication_types.go
mpryc: Yeah it was moved in upstream to use from kube https://github.com/openshift/velero/commit/26cc41f26daad20bd9a2c87ae408c240b28e1165#diff-314c2d754eaafb3e32404e38549140072c0b258108b1befe8804c6ca5978bdb0

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -848,6 +862,23 @@ func (dpa *DataProtectionApplication) AutoCorrect() {
 
 	dpa.Spec.Configuration.Velero.DefaultPlugins = common.RemoveDuplicateValues(dpa.Spec.Configuration.Velero.DefaultPlugins)
 	dpa.Spec.Configuration.Velero.FeatureFlags = common.RemoveDuplicateValues(dpa.Spec.Configuration.Velero.FeatureFlags)
+
+	// Auto correct nodeAffinity for the node agent, but only if the new schema is not used
+	// The new schema will be used instead of the dpa.Spec.Configuration.NodeAgent.PodConfig
+	// There is need to translate map of labels to a LabelSelector with MatchLabels
+	if dpa.Spec.Configuration.NodeAgent != nil && dpa.Spec.Configuration.NodeAgent.PodConfig != nil && dpa.Spec.Configuration.NodeAgent.PodConfig.NodeSelector != nil {
```
mateusoliveira43, line 869: I was worried about this

example: 
> I create DPA using old style (NodeAgent.PodConfig.NodeSelector), code will make me use new style NodeAgent.LoadAffinityConfig
>
> But once I update DPA NodeAgent.PodConfig.NodeSelector, NodeAgent.LoadAffinityConfig will get out of date, right? and then node agent will not have node selector user wants. Can this create problems?

but the validator controller errors if both objects are not equal, right?

internal/controller/validator_test.go
```diff
@@ -1998,6 +1998,335 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
...
+								},
+							},
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "[valid] All Labels from the LoadAffinityConfig are present in the PodConfig, and the PodConfig has more labels than the LoadAffinityConfig",
```
mateusoliveira43, line 2062: is not this an expected error? related to https://github.com/openshift/oadp-operator/pull/1683/files#r2035458769

mpryc: No, one is a subset of other.

api/v1alpha1/dataprotectionapplication_types.go
mateusoliveira43: but if user update is to add a new label to old field, then ConfigMap will not reflect what it is expecting, right? like the test I got worried abot

mpryc: No, the test involves combining both old and new configurations.

In this path we're only focusing on PodConfig.

Could you please provide the steps to reproduce the potential issue you're thinking about, so I can better understand what you're referring to, as I might have overlooked something.

api/v1alpha1/dataprotectionapplication_types.go
mpryc: Checked this and it is correctly pointing to:
https://github.com/openshift/velero/blob/584cf1148a746838ee67aa27e3e4e0ded1f5c069/pkg/nodeagent/node_agent.go#L60-L63

We can't use directly the L95 due to the fact the type *kube.LoadAffinity has no field or method DeepCopyInto.

api/v1alpha1/dataprotectionapplication_types.go
mateusoliveira43: Could not reproduce problem in cluster

internal/controller/nodeagent.go
```diff
@@ -470,6 +461,7 @@ func (r *DataProtectionApplicationReconciler) customizeNodeAgentDaemonset(ds *ap
 	if dpa.Spec.Configuration.NodeAgent.PodConfig != nil {
 		ds.Spec.Template.Spec.Tolerations = dpa.Spec.Configuration.NodeAgent.PodConfig.Tolerations
 		if len(dpa.Spec.Configuration.NodeAgent.PodConfig.NodeSelector) != 0 {
```
mateusoliveira43, line 463: do we need to remove this logic of adding NodeSelector to DaemonSet?

new configMap approach does not take care of this case?

api/v1alpha1/dataprotectionapplication_types.go
mpryc: I was testing this scenario with updating and for me was ok, we are using patch of configmap. We should be good here, right?

internal/controller/validator_test.go
mateusoliveira43: I am still confused why this is not an error

internal/controller/nodeagent.go
mpryc: There are two problems we need to solve:
 1. We want to limit where the node-agent is scheduled
 2. We need to limit where the data mover pod is scheduled

The second must run on the node where node-agent is scheduled, that is why the second must be a subset of first.

Two of those are very much separate workflows and logic, because node-agent process uses CM and the node-agent pod uses affinity within DS.
We cannot have one without another, so now we do not want to remove it.

internal/controller/validator_test.go
mpryc: @mateusoliveira43 you are very much correct here - the condition should be vice-versa. I will add separate test to cover both conditions.

The labels within PodConfig are using AND operator, so it's a subset but other way around. The LoadAffinityConfig may have more labels then node-agent.

```
[valid] All Labels from the PodConfig are present in the LoadAffinityConfig, and the LoadAffinityConfig has more labels than the PodConfig

[invalid] All Labels from the LoadAffinityConfig are present in the PodConfig, and the PodConfig has more labels than the LoadAffinityConfig
```

mpryc: Should be fixed in the new commit (made it separate commit for easier review)

sseago (approved):

weshayutin (approved):

shubham-pampattiwar (approved):

weshayutin (approved):
/LGTM

PR #1689, opened 2025-04-02
Updating image tags for oadp-1.5 branch

sseago:
velero-restore-helper is not updated because it will be removed in a follow-on PR


Makefile
```diff
@@ -3,7 +3,7 @@
 # To re-generate a bundle for another specific version without changing the standard setup, you can:
 # - use the VERSION as arg of the bundle target (e.g make bundle VERSION=0.0.2)
 # - use environment variables to overwrite this value (e.g export VERSION=0.0.2)
-DEFAULT_VERSION := 99.0.0
+DEFAULT_VERSION := 1.5.0
 VERSION ?= $(DEFAULT_VERSION)
```
shubham-pampattiwar, line 8: Also update CHANNELS and DEFAULT_CHANNEL to stable-1.5 ?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
mateusoliveira43, line 1: need to update this as well, no? https://github.com/openshift/oadp-operator/blob/master/pkg/common/common.go#L68

Makefile
mateusoliveira43: I think we want stable here to match PROW and prod

sseago: @shubham-pampattiwar Should just be stable, since we're using stable for this release rather than stable-1.5 -- we had a discussion yesterday on slack about that. @weshayutin Right?

bundle/manifests/oadp-operator.clusterserviceversion.yaml
shubham-pampattiwar: yeah we stopped updating those afaict. CSV usually has the `RELATED_IMAGES_*`

sseago: @mateusoliveira43 we didn't for oadp-1.4: https://github.com/openshift/oadp-operator/blob/oadp-1.4/pkg/common/common.go#L64

If I'm remembering correctly, those are just defaults, overridden by env vars/Makefile if present there.

Makefile
rayfordj: `stable`

kaovilai: Yes. `stable` here reflects shipping channel. Although this really only affects `make deploy-olm` operator.

shubham-pampattiwar (approved):

weshayutin (approved):
/LGTM

kaovilai (approved):

PR #1695, opened 2025-04-04
remove velero-restore-helper image

sseago:
Remove velero-restore-helper image references and
delete fs-restore-action-config configmap if present.

## Why the changes were made

As of Velero 1.16, the velero image includes the restore helper binary, so no separate container image is required. By default, Velero uses the same image for the restore helper initcontainer as is used for velero. As a result, we no longer need the configmap to point to a separate image.

Verify that with this change, the `fs-restore-action-config` configmap no longer exists in the oadp namespace.

Additionally, when restoring via fs-backup, verify that the restore-wait initcontainer uses the regular velero image and not the velero-restore-helper image:
```
    initContainers:
    - args:
      - 24e069f4-fbaf-4952-a5c2-9df3890870dd
      command:
      - /velero-restore-helper
      env:
      - name: POD_NAMESPACE
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.namespace
      - name: POD_NAME
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.name
      image: quay.io/konveyor/velero:latest
      imagePullPolicy: Always
      name: restore-wait
```


sseago:
I also verified that if I create the `fs-restore-action-config` configmap, the oadp operator removes it, which means that on upgrade, this cleanup should also happen properly.

internal/controller/nodeagent.go
```diff
@@ -628,56 +628,25 @@ func (r *DataProtectionApplicationReconciler) customizeNodeAgentDaemonset(ds *ap
 }
 
 func (r *DataProtectionApplicationReconciler) ReconcileFsRestoreHelperConfig(log logr.Logger) (bool, error) {
```
mateusoliveira43, line 630: would add a comment that this is for 1.4 and lower (and we can remove in 1.6, right?)

sseago: I'll add the comment. As for removing in 1.6, that depends on whether there will ever be a case where someone upgrades from 1.4 to 1.6 without running the operator at 1.5 in-between.

mateusoliveira43 (approved):

kaovilai (approved):

PR #1696, opened 2025-04-04
[oadp 1.5] remove velero-restore-helper image

sseago:
Remove velero-restore-helper image references and
delete fs-restore-action-config configmap if present.

cherry-pick of https://github.com/openshift/oadp-operator/pull/1695
## Why the changes were made
As of Velero 1.16, the velero image includes the restore helper binary, so no separate container image is required. By default, Velero uses the same image for the restore helper initcontainer as is used for velero. As a result, we no longer need the configmap to point to a separate image.

Verify that with this change, the `fs-restore-action-config` configmap no longer exists in the oadp namespace.

Additionally, when restoring via fs-backup, verify that the restore-wait initcontainer uses the regular velero image and not the velero-restore-helper image:
```
    initContainers:
    - args:
      - 24e069f4-fbaf-4952-a5c2-9df3890870dd
      command:
      - /velero-restore-helper
      env:
      - name: POD_NAMESPACE
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.namespace
      - name: POD_NAME
        valueFrom:
          fieldRef:
            apiVersion: v1
            fieldPath: metadata.name
      image: quay.io/konveyor/velero:latest
      imagePullPolicy: Always
      name: restore-wait
```


rayfordj (approved):
/lgtm
/approve

mateusoliveira43 (approved):

PR #1705, opened 2025-04-09
OADP-5900: DPA config for backup repo cache and full maintenance interval

sseago:
## Why the changes were made
Enable backup repository configuration via DPA. Upstream docs are here: https://velero.io/docs/v1.16/backup-repository-configuration/

This PR adds two new DPA fields to NodeAgentConfig to set `cacheLimitMB` and `fullMaintenanceInterval`

Changes to `cacheLimitMB` will be reflected in the conf file under $HOME/udmrepo in the velero pod. The number doesn't translate exactly to a single value seen, but for example setting it to 8000 produces these numbers:
```
  "caching": {
    "cacheDirectory": "../.cache/kopia/2123fc541362ef5f",
    "maxCacheSize": 5368709120,
    "contentCacheSizeLimitBytes": 6710886400,
    "maxMetadataCacheSize": 1342177280,
    "metadataCacheSizeLimitBytes": 1677721600,
    "maxListCacheDuration": 30000000000
  },
Setting it to 4000 will drop all of those numbers except for `maxListCacheDuration` in half.

Also, verify that the `backup-repository-dpa-name` configmap has both fields set (if you set both) and the values match the DPA field valuess:
```
apiVersion: v1
data:
  kopia: '{"cacheLimitMB":2000,"fullMaintenanceInterval":"fastGC"}'
kind: ConfigMap
```
and that the configmap is passed as an arg to the velero pod:
```
        - --backup-repository-configmap=backup-repository-velero-sample
```
<!-- Explain how the changes introduced by this PR can be tested and verified, with commands and pictures -->

### CODE REVIEW

File: api/v1alpha1/dataprotectionapplication_types.go [hunk 1]
```diff
@@ -433,8 +433,28 @@ type NodeAgentConfig struct {
...
+	// Embedding KopiaRepoOptions
+	// +optional
+	KopiaRepoOptions `json:",inline"`
+}
+
+type KopiaRepoOptions struct {
+	// +kubebuilder:validation:Minimum=0
+	// +optional
+	CacheLimitMB *int64 `json:"cacheLimitMB,omitempty"`
+	// +kubebuilder:validation:Enum=normalGC;fastGC;eagerGC
```
mateusoliveira43, line 445: I think description would be good here, I do not know what the available values mean

api/v1alpha1/dataprotectionapplication_types.go
```diff
@@ -433,8 +433,33 @@ type NodeAgentConfig struct {
 	// Embedding NodeAgentConfigMapSettings
 	// +optional
 	NodeAgentConfigMapSettings `json:",inline"`
+	// Embedding KopiaRepoOptions
+	// +optional
+	KopiaRepoOptions `json:",inline"`
+}
+
+type KopiaRepoOptions struct {
```
mateusoliveira43, line 441: the idea to have a separate struct is because this would be datamover options and not FSB options? (I think I heard some discussion around this)

mpryc: I think separate to use marshal:
https://github.com/openshift/oadp-operator/pull/1705/files/2e2b78de35561c27a6dd42af95d88e012eb879a2#diff-3d7a2c405414ff19b6660300351bf73aa3c836948cfe76a80f4e9e4d7b689d07R31

mateusoliveira43: alright

shubham-pampattiwar (approved):

mateusoliveira43 (approved):

PR #1722, opened 2025-04-18
OADP-5960: Don't mount default creds if no BSLs or VSLs for provider

sseago:
## Why the changes were made

OADP should only mount the default creds if they're needed for a BSL or VSL. Current logic is mounting default creds if either the provider has no BSLs/VSLs or there's at least one BSL/VSL whihc needs it. This PR drops the mount if there are no BSLs or VSLs at all.

Fixes [OADP-5960](https://issues.redhat.com//browse/OADP-5960)

Create a DPA with a single AWS backupLocation that uses named credentials. Add both aws and azure under defaultPlugins. Without this fix, /cloud-credentials-azure is mounted as a volume. With the fix, we won't get the mount.
To confirm that the fix doesn't break the "needs default creds" case, add an aws BSL with no credentials specified, and then you should see /cloud-credentials mounted.
(make sure cloud-credentials and cloud-credentials-azure secrets exist in order to test whether or not they mount properly)

### DIFF
--- a/internal/controller/validator.go
+++ b/internal/controller/validator.go
@@ -318,24 +318,16 @@ func (r *DataProtectionApplicationReconciler) ValidateVeleroPlugins() (bool, err
 
 	dpa := r.dpa
 
-	providerNeedsDefaultCreds, hasCloudStorage, err := r.noDefaultCredentials()
+	providerNeedsDefaultCreds, err := r.noDefaultCredentials()
 	if err != nil {
 		return false, err
-	}
-
-	snapshotLocationsProviders := make(map[string]bool)
-	for _, location := range dpa.Spec.SnapshotLocations {
-		if location.Velero != nil {
-			provider := strings.TrimPrefix(location.Velero.Provider, veleroIOPrefix)
-			snapshotLocationsProviders[provider] = true
-		}
 	}
 
 	foundAWSPlugin := false
 	foundLegacyAWSPlugin := false
 	for _, plugin := range dpa.Spec.Configuration.Velero.DefaultPlugins {
 		pluginSpecificMap, ok := credentials.PluginSpecificFields[plugin]
-		pluginNeedsCheck, foundInBSLorVSL := providerNeedsDefaultCreds[pluginSpecificMap.ProviderName]
+		pluginNeedsCheck := providerNeedsDefaultCreds[pluginSpecificMap.ProviderName]
 
 		// "aws" and "legacy-aws" cannot both be specified
 		if plugin == oadpv1alpha1.DefaultPluginAWS {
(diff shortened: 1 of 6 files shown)


weshayutin:
thanks @sseago !

sseago:
@mpryc That is for mounting /tmp/credentials used for the per-BSL creds. This PR is dealing with avoiding the default credentials mount at /cloud-credentials, /cloud-credentials-azure, and /cloud-credentials-gcp when we don't have any BSLs or VSLs that need it.

internal/controller/validator.go
```diff
@@ -349,12 +349,6 @@ func (r *DataProtectionApplicationReconciler) ValidateVeleroPlugins() (bool, err
 		if plugin == oadpv1alpha1.DefaultPluginVSM {
 			return false, errors.New("Delete vsm from spec.configuration.velero.defaultPlugins and dataMover object from spec.features. Use Velero Built-in Data Mover instead")
 		}
-		if foundInVSL := snapshotLocationsProviders[string(plugin)]; foundInVSL {
```
mateusoliveira43, line 352: we are removing this because it is already handled here?
https://github.com/openshift/oadp-operator/blob/b583cb79233e738308ec16d8979ea8805976862d/internal/controller/velero.go#L810-L823



we should also remove this
https://github.com/openshift/oadp-operator/blob/b583cb79233e738308ec16d8979ea8805976862d/internal/controller/validator.go#L326-L332
since it is not used anymore

internal/controller/validator.go
```diff
@@ -349,12 +349,6 @@ func (r *DataProtectionApplicationReconciler) ValidateVeleroPlugins() (bool, err
...
-			pluginNeedsCheck = true
-		}
-		if !foundInBSLorVSL && !hasCloudStorage {
-			pluginNeedsCheck = true
-		}
 		if ok && pluginSpecificMap.IsCloudProvider && pluginNeedsCheck && !dpa.Spec.Configuration.Velero.NoDefaultBackupLocation && !dpa.Spec.Configuration.Velero.HasFeatureFlag("no-secret") {
```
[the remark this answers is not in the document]
sseago: I don't think we should change the NoDefaultBackupLocation check. We still want to avoid validation there. The only intended change here is to *not* mount the default secret if there are no BSLs or VSLs at all for the provider.

internal/controller/validator.go
sseago: We're removing it because we don't want to mount the secret if no locations are found, so the "found" logic isn't used for anything else here, it can be removed.

shubham-pampattiwar (approved):

weshayutin (approved):

PR #1726, opened 2025-04-23
fix: must-gather standards

mateusoliveira43:
## Why the changes were made

Make must-gather follow standards https://github.com/openshift/enhancements/blob/995b620cb04c030bf62c908e188472fe7031a704/enhancements/oc/must-gather.md?plain=1#L94-L104

Check if standards are being followed.

### DIFF
--- a/must-gather/README.md
+++ b/must-gather/README.md
@@ -75,3 +75,33 @@ go mod verify
 ## Deprecated folder
 
 Scripts under `deprecated/` folder are for backwards compatibility with old OADP Must-gather shell script. Users should use new OADP Must-gather Go script, as highlighted in product documentation.
+
+## Standards
+
+OADP Must-gather must comply with https://github.com/openshift/enhancements/blob/995b620cb04c030bf62c908e188472fe7031a704/enhancements/oc/must-gather.md?plain=1#L94-L104
+
+>1. Must have a zero-arg, executable file at `/usr/bin/gather` that does your default gathering
+
+OADP Must-gather binary can be called without args. All OADP Must-gather binary args are optional
+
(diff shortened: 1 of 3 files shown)


must-gather/README.md
```diff
@@ -75,3 +75,33 @@ go mod verify
...
+
+>1. Must have a zero-arg, executable file at `/usr/bin/gather` that does your default gathering
+
+OADP Must-gather binary can be called without args. All OADP Must-gather binary args are optional
+
+>2. Must produce data to be copied back at `/must-gather`. The data must not contain any sensitive data. We don't string PII information, only secret information.
+
+OADP Must-gather collected data is stored at `/must-gather` folder in the same path the binary was called.
+
+Most of the data is collected through `oc adm inspect` command (including Secrets). The other data are cluster information, OADP related information (CRDs and CRs) and storage information (StorageClasses, VolumeSnapshotClasses and CSIDrivers CRDs and CRs). These objects should not contain any sensitive data.
```
kaovilai, line 91: What mechanisms if any are we using to prevent sensitive data?

weshayutin: @kaovilai so do the following.. run the must-gather against an active oadp deployment.  Check the secrets for example.  When @mateusoliveira43 looked at it the secrets could not be base64 decoded or read.  You double checking that would be appreciated.   AFAICT it's built into the gather tooling already sir.

weshayutin (approved):

must-gather/README.md
mateusoliveira43: just good sense

for example, when we need data for DPA, we ask for a Secret name and not directly to sensitive data to be added to DPA spec

but we need to be careful to always do this approach

kaovilai: ok got it.. I thought we have some secret stripping mech to get secret format without exposing actual keys etc.. resolving.

PR #1739, opened 2025-05-01
use golang 1.23.x for 1.3

weshayutin:
## Why the changes were made

We accidentally used golang 1.24 for the mustgather go mod and did a crappy job reviewing :)
Including myself in that.

### DIFF
--- a/must-gather/go.mod
+++ b/must-gather/go.mod
@@ -1,8 +1,8 @@
 module github.com/openshift/oadp-operator/must-gather
 
-go 1.24.0
+go 1.23.1
 
-toolchain go1.24.2
+toolchain go1.23.4
 
 require (
 	github.com/kubernetes-csi/external-snapshotter/client/v4 v4.2.0


must-gather/go.mod
```diff
@@ -1,8 +1,8 @@
 module github.com/openshift/oadp-operator/must-gather
 
-go 1.24.0
+go 1.23.1
```
shubham-pampattiwar, line 3: 1.23.0 ?

must-gather/go.mod
```diff
@@ -1,8 +1,8 @@
...
 
-toolchain go1.24.2
+toolchain go1.23.4
```
shubham-pampattiwar, line 5: 1.23.6 ?

weshayutin: I agree but go look at what's merged on 1.4 :) and these don't really matter for downstream as long as it's NOT HIGHER.  @rayfordj right?

shubham-pampattiwar: I think I check 1.5 but no worries if does not matter, Acking the PR

shubham-pampattiwar (approved):

must-gather/go.mod
rayfordj: yes-sir, as long as we're never more than the latest builder version we have available to use -- currently, 1.23.6.  :sunglasses:

kaovilai (approved):

must-gather/go.mod
sseago: Actually, Velero 1.16 has toolchain 1.23.7 which is higher than our builder, but @rayfordj mentioned to me when this came up earlier that it's OK, since we don't use that value for specifying the toolchain we build with. Hopefully I got that right, since for openshift/velero OADP 1.5 the value *is* higher than the downstream builder version.

sseago (approved):

PR #1752, opened 2025-05-12
OADP-5275 init plugin containers to have read only fs

mpryc:
This adds ReadOnlyRootFilesystem to the init containers (plugins)

Also additional security context flags are applied to ensure the init containers are using:
 - ReadOnlyRootFilesystem: true
 - Privileged: false
 - AllowPrivilegeEscalation: false

## Why the changes were made

Fix [OADP-5275](https://issues.redhat.com//browse/OADP-5275)

## How to test the changes made

Deploy operator, run DPA with plugins, check the yaml of the init containers to set the above flags.

PR #1755, opened 2025-05-13
Fix velero container to have explicit ReadOnlyRootFilesystem

mpryc:
Depends-on #1752 

## Why the changes were made

Similar to https://github.com/openshift/oadp-operator/pull/1752

This PR is for the Velero container.

Deploy, check YML of the velero container as well sh to the running velero container and see:
```shell
sh-5.1$ touch /testme
touch: cannot touch '/testme': Read-only file system
sh-5.1$ touch /plugins/testme
touch: cannot touch '/plugins/testme': Read-only file system
sh-5.1$ ls /plugins
velero-plugin-for-aws  velero-plugins
sh-5.1$ touch ~/testme
sh-5.1$ ls -lah ~/testme
-rw-r--r--. 1 1000660000 1000660000 0 May 13 08:08 /home/velero/testme
```

### DIFF
--- a/internal/controller/velero.go
+++ b/internal/controller/velero.go
@@ -272,6 +272,18 @@ func (r *DataProtectionApplicationReconciler) customizeVeleroDeployment(veleroDe
 				EmptyDir: &corev1.EmptyDirVolumeSource{},
 			},
 		},
+		corev1.Volume{
+			Name: "tmp",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
+		corev1.Volume{
+			Name: "home",
+			VolumeSource: corev1.VolumeSource{
+				EmptyDir: &corev1.EmptyDirVolumeSource{},
+			},
+		},
 		// used for short-lived credentials, inert if not used
 		corev1.Volume{
 			Name: "bound-sa-token",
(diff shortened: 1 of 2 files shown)


weshayutin:
@sseago @shubham-pampattiwar @kaovilai please have a look

kaovilai:
Can you link jira?

kaovilai (commented):
As far as velero container goes.. this lgtm. Didn't see any reference in velero container codes using other dirs.

kaovilai (approved):

shubham-pampattiwar (approved):

kaovilai:
## Azure Workload Identity Investigation Required

During testing of the Azure workload identity implementation, we encountered an authentication error that requires further investigation:

```
WorkloadIdentityCredential authentication failed. FromAssertion(): http call(https://login.microsoftonline.com/.../oauth2/v2.0/token)(POST) error: reply status code was 400:
{"error":"unauthorized_client","error_description":"AADSTS700016: Application with identifier 'https://[storage-account].blob.core.windows.net/[container]' was not found in the directory..."}
```

### Error Details
- **Error Code**: AADSTS700016
- **Issue**: The application identifier being used appears to be the blob storage URL rather than the expected Azure AD application (client) ID
- **Impact**: BackupStorageLocation is marked as unavailable

### Current Implementation
The PR includes:
- ✅ Service account annotation with `azure.workload.identity/client-id`
- ✅ Deployment label `azure.workload.identity/use: "true"`
- ✅ Proper Azure credentials in the secret

### Investigation Needed
1. The error suggests the workload identity is using the storage account URL as the application identifier instead of the client ID
2. Need to verify how the Azure plugin interprets the credentials and constructs the authentication request
3. May need to investigate if additional configuration is required for the Azure plugin to properly use workload identity

### Next Steps
- Review Azure plugin's workload identity implementation
- Verify the expected format of credentials for workload identity
- Test with different credential configurations
- Check if additional environment variables or configuration is needed

PR #1767, opened 2025-05-28
Made the TTL duration configurable in the Makefile.

kaovilai:
Here's what was changed:

  1. Added a TTL_DURATION ?= 1h variable near the top of the Makefile (line 62) with helpful comments explaining it's configurable and has a maximum of 24h.
  2. Updated all 7 instances of hardcoded :1h to use $(TTL_DURATION) variable:
    - Lines 448-449: deploy-olm target
    - Lines 473-477: catalog-test-upgrade target

  Now you can override the TTL duration by setting the environment variable:
  export TTL_DURATION=4h
  make deploy-olm

  Or pass it directly to make:
  make deploy-olm TTL_DURATION=8h

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,6 +55,12 @@ OPERATOR_SDK_VERSION ?= v1.34.2
 
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/konveyor/oadp-operator:latest
+
+# TTL_DURATION defines the time-to-live for temporary images pushed to ttl.sh
+# The maximum allowed value by ttl.sh is 24h. Default is 1h.
+# You can override this with environment variable (e.g., export TTL_DURATION=4h)
+TTL_DURATION ?= 1h
+
 # ENVTEST_K8S_VERSION refers to the version of kubebuilder assets to be downloaded by envtest binary.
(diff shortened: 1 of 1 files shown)


weshayutin (approved):
nice!

Makefile
```diff
@@ -55,6 +55,12 @@ OPERATOR_SDK_VERSION ?= v1.34.2
 
 # Image URL to use all building/pushing image targets
 IMG ?= quay.io/konveyor/oadp-operator:latest
+
+# TTL_DURATION defines the time-to-live for temporary images pushed to ttl.sh
+# The maximum allowed value by ttl.sh is 24h. Default is 1h.
+# You can override this with environment variable (e.g., export TTL_DURATION=4h)
+TTL_DURATION ?= 1h
```
mpryc, line 62: how about naming it:

And then check if the number is within 1-24?

```Makefile
TTL_DURATION_HOURS ?= 1

ifeq (,$(filter $(TTL_DURATION_HOURS),1 2 3 4 5 6 7 8 9 10 11 12 13 14 15 16 17 18 19 20 21 22 23 24))
$(error TTL_DURATION_HOURS must be between 1 and 24 (got: $(TTL_DURATION_HOURS)))
endif
```

And then when you use it simply add 'h' as the suffix. 

?

kaovilai: well that is limiting. maybe I want 6h30s

mpryc (approved):

PR #1769, opened 2025-05-29
allow user to skip must-gather

weshayutin:
## Why the changes were made

If using amd64 via laptop, and arm64 cluster
the must-gather will panic and fail.

export SKIP_MUST_GATHER=true

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -562,6 +562,7 @@ test-e2e-setup: login-required build-must-gather
 	VSL_REGION="$(VSL_REGION)" \
 	BSL_REGION="$(BSL_REGION)" \
 	BSL_AWS_PROFILE="$(BSL_AWS_PROFILE)" \
+        SKIP_MUST_GATHER="$(SKIP_MUST_GATHER)" \
 	/bin/bash "tests/e2e/scripts/$(CLUSTER_TYPE)_settings.sh"
 
 VELERO_INSTANCE_NAME ?= velero-test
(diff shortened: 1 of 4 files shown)


weshayutin:
> Just always use multiarch mustgather.. problem solved?

@kaovilai it's not using a container :) , so no that would not work in this case.

Makefile
```diff
@@ -601,6 +603,7 @@ test-e2e: test-e2e-setup install-ginkgo ## Run E2E tests against OADP operator i
 	-artifact_dir=$(ARTIFACT_DIR) \
 	-kvm_emulation=$(KVM_EMULATION) \
 	-hco_upstream=$(HCO_UPSTREAM) \
+        -skipMustGather=$(SKIP_MUST_GATHER) \
```
mpryc, line 606: How about keeping same naming:

-skip_must_gather

shubham-pampattiwar:
@weshayutin I kind of agree with @kaovilai We need both - the configurability to skip running must-gather and also have the image multi-arch.
We might need to update the image used in RunMustGather func:
```
func RunMustGather(oc_cli string, artifact_dir string) error {
	ocClient := oc_cli
	ocAdmin := "adm"
	mustGatherCmd := "must-gather"
	mustGatherImg := "--image=quay.io/konveyor/oadp-must-gather:latest". <<<<------- This image ! 
	destDir := "--dest-dir=" + artifact_dir

	cmd := exec.Command(ocClient, ocAdmin, mustGatherCmd, mustGatherImg, destDir)
	_, err := cmd.Output()
	return err
}
```

weshayutin:
> quay.io/konveyor/oadp-must-gathe
@shubham-pampattiwar the must-gather image was specifically avoided so that changes to must-gather itself would be incorporated to the pr.   This was discussed in scrum.   Am I missing something?   We could add the option to pull the image vs. execute locally, but why?

I don't really see the value in running must-gather during our local `make test-e2e` anyway.

kaovilai (approved):

shubham-pampattiwar (approved):

PR #1791, opened 2025-06-20
NO-JIRA: Update Konflux new images for HCP OADP Plugin

jparrill:
## Why the changes were made
- Need to update the images generated by Konflux in the new tenant for the Hypershift OADP plugin


kaovilai:
same as https://github.com/openshift/oadp-operator/pull/1790 ?

PR #1813, opened 2025-06-30
Add design doc for wildcard namespaces

Joeavaikath:
## Why the changes were made

Adding design doc for wildcard namespace enumeration and selection support

Design doc

### DIFF
--- a/docs/design/wildcard-namespace-design.md
+++ b/docs/design/wildcard-namespace-design.md
@@ -95,3 +95,4 @@ Estimated timeline is 2-3 development cycles with proper testing and validation.
 
 Error handling strategy for cases where wildcard patterns match no namespaces needs clarification - should this be treated as an error or proceed with empty namespace list.
 Performance impact on very large clusters (1000+ namespaces) requires testing and potential optimization. 
+- Potentially, could have an extra step where we list the namespaces that would be effectively backed up if the user supplies wildcard namespaces. It's a big change from the usual flow.


docs/design/wildcard-namespace-design.md
```diff
@@ -0,0 +1,97 @@
...
+
+Velero's backup CLI currently accepts wildcard patterns for namespace inclusion and exclusion but has incomplete implementation.
+When users run commands like `velero backup create test --include-namespaces "prod-*"`, the wildcard pattern is stored literally in the backup specification without being resolved to actual namespace names.
+
+This causes backup failures because the system attempts to list resources in namespaces literally named `"prod-*"` which don't exist.
+The disconnect occurs between namespace collection (which uses proper glob matching) and resource collection (which uses literal patterns), resulting in inconsistent behavior and silent failures.
+
+## Goals
+
+- Enable wildcard patterns in backup CLI commands to work correctly by resolving them to actual namespace names
```
sseago, line 18: We probably need the wildcard expansion to happen in the Backup schedule as well. Basically whatever logic we put in the CLI to generate the backup.spec.Included/ExcludedNamespaces field values should also be added to the schedule controller where that controller creates a Velero backup CR from the Schedule template. This still preserves the existing backup functionality, since backup processing itself won't be doing the expansion.

docs/design/wildcard-namespace-design.md
```diff
@@ -0,0 +1,98 @@
...
+## Non Goals
+
+- Wildcard support for restore operations (future enhancement)
+- Complex regex patterns beyond standard glob patterns
+- Dynamic namespace matching during backup execution (patterns resolved at CLI time only)
+
+## High-Level Design
+
+The solution resolves wildcard patterns during CLI command processing before creating backup objects.
+When a user specifies patterns like `prod-*`, the CLI will query the Kubernetes API to fetch all active namespaces, apply glob pattern matching, and store only the resolved actual namespace names in the backup specification.
```
shubham-pampattiwar, line 31: Why not keep the queried NS/matched NS in-memory instead of patching the backup spec ?

Joeavaikath: When you say in-memory, do you mean that the spec would keep the wildcard namespaces, and that it would resolve at the time of backup creation, rather than us manually enumerating, finding and filling those namespaces in?
My approach was more find at cli command runtime and substitute. That makes it deterministic but then its not dynamic, like schedules picking up new namespaces matching the pattern

docs/design/wildcard-namespace-design.md
```diff
@@ -0,0 +1,98 @@
...
+velero backup create test --include-namespaces "prod-*"
+# Backup fails - tries to find resources in namespace literally named "prod-*"
+```
+
+**New behavior**:
+```bash
+velero backup create test --include-namespaces "prod-*"
+# CLI resolves to: ["prod-app", "prod-db", "prod-cache"]
+# Backup succeeds with all resources from matching namespaces
+```
```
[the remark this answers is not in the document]
kaovilai: Or is this separate from the oc plugin work?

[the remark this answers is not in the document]
kaovilai: Any reason to not propose in https://github.com/vmware-tanzu/velero/tree/main/design ?

docs/design/wildcard-namespace-design.md
sseago: This isolates the wildcard expansion to backup creation time (for CLI and hopefully for schedules too)

docs/design/wildcard-namespace-design.md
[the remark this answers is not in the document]
Joeavaikath: Will close this and start one on upstream

PR #1844, opened 2025-07-28
Propagate errors from HCP deletion checks

kaovilai:
## Why the changes were made

Previously, IsHCDeleted and IsHCPDeleted only returned a boolean value, causing actual API errors to be lost and reported as generic "context deadline exceeded" errors. This made debugging difficult when deletion checks failed due to API errors (network issues, permission problems, etc.) rather than the resource still existing.

This change improves error visibility by propagating the actual error from the Kubernetes API calls, helping operators troubleshoot HCP deletion timeouts more effectively.

1. Run the unit tests:
   ```bash
   go test ./tests/e2e/lib/hcp/... -v
   ```

2. To see the improved error reporting in action, you can simulate an API error scenario where the deletion check fails. The error will now show the actual API error instead of just "context deadline exceeded".

3. E2E tests should continue to pass as before, but with better error messages when deletion checks fail.


weshayutin:
@kaovilai ++

PR #1827, opened 2025-07-09
Add design for managing Kopia repositories via BSL and new BSLR

mpryc:
This design introduces the BackupStorageLocationRepository (BSLR) as a new custom resource that models and manages Kopia repositories on a per-BSL basis.

## Why the changes were made

Propose new design that provides a clear separation between BSL config and repository state, and enables early provisioning of Kopia repos before backups run. This in the future will allow to create BSL Server on top of BSLR.

## How to test the changes made

Read the design.

PR #1830, opened 2025-07-15
Add design for the Backup Storage Location Server

mpryc:
Introduces the BSLS design to enable backup and restore operations through a proxy service managed by the OADP Operator.

## Why the changes were made

This is complementary design to the https://github.com/openshift/oadp-operator/pull/1827

To enable backup and restore operations via a proxy service managed by the OADP Operator, improving flexibility and management of backup workflows.

## How to test the changes made

Read the design.

PR #1845, opened 2025-07-29
Add design for the OADP Virtual Machine Data Protection (VMDP)

mpryc:
New design to allow user based backup and restore through the kopia server deployed as part of the OADP operator.

This design replaces the following two designs, which after agreeing can be closed:
 https://github.com/openshift/oadp-operator/pull/1827
 https://github.com/openshift/oadp-operator/pull/1830

## Why the changes were made

To enable backup and restore operations via a proxy service managed by the OADP Operator, improving flexibility and management of backup workflows.

Read the design.

AI assisted: Cursor (default model).

### DIFF
--- a/docs/design/vm_data_protection-design.md
+++ b/docs/design/vm_data_protection-design.md
@@ -168,7 +168,7 @@ There are three distinct types of credentials involved in the system:
    The user uses the Kopia compatible CLI inside the VM to connect to the Kopia server (Backup Service) running in the cluster. This step establishes a trusted session and configures the local Kopia client to communicate with the remote repository.
 
 2. **Perform Backup**  
-   Once connected, the user initiates a backup by running the `vmdp backup create /path/to/backup` command (or similar), specifying the files or directories to back up. Kopia scans the specified paths, deduplicates data, encrypts it, and uploads the backup to the repository via the backup server. The backup operation can be repeated as needed, and incremental backups are supported by default. All the encryption of files and folders happens on the client side inside the VM.
+   Once connected, the user initiates a backup by running the `vmdp backup create /path/to/file_to_backup` command (or similar), specifying the files or directories to back up. Kopia scans the specified paths, deduplicates data, encrypts it, and uploads the backup to the repository via the backup server. The backup operation can be repeated as needed, and incremental backups are supported by default. All the encryption of files and folders happens on the client side inside the VM.
 
 3. **Session Persistence**
    The connection and authentication information is typically cached locally (e.g., in the user's Kopia config directory), so future backup or restore operations do not require re-authentication unless the session is reset, credentials change or config files deleted.
--- a/docs/design/vm_data_protection-design.md
+++ b/docs/design/vm_data_protection-design.md
@@ -1,0 +1,502 @@
+# OADP Virtual Machine Data Protection (VMDP): A Design for User-Driven, File-Level Backups
+
+## Abstract
+
+The OpenShift API for Data Protection (OADP) delivers comprehensive, snapshot-based backup and restore capabilities for OpenShift Virtual Machines (VMs), providing a reliable solution for full disaster recovery scenarios. However, current OADP functionality does not extend to granular, file-level backup and restore operations initiated from within the guest operating system—such as safeguarding application configurations or user-generated data.
+
+This design introduces a Kubernetes-native, client-server architecture that enables secure, file-level backup and restore directly from within VMs. This approach empowers users with a complementary mechanism to perform targeted backups and recover individual files, including those residing on external network file systems (e.g., NFS) mounted within the guest OS.
+
+
+## Background
+
+The OpenShift API for Data Protection (OADP) is designed to provide full-system backup and restore for OpenShift Virtual Machines (VMs), with workflows typically managed by cluster or namespace administrators. While this approach is effective for disaster recovery and infrastructure-level protection, it does not natively support granular, file-level backup and restore operations.
+
+In practice, especially in multi-tenant or self-service OpenShift Virtualization environments, VM users and application owners often require the ability to back up or restore specific files and directories—such as application configuration files, logs, or user data. These users generally do not have access to storage backends or the elevated privileges needed for full volume snapshot or restore operations.
+
+To address this limitation, we propose a Kubernetes-native, client-server solution that operates from within the guest OS. This design enables secure, authenticated, and user-driven file-level backup and restore, without exposing sensitive infrastructure details or requiring cluster-level privileges. By complementing OADP’s snapshot-based approach, this new architectural component empowers end users to manage their own data protection needs, bridging the gap between infrastructure-level and application-level protection in OpenShift environments.
+
+## Scope
+
+The scope of the **Virtual Machine Data Protection (VMDP)** feature is to introduce a secure, user-driven mechanism for file-level backup and restore operations originating from within a guest VM. This functionality is intended to complement - NOT to replace - OADP’s existing snapshot-based backup and restore capabilities for full VMs and persistent volumes. Additionally, the new mechanism extends support to file-level operations on network-mounted file systems (e.g., NFS, CIFS) accessible from the guest OS, offering users greater flexibility and control over their data protection needs.
+
+### In Scope
+
+The following components and functionalities are considered within scope for this design and its initial implementation:
+
+- **Core Functionality**: A client-server model that enables backup and restore of specific files and directories from within a guest operating system.
+- **User Interaction**: A command-line interface (CLI) available on Linux, macOS, and Windows, allowing VM users to initiate backup and restore operations.
+- **Control Plane Integration**: OpenShift-native management using Custom Resource Definitions (CRDs) and controllers to handle and track file-level backup/restore requests.
+- **Authentication and Security**: User-specific credentials are provisioned using OpenShift `Secrets` and `ConfigMaps`, and synchronized with the Kopia repository server. These credentials allow the in-guest client to authenticate directly with the Kopia server to perform backup and restore operations.
+- **Storage Integration**: Use of OADP’s existing `BackupStorageLocation` (BSL) abstraction to define and manage backup targets.
+- **Supported File Systems**: Backup and restore operations on any standard, mounted file system accessible from within the guest OS, including network file systems such as NFS and CIFS.
+- **Repository Management**: Support for multiple logical backup repositories within a single namespace, associated with a shared BSL.
+
+### Out of Scope
+
+The following are explicitly out of scope for this project phase:
+
+- **Full VM Protection**: This design does not modify or replace OADP’s snapshot-based backup and restore for entire VMs or persistent volumes.
+- **Application Quiescing**: The system will not implement application-level consistency mechanisms (e.g., pausing databases). Ensuring data consistency remains the user's responsibility.
+- **Block-Level Operations**: Backup or restore of raw block devices, unmounted partitions, or disk images is not supported.
+- **Graphical User Interface (UI)**: No integration with the OpenShift Console or other web-based user interfaces is included.
+- **Client Lifecycle Management**: Automated deployment, installation, or upgrading of the in-guest client software is out of scope.
+- **External VM Access**: The design does not introduce new mechanisms for cluster-initiated access to guest VMs (e.g., SSH injection, `virt`-based access).
+- **New Storage Backends**: This effort does not introduce support for new storage backend types. Only those already supported by OADP will be used.
+- **Advanced Observability**: Detailed metrics, dashboards, or alerting features for file-level backup operations are not addressed in this phase.
+
+## Goals
+
+By delivering the features defined in this scope, the VMDP design aims to achieve the following strategic objectives:
+
+- **Empower End Users**: Provide VM users with secure, self-service capabilities to back up and restore their own application data, without needing elevated privileges or administrator intervention.
+- **Enhance Data Protection Granularity**: Address a critical gap between infrastructure-level recovery and application-level data needs by supporting fine-grained, file-specific protection.
+- **Maintain a Secure, Zero-Trust Model**: Ensure that all client-server interactions are strongly authenticated and authorized using OpenShift-native mechanisms, minimizing risk and privilege exposure.
+- **Integrate Seamlessly with OADP**: Leverage existing OADP concepts and APIs (such as `BackupStorageLocation`) to ensure consistent behavior and ease of adoption by administrators.
+- **Deliver a Platform-Agnostic Solution**: Support major operating systems commonly used within guest VMs to maximize utility across enterprise environments.
+
+## High-Level Design
+
+The **Virtual Machine Data Protection (VMDP)** feature introduces a OpenShift-native, client-server architecture that enables file-level backup and restore operations initiated from within OpenShift Virtual Machines. The system integrates seamlessly with OpenShift APIs and the existing OADP infrastructure, while preserving a clear separation of responsibilities between cluster administrators and VM users.
+
+### Architecture Overview
+
+The solution consists of the following core components:
+
+1. **In-Guest Backup Client**  
+   A lightweight, statically linked Go-based CLI tool, built from the standard Kopia CLI source code and fully compatible with Kopia APIs. It is designed to run entirely in user space (without requiring elevated privileges) and can be easily downloaded and executed—even outside standard installation paths. This client allows users to perform file-level backup and restore operations from within the guest VM, authenticating with the Kopia server using user-specific credentials.
+
+2. **Backup Service (Kopia Server)**  
+   A multi-user backup repository server deployed within the OpenShift cluster. It handles file uploads, downloads, metadata management, and repository access via authenticated client connections. The server uses per-user credentials, provisioned and synchronized from OpenShift `Secrets` and `ConfigMaps`, to authorize operations.
+
+   Multiple Backup Services can be deployed per each BackupStorageLocation (BSL), offering flexibility in deployment strategies. It is up to the cluster administrator to decide whether to share a single Backup Service instance among multiple users and VMs or to provision a dedicated instance per VM. Both approaches are secure: shared services rely on strong user isolation and credential scoping within the Kopia server, while per-VM deployments offer additional isolation where needed. This flexibility allows administrators to tailor deployments to different performance, tenancy, or security requirements.
+
+3. **Custom Resource Definitions (CRDs) and Controller**  
+   This component extends the existing OADP operator by adding a new controller and introducing two new CRDs:
+
+   - **BackupStorageLocationServer (BSLS)** — represents the deployment and configuration of a Kopia Backup Service instance within the OpenShift cluster. The BSLS points either directly to a Velero `BackupRepository` or to a new CRD `BackupStorageLocationRepository`.
+
+   - **BackupStorageLocationRepository (BSLR)** — represents the Kopia repository backing the file-level backups for a given Backup Storage Location. The BSLR points back to a `BackupStorageLocation` (BSL). The BSLR concept is similar to a Kopia BackupRepository but is managed by OADP.
+
+   This design separates the management of backup repositories from the storage locations themselves, making it easy to map Kopia repositories to OpenShift BackupStorageLocations in a flexible way. Administrators can control exactly how repositories are created, assigned, and managed for each user or VM. By building on top of the existing OADP operator and using familiar resources like `BackupStorageLocation` (BSL), the system fits naturally into OpenShift’s data protection workflows. This approach makes it straightforward to perform detailed, file-level backup and restore operations, while keeping the process consistent and easy to understand.
+
+4. **Repository Provisioning & Management**  
+   Logical backup repositories are created and scoped per user, per VM, or shared among multiple users and VMs. These repositories are linked to existing `BackupStorageLocation` (BSL) definitions to leverage OADP-supported storage backends such as S3, GCS, or Azure Blob.
+
+   - A **Velero Backup Repository (BR)** stores its metadata (e.g., snapshots, indexes) directly within the backup storage backend (e.g., an S3 bucket). This means the repository metadata is part of the stored content and can be automatically synchronized or recovered when a new `BackupStorageLocation` (BSL) points to the same storage location.
+
+   - In contrast, a **BackupStorageLocationRepository (BSLR)** is a OpenShift-managed resource that represents a Kopia repository with metadata managed primarily inside OpenShift (via the CRD). The BSLR’s metadata is **not stored in the backup storage backend** (e.g., S3). As a result, repository metadata for BSLRs is **not automatically synced or restored** when creating or switching to a new BSL pointing to the same storage location.
+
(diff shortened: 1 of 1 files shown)


docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+- **Control Plane Integration**: OpenShift-native management using Custom Resource Definitions (CRDs) and controllers to handle and track file-level backup/restore requests.
+- **Authentication and Security**: User-specific credentials are provisioned using OpenShift `Secrets` and `ConfigMaps`, and synchronized with the Kopia repository server. These credentials allow the in-guest client to authenticate directly with the Kopia server to perform backup and restore operations.
+- **Storage Integration**: Use of OADP’s existing `BackupStorageLocation` (BSL) abstraction to define and manage backup targets.
+- **Supported File Systems**: Backup and restore operations on any standard, mounted file system accessible from within the guest OS, including network file systems such as NFS and CIFS.
+- **Repository Management**: Support for multiple logical backup repositories within a single namespace, associated with a shared BSL.
+
+### Out of Scope
+
+The following are explicitly out of scope for this project phase:
+
```
weshayutin, line 37: Do we have anything we can do cap or limit the space used in s3? If not let's mark it out of scope.   Customers will want this though.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+- **Full VM Protection**: This design does not modify or replace OADP’s snapshot-based backup and restore for entire VMs or persistent volumes.
+- **Application Quiescing**: The system will not implement application-level consistency mechanisms (e.g., pausing databases). Ensuring data consistency remains the user's responsibility.
+- **Block-Level Operations**: Backup or restore of raw block devices, unmounted partitions, or disk images is not supported.
+- **Graphical User Interface (UI)**: No integration with the OpenShift Console or other web-based user interfaces is included.
```
weshayutin, line 41: I think the only reason we're avoiding a ui is because of the CVE impact, is that correct?  Just want to make sure we're thinking about it in the same way.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+- **Client Lifecycle Management**: Automated deployment, installation, or upgrading of the in-guest client software is out of scope.
+- **External VM Access**: The design does not introduce new mechanisms for cluster-initiated access to guest VMs (e.g., SSH injection, `virt`-based access).
+- **New Storage Backends**: This effort does not introduce support for new storage backend types. Only those already supported by OADP will be used.
+- **Advanced Observability**: Detailed metrics, dashboards, or alerting features for file-level backup operations are not addressed in this phase.
```
weshayutin, line 45: @shubham-pampattiwar sorry but w/ my PM hat on.  I wonder if the DPT could potentially be used to report on the size of s3 buckets.   Do a little poor man's s3 reporting, just a thought and not a todo.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+## Goals
```
weshayutin, line 47: +1 this is nice

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+4. **Repository Provisioning & Management**  
+   Logical backup repositories are created and scoped per user, per VM, or shared among multiple users and VMs. These repositories are linked to existing `BackupStorageLocation` (BSL) definitions to leverage OADP-supported storage backends such as S3, GCS, or Azure Blob.
+
+   - A **Velero Backup Repository (BR)** stores its metadata (e.g., snapshots, indexes) directly within the backup storage backend (e.g., an S3 bucket). This means the repository metadata is part of the stored content and can be automatically synchronized or recovered when a new `BackupStorageLocation` (BSL) points to the same storage location.
+
+   - In contrast, a **BackupStorageLocationRepository (BSLR)** is a OpenShift-managed resource that represents a Kopia repository with metadata managed primarily inside OpenShift (via the CRD). The BSLR’s metadata is **not stored in the backup storage backend** (e.g., S3). As a result, repository metadata for BSLRs is **not automatically synced or restored** when creating or switching to a new BSL pointing to the same storage location.
```
weshayutin, line 87: hrm.. I hadn't thought of the bslr not being protected the sync controller.   Would we be leaking creds or exposing anything if ( wave's magic wand ) and BSLR's were added to the sync controller list of objects?

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+   Multiple Backup Services can be deployed per each BackupStorageLocation (BSL), offering flexibility in deployment strategies. It is up to the cluster administrator to decide whether to share a single Backup Service instance among multiple users and VMs or to provision a dedicated instance per VM. Both approaches are secure: shared services rely on strong user isolation and credential scoping within the Kopia server, while per-VM deployments offer additional isolation where needed. This flexibility allows administrators to tailor deployments to different performance, tenancy, or security requirements.
+
+3. **Custom Resource Definitions (CRDs) and Controller**  
+   This component extends the existing OADP operator by adding a new controller and introducing two new CRDs:
+
+   - **BackupStorageLocationServer (BSLS)** — represents the deployment and configuration of a Kopia Backup Service instance within the OpenShift cluster. The BSLS points either directly to a Velero `BackupRepository` or to a new CRD `BackupStorageLocationRepository`.
```
weshayutin, line 76: The BSLS config is also something what wouldn't be synced by the sync-controller right?  Also interested in hearing thoughts about that.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+  - Repository Credentials
+  - User Access Credentials
+
+#### Backup Flow
+
+1. **Connect to the Backup Server**  
+   The user uses the Kopia compatible CLI inside the VM to connect to the Kopia server (Backup Service) running in the cluster. This step establishes a trusted session and configures the local Kopia client to communicate with the remote repository.
+
+2. **Perform Backup**  
+   Once connected, the user initiates a backup by running the `vmdp backup create /path/to/backup` command (or similar), specifying the files or directories to back up. Kopia scans the specified paths, deduplicates data, encrypts it, and uploads the backup to the repository via the backup server. The backup operation can be repeated as needed, and incremental backups are supported by default. All the encryption of files and folders happens on the client side inside the VM.
```
weshayutin:
```suggestion
   Once connected, the user initiates a backup by running the `vmdp backup create /path/to/file_to_backup` command (or similar), specifying the files or directories to back up. Kopia scans the specified paths, deduplicates data, encrypts it, and uploads the backup to the repository via the backup server. The backup operation can be repeated as needed, and incremental backups are supported by default. All the encryption of files and folders happens on the client side inside the VM.
```

weshayutin (commented):
Looks great.. couple questions about
* sync controller and new crds, whether or not it makes sense or is possible to sync these objects.
* I would like to better understand the maintenance jobs for BSLR's.  Is it the same as any maint job and does the oadp dpa config handle that?

shawn-hurley (commented):
IIUC any user that is able to exec into the vm, would be able to restore/backup files through the BSLS by running the kopia commands. 

Is that a fair assumption?


Have we considered using oauth tokens to do the auth/z with kubernetes rbac? 

I am imagining something like, user execs into the VM -> user calls the kopia command and passes in or authenticates with the cluster oauth server -> that credential is then used by the BSLS to create a subjectaccessreview  -> we design some RBAC rules to verify that the authenticated user can do the action.

weshayutin:
@shawn-hurley @shubham-pampattiwar any comments?

shawn-hurley:
Any comments that I have had, have been addressed offline. 

Overall, this is an excellent idea for Tech Preview to kick the tires of it. 

I am also thinking about a world where this is out of tech preview and BSLR might be an interesting concept to add to non-admin(?) but that is a wild thought that I just had.

weshayutin:
@shawn-hurley thanks! I agree the BSLR concept is going to be a valuable tool imho, excited to see how we can utilize that.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+   - **BackupStorageLocationRepository (BSLR)** — represents the Kopia repository backing the file-level backups for a given Backup Storage Location. The BSLR points back to a `BackupStorageLocation` (BSL). The BSLR concept is similar to a Kopia BackupRepository but is managed by OADP.
+
+   This design separates the management of backup repositories from the storage locations themselves, making it easy to map Kopia repositories to OpenShift BackupStorageLocations in a flexible way. Administrators can control exactly how repositories are created, assigned, and managed for each user or VM. By building on top of the existing OADP operator and using familiar resources like `BackupStorageLocation` (BSL), the system fits naturally into OpenShift’s data protection workflows. This approach makes it straightforward to perform detailed, file-level backup and restore operations, while keeping the process consistent and easy to understand.
```
shubham-pampattiwar, line 80: s/ OpenShift BackupStorageLocations/ OADP-Velero BackupStorageLocations

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+## High-Level Design
+
+The **Virtual Machine Data Protection (VMDP)** feature introduces a OpenShift-native, client-server architecture that enables file-level backup and restore operations initiated from within OpenShift Virtual Machines. The system integrates seamlessly with OpenShift APIs and the existing OADP infrastructure, while preserving a clear separation of responsibilities between cluster administrators and VM users.
+
+### Architecture Overview
+
+The solution consists of the following core components:
+
+1. **In-Guest Backup Client**
```
shubham-pampattiwar, line 65: I notice client installation is currently out of scope, would it make sense to include a brief installation framework in this design, or is this planned as a separate follow-up? Since installation method can influence architectural decisions, considering both together might help avoid future design constraints.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+The **Virtual Machine Data Protection (VMDP)** feature introduces a OpenShift-native, client-server architecture that enables file-level backup and restore operations initiated from within OpenShift Virtual Machines. The system integrates seamlessly with OpenShift APIs and the existing OADP infrastructure, while preserving a clear separation of responsibilities between cluster administrators and VM users.
+
+### Architecture Overview
+
+The solution consists of the following core components:
+
+1. **In-Guest Backup Client**  
+   A lightweight, statically linked Go-based CLI tool, built from the standard Kopia CLI source code and fully compatible with Kopia APIs. It is designed to run entirely in user space (without requiring elevated privileges) and can be easily downloaded and executed—even outside standard installation paths. This client allows users to perform file-level backup and restore operations from within the guest VM, authenticating with the Kopia server using user-specific credentials.
+
+2. **Backup Service (Kopia Server)**
```
shubham-pampattiwar, line 68: 1. Ok so this backup server acts as a proxy between the VM and the object storage, right ?
2. Also, I see that there are no constraints on the number of backup servers being deployed ? (IMO we should keep it simple as one per cluster or one per namespace or one per VM, some kind of policy in place would be helpful in monitoring and managing things/workflows)
3. Where will the backup server be deployed ? Which node and which namespace ?

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+3. **Custom Resource Definitions (CRDs) and Controller**  
+   This component extends the existing OADP operator by adding a new controller and introducing two new CRDs:
+
+   - **BackupStorageLocationServer (BSLS)** — represents the deployment and configuration of a Kopia Backup Service instance within the OpenShift cluster. The BSLS points either directly to a Velero `BackupRepository` or to a new CRD `BackupStorageLocationRepository`.
+
+   - **BackupStorageLocationRepository (BSLR)** — represents the Kopia repository backing the file-level backups for a given Backup Storage Location. The BSLR points back to a `BackupStorageLocation` (BSL). The BSLR concept is similar to a Kopia BackupRepository but is managed by OADP.
```
shubham-pampattiwar, line 78: Hmm So we are introducing BSLR so as to overcome the limitation of BR being NS scoped and one particular ns can have multiple VMs ? Thus giving more fine grained control to users (vm admins/users)

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
```
shubham-pampattiwar, line 79: Have we explored the possibility of overcoming the limitations Velero BR ? **(I know that would be an upstream task but just wanted to know your thoughts on this, I am thinking on the lines where we could directly leverage the Velero BR to enable file level B/R on Velero backups directly)**

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+5. **Credential Management**  
+There are three distinct types of credentials involved in the system:
+
+   - **Repository Credentials:** These are the passwords required to open and operate a specific Kopia repository. They are securely stored as OpenShift `Secrets` and are associated with either a *Velero Backup Repository (BR)* or a *BackupStorageLocationRepository (BSLR)*. Possession of these credentials enables clients to unlock and access the repository’s metadata and content.
+
+   - **Storage Backend Credentials:** Credentials (e.g., S3 access keys) required by Kopia to access the object storage where the repository resides. These are associated with the *BackupStorageLocation (BSL)* and are needed for repository creation and operation. They are not exposed to end-users, as the Kopia Server (BSLS) acts as a secure proxy, handling storage access internally.
+
+   - **User Access Credentials:** These credentials are linked to the *BackupStorageLocationServer (BSLS)* and represent user-specific accounts on the Kopia server. They authenticate users connecting to the Kopia service (BSLS service), enforcing isolation and access control so users can only access their data.
```
shubham-pampattiwar, line 98: Ok so we are having user access credentials for authentication to backup server from VM then why not keep only one backup server for the whole cluster, deployed in OADP install NS ? (I am thinking of easier management and lower resource consumption too).

shubham-pampattiwar, line 98: So cluster admin creates the secrets consisting of user creds in OADP install NS ? And then the cluster admin shares these creds with the VM user/VM admin via some other channel ? Lets clarify this in the design.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+The key difference between a Velero Backup Repository (BR) and a BackupStorageLocationRepository (BSLR) lies in where their metadata is managed. BR stores repository metadata directly in the backup storage backend (e.g., S3), enabling automatic synchronization and recovery by simply reusing the storage location. In contrast, BSLR manages metadata within OpenShift via Kubernetes CRDs, decoupling it from the storage backend. This design provides more explicit control over repository provisioning, lifecycle, and security.
+
+5. **Credential Management**  
+There are three distinct types of credentials involved in the system:
+
+   - **Repository Credentials:** These are the passwords required to open and operate a specific Kopia repository. They are securely stored as OpenShift `Secrets` and are associated with either a *Velero Backup Repository (BR)* or a *BackupStorageLocationRepository (BSLR)*. Possession of these credentials enables clients to unlock and access the repository’s metadata and content.
```
shubham-pampattiwar, line 94: 1. Who creates the Repository credentials ? I am assuming cluster admins ? 
2. Also where do the BSLR CR reside ? (which NS)
3. The repo creds and BSLR would be in the same NS, right ?

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+4. **Credential Synchronization**  
+   Sync user credentials stored as Kubernetes Secrets to the Kopia server, enabling authenticated access for VM users.
+
+5. **Configuration Updates**  
+   Detect changes in the BSLS or BSLR resources and apply updates to the Kopia server configuration accordingly.
+
+Through these reconciliation steps, the BSLS controller ensures the Kopia server is securely configured, operational, and aligned with cluster resource definitions.
+
+### BackupStorageLocationServer CRD
```
shubham-pampattiwar, line 273: Should we prefix all the VMDP related CRDs with **VM** as a prefix ?

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+      secretUsers --> kopiaService
+
+      %% Kopia Server accesses repository in Object Storage
+      kopiaService --> repoData
+
+      %% User authentication happens locally on Kopia Server using Secrets
+      vmUser --> kopiaService
+   ```
+
+### Backup and Restore Workflow Summary
```
shubham-pampattiwar, line 154: I think the design should clarify what are the VM admin/user duties and Cluster admin duties ?
 - Who manages what in the VMDP lifecycle
  - Permission boundaries between different admin roles
  - Operational responsibilities for troubleshooting
  - Self-service capabilities vs admin-required tasks

IMHO just adding a simple step-by-step workflow starting from installation through end-to-end use cases, clearly identifying actors (VM users vs cluster admins) at each step would go a long for this design doc 😄

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+
+#### Prerequisites
+
+- Internal OpenShift networking must allow VMs to connect to the Backup Service deployed in the OADP namespace.
```
shubham-pampattiwar, line 158: Should we include preflight connectivity checks to validate VM-to-backup-server communication before backup operations? Consider:
  - Network connectivity tests (ping/telnet to backup server)
  - Service availability validation
  - Authentication pre-verification
  - Performance/bandwidth checks
  - DNS resolution validation
This would help troubleshoot deployment issues and provide better user experience when backup operations fail due to connectivity problems.

Hmm... Could be implemented as a `vmdp test-connection` command or automated health checks.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+| `endpoint`         | string  | Network address to reach the Kopia Server.  |
+| `tlsFingerprint`   | string  | SHA-256 fingerprint of the server TLS cert. |
+| `conditions`       | array   | List of current status conditions/events.   |
+
+---
+
+```yaml
+# Example BSLS CRD specification
+apiVersion: oadp.openshift.io/v1alpha1
+kind: BackupStorageLocationServer
```
shubham-pampattiwar, line 310: What container image will be used for this backup server ? Official kopia image or some custom OADP extended image ? 
I think we also add an image override via DPA for this server.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+- **Maintain a Secure, Zero-Trust Model**: Ensure that all client-server interactions are strongly authenticated and authorized using OpenShift-native mechanisms, minimizing risk and privilege exposure.
+- **Integrate Seamlessly with OADP**: Leverage existing OADP concepts and APIs (such as `BackupStorageLocation`) to ensure consistent behavior and ease of adoption by administrators.
+- **Deliver a Platform-Agnostic Solution**: Support major operating systems commonly used within guest VMs to maximize utility across enterprise environments.
+
+## High-Level Design
+
+The **Virtual Machine Data Protection (VMDP)** feature introduces a OpenShift-native, client-server architecture that enables file-level backup and restore operations initiated from within OpenShift Virtual Machines. The system integrates seamlessly with OpenShift APIs and the existing OADP infrastructure, while preserving a clear separation of responsibilities between cluster administrators and VM users.
+
+### Architecture Overview
```
shubham-pampattiwar, line 61: How do administrators enable VMDP? Should this integrate with DPA configuration or be standalone CRD management? Please clarify this aspect in the design.

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+    limits:
+      cpu: "1"
+      memory: "2Gi"
+status:
+  phase: Ready
+  endpoint: bsls-vm-backup-server.openshift-adp.svc.cluster.local:51515
+  tlsFingerprint: sha256:abcd1234ef5678...
+  conditions: []
+```
+
```
shubham-pampattiwar, line 338: Is this auth flow correct ? 
VM User -> [username/password] -> Backup Server -> [repo password] -> Repository

shubham-pampattiwar (commented):
Awesome work @mpryc ! Solid architecture and impressive design, this feature is gonna be GOLD. Added a couple of comments here and there. Thank you !!

docs/design/vm_data_protection-design.md
mpryc: Will add to the design as out of scope: 
```
No native mechanisms for enforcing storage quotas or caps at the repository level. This functionality may be addressed in future phases through cloud provider integration or Kopia enhancements.
```

docs/design/vm_data_protection-design.md
mpryc: CVE impact is one factor, but not the only reason for limiting UI use.
 - Web UI may create repository management issues across users. UI operates on repository-level access, not individual user-level access, so one user could disconnect or misconfigure a repository, affecting all other users. Or we would need to have some extra logic to disallow reconfiguring it which would introduce more development and testing resources. We may plan for it after this initial tech-preview.
 - Web UIs increase the attack surface and require ongoing security maintenance.
 - A CLI-first approach better supports self-service and automation, aligning with VM-level operations.

docs/design/vm_data_protection-design.md
mpryc: Interesting, we could have some data, but I think it would be more into two places 
 - kopia's ability to report size of individual repository. 
 - s3 size of a given bucket with prefix (similar to what example 5 in here has: https://docs.aws.amazon.com/cli/latest/reference/s3/ls.html)


This imo is outside of this design and can be a nice enhancement to the overall status of unified repo.

weshayutin (approved):

shubham-pampattiwar (approved):

docs/design/vm_data_protection-design.md
```diff
@@ -0,0 +1,502 @@
...
+    CLI->>CLI: Decrypt and reconstruct files on VM
+```
+
+## BackupStorageLocationServer (BSLS) Design and Reconciliation
+
+### BackupStorageLocationServer Overview
+
+The BackupStorageLocationServer (BSLS) is a Kubernetes Custom Resource that manages Kopia backup servers within the cluster, serving as a secure gateway between virtual machine (VM) users and the Kopia repository. 
+
+BSLS itself does not handle encryption; instead, encryption is performed client-side within each VM using Kopia’s client tools. Encrypted data is then transmitted securely over TLS to the Kopia server managed by BSLS. This approach ensures that all backup data remains encrypted in transit and at rest on the repository server.
```
kaovilai, line 248:
```suggestion
BSLS itself does not handle kopia snapshot backup file encryption; instead, encryption is performed client-side within each VM using Kopia’s client tools. Encrypted data is then transmitted securely over TLS to the Kopia server managed by BSLS. This approach ensures that all backup data remains encrypted in transit and at rest on the repository server.
```

kaovilai (approved):

PR #1867, opened 2025-08-04
Fix OADP-6459: Adding a backupLocation to DPA with an existing backupLocation name is not rejected

shubham-pampattiwar:
## Why the changes were made

The `ValidateBackupStorageLocations()` function was missing validation to detect duplicate backup location names before BSL object creation.

Added early validation in the BSL validation process to detect and reject duplicate backup location names, preventing the problematic state from occurring.

-  Create a DPA with duplicate backup location names to verify validation:
```yaml
  metadata:
    name: test-dpa
    namespace: openshift-adp
  spec:
    configuration:
      velero:
        defaultPlugins:
          - aws
          - azure
    backupLocations:
      - name: "duplicate-name"  # First location
        velero:
          provider: aws
          default: true
          objectStorage:
            bucket: test-aws-bucket
            prefix: velero/backups
          config:
            region: us-east-1
          credential:
            name: cloud-credentials
            key: cloud
      - name: "duplicate-name"  # Duplicate name - should fail
        velero:
          provider: azure
          default: false
          objectStorage:
            bucket: test-azure-bucket
            prefix: velero/backups
          config:
            resourceGroup: test-rg
            storageAccount: test-sa
          credential:
            name: azure-credentials
            key: cloud
```
  Expected behavior:
  - DPA creation/update should be rejected with validation error
  - Error message: backup location name 'duplicate-name' is duplicated. Backup location names must be unique
  - No BackupStorageLocation objects should be created
    
🤖 Generated with https://claude.ai/code

Co-Authored-By: Claude noreply@anthropic.com

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -19,6 +19,15 @@ import (
 	"github.com/openshift/oadp-operator/pkg/storage/aws"
 )
 
+// getBSLName generates the BackupStorageLocation name for a given backup location spec and index.
+// It returns the user-provided name if specified, otherwise generates a name using the DPA name and index.
+func (r *DataProtectionApplicationReconciler) getBSLName(bslSpec *oadpv1alpha1.BackupLocation, index int) string {
+	if bslSpec.Name != "" {
+		return bslSpec.Name
+	}
+	return fmt.Sprintf("%s-%d", r.NamespacedName.Name, index+1)
+}
+
 func (r *DataProtectionApplicationReconciler) ValidateBackupStorageLocations() (bool, error) {
 	// Ensure BSL is a valid configuration
 	// First, check for provider and then call functions based on the cloud provider for each backupstoragelocation configured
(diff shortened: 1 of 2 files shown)


kaovilai (commented):
1. Consider extracting BSL name generation logic: The BSL naming logic is
   duplicated between ValidateBackupStorageLocations and
  ReconcileBackupStorageLocations. Consider extracting this into a helper
  function:
```go
  func (r *DataProtectionApplicationReconciler) getBSLName(bslSpec 
  *BackupLocation, index int) string {
      if bslSpec.Name != "" {
          return bslSpec.Name
      }
      return fmt.Sprintf("%s-%d", r.NamespacedName.Name, index+1)
  }
```
  2. Add validation for empty names: While the current logic handles empty
  names by generating them, consider validating that user-provided names
  are not empty strings to avoid confusion.

kaovilai (approved):

weshayutin (approved):

PR #1897, opened 2025-08-07
OADP-6500: allow nonadmin.enabled.false

weshayutin:
## Why the changes were made

Only one dpa w/ nonadmin.enabled true is allowed.  Other DPA's should be allowed to have nonadmin.enabled.false w/o error

* install 2 oadp's in different namespaces
* set the first to nonadmin.enabled: true
* Create the second dpa w/ nonadmin.enabled.false
  * you should not see an error.

### DIFF
--- a/internal/controller/validator_test.go
+++ b/internal/controller/validator_test.go
@@ -1508,7 +1508,7 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
 			messageErr: "only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is already configured and installed in test-another-ns namespace",
 		},
 		{
-			name: "[valid] DPA CR: NonAdmin.Enable is false with existing NAC deployment",
+			name: "[valid] DPA CR: NonAdmin.Enable is true with another DPA having NonAdmin.Enable false",
 			dpa: &oadpv1alpha1.DataProtectionApplication{
 				ObjectMeta: metav1.ObjectMeta{
 					Name:      "test-DPA-CR",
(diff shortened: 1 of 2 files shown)


internal/controller/validator_test.go
```diff
@@ -1507,6 +1507,49 @@ func TestDPAReconciler_ValidateDataProtectionCR(t *testing.T) {
 			wantErr:    true,
 			messageErr: "only a single instance of Non-Admin Controller can be installed across the entire cluster. Non-Admin controller is already configured and installed in test-another-ns namespace",
 		},
+		{
+			name: "[valid] DPA CR: NonAdmin.Enable is false with existing NAC deployment",
```
kaovilai:
```suggestion
			name: "[valid] DPA CR: NonAdmin.Enable is true with another DPA having NonAdmin.Enable false",
```

kaovilai (changes requested):

shubham-pampattiwar (approved):

kaovilai (approved):
/lgtm

weshayutin:
hrm.. hcp failing w/ 
```
          s: "failed to install operator multicluster-engine: failed to get PackageManifest for operator multicluster-engine: packagemanifests.packages.operators.coreos.com \"multicluster-engine\" not found",
```

Ah.. this is an optional job anyway :)
https://github.com/openshift/release/blob/master/ci-operator/config/openshift/oadp-operator/openshift-oadp-operator-oadp-dev__4.20.yaml#L140-L141

mpryc (approved):

mpryc:
This may not entirely fix the cli-aws job, but at least it won't fail on velero namespace check:
  https://github.com/openshift/oadp-operator/pull/1904

kaovilai:
> This may not entirely fix the cli-aws job, but at least it won't fail on velero namespace check: #1904

The "failing namespace check" is an err that is printed to stdout and does not cause `exit 1`, and is not the root cause of ci failure.
https://github.com/openshift/oadp-operator/pull/1899 fixes cli-aws job.

PR #1901, opened 2025-08-11
make bundle and VERSION update - after rebase

mpryc:
Updated VERSION in the Makefile to 1.5.1
Ran: make bundle

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -295,7 +295,7 @@ metadata:
     operatorframework.io/arch.arm64: supported
     operatorframework.io/arch.ppc64le: supported
     operatorframework.io/arch.s390x: supported
-  name: oadp-operator.v1.5.0
+  name: oadp-operator.v1.5.1
   namespace: openshift-adp
 spec:
   apiservicedefinitions: {}
@@ -864,13 +864,6 @@ spec:
           - update
           - watch
         - apiGroups:
-          - ""
-          resources:
-          - persistentvolumerclaims
-          - persistentvolumes
-          verbs:
-          - get
-        - apiGroups:
           - apps
           resources:
           - daemonsets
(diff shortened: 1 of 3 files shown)


bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -1015,19 +1008,6 @@ spec:
           - velero.io
           resources:
           - '*'
-          - backuprepositories
```
mpryc, line 1018: The '*' which is one line above should cover the removals.

config/rbac/role.yaml
```diff
@@ -187,19 +180,6 @@ rules:
   - velero.io
   resources:
   - '*'
-  - backuprepositories
```
mpryc, line 190: The '*' above should cover all the removals.

config/rbac/role.yaml
```diff
@@ -35,13 +35,6 @@ rules:
   - patch
   - update
   - watch
-- apiGroups:
-  - ""
-  resources:
-  - persistentvolumerclaims
```
mpryc, line 41: persistentvolumerclaims - I think this could have been mistake

mpryc: persistentvolumes is something more problematic as it's not covered by other rbac privs here and I am not sure if we should give that somewhere.

mpryc:
The error in the `unit-tests` comes from the `make bundle-isupdated` target, but I am little confused why this happens. I've tried to reproduce the issue on my local computer with the same exact source code `gh pr checkout 1901` and the `make bundle-isupdated` does not give me this error.

mpryc:
Note: non-admin-backup has nothing to do with the bundle manifest change.

config/rbac/role.yaml
mpryc: Those are not present in the oadp-dev branch of oadp-operator directory.

mpryc:
Fixed in https://github.com/openshift/oadp-operator/pull/1903

PR #1899, opened 2025-08-08
ci change to avoid interactive prompt

Joeavaikath:
## Why the changes were made

Moves the install to within the cli e2e test suite.

## How to test the changes made

cli e2e should run without any issues. ci-Dockerfile only has kubectl install now.

PR #1904, opened 2025-08-14
OADP CLI check - do not attempt to connect to server

mpryc:
Use `--client` with `kubectl oadp version` in the CLI availability check to ensure it only reports the client version without attempting to connect to the cluster. This prevents errors when the `velero` namespace is not yet present during test setup.

The example error:
 https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/1897/pull-ci-openshift-oadp-operator-oadp-dev-4.19-e2e-test-cli-aws/1955901235508809728/artifacts/e2e-test-cli-aws/e2e/build-log.txt

### DIFF
--- a/tests/e2e/backup_restore_cli_suite_test.go
+++ b/tests/e2e/backup_restore_cli_suite_test.go
@@ -198,7 +198,7 @@ var _ = ginkgo.Describe("Backup and restore tests via OADP CLI", ginkgo.Label("c
 	ginkgo.BeforeAll(func() {
 		// Verify OADP CLI is available (should be installed in Docker image)
 		log.Print("Verifying OADP CLI is available...")
-		cmd := exec.Command("kubectl", "oadp", "version")
+		cmd := exec.Command("kubectl", "oadp", "version", "--client-only")
 		output, err := cmd.CombinedOutput()
 		if err != nil {
 			ginkgo.Fail(fmt.Sprintf("OADP CLI not available: %v, output: %s", err, string(output)))


tests/e2e/backup_restore_cli_suite_test.go
```diff
@@ -198,7 +198,7 @@ var _ = ginkgo.Describe("Backup and restore tests via OADP CLI", ginkgo.Label("c
 	ginkgo.BeforeAll(func() {
 		// Verify OADP CLI is available (should be installed in Docker image)
 		log.Print("Verifying OADP CLI is available...")
-		cmd := exec.Command("kubectl", "oadp", "version")
+		cmd := exec.Command("kubectl", "oadp", "version", "--client-only")
```
mpryc, line 201: It's how currently oadp client is implemented but IMO the flag should be `--client` to match what is in the kubectl:

https://kubernetes.io/docs/reference/kubectl/#operations

kaovilai: It's more a parallel to how velero CLI is doing, not exactly kubectl

kaovilai: I think keep it same as velero CLI, and propose velero CLI upstream match kubectl in the future.

kaovilai:
From reviewing recent CLI e2e PR, failures are from a path issue and solved in https://github.com/openshift/oadp-operator/pull/1899. Even with velero version trying to connect to server, it is only printing as error, but is not the root cause of recent cli CI failures.

We can leave this PR open for discussion if we rather this check be purely client-only for cleaner logs/semantics.

weshayutin:
```
whayutin@fedora:~/OPENSHIFT/git/OADP/oadp-operator$ remote_velero version
Client:
	Version: konveyor-dev
	Git commit: -
Server:
	Version: konveyor-dev
whayutin@fedora:~/OPENSHIFT/git/OADP/oadp-operator$ oc oadp version
Client:
	Version: 
	Git commit: -
Server:
	Version: konveyor-dev
```

PR #1929, opened 2025-08-25
OADP-1.5: Makefile: ensure the various versions are very visible

weshayutin:
## Why the changes were made

Make the versions of tools we use to build more visible

make versions
make deploy-olm

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,7 +3,7 @@
 DEFAULT_VERSION := 1.5.0
 VERSION ?= $(DEFAULT_VERSION)
 OPERATOR_SDK_VERSION ?= v1.34.2
-ENVTEST_K8S_VERSION = 1.32 # Kubernetes version from OpenShift 4.19.x
+ENVTEST_K8S_VERSION = 1.32 #refers to the version of kubebuilder assets to be downloaded by envtest binary # Kubernetes version from OpenShift 4.19.x
 GOLANGCI_LINT_VERSION ?= v1.55.2
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.16.5
--- a/Makefile
+++ b/Makefile
@@ -8,6 +8,7 @@ GOLANGCI_LINT_VERSION ?= v1.55.2
 KUSTOMIZE_VERSION ?= v5.2.1
 CONTROLLER_TOOLS_VERSION ?= v0.16.5
 OPM_VERSION ?= v1.23.0
+BRANCH_VERSION = oadp-1.5
 PREVIOUS_CHANNEL ?= oadp-1.4
 PREVIOUS_CHANNEL_GO_VERSION ?= 1.22
 # Extract the toolchain directive from go.mod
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -434,6 +445,7 @@ deploy-olm: THIS_OPERATOR_IMAGE?=ttl.sh/oadp-operator-$(GIT_REV):1h # Set target
 deploy-olm: THIS_BUNDLE_IMAGE?=ttl.sh/oadp-operator-bundle-$(GIT_REV):1h # Set target specific variable
 deploy-olm: DEPLOY_TMP:=$(shell mktemp -d)/ # Set target specific variable
 deploy-olm: undeploy-olm ## Build current branch operator image, bundle image, push and install via OLM. For more information, check docs/developer/install_from_source.md
+	@make versions
```
kaovilai, line 526: I would put under test target as well.

weshayutin: k.. will do

Makefile
```diff
@@ -108,6 +109,22 @@ all: build
...
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "VERSION" "$(VERSION)" "bundle, catalog-build, deploy-olm, undeploy-olm"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "OPERATOR_SDK_VERSION" "$(OPERATOR_SDK_VERSION)" "operator-sdk"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "ENVTEST_K8S_VERSION" "$(ENVTEST_K8S_VERSION)" "test"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "GOLANGCI_LINT_VERSION" "$(GOLANGCI_LINT_VERSION)" "golangci-lint"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "KUSTOMIZE_VERSION" "$(KUSTOMIZE_VERSION)" "kustomize"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "CONTROLLER_TOOLS_VERSION" "$(CONTROLLER_TOOLS_VERSION)" "controller-gen"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "OPM_VERSION" "$(OPM_VERSION)" "opm"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "PREVIOUS_CHANNEL" "$(PREVIOUS_CHANNEL)" "catalog-test-upgrade"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "PREVIOUS_CHANNEL_GO_VERSION" "$(PREVIOUS_CHANNEL_GO_VERSION)" "catalog-test-upgrade"
+	@printf "\033[36m%-30s\033[0m %-20s %s\n" "GO_TOOLCHAIN_VERSION" "$(GO_TOOLCHAIN_VERSION)" "(informational only)"
```
kaovilai: seems this only display versions var used.
the crucial mismatch is local bin. in the next gen of kubebuilder what they do is ./bin/cliname-versionString instead of ./bin/localbinNoVersion.

We should be running version command against local bin available

weshayutin: ok.. I'll do that now vs. follow up :)

Makefile
```diff
@@ -104,10 +106,50 @@ all: build
...
+define CHECK_TOOL_VERSION
+	@printf "\n\033[1m$(1) Version Check:\033[0m\n"
+	@if [ -f "$(2)" ]; then \
+		INSTALLED_VERSION=$$($(3) || echo "unknown"); \
+		EXPECTED_VERSION="$(4)"; \
+		if [ "$$INSTALLED_VERSION" = "$$EXPECTED_VERSION" ]; then \
+			printf "\033[32m%-30s\033[0m %-20s %s\n" "$(6)" "$$INSTALLED_VERSION" "✓ matches Makefile"; \
+		else \
+			printf "\033[33m%-30s\033[0m %-20s %s\n" "$(6)" "$$INSTALLED_VERSION" "⚠ differs from Makefile ($$EXPECTED_VERSION)"; \
+			exit 1; \
```
kaovilai: when differ can we just download correct version for dev?

weshayutin: I'm happy to change this to what the team wants.  instead of exit 1, we would add $(MAKE) $(5) there as well.   I think that makes sense.  If someone wants a diff version it's easy and clear to change at the top of the makefile or export.

kaovilai (approved):

sseago (approved):

PR #1935, opened 2025-09-03
Update CSV for 1.5.1

rayfordj:
With the move to Konflux, OADP will need to keep better update of CSV because we don't have the same flexibility that we did using CPaaS's render_templates.

This change aligns CSV with what's currently rendered via template during CPaaS build-pipeline job downstream.

### DIFF
--- a/bundle/manifests/oadp-operator.clusterserviceversion.yaml
+++ b/bundle/manifests/oadp-operator.clusterserviceversion.yaml
@@ -268,7 +268,7 @@ metadata:
       Cloud Provider,Developer Tools,Modernization & Migration,OpenShift Optional,Storage
     certified: "false"
     containerImage: quay.io/konveyor/oadp-operator:oadp-1.5
-    createdAt: "2025-02-28T20:03:54Z"
+    createdAt: "2025-09-03T00:37:40Z"
     description: OADP (OpenShift API for Data Protection) operator sets up and installs
       Data Protection Applications on the OpenShift platform.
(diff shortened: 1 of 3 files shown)


weshayutin:
> /test 4.20-e2e-test-hcp-aws

@rayfordj that test is optional, fyi

weshayutin (approved):

bundle/manifests/oadp-operator.clusterserviceversion.yaml
```diff
@@ -1287,4 +1287,5 @@ spec:
     name: mustgather
   - image: quay.io/konveyor/oadp-non-admin:oadp-1.5
     name: non-admin-controller
-  version: 1.5.0
+  replaces: oadp-operator.v1.5.0
```
shubham-pampattiwar, line 1290: Hmm, Did we not just rely on `olm.skipRange` ? Curious why we need to add the `replaces` spec.

kaovilai: must be konflux

kaovilai (approved):

bundle/manifests/oadp-operator.clusterserviceversion.yaml
rayfordj: Downstream, we've always used the combination of `olm.skipRange` & `replaces` for OADP (MTC uses `olm.skipRange` & `skips`)

PR #1936, opened 2025-09-03
OADP-6652: Fix unnecessary secret updates and logging in STS flow

kaovilai:
The operator was repeatedly logging "Secret already exists, updating"
and "Following standardized STS workflow, secret created successfully"
even when the secret content hadn't changed. This was happening because
the CloudStorage controller calls STSStandardizedFlow() on every
reconciliation, which always attempted to create the secret first,
then caught the AlreadyExists error and performed an update.

Changed the approach to:
- First check if the secret exists
- Compare existing data with desired data
- Only update when there are actual differences
- Skip updates and avoid logging when content is identical
- Changed CloudStorage controller to use Debug level and more accurate
  message when STS secret is available (not necessarily created)

This eliminates unnecessary API calls to the Kubernetes cluster and
reduces noise in the operator logs.

### DIFF
--- a/pkg/credentials/stsflow/stsflow.go
+++ b/pkg/credentials/stsflow/stsflow.go
@@ -216,7 +216,7 @@ func CreateOrUpdateSTSSecretWithClients(setupLog logr.Logger, secretName string,
 func CreateOrUpdateSTSSecretWithClientsAndWait(setupLog logr.Logger, secretName string, credStringData map[string]string, secretNS string, clientInstance client.Client, clientset kubernetes.Interface, waitForSecret bool) error {
 	// Create a secret with the appropriate credentials format for STS/WIF authentication
 	// Secret format follows standard patterns used by cloud providers
-	secret := corev1.Secret{
+	desiredSecret := corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      secretName,
 			Namespace: secretNS,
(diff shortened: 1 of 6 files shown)


pkg/credentials/stsflow/stsflow.go
```diff
@@ -226,48 +226,88 @@ func CreateOrUpdateSTSSecretWithClientsAndWait(setupLog logr.Logger, secretName
...
+
+		// Compare each key in credStringData
+		for key, desiredValue := range credStringData {
+			if existingValue, exists := existingData[key]; !exists || existingValue != desiredValue {
+				needsUpdate = true
+				break
+			}
+		}
+
+		if needsUpdate {
```
weshayutin, line 285: these statements look like a nice add.. pretty clear :)

weshayutin (dismissed):
/LGTM /approve

pkg/credentials/stsflow/stsflow.go
```diff
@@ -226,48 +226,88 @@ func CreateOrUpdateSTSSecretWithClientsAndWait(setupLog logr.Logger, secretName
...
+			// Some other error occurred while getting the secret
+			setupLog.Error(err, "unable to get secret resource")
+			return err
+		}
+	} else {
+		// Secret exists, check if update is needed
+		needsUpdate := false
+
+		// Check if labels need updating
+		if existingSecret.Labels == nil || existingSecret.Labels["oadp.openshift.io/secret-type"] != "sts-credentials" {
```
shubham-pampattiwar: Lets use constants here

shubham-pampattiwar (approved):

weshayutin (approved):

PR #1942, opened 2025-09-05
OADP-6669: Use CloudStorage creationSecret and config as fallback for BSL

kaovilai:
## Description

This PR enhances the CloudStorage integration in DataProtectionApplication (DPA) resources by implementing automatic fallback mechanisms for both credentials and configuration when using CloudStorage CRs.

### 1. Credential Fallback (commit aa93572)
When a DPA references a CloudStorage CR without explicit credentials, the BSL controller now automatically uses the CloudStorage's `creationSecret` as a fallback for authentication.

**Benefits:**
- Eliminates credential duplication between CloudStorage and DPA resources
- Simplifies configuration when using CloudStorage CRs
- Maintains backward compatibility with existing configurations

### 2. Configuration and Region Fallback (commit 345b342)
The BSL now inherits configuration values from the CloudStorage CR as fallback, including automatic region propagation.

**Implementation details:**
- CloudStorage CR's `config` field values are used as base configuration
- CloudStorage CR's `region` field is automatically added to BSL config
- DPA's CloudStorageLocation.Config values override CloudStorage CR values (higher priority)
- Supports all provider-specific configurations (AWS, Azure, GCP)

**Example use case:**
```yaml
# CloudStorage CR defines base configuration
kind: CloudStorage
metadata:
  name: my-storage
spec:
  provider: aws
  region: us-west-2
  config:
    serverSideEncryption: "AES256"
    s3ForcePathStyle: "true"
  creationSecret:
    name: cloud-creds
    key: credentials

---
# DPA can reference CloudStorage without duplicating config
metadata:
  name: my-dpa
spec:
  backupLocations:
    - cloudStorage:
        cloudStorageRef:
          name: my-storage
        # No need to specify credential or config - inherited from CloudStorage
        # Can optionally override specific values:
        config:
          profile: "custom"  # This overrides while keeping other CloudStorage configs
```

## Testing

- Added comprehensive test coverage for both credential and config fallback scenarios
- Updated existing tests to account for inherited region configuration
- All tests passing: `go test ./internal/controller/ -run TestDPAReconciler_ReconcileBackupStorageLocations`

## Impact

This change is backward compatible and improves the user experience by:
1. Reducing configuration duplication
2. Centralizing cloud storage settings in CloudStorage CRs
3. Maintaining flexibility to override at the DPA level when needed

Fixes #[issue_number] (if applicable)

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -182,7 +182,18 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 					}
 					bsl.Spec.Config["enableSharedConfig"] = "true"
 				}
-				bsl.Spec.Credential = bslSpec.CloudStorage.Credential
+				// Use DPA's CloudStorage credential if provided, otherwise fallback to CloudStorage's creationSecret
+				if bslSpec.CloudStorage.Credential != nil {
+					bsl.Spec.Credential = bslSpec.CloudStorage.Credential
+				} else {
+					// Use CloudStorage's creationSecret as the BSL credential
+					bsl.Spec.Credential = &corev1.SecretKeySelector{
+						LocalObjectReference: corev1.LocalObjectReference{
+							Name: bucket.Spec.CreationSecret.Name,
+						},
+						Key: bucket.Spec.CreationSecret.Key,
+					}
+				}
 				bsl.Spec.Default = bslSpec.CloudStorage.Default
 				bsl.Spec.ObjectStorage = &velerov1.ObjectStorageLocation{
 					Bucket: bucket.Spec.Name,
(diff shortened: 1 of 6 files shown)


shubham-pampattiwar (approved):

sseago (approved):

internal/controller/bsl.go
```diff
@@ -175,14 +175,50 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
...
+
+				// Add region from CloudStorage CR if specified
+				if bucket.Spec.Region != "" && bsl.Spec.Config == nil {
+					bsl.Spec.Config = make(map[string]string)
+				}
+				if bucket.Spec.Region != "" {
+					bsl.Spec.Config["region"] = bucket.Spec.Region
+				}
+
+				// Override with DPA's CloudStorageLocation config (higher priority)
```
mpryc, line 195: What happens when the Velero controller syncs back the bsl from the s3 metadata? Won't this cause conflicts in who is updating it?
Scenario:
 - bsl exists configured via CloudStorageconfig
 - dpa updates the CloudStorageconfigLocation
 
At this point velero may override the CloudStorageconfigLocation from DPA from previously configured CloudStorageconfig?

kaovilai: The controller writes to Spec.Default field in the ensureSingleDefaultBSL
  function (lines 222-270). Specifically:

  1. Lines 257-260: When there are multiple BSLs marked as default, the
  controller unsets the Default flag for all but the most recently created
  one:
  bsl.Spec.Default = false
  if err := r.client.Update(r.ctx, bsl); err != nil {
      return defaultFound, errors.Wrapf(err, "failed to unset default backup 
  storage location %q", bsl.Name)
  }

  All other interactions are read-only or status updates:
  - Lines 124-138: Only updates Status fields (LastValidationTime, Phase,
  Message)
  - Lines 141-152: Reads spec to validate the backup store
  - Lines 231-233: Reads Spec.Default to find default BSLs

So generally not a concern.

kaovilai: It's not new that we reconcile backupStorageLocation continuously from oadp-operator controller to ensure spec compliance with dpa.

mpryc (approved):
/lgtm

PR #1992, opened 2025-10-13
Virtual Machine File Restore design

mpryc:
This design presents a Kubernetes-native solution for recovering individual files from KubeVirt Virtual Machine backups created with OADP in OpenShift.

### DIFF
--- a/docs/design/vm_file_restore-design.md
+++ b/docs/design/vm_file_restore-design.md
@@ -1,0 +1,813 @@
+# File-Level Restore for Virtual Machines in OADP
+
+## Abstract
+
+This design presents a Kubernetes-native solution for recovering individual files from KubeVirt Virtual Machine backups created with OADP in OpenShift.
+
+Instead of restoring an entire VM to retrieve a single file, users can browse multiple backups and restore only the files they need - directly within Kubernetes, using familiar tools such as a web browser or command-line utilities like rsync.
+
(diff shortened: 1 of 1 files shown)


docs/design/vm_file_restore-design.md
```diff
@@ -0,0 +1,813 @@
...
+
+### Filesystem Access Mechanisms
+
+The controller uses a specialized container to mount and expose VM disk images from restored PVCs, making the VM’s filesystem contents available to file-serving sidecars in a consistent, read-only structure.
+
+#### Container Implementation
+
+**Base Technology**
+- **libguestfs** — Provides access to VM disk images without booting the virtual machine.  
+- **FUSE (Filesystem in Userspace)** — Allows safe, unprivileged filesystem mounting inside containers without access to host `/dev/kvm` or escalated pod privileges.
```
mpryc, line 499: @shubham-pampattiwar thank you for pointing offline this is not true.

FUSE enables mounting filesystems from userspace, but each mount(2) or umount(2) call is a kernel-level operation. Kubernetes enforces security restrictions on such operations, so mounting a FUSE filesystem is not permitted in a container unless it runs with elevated privileges (e.g., CAP_SYS_ADMIN or privileged: true).

Implementing this requires running the system in a privileged container, whether as an init container or the main container, which necessitates changes to the current design. Once privileged operations are needed, it is generally preferable to adopt a more performant and approach.

weshayutin (approved):
Nice, but very long :)

shubham-pampattiwar (approved):

sseago (approved):

PR #1998, opened 2025-10-20
oadp-1.5: fix(controller): prevent NodeAgent restarts from ESO metadata updates

kaovilai:
Fixes [ESO-234](https://issues.redhat.com//browse/ESO-234): NodeAgent was restarting every ~30s when External Secrets
Operator managed the cloud-credentials secret. ESO's metadata-only updates
were triggering unnecessary DPA reconciliations.

Changes:
- Updated labelHandler.Update() to skip reconciliation for Secret objects
  when only metadata changes (ResourceVersion, annotations, etc.)
- Added comprehensive unit tests for labelHandler covering all scenarios
- Maintains backward compatibility for non-Secret resources

This prevents unnecessary NodeAgent daemonset updates while preserving
reconciliation for actual data or label changes.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude <noreply@anthropic.com>

PR #1999, opened 2025-10-21
fix(controller): prevent NodeAgent restarts from ESO metadata updates

kaovilai:
Fixes [ESO-234](https://issues.redhat.com//browse/ESO-234): NodeAgent was restarting every ~30s when External Secrets
Operator managed the cloud-credentials secret. ESO's metadata-only updates
were triggering unnecessary DPA reconciliations.

Changes:
- Updated labelHandler.Update() to skip reconciliation for Secret objects
  when only metadata changes (ResourceVersion, annotations, etc.)
- Added comprehensive unit tests for labelHandler covering all scenarios
- Maintains backward compatibility for non-Secret resources

This prevents unnecessary NodeAgent daemonset updates while preserving
reconciliation for actual data or label changes.

Continuation of #1998 on a thawed oadp-dev branch.

### DIFF
--- a/internal/controller/dataprotectionapplication_controller.go
+++ b/internal/controller/dataprotectionapplication_controller.go
@@ -19,6 +19,7 @@ package controller
 import (
 	"context"
 	"os"
+	"reflect"
 
 	"github.com/go-logr/logr"
 	routev1 "github.com/openshift/api/route/v1"
@@ -193,6 +194,21 @@ func (l *labelHandler) Update(ctx context.Context, evt event.TypedUpdateEvent[cl
 	if evt.ObjectNew.GetLabels()[oadpv1alpha1.OadpOperatorLabel] == "" || dpaname == "" {
 		return
 	}
+
+	// For Secrets, check if only metadata changed (e.g., ResourceVersion updates from ESO)
+	// This prevents unnecessary reconciliations when external-secrets-operator updates metadata
+	if oldSecret, ok := evt.ObjectOld.(*corev1.Secret); ok {
+		if newSecret, ok := evt.ObjectNew.(*corev1.Secret); ok {
+			// Skip reconciliation if data, stringData, and labels haven't changed
+			// This filters out metadata-only updates (ResourceVersion, ManagedFields, etc.)
+			if reflect.DeepEqual(oldSecret.Data, newSecret.Data) &&
+				reflect.DeepEqual(oldSecret.StringData, newSecret.StringData) &&
+				reflect.DeepEqual(oldSecret.Labels, newSecret.Labels) {
+				return
+			}
+		}
+	}
+
 	q.Add(reconcile.Request{NamespacedName: types.NamespacedName{
 		Name:      dpaname,
 		Namespace: namespace,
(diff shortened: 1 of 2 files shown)


weshayutin (approved):
thanks @kaovilai !  Please do not cherry pick until 1.5.3. is GA :)

internal/controller/dataprotectionapplication_controller.go
```diff
@@ -193,6 +194,21 @@ func (l *labelHandler) Update(ctx context.Context, evt event.TypedUpdateEvent[cl
...
+
+	// For Secrets, check if only metadata changed (e.g., ResourceVersion updates from ESO)
+	// This prevents unnecessary reconciliations when external-secrets-operator updates metadata
+	if oldSecret, ok := evt.ObjectOld.(*corev1.Secret); ok {
+		if newSecret, ok := evt.ObjectNew.(*corev1.Secret); ok {
+			// Skip reconciliation if data, stringData, and labels haven't changed
+			// This filters out metadata-only updates (ResourceVersion, ManagedFields, etc.)
+			if reflect.DeepEqual(oldSecret.Data, newSecret.Data) &&
+				reflect.DeepEqual(oldSecret.StringData, newSecret.StringData) &&
+				reflect.DeepEqual(oldSecret.Labels, newSecret.Labels) {
```
kaovilai, line 206: @weshayutin although.. do we know if they are changing label every 30 seconds? are we supposed to ignore label changes?

weshayutin: Couple notes that I have would be.
1. I could be wrong but it's my understanding the external secrets operator is tech-preview.
2. I'm not the smartest guy in the world but it seems to me that ESO should check the secret but if no change is required, I think ANY metadata on the secret should remain UNCHANGED, hence why I moved the bug to ESO. 
3. I DO like being defensive in this case, and we either need to set it up and try/test or enquire w/ the ESO team re: labels.  
4. I don't think this is a priority for our attention atm based on my above understanding.  I could be wrong though.

kaovilai:
/retest

ai-retester: The end-to-end test `e2e-test-aws` failed because the `MySQL application KOPIA` test timed out. Specifically the `todolist` container in pod `todolist-1-vcz9l` showed `ContainersNotReady` condition, failing to start after an extended period. This seems to be the primary cause, even though the *mysql container triggered warning messages regarding failed liveness probes too.

kaovilai:
/retest

ai-retester: The e2e test failed because the `MySQL application KOPIA` test timed out and ultimately failed due to the `todolist` container in pod `todolist-1-9znqw` repeatedly failing readiness checks (containers not ready, "PodInitializing"). The underlying issues may have been with the liveness probe failing for the mysql database indicating i/o timeout .

internal/controller/dataprotectionapplication_controller.go
shubham-pampattiwar, line 206: Should we also check for annotations ?

kaovilai: We are explicitly avoiding reconcile on annotation changes.

shubham-pampattiwar (approved):

internal/controller/dataprotectionapplication_controller.go
kaovilai: eg. avoiding reconcile on https://github.com/external-secrets/external-secrets/blob/579135b5bb558b8b79f71bfc57f1a852968abac4/docs/provider/ibm-secrets-manager.md?plain=1#L283-L284

PR #2037, opened 2025-12-02
Design for virt qcow2 datamover

sseago:
## Why the changes were made

This PR adds a design document for the VM qcow2 incremental backup feature.

<wes> Background:
* https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md
* https://github.com/kubevirt/kubevirt/pull/14772
*

### DIFF
--- a/docs/design/kubevirt-datamover.md
+++ b/docs/design/kubevirt-datamover.md
@@ -1,0 +1,211 @@
+# Kubevirt incremental datamover
+
+In a kubevirt environment, one possibility for improving volume backup performance to allow more frequent backups would be to take incremental qcow2 backups using libvirt tools. Integrating this into OADP/Velero backups will require some new components to OADP (plugins, controller), as well as some modifications to the velero codebase.
+
+## Background
+
+Taking a VolumeSnapshot and then using kopia to process the entire volume and copy the required incremental changes into the Backup Storage Location (BSL) is a heavyweight process. Creating an incremental qcow2 backup for the same volume is generally a much more lightweight action. We want to make use of the existing Velero backup/restore process, with actual kubevirt incremental backup/restore happening via a new controller. For the moment, this will be referred to as the Kubevirt Datamover Controller. This action will be coordinated with Velero via existing infrastructure -- BackupItemActions (BIAs), RestoreItemActions (RIAs) and the DataUpload/DataDownload CRs. Initial implementation should require minimal changes to Velero, since Velero currently ignores DataUploads with `Spec.DataMover` set to something other than `velero`.
+
+## Goals
+
+- Back up and restore VM volumes using kubevirt incremental backup instead of Velero's built-in CSI datamover
+- Use existing Velero infrastructure to integrate this feature into regular velero backup and restore
+- Implementation based on kubevirt enhancement defined at <https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md>
+  - There is a follow-on design PR at <https://github.com/kubevirt/enhancements/pull/126/changes> although this mainly discusses push-mode, which is out of scope for the current design.
+
+## Non goals
+- Deep integration with velero data mover pluggability (this could be considered in the long-term though, which would minimize duplication of effort and enhance maintainability)
+- Using push mode with kubevirt.
+
+## Use cases
+- As a user I want to use OADP to trigger backups that will back up volume data using kubevirt tooling rather than CSI snapshots
+  - Volume backups will be incremental when possible (first backup for a given volume will be full, subsequent backups for that same volume will be incremental)
+  - Assuming that kubevirt incremental volume backups should be much faster than CSI snapshots followed by incremental kopia snapshot copy, the expectation is that users might run kubevirt datamover backups more frequently than they would for CSI-based backups.
+  - If users are backing up more frequently with this method, they should ensure that they are not using CSI snapshots or fs-backup via the resource/volume policy configuration in the backup.
+
+```
+  Key Differences between CSI and KubeVirt incremental snapshots
+  ┌────────────────────┬──────────────────────────────────────────┬────────────────────────────────────────────┐
+  │       Aspect       │               CSI Approach               │          KubeVirt qcow2 Approach           │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ Layer              │ Storage (CSI driver)                     │ Hypervisor (QEMU/libvirt)                  │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ Snapshot mechanism │ CSI VolumeSnapshot                       │ VirtualMachineBackup CR                    │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ Incremental        │ Kopia deduplication (scans whole volume) │ True block-level CBT (only changed blocks) │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ Data mover         │ Velero's node-agent + kopia              │ New OADP controller + qemu-img             │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ VM awareness       │ None (just sees PVC)                     │ Full (knows it's a VM disk)                │
+  └────────────────────┴──────────────────────────────────────────┴────────────────────────────────────────────┘
+```
+
+## High-Level design
+
+### Upstream velero changes
+- Update velero volume policy model to allow unrecognized volume policy actions. Velero would treat unrecognized actions as "skip" (i.e. no snapshot or fs-backup), but the kubevirt datamover could only act if the policy action is "kubevirt".
+- In `pkg/restore/actions/dataupload_retrieve_action.go` and in `DataDownload` we need to add SnapshotType.
+
+### BackupItemAction/RestoreItemAction plugins
+- VirtualMachine BIA plugin
+  - The plugin will check whether the VirtualMachine's `status.ChangedBlockTracking` is `Enabled`
+  - The plugin must also determine whether the VM is running, since offline backup is not supported in the initial release.
+    - If QEMU backup is enabled, the next question is whether the user wants to use the kubevirt datamover for this VM's volumes. We will use volume policies for this, although it's a bit more complicated since a VM could have multiple PVCs. If at least one PVC for the VM has the "kubevirt" policy action specified, and no PVCs in this VM have other non-skip policies (i.e. "snapshot", etc.) then we'll use the kubevirt datamover
+	- In addition, SnapshotMoveData must be true
+    - Iterate over all PVCs for the VM
+    - If any PVC has an action other than "kubevirt" or "skip", exit without action
+    - If at least one PVC has an action of "kubevirt", then use the kubevirt datamover
+  - This plugin will create a DataUpload with `Spec.SnapshotType` set to "kubevirt" and `Spec.DataMover` set to "kubevirt"
+    - Note that unlike the built-in datamover, this DataUpload is tied to a VM (which could include multiple PVCs) rather than to a single PVC.
+  - An annotation will be added to the DataUpload identifying the VirtualMachine we're backing up.
+  - Add `velerov1api.DataUploadNameAnnotation` to VirtualMachine
+  - OperationID will be created and returned similar to what's done with the CSI PVC plugin, and the async operation Progress method will report on progress based on the DU status (similar to CSI PVC plugin)
+- VirtualMachine RIA plugin
+  - (further investigation still needed on exactly what's needed here)
+  - Similar in functionality to csi PVC restore action
+  - Create temporary PVC (needed, or is this handled by controller?)
+  - Create DD based on DU annotation and DU ConfigMap (this may be handled already by existing DU RIA)
+- VirtualMachineBackup/VirtualMachineBackupTracker RIA plugin
+  - Simple RIA that discards VMB/VMBT resources on restore
+  - We don't want to restore these because they would kick off another VMBackup action.
+
+### Kubevirt Datamover Controller
+- Responsible for reconciling DataUploads/DataDownloads where `Spec.DataMover` is "kubevirt"
+- Configurable concurrency limits: concurrent-vm-backups and concurrent-vm-datauploads
+- DataUpload reconciler (backup):
+  - create the (temporary) PVC.
+  - identify the VirtualMachine from the PVC metadata.
+  - From the BSL, identify the latest checkpoint for this VM and confirm that all of the related qcow2 files are available.
+  - create/update the VirtualMachineBackupTracker CR with `source` set to the VirtualMachine, and `status.latestCheckpoint.name` set to match the latest checkpoint we have for this vm. Note that it might be better if we could avoid hacking the VMBT status and instead have a field on the VM -- something like `spec.overrideCheckpoint` to use a specific checkpoint if the one in the VMBT doesn't match the latest we have in the BSL.
+  - create the VirtualMachineBackup CR with `source` set to the VirtualMachineBackupTracker, `pvcName` set to the temporary PVC, and (optionally)`forceFullBackup`, set to `true` to force a full backup
+  - Wait for VMBackup to complete (monitoring status)
+  - Launch kubevirt datamover pod mounting the temporary PVC with the qcow2 file(s) from the backup.
+    - This pod needs to be running a command that will do the datamover operation from pvc to object storage
+	- The datamover pod functionality should be built into the same image as the kubevirt-datamover-controller pod image.
+  - Copy the new file to object storage (see [Where to store qcow2 files](#wherehow-to-store-qcow2-files-and-metadata) below)
+  - Save any required metadata to identify the stored data (collection of qcow2 pathnames/checkpoints, etc.), along with identifying the backup and VirtualMachine they're associated with. Save this metadata file as well (see [Where to store qcow2 files](#wherehow-to-store-qcow2-files-and-metadata) below)
+    - We need to properly handle cases where we attempt an incremental backup but a full backup is taken instead (checkpoint lost, CSI snapshot restore since last checkpoint, VM restart, etc.)
+    - Aborted backups also need to be handled (resulting in a failed PVC backup on the Velero side)
+- DataDownload reconciler (restore)
+  - (this area is less well defined so far, since the kubevirt enhancement doesn't go into as much detail on restore)
+  - We will need a temporary PVC for pulling qcow2 images from object store (if we're restoring the empty temp PVC from backup, that might work here)
+  - We also need PVCs created for each VM disk we're restoring from qcow2 images.
+  - We'll need to create another datamover pod here which will do the following:
+    - pod will have temp PVC mounted, as well as PVCs mounted for each vm disk we're creating.
+    - pod running command/image will first get the list of qcow2 files to pull from object storage
+    - once we have the qcow2 full backup and incremental files from object store, repeatedly call `qemu-img rebase -b fullbackup.qcow2 -f qcow2 -u incrementalN.qcow2` for each incremental backup, in order
+    - Then, convert qcow2 image to raw disk image: `qemu-img convert -f qcow2 -O raw fullbackup.qcow2 restored-raw.img`
+    - Finally, write this image to the PVC which contains the VM disk
+    - (repeat process for each disk if VM has multiple disks to restore)
+    - Note that the various `qemu-img` actions might eventually be combined into a single kubevirt API call, but for the moment this would need to be done manually.
+  - Once datamover pod has restored the VM disks, it will exit and the VirtualMachine can launch with these disks (following the velero datamover model where the temporary PVC is deleted, retaining the PV, which then binds to the VM's PVCs may work here). The temporary PVC (containing the qcow2 images, not the restored VM disk image) should be completely removed at this point, including PV content.
+
+### Kubevirt datamover backup data/metadata
+
+The kubevirt datamover data will be stored in the BSL using a prefix derived from `BSLPrefix+"-kubevirt-datamover"`.
+
+The directory structure will be as follows:
+```
+  <bsl-prefix>-kubevirt-datamover/
+  ├── manifests/
+  │   └── <velero-backup-name>/
+  │       ├── <vm-name>.json                       # Per-backup-per-vm manifest
+  │       └── index.json                           # Per-backup manifest
+  └── checkpoints/
+      └── <namespace>/
+          └── <vm-name>/
+              ├── <checkpoint-id>/                 # checkpoint dir
+              │   └── <vmb-name>-<disk-name>.qcow2 # Actual qcow2 files
+              └── index.json                       # Per-VM index file
+```
+Example of a Per-VM Index file:
+```
+Per-VM Index (checkpoints/<ns>/<vm-name>/index.json):
+
+{
+  "vmName": "my-vm",
+  "namespace": "default",
+  "checkpoints": [
+    {
+      "id": "cp-001",
+      "type": "full",
+      "timestamp": "2025-01-10T10:00:00Z",
+      "vmBackup": "vmb-001",
+      "files": ["vmb-001-disk0.qcow2", "vmb-001-disk1.qcow2"],
+      "pvcs": ["my-vm-pvc-0", "my-vm-pvc-1"],
+      "referencedBy": ["backup-2025-01-10", "backup-2025-01-11", "backup-2025-01-12"]
+    },
+    {
+      "id": "cp-002",
+      "type": "incremental",
+      "parent": "cp-001",
+      "timestamp": "2025-01-11T10:00:00Z",
+      "vmBackup": "vmb-002",
+      "files": ["vmb-002-disk0.qcow2", "vmb-002-disk1.qcow2"],
+      "pvcs": ["my-vm-pvc-0", "my-vm-pvc-1"],
+      "referencedBy": ["backup-2025-01-11", "backup-2025-01-12"]
+    },
+    {
+      "id": "cp-003",
+      "type": "incremental",
+      "parent": "cp-002",
+      "timestamp": "2025-01-12T10:00:00Z",
+      "vmBackup": "vmb-003",
+      "files": ["vmb-003-disk0.qcow2", "vmb-003-disk1.qcow2"],
+      "pvcs": ["my-vm-pvc-0", "my-vm-pvc-1"],
+      "referencedBy": ["backup-2025-01-12"]
+    }
+  ]
+}
+```
+Example of a per-backup manifest:
+```
+Per-Backup Manifest (manifests/<backup-name>/index.json):
+
+{
+  "backupName": "backup-2025-01-12",
+  "timestamp": "2025-01-12T10:00:00Z",
+}
+```
+Example of a per-backup manifest:
+```
+Per-Backup-oer-vm Manifest (manifests/<backup-name>/<vm-name>.json):
+
+{
+  "namespace": "default",
+  "name": "my-vm",
+  "checkpointChain": ["cp-001", "cp-002", "cp-003"]
+}
+```
+
+
+### Where/how to store qcow2 files and metadata
+- Current approach:
+  - Use the Velero object store plugin API but not the velero-specific logic in `persistence/object-store.go`
+  - Create a top-level dir in the BSL (under the BSL prefix, parallel to backups/restores/kopia) for kubevirt datamover.
+    - Actually, this may have to be outside the prefix (i.e. if prefix is "data" then we may need to create a parallel dir "data-kubevirt" or something similar, since I think Velero allows only its own files under the prefix)
+  - Copy individual qcow2 files and metadata files identifying the required qcow2 checkpoints. We may want a subdir per VirtualMachine for qcow2 files. For metadata files, these should probably be organized by velero backup.
+  - We need to manage storage usage on backup deletion -- when deleting a backup, we should delete any qcow2 files no longer referenced by any still-existing backups.
+- Other approaches:
+  - On volume in cluster
+    - Likely the simplest approach
+    - Volume must be mounted by the controller pod
+    - Will require its own periodic velero backups (less frequently than the VM incremental backups) for disaster recovery purposes
+  - Using kopia
+    - We could use kopia on top of the object storage API, but it is not clear that this will provide any real benefits, since we're already working with files that represent just the data diff we need. We can just manage them as individual objects.
+    - This will also require additional overhead around kopia maintenance, and we still may need to manage qcow2 file deletion manually.
+
+### Open questions
+- How to determine PVC size?
+  - user-configurable? configmap or annotation?
+  - From the kubevirt enhancement: "Before the process begins, an estimation of the required backup size will be performed. If the provided PVC size is insufficient, an error will be returned"
+  - If the PVC is too small, we need a clear error on the backup indicating that it failed due to insufficient PVC space.
+  - Since controller is responsible for PVC creation rather than plugin, the controller may be able to respond to PVC too small errors by retrying with a larger PVC.
+- The kubevirt datamover controller will be responsible for deleting the `VirtualMachineBackup` resource once it's no longer needed. When should this happen? Upon velero backup deletion? This would enable debugging in the case of failed operations. If we delete it immediately, that will make troubleshooting more difficult. If on backup deletion, we'll need to write a `DeleteItemAction` plugin.
+- Do we need an option to force full backups? If we're always doing incremental, eventually the incremental backup list becomes really long, requiring applying possibly hundreds of incremental files for a single restore.
+  - For initial release, we can add a force-full-virt-backup annotation on the velero backup. Longer-term, we can push for a general datamover feature in velero which could force full backups for both fs-backup and velero datamover if backup.Spec.ForceFullVolumeBackup is true, and once implemented, the qcow2 datamover can use this as well.
+
+### General notes
+- SnapshotMoveData must be true on the backup or DU/DD processing won't work properly
+- Longer-term, we can probably eliminate some of the custom code in the new controller by refactoring the velero datamover pluggability features, allowing the node agent to orchestrate this (with a custom image and configuration for datamover pods, etc.)
+- The kubevirt enhancement references both "push mode" and "pull mode" -- initial implementation on the kubevirt side will be push mode only. This OADP proposal is also push mode only for the initial implementation


docs/design/virt-qcow2-datamover.md
```diff
@@ -0,0 +1,113 @@
...
+
+In a kubevirt environment, one possibility for improving volume backup performance to allow more frequent backups would be to take incremental qcow2 backups using libvirt tools. Integrating this into OADP/Velero backups will require some new components to OADP (plugins, controller), as well as some modifications to the velero codebase.
+
+## Background
+
+Taking a VolumeSnapshot and then using kopia to process the entire volume and copy the required incremental changes into the Backup Storage Location (BSL) is a heavyweight process. Creating an incremental qcow2 backup for the same volume is generally a much more lightweight action. We want to make use of the existing Velero backup/restore process, with actual libvirt backup/restore happening via a new controller. For the moment, this will be referred to as the Virt Backup Controller. This action will be coordinated with Velero via existing infrastructure -- BackupItemActions (BIAs), RestoreItemActions (RIAs) and the DataUpload/DataDownload CRs. Initial implementation should require minimal changes to Velero, since Velero currently ignores DataUploads with `Spec.DataMover` set to something other than `velero`.
```
weshayutin: I think we still need to leave the door open to something other than kopia..  perhaps s/kopia/kopia (or other file mover).   Specially not using the term data mover as that is a larger set of function.

docs/design/virt-qcow2-datamover.md
```diff
@@ -0,0 +1,113 @@
...
+
+## Goals
+
+- Back up and restore VM volumes using libvirt incremental backup instead of Velero's built-in CSI datamover
+- Use existing Velero infrastructure to integrate this feature into regular velero backup and restore
+- Implementation based on kubevirt enhancement defined at https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md
```
weshayutin: let's add more context.
https://github.com/kubevirt/enhancements/pull/126/files

docs/design/virt-qcow2-datamover.md
```diff
@@ -0,0 +1,113 @@
...
+- As a user I want to use OADP to trigger backups that will back up volume data using libvirt tooling rather than CSI snapshots
+    - Volume backups will be incremental when possible (first backup for a given volume will be full, subsequent backups for that same volume will be incremental)
+    - Based on the assumption that libvirt incremental volume backups should be much faster than CSI snapshots followed by incremental kopia snapshot copy, the expectation is that users might run libvirt-based OADP backups more frequently than they would for CSI-based backups.
+    - If users are backing up more frequently with this method, they should ensure that they are not using CSI snapshots or fs-backup via the resource/volume policy configuration in the backup.
+
+
+## High-Level design
+
+### Upstream velero changes
+- Update velero volume policy model to allow unrecognized volume policy actions. Velero would treat unrecognized actions as "skip" (i.e. no snapshot or fs-backup), but the libvirt datamover could only act if the policy action is "virt".
```
weshayutin: meh.. naming is hard.  Since this is CNV specific and it's a possible upstream name.. I suggest s/virt/cnv  but that's just my opinion.  Maybe less offensive to the powers that be.

docs/design/virt-qcow2-datamover.md
sseago: @weshayutin That's the whole point of this enhancement -- to use something other than kopia (and than CSI snapshots). This is the background section, so "kopia" is correct here as that is the current heavyweight process that we are providing an alternative for.

docs/design/virt-qcow2-datamover.md
sseago: "virt" or "cnv" won't go upstream. The upstream change is to ignore any unrecognized policy action, so upstream will treat "virt", "cnv", or even "weshayutin" all the same way. That being said, we do need to decide what string to use for the policy. Should we use "cnv" instead? I think that's probably better than virt here anyway. Also, overall, should we call this feature "CNV incremental datamover"? Related, we've already created migtools repos: "kubevirt-datamover-plugin" and "kubevirt-datamover-controller" -- should those names also use cnv instead of kubevirt?

sseago: I'm changing the design to use "cnv" for the datamover type field in the DU/DD as well as for the volume policy action, instead of "virt" for both. For the moment, I have not changed any other usage of "virt" or "libvirt" to "cnv". We should probably have a more general naming discussion to work out the related issues.

docs/design/virt-qcow2-datamover.md
sseago: Adding the other URL for context, but noting that using pull mode is out of scope for this design.

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
+# Kubevirt incremental datamover
+
+In a kubevirt environment, one possibility for improving volume backup performance to allow more frequent backups would be to take incremental qcow2 backups using libvirt tools. Integrating this into OADP/Velero backups will require some new components to OADP (plugins, controller), as well as some modifications to the velero codebase.
```
mpryc, line 3: The `VirtualMachineInstance` supports both `qcow2` and `raw` image types, does this design cover qcow2 or both?

sseago: If we don't have a qcow disk, the kubevirt side won't work. But that should be covered by the checks described in the kubevirt enhancement on how to check whether incremental backups are enabled.

weshayutin: QEMU backup with CBT is only supported with QCOW2 images
https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+## Non goals
+- Deep integration with velero data mover pluggability (this could be considered in the long-term though, which would minimize duplication of effort and enhance maintainability)
+- Using push mode with kubevirt.
+
+## Use cases
+- As a user I want to use OADP to trigger backups that will back up volume data using kubevirt tooling rather than CSI snapshots
+  - Volume backups will be incremental when possible (first backup for a given volume will be full, subsequent backups for that same volume will be incremental)
+  - Assuming that kubevirt incremental volume backups should be much faster than CSI snapshots followed by incremental kopia snapshot copy, the expectation is that users might run kubevirt datamover backups more frequently than they would for CSI-based backups.
+  - If users are backing up more frequently with this method, they should ensure that they are not using CSI snapshots or fs-backup via the resource/volume policy configuration in the backup.
+
```
weshayutin, line 25: So from what I'm reading users would have a number of steps to backup a vm w/ qcow2 incremental.  @sseago have we considered a workflow where OADP creates the VirtualMachineBackup CR's for the VM, qcow2 backup taken, and then OADP sends the snapshot off cluster to s3?  

Maybe I'm misunderstanding..  A workflow process doc may help illustrate

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - Create DD based on DU annotation and DU ConfigMap
+
+### Kubevirt Datamover Controller
+- Responsible for reconciling DataUploads/DataDownloads where `Spec.DataMover` is "kubevirt"
+- DataUpload reconciler (backup):
+  - identify the (temporary) PVC from the DU metadata
+  - identify the VirtualMachineInstance from the PVC metadata.
+  - identify the already-existing qcow2 files that this incremental backup will be based on
+  - create the VirtualMachineBackup CR with `source` set to the VirtualMachineInstance, `pvcName` set to the temporary PVC, and `backupTracker` set to the name of a `VirtualMachineBackupTracker` referencing the most recent backup checkpoint, and (optionally)`forceFullBackup`, set to `true` to force a full backup
+  - Wait for VMBackup to complete (monitoring status)
```
weshayutin: ah.. this sounds like what I'm talking about in my earlier comment..  still an AI gen workflow may help :)

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - Launch kubevirt datamover pod mounting the temporary PVC with the qcow2 file(s) from the backup.
+    - This pod needs to be running a command that will do the datamover operation from pvc to object storage
+    - Will there be multiple qcow2 files if there are multiple disks with CBT enabled? Will these all use the same checkpoint name?
+  - Copy the new file to object storage (see [Where to store qcow2 files](#wherehow-to-store-qcow2-files-and-metadata) below)
+  - Save any required metadata to identify the stored data (collection of qcow2 pathnames/checkpoints, etc.), along with identifying the backup and VirtualMachineInstance they're associated with. Save this metadata file as well (see [Where to store qcow2 files](#wherehow-to-store-qcow2-files-and-metadata) below)
+    - The VirtualMachineBackupTracker CR will be helpful here.
```
weshayutin: seems to me we would want to send the VirtualMachineBackupTracker CR instance to s3, given that w/o it full backups are required

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+    - We need to properly handle cases where we attempt an incremental backup but a full backup is taken instead (checkpoint lost, CSI snapshot restore since last checkpoint, VM restart, etc.)
+    - Aborted backups also need to be handled (resulting in a failed PVC backup on the Velero side)
+- DataDownload reconciler (restore)
+  - (this area is less well defined so far, since the kubevirt enhancement doesn't go into as much detail on restore)
```
weshayutin, line 90: agree, IMHO it seems like OADP would only be responsible for ensuring the pvc's, VMbackup cr and tracker cr are restored to the namespace such that the user could then run qemu commands.  Not ideal that CNV spec does not handle restores :/

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - We will need a temporary PVC for pulling qcow2 images from object store (if we're restoring the empty temp PVC from backup, that might work here)
+  - We also need PVCs created for each VM disk we're restoring from qcow2 images.
+  - We'll need to create another datamover pod here which will do the following:
+    - pod will have temp PVC mounted, as well as PVCs mounted for each vm disk we're creating.
+    - pod running command/image will first get the list of qcow2 files to pull from object storage
+    - once we have the qcow2 full backup and incremental files from object store, repeatedly call `qemu-img rebase -b fullbackup.qcow2 -f qcow2 -u incrementalN.qcow2` for each incremental backup, in order
+    - Then, convert qcow2 image to raw disk image: `qemu-img convert -f qcow2 -O raw fullbackup.qcow2 restored-raw.img`
+    - Finally, write this image to the PVC which contains the VM disk
+    - (repeat process for each disk if VM has multiple disks to restore)
```
weshayutin, line 99: lines 77 - 89, is this something we would code or users are meant to execute?

weshayutin: ah, line 81..

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+    - Actually, this may have to be outside the prefix (i.e. if prefix is "data" then we may need to create a parallel dir "data-kubevirt" or something similar, since I think Velero allows only its own files under the prefix)
+  - Copy individual qcow2 files and metadata files identifying the required qcow2 checkpoints. We may want a subdir per VirtualMachineInstance for qcow2 files. For metadata files, these should probably be organized by velero backup.
+  - We need to manage storage usage on backup deletion -- when deleting a backup, we should delete any qcow2 files no longer referenced by any still-existing backups.
+- Other approaches:
+  - On volume in cluster
+    - Likely the simplest approach
+    - Volume must be mounted by the controller pod
+    - Will require its own periodic velero backups (less frequently than the VM incremental backups) for disaster recovery purposes
+  - Using kopia
+    - We could use kopia on top of the object storage API, but it is not clear that this will provide any real benefits, since we're already working with files that represent just the data diff we need. We can just manage them as individual objects.
```
weshayutin, line 195: avoiding kopia for this work is a must I think

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+- The kubevirt datamover controller will be responsible for deleting the `VirtualMachineBackup` resource once it's no longer needed. When should this happen? Upon velero backup deletion? This would enable debugging in the case of failed operations. If we delete it immediately, that will make troubleshooting more difficult. If on backup deletion, we'll need to write a `DeleteItemAction` plugin.
+- When a backup is deleted, its snapshots may be needed by more recent incremental backups. What ever means we're using to map snapshots to backups will need to update this mapping to associate the deleted backup's snapshots with the next most recent backup that relies on these files. This means that for a given backup and volume we could have more than one snapshot file. We need to handle this mapping, including the order of application.
+- Do we need an option to force full backups? If we're always doing incremental, eventually the incremental backup list becomes really long, requiring applying possibly hundreds of incremental files for a single restore.
```
weshayutin, line 205: forcing a full backup would just be removing the VirtualMachineBackupTracker instance for the VM right?

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - For initial release, we can add a force-full-virt-backup annotation on the velero backup. Longer-term, we can push for a general datamover feature in velero which could force full backups for both fs-backup and velero datamover if backup.Spec.ForceFullVolumeBackup is true, and once implemented, the qcow2 datamover can use this as well.
+
+### General notes
+- SnapshotMoveData must be true on the backup or DU/DD processing won't work properly
+- Longer-term, we can probably eliminate some of the custom code in the new controller by refactoring the velero datamover pluggability features, allowing the node agent to orchestrate this (with a custom image and configuration for datamover pods, etc.)
+- The kubevirt enhancement references both "push mode" and "pull mode" -- initial implementation on the kubevirt side will be push mode only. This OADP proposal is also push mode only for the initial implementation
```
weshayutin, line 211: AFAICT, they are know going to deliver BOTH push and pull.  We should double check :)

docs/design/kubevirt-datamover.md
mpryc: Kopia could be beneficial here by reducing storage usage through deduplication, particularly for VMs that contain identical or highly similar data chunks, avoiding 1-to-1 storage of duplicated data. For subsequent incremental backups, where only modified blocks are persisted, the additional benefit from deduplication will be smaller and may be limited without additional benefit.

weshayutin: The qcow2 snapshot should already be deduped and changed blocks only, right?

docs/design/kubevirt-datamover.md
sseago: All of the things you mention will be done by OADP (via velero, our custom BIA, and our controller).  The user just needs to:
1) enable incremental backup in the VM
2) create a Velero backup with the kubevirt volume policy matching the supported PVCs.

docs/design/kubevirt-datamover.md
sseago: We either want to send this CR to s3, or our own custom json/yaml -- we actually may need both. We need to track the full chain of full+incremental backups for each VM in the backup, which is more than the VMBackupTracker CR will have.

docs/design/kubevirt-datamover.md
sseago: I don't think we want to make the user do all that manually. Why shouldn't the controller run the commands? If th user runs them, we have to spit out an output file listing the series of commands the user must then run in-order (i.e. the qemu commands to restore the initial full backup, then apply the incremental backups, in order, repeat for each VM. If we have to generate this for the user anyway, why can't we have a controller action that actually executes the commands? Otherwise this feature will be a usability nightmare. User expectation is that if `velero backup create` runs the full backup, `velero restore create` will run the full restore.

docs/design/kubevirt-datamover.md
sseago: I had envisioned the controller doing all of this. The user should just be running `velero restore create...`

docs/design/kubevirt-datamover.md
sseago: Yeah, I don't think there will be much to  dedup when we're uploading a series of incremental backups. I'm just not seeing that kopia really solves any problems for us here.

docs/design/kubevirt-datamover.md
sseago: Well forcing it or setting the right field in the tracker. Also, whatever work we do to track the full chain of incremental backups that will be needed to restore this backup might have some additional work needed.

docs/design/kubevirt-datamover.md
sseago: this mapping of full and incremental backup files to velero backups is also needed so when a backup is deleted, our DeleteItemAction can figure out which incremental files are no longer needed and delete them from s3. i.e. if we have a single VM and have one full and 5 incremental backups, we still need all the files, even if the earlier velero backups have expired. However, if we force a full backup after incremental backup 5, then once all backups from incremental 5 and earlier have been deleted, all of these qcow2 files must be deleted from s3.

docs/design/kubevirt-datamover.md
sseago: My understanding is push mode is first, followed by pull. In any case, I think from the OADP side, we should focus on push mode only initially. Then we can decide:
1) Does pull mode make sense at all for velero/OADP integration?
2) If so, when do we deliver it?

If we're already tight for time here, then I think deciding to do push and pull at the same time might put 1.6 timeline at risk.

docs/design/kubevirt-datamover.md
weshayutin: OK.. then I'm good.  Thanks @sseago

docs/design/kubevirt-datamover.md
weshayutin: ok.. that's great if we can fully automate it :)

docs/design/kubevirt-datamover.md
weshayutin: agree, we can revisit this bit

docs/design/kubevirt-datamover.md
weshayutin: agree, one rock solid way is better than going after both.  I do think it's worth checking in w/ Shelly about it for clarification.

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+
+### BackupItemAction/RestoreItemAction plugins
+- VirtualMachineInstance BIA plugin
+  - This plugin will check to see whether QEMU backup is enabled in any of the disks -- whether `Spec.Domain.Devices.Disk[i].CBT` is `true` for at least one disk. If none are enabled, then the plugin exits without action (question: is this still valid, or is it replaced by the next?)
+  - The plugin will check whether the VirtualMachineInstance's `status.ChangedBlockTracking` is `Enabled`
+  - The plugin must also determine whether the VM is running, since offline backup is not supported in the initial release.
+    - If QEMU backup is enabled, the next question is whether the user wants to use the kubevirt datamover for this VM's volumes. We will use volume policies for this, although it's a bit more complicated since a VM could have multiple PVCs. If at least one PVC for the VM has the "kubevirt" policy action specified, and no PVCs in this VM have other non-skip policies (i.e. "snapshot", etc.) then we'll use the kubevirt datamover
+    - Iterate over all PVCs for the VM
+    - If any PVC has an action other than "kubevirt" or "skip", exit without action
```
shubham-pampattiwar, line 56: If we are going to introduce a new action then lets stick just with `kubevirt` instead of `skip`

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+- Deep integration with velero data mover pluggability (this could be considered in the long-term though, which would minimize duplication of effort and enhance maintainability)
+- Using push mode with kubevirt.
+
+## Use cases
+- As a user I want to use OADP to trigger backups that will back up volume data using kubevirt tooling rather than CSI snapshots
+  - Volume backups will be incremental when possible (first backup for a given volume will be full, subsequent backups for that same volume will be incremental)
+  - Assuming that kubevirt incremental volume backups should be much faster than CSI snapshots followed by incremental kopia snapshot copy, the expectation is that users might run kubevirt datamover backups more frequently than they would for CSI-based backups.
+  - If users are backing up more frequently with this method, they should ensure that they are not using CSI snapshots or fs-backup via the resource/volume policy configuration in the backup.
+
+
```
shubham-pampattiwar, line 42: The design might benefit from something like a comparison table for CSI DM vs QCOW2 DM (just a thought):
```
  Key Differences
  ┌────────────────────┬──────────────────────────────────────────┬────────────────────────────────────────────┐
  │       Aspect       │               CSI Approach               │          KubeVirt qcow2 Approach           │
  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ Layer              │ Storage (CSI driver)                     │ Hypervisor (QEMU/libvirt)                  │
  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ Snapshot mechanism │ CSI VolumeSnapshot                       │ VirtualMachineBackup CR                    │
  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ Incremental        │ Kopia deduplication (scans whole volume) │ True block-level CBT (only changed blocks) │
  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ Data mover         │ Velero's node-agent + kopia              │ New OADP controller + qemu-img             │
  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
  │ VM awareness       │ None (just sees PVC)                     │ Full (knows it's a VM disk)                │
  └────────────────────┴──────────────────────────────────────────┴────────────────────────────────────────────┘
```

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - We need to manage storage usage on backup deletion -- when deleting a backup, we should delete any qcow2 files no longer referenced by any still-existing backups.
+- Other approaches:
+  - On volume in cluster
+    - Likely the simplest approach
+    - Volume must be mounted by the controller pod
+    - Will require its own periodic velero backups (less frequently than the VM incremental backups) for disaster recovery purposes
+  - Using kopia
+    - We could use kopia on top of the object storage API, but it is not clear that this will provide any real benefits, since we're already working with files that represent just the data diff we need. We can just manage them as individual objects.
+    - This will also require additional overhead around kopia maintenance, and we still may need to manage qcow2 file deletion manually.
+
```
shubham-pampattiwar, line 197: The backup and restore flows involve multiple components (BIA/RIA plugins, DataUpload/DataDownload, controller, VirtualMachineBackup CR, datamover pods). Visual diagrams would help readers follow the sequence. Here are suggested mermaid diagrams(please let me know if they are accurate or change them as needed to add into the design doc):
- **Backup Flow Diagram:**
```mermaid
  sequenceDiagram
      participant Velero
      participant VMI_BIA as VMI BIA Plugin
      participant DU as DataUpload CR
      participant Controller as Kubevirt Datamover Controller
      participant VMBackup as VirtualMachineBackup CR
      participant KubeVirt
      participant TempPVC as Temporary PVC
      participant Pod as Datamover Pod
      participant BSL as Object Storage (BSL)

      Velero->>VMI_BIA: Backup VirtualMachineInstance
      VMI_BIA->>VMI_BIA: Check CBT enabled & volume policies
      VMI_BIA->>TempPVC: Create temporary PVC
      VMI_BIA->>DU: Create DataUpload (DataMover: "kubevirt")

      Controller->>DU: Reconcile DataUpload
      Controller->>VMBackup: Create VirtualMachineBackup CR
      VMBackup->>KubeVirt: Trigger backup
      KubeVirt->>KubeVirt: Extract changed blocks via libvirt/QEMU
      KubeVirt->>TempPVC: Write qcow2 file(s)
      KubeVirt->>VMBackup: Update status: Complete

      Controller->>Pod: Launch datamover pod
      Pod->>TempPVC: Mount & read qcow2 files
      Pod->>BSL: Upload qcow2 + metadata
      Pod->>Controller: Exit success
      Controller->>DU: Update status: Complete
```

- **Restore Flow Diagram**

```mermaid
  sequenceDiagram
      participant Velero
      participant VMI_RIA as VMI RIA Plugin
      participant DD as DataDownload CR
      participant Controller as Kubevirt Datamover Controller
      participant TempPVC as Temporary PVC
      participant TargetPVC as Target PVC(s)
      participant Pod as Datamover Pod
      participant BSL as Object Storage (BSL)

      Velero->>VMI_RIA: Restore VirtualMachineInstance
      VMI_RIA->>TempPVC: Create temporary PVC
      VMI_RIA->>DD: Create DataDownload (DataMover: "kubevirt")

      Controller->>DD: Reconcile DataDownload
      Controller->>TargetPVC: Create target PVC(s) for VM disks
      Controller->>Pod: Launch datamover pod

      Pod->>BSL: Download qcow2 files (full + incrementals)
      Pod->>TempPVC: Write qcow2 files
      Pod->>Pod: qemu-img rebase (chain incrementals)
      Pod->>Pod: qemu-img convert -f qcow2 -O raw
      Pod->>TargetPVC: Write raw disk image
      Pod->>Controller: Exit success

      Controller->>DD: Update status: Complete
      Controller->>TempPVC: Delete temporary PVC
      Note over TargetPVC: VM can now boot with restored disk
```

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - Create temporary PVC
+  - Create DD based on DU annotation and DU ConfigMap
+
+### Kubevirt Datamover Controller
+- Responsible for reconciling DataUploads/DataDownloads where `Spec.DataMover` is "kubevirt"
+- DataUpload reconciler (backup):
+  - identify the (temporary) PVC from the DU metadata
+  - identify the VirtualMachineInstance from the PVC metadata.
+  - identify the already-existing qcow2 files that this incremental backup will be based on
+  - create the VirtualMachineBackup CR with `source` set to the VirtualMachineInstance, `pvcName` set to the temporary PVC, and `backupTracker` set to the name of a `VirtualMachineBackupTracker` referencing the most recent backup checkpoint, and (optionally)`forceFullBackup`, set to `true` to force a full backup
```
shubham-pampattiwar: Would a sentence or two explaining what happens when this CR is created would be help here ? (I know the background design explains it but just for context i.e., KubeVirt uses libvirt APIs to extract changed blocks from the running VM and writes them as qcow2 files).

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+  - Wait for VMBackup to complete (monitoring status)
+  - Launch kubevirt datamover pod mounting the temporary PVC with the qcow2 file(s) from the backup.
```
shubham-pampattiwar, line 82: What image will the datamover pod run? Is this a new binary that needs to be built, or leveraging existing tooling (e.g., qemu-img container)?

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+    - once we have the qcow2 full backup and incremental files from object store, repeatedly call `qemu-img rebase -b fullbackup.qcow2 -f qcow2 -u incrementalN.qcow2` for each incremental backup, in order
+    - Then, convert qcow2 image to raw disk image: `qemu-img convert -f qcow2 -O raw fullbackup.qcow2 restored-raw.img`
+    - Finally, write this image to the PVC which contains the VM disk
+    - (repeat process for each disk if VM has multiple disks to restore)
+    - Note that the various `qemu-img` actions might eventually be combined into a single kubevirt API call, but for the moment this would need to be done manually.
+  - Once datamover pod has restored the VM disks, it will exit and the VirtualMachineInstance can launch with these disks (following the velero datamover model where the temporary PVC is deleted, retaining the PV, which then binds to the VM's PVCs may work here). The temporary PVC (containing the qcow2 images, not the restored VM disk image) should be completely removed at this point, including PV content.
+
+
+### Where/how to store qcow2 files and metadata
```
shubham-pampattiwar, line 182: Would the storage path in s3 be something like:
```  
  <bsl-prefix>/kubevirt-datamover/
  ├── manifests/
  │   └── <velero-backup-name>.json       # Per-backup manifest
  └── checkpoints/
      └── <namespace>/
          └── <vmi-name>/
              ├── <checkpoint-id>.qcow2   # Actual qcow2 files
              └── index.json              # Per-VMI checkpoint chain
```  
Or organized differently?

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+- Current approach:
+  - Use the Velero object store plugin API but not the velero-specific logic in `persistence/object-store.go`
+  - Create a top-level dir in the BSL (under the BSL prefix, parallel to backups/restores/kopia) for kubevirt datamover.
+    - Actually, this may have to be outside the prefix (i.e. if prefix is "data" then we may need to create a parallel dir "data-kubevirt" or something similar, since I think Velero allows only its own files under the prefix)
+  - Copy individual qcow2 files and metadata files identifying the required qcow2 checkpoints. We may want a subdir per VirtualMachineInstance for qcow2 files. For metadata files, these should probably be organized by velero backup.
```
shubham-pampattiwar: What fields are in the metadata files? How do we map Velero backups to their required checkpoints?

shubham-pampattiwar: Also, lets say we have Backup A and Backup B
Backup A: uses checkpoints 1, 2, 3
Backup B: uses checkpoints 1, 2, 3, 4, 5

Delete Backup A → can't delete 1, 2, 3 (still needed by B)
Delete Backup B → now can delete 4, 5, and also 1, 2, 3

How is this reference counting implemented? We will have to store this mapping and We would need some structure for this.

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+- DataUpload reconciler (backup):
+  - identify the (temporary) PVC from the DU metadata
+  - identify the VirtualMachineInstance from the PVC metadata.
+  - identify the already-existing qcow2 files that this incremental backup will be based on
+  - create the VirtualMachineBackup CR with `source` set to the VirtualMachineInstance, `pvcName` set to the temporary PVC, and `backupTracker` set to the name of a `VirtualMachineBackupTracker` referencing the most recent backup checkpoint, and (optionally)`forceFullBackup`, set to `true` to force a full backup
+  - Wait for VMBackup to complete (monitoring status)
+  - Launch kubevirt datamover pod mounting the temporary PVC with the qcow2 file(s) from the backup.
+    - This pod needs to be running a command that will do the datamover operation from pvc to object storage
+    - Will there be multiple qcow2 files if there are multiple disks with CBT enabled? Will these all use the same checkpoint name?
+  - Copy the new file to object storage (see [Where to store qcow2 files](#wherehow-to-store-qcow2-files-and-metadata) below)
```
shubham-pampattiwar, line 85: How will the pod receive BSL configuration and credentials? For example, will the controller inject the BSL details via environment variables and mount the credential secret?

docs/design/kubevirt-datamover.md
shubham-pampattiwar: @sseago I am thinking something like this:
 - Per-VMI Index (`checkpoints/<ns>/<vmi>/index.json`):
```json  
{
    "vmiName": "my-vm",
    "namespace": "default",
    "pvcs": ["my-vm-disk-0", "my-vm-disk-1"],
    "checkpoints": [
      {
        "id": "cp-001",
        "type": "full",
        "timestamp": "2025-01-10T10:00:00Z",
        "files": ["cp-001-disk0.qcow2", "cp-001-disk1.qcow2"],
        "referencedBy": ["backup-2025-01-10"]
      },
      {
        "id": "cp-002",
        "type": "incremental",
        "parent": "cp-001",
        "timestamp": "2025-01-11T10:00:00Z",
        "files": ["cp-002-disk0.qcow2", "cp-002-disk1.qcow2"],
        "referencedBy": ["backup-2025-01-11"]
      },
      {
        "id": "cp-003",
        "type": "incremental",
        "parent": "cp-002",
        "timestamp": "2025-01-12T10:00:00Z",
        "files": ["cp-003-disk0.qcow2", "cp-003-disk1.qcow2"],
        "referencedBy": ["backup-2025-01-12"]
      }
    ]
  }
```

- Per-Backup Manifest (`manifests/<backup-name>.json`):
```json  
{
    "backupName": "backup-2025-01-12",
    "timestamp": "2025-01-12T10:00:00Z",
    "vmis": [
      {
        "namespace": "default",
        "name": "my-vm",
        "checkpointChain": ["cp-001", "cp-002", "cp-003"]
      }
    ]
  }
```

shubham-pampattiwar: When deleting backup-2025-01-11:
  1. Read its manifest to find referenced checkpoints (cp-001, cp-002)
  2. For each checkpoint, remove this backup from referencedBy in the VMI index
  3. If referencedBy becomes empty AND no later checkpoint depends on it as parent, delete the qcow2 files
  4. Delete the backup manifest

This handles the reference counting problem - cp-001 and cp-002 would still be referenced by backup-2025-01-12 (since it needs the full chain to restore).

shubham-pampattiwar (commented):
Great design, @sseago ! Excited to see this move forward, qcow2 incremental backups will be a big win for VM backup performance. Left a few comments with questions and suggestions.

docs/design/kubevirt-datamover.md
shubham-pampattiwar: Added suggestion: https://github.com/openshift/oadp-operator/pull/2037#discussion_r2683989910 and https://github.com/openshift/oadp-operator/pull/2037#discussion_r2683992153

docs/design/kubevirt-datamover.md
sseago: We are sticking with `kubevirt` -- the intent here is that if a VM has 2 PVCs and one has `kubevirt` and one has `snapshot`, then we will *not* create a kubevirt DU and will log a warning. However, if one has `kubevirt` and one has `skip`, then we'll go ahead and back up this VM. Basically, one `kubevirt` pvc for the VM is sufficient for kubevirt backup, as long as the user has not specified `snapshot` or `fs-backup` for any of the VM PVCs.

docs/design/kubevirt-datamover.md
sseago: @weshayutin has created a backup flow diagram on a wiki page. We can add a similar restore one -- comparing with these to make sure we've gotten it all right.

docs/design/kubevirt-datamover.md
sseago: Sure, that makes sense. I can add that, just making it clear that this workflow isn't something we're creating ourselves,

docs/design/kubevirt-datamover.md
sseago: I'm thinking we can use the same image for the datamover pods and the new controller -- similar to what Velero does with mover pods and node agent.

docs/design/kubevirt-datamover.md
sseago: It won't be directly under bsl-prefix since Velero complains if there are non-velero dirs there. We'd probably need to combine `<bsl-prefix>` and kubevirt-datamover into a single directory name: `<bsl-prefix>-kubevirt` or something similar. But yes, something like this. However, we also need to make sure we capture this per-backup. i.e. for a given vmi, backup1 will include checkpoint1.qcow2, backup2 will have checkppoint1.qcow2 *and* checkpoint2.qcow2 (in that order). But maybe that's all captured in index.json. Although we may need to have index.json which references a list of backups and then <backup-name>.json with that backup's list of checkpoints and qcow2 files. Also, checkpoint name and qcow2 filename may not be guaranteed to be the same, in hich case the various json files will need to make the connection explicit. The json files also probably need to list the PVCs.

docs/design/kubevirt-datamover.md
sseago: See my comments above where you reference the dir structure. The vmi->backup(s)->checkpoints connections will definitely need to be part of the definition of what files we store and what's in them.

docs/design/kubevirt-datamover.md
sseago: Should be similar to what datamover pods do I'd imagine. I haven't looked at that recently, but presumably DU links to Backup which then lists BSL?

sseago: And yes, mounting/accessing the credentials is something we can probably handle in the same way we do for datamover and fs-backup pods upstream.

docs/design/kubevirt-datamover.md
sseago: @shubham-pampattiwar yes, I think something like this works. If we're managing the references in both directions, this makes restore easier (using checkpointChain), and also deletion (using checkpointChain *and* referencedBy)

docs/design/kubevirt-datamover.md
sseago: ok, So some of my suggestions above are handled better by your later examples.

weshayutin:
@kaovilai @mpryc @shubham-pampattiwar ( nice comments ) keep pushing on this until we're ready to merge.  Thanks!

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+
+## High-Level design
+
+### Upstream velero changes
+- Update velero volume policy model to allow unrecognized volume policy actions. Velero would treat unrecognized actions as "skip" (i.e. no snapshot or fs-backup), but the kubevirt datamover could only act if the policy action is "kubevirt".
+- In `pkg/restore/actions/dataupload_retrieve_action.go` and in `DataDownload` we need to add SnapshotType.
+
+### BackupItemAction/RestoreItemAction plugins
+- VirtualMachineInstance BIA plugin
+  - This plugin will check to see whether QEMU backup is enabled in any of the disks -- whether `Spec.Domain.Devices.Disk[i].CBT` is `true` for at least one disk. If none are enabled, then the plugin exits without action (question: is this still valid, or is it replaced by the next?)
```
Joeavaikath: I think it is
status.ChangedBlockTracking = Enabled should tell me if the VMI is using CBT, as stated in the VEP

sseago: @Joeavaikath The line below this has exactly what you propose. The question is whether this is needed at all, or needed in addition to the other. But I see no reference to this in the VEP now. I suspect this was left over from an earlier iteration of the proposal. I'll remove it.

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,115 @@
...
+    - pod running command/image will first get the list of qcow2 files to pull from object storage
+    - once we have the qcow2 full backup and incremental files from object store, repeatedly call `qemu-img rebase -b fullbackup.qcow2 -f qcow2 -u incrementalN.qcow2` for each incremental backup, in order
+    - Then, convert qcow2 image to raw disk image: `qemu-img convert -f qcow2 -O raw fullbackup.qcow2 restored-raw.img`
+    - Finally, write this image to the PVC which contains the VM disk
+    - (repeat process for each disk if VM has multiple disks to restore)
+    - Note that the various `qemu-img` actions might eventually be combined into a single kubevirt API call, but for the moment this would need to be done manually.
+  - Once datamover pod has restored the VM disks, it will exit and the VirtualMachineInstance can launch with these disks (following the velero datamover model where the temporary PVC is deleted, retaining the PV, which then binds to the VM's PVCs may work here). The temporary PVC (containing the qcow2 images, not the restored VM disk image) should be completely removed at this point, including PV content.
+
```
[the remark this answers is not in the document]
sseago: The `-u` flag was in the proposal in VEP 25. Leaving this for now but we can revisit when we get to the detailed restore design:
https://github.com/kubevirt/enhancements/blob/main/veps/sig-storage/incremental-backup.md#restore

shubham-pampattiwar (approved):

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,211 @@
...
+	- In addition, SnapshotMoveData must be true
+    - Iterate over all PVCs for the VM
+    - If any PVC has an action other than "kubevirt" or "skip", exit without action
+    - If at least one PVC has an action of "kubevirt", then use the kubevirt datamover
+  - This plugin will create a DataUpload with `Spec.SnapshotType` set to "kubevirt" and `Spec.DataMover` set to "kubevirt"
+    - Note that unlike the built-in datamover, this DataUpload is tied to a VM (which could include multiple PVCs) rather than to a single PVC.
+  - An annotation will be added to the DataUpload identifying the VirtualMachine we're backing up.
+  - Add `velerov1api.DataUploadNameAnnotation` to VirtualMachine
+  - OperationID will be created and returned similar to what's done with the CSI PVC plugin, and the async operation Progress method will report on progress based on the DU status (similar to CSI PVC plugin)
```
Joeavaikath, line 62: Just to improve my understanding, this would be:
KubeVirt VM plugin working at the VM level possibly hitting multiple PVCs
analogous to the existing 
CSI PVC plugin which is PVC specific

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,211 @@
...
+  - We don't want to restore these because they would kick off another VMBackup action.
+
+### Kubevirt Datamover Controller
+- Responsible for reconciling DataUploads/DataDownloads where `Spec.DataMover` is "kubevirt"
+- Configurable concurrency limits: concurrent-vm-backups and concurrent-vm-datauploads
+- DataUpload reconciler (backup):
+  - create the (temporary) PVC.
+  - identify the VirtualMachine from the PVC metadata.
+  - From the BSL, identify the latest checkpoint for this VM and confirm that all of the related qcow2 files are available.
+  - create/update the VirtualMachineBackupTracker CR with `source` set to the VirtualMachine, and `status.latestCheckpoint.name` set to match the latest checkpoint we have for this vm. Note that it might be better if we could avoid hacking the VMBT status and instead have a field on the VM -- something like `spec.overrideCheckpoint` to use a specific checkpoint if the one in the VMBT doesn't match the latest we have in the BSL.
```
Joeavaikath, line 79: On the first ever backup, we create VirtualMachineBackupTracker (VMBT) with source set to a VirtualMachine resource, and on backup completion `status.latestCheckpoint` is updated by KubeVirt.
VMB uses it, usual ops follow
On backup completion, we cleanup the VMB resource

Generally, for a VMB backup, before we create a VMBT:
- Check if VMBT for the VM exists (labels: oadp, vm resource name)
- If the checkpoint data matches BSL (or the spec.overrideCheckpoint field, do we assume this will be the right value always)
  - If not, update to ensure it is correct
    - If this update somehow fails (checkpoint missing), do a full backup 

Remaining ops proceed

docs/design/kubevirt-datamover.md
sseago: Yes. In both cases, we have a plugin which creates a DataUpload and then in its async Progress method, monitors the DU state until it's done.

docs/design/kubevirt-datamover.md
sseago: Yes, that's right, with one open question in my mind:

"On backup completion, we cleanup the VMB resource"
Probably? It's also possible that we'll want to keep it around and clean it up upon backup deletion (similar to what we do for PodVolumeBackup, DataUpload, etc.) -- depends on whether we think VMB will ever be useful for debug purposes.

docs/design/kubevirt-datamover.md
```diff
@@ -0,0 +1,211 @@
...
+  │ Data mover         │ Velero's node-agent + kopia              │ New OADP controller + qemu-img             │
+  ├────────────────────┼──────────────────────────────────────────┼────────────────────────────────────────────┤
+  │ VM awareness       │ None (just sees PVC)                     │ Full (knows it's a VM disk)                │
+  └────────────────────┴──────────────────────────────────────────┴────────────────────────────────────────────┘
+```
+
+## High-Level design
+
+### Upstream velero changes
+- Update velero volume policy model to allow unrecognized volume policy actions. Velero would treat unrecognized actions as "skip" (i.e. no snapshot or fs-backup), but the kubevirt datamover could only act if the policy action is "kubevirt".
```
kaovilai, line 46: Have we created an issue for this in velero? Do you think this will get pushed back? ie. we only allow recognizable policy actions.

sseago: No issue created upstream yet. The issue and PR for this is one of the tasks I'll be working on this week. And yes, pushback wouldn't completely surprise me. Hopefully we can convince the other maintainers that this is a good thing to allow for 3rd party plugins, etc. but we may have to carry the commit in our fork if we are unsuccessful there.

shubham-pampattiwar: We have a strong argument here. Our goal is to make Velero more portable and usable for third-party datamovers, and we deliberately want to avoid introducing custom, vendor-specific logic like the vSphere-specific code that exists in some parts of Velero today: https://github.com/search?q=repo%3Avmware-tanzu%2Fvelero%20vsphere&type=code

weshayutin:
@mpryc @kaovilai @Joeavaikath if time and duties align, please try to have your final thoughts and reviews in by tomorrow.   In scrum @sseago will be walking the team through https://github.com/orgs/migtools/projects/7/views/20 and expanding, explaining, breaking up issues into smaller tasks. 

Thanks all 0/

shubham-pampattiwar:
If there are no remaining unresolved comments, questions, or design discussions, this PR appears to be in a merge-ready state. Lets Go 🚀

sseago:
@shubham-pampattiwar just needs lgtm

docs/design/kubevirt-datamover.md
Joeavaikath: Right, so it depends on what information would the user would access
My concern was around the VMBT finding/verifying bit, because it wasn't as clear to me. I'll keep an eye out during implementation

weshayutin:
@kaovilai @mpryc keep pushing on this until we're ready to merge. Thanks!

mpryc (approved):

PR #2042, opened 2025-12-09
Ensure that spec.nonAdmin and spec.vmFileRestore enable/disable have basic tests

weshayutin:
## Why the changes were made

because the lord knows we NEED tests

 make test-e2e TEST_FILTER="" GINKGO_ARGS="--focus='Should enable and disable VMFileRestore'"

### DIFF
--- a/tests/e2e/dpa_deployment_suite_test.go
+++ b/tests/e2e/dpa_deployment_suite_test.go
@@ -33,6 +33,8 @@ type TestDPASpec struct {
 	NoS3ForcePathStyle      bool
 	NoRegion                bool
 	DoNotBackupImages       bool
+	EnableVMFR              bool
+	EnableNonAdmin          bool
 }
 
 func createTestDPASpec(testSpec TestDPASpec) *oadpv1alpha1.DataProtectionApplicationSpec {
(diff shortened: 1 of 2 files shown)


kaovilai (approved):

tests/e2e/dpa_deployment_suite_test.go
```diff
@@ -374,6 +386,52 @@ var _ = ginkgo.Describe("Configuration testing for DPA Custom Resource", func()
...
+		testSpec := TestDPASpec{
+			BSLSecretName:  bslSecretName,
+			EnableNonAdmin: true,
+		}
+		dpaSpec := createTestDPASpec(testSpec)
+		err := dpaCR.CreateOrUpdate(dpaSpec)
+		gomega.Expect(err).ToNot(gomega.HaveOccurred())
+
+		log.Printf("Waiting for DPA to be reconciled with NonAdmin enabled")
+		gomega.Eventually(dpaCR.IsReconciledTrue(), time.Minute*2, time.Second*5).Should(gomega.BeTrue())
```
shubham-pampattiwar, line 400: Should we also add check for pods ?

weshayutin: I would like to yes, but was thinking about a follow up patch after discussing with you.  I would like the operator to NOT show reconciled if the expected pods like oadp-non-admin and vmfr pod are not running.  I'm not an expert but in my unexpert opinion that is what operators are for.   This test could also check for the pod running, however if the operator properly does that, reconciled check would be sufficient.  @shubham-pampattiwar you know better than I do, if you want to please take over this PR or create a fresh one.   ping me if you want.

shubham-pampattiwar (approved):

PR #2049, opened 2025-12-17
Add timeout and retry to CLI log/describe commands in e2e tests

mpryc:
Prevents tests from hanging indefinitely when streaming logs from object storage. Uses 5-minute timeout with 3 retries.

### DIFF
--- a/tests/e2e/lib/cli_common.go
+++ b/tests/e2e/lib/cli_common.go
@@ -1,12 +1,23 @@
 package lib
 
 import (
+	"context"
 	"fmt"
 	"log"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
+)
+
+// Default timeout for CLI commands that may hang (e.g., log streaming)
+const DefaultCLITimeout = 5 * time.Minute
+
+// Default retry settings for CLI commands
+const (
+	DefaultCLIRetries    = 3
+	DefaultCLIRetryDelay = 10 * time.Second
 )
 
 type CLICommand struct {
@@ -14,6 +25,7 @@ type CLICommand struct {
 	Action   string // "create", "get", "delete", etc.
 	Name     string
 	Options  []string
+	Timeout  time.Duration // Optional timeout for commands that may hang
 }
 
 func (c *CLICommand) Execute() ([]byte, error) {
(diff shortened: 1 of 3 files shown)


kaovilai:
if its 5 minutes here you're suggesting CLI itself should use even lower timeout? or is this a catchall if cli does not implement timeout

mpryc:
> if its 5 minutes here you're suggesting CLI itself should use even lower timeout? or is this a catchall if cli does not implement timeout

for CI I think we are fine 5min as we know the download speed is pretty fast, but on the CLI side I would not assume so, so timeout here is separate to CLI itself and we do not look here at the CLI. Another option would be to use env from CLI here, but than we need to have CLI change in and new CLI available

kaovilai:
So should CLI have default timeout higher than 5?

mpryc:
@kaovilai it does: https://github.com/migtools/oadp-cli/pull/100/files#diff-bbf899c050411e5818783e2c2a8a7d6c6130576814c10d7c8c47772899899101R38

tests/e2e/lib/cli_common.go
```diff
@@ -40,6 +52,103 @@ func (c *CLICommand) ExecuteOutput() ([]byte, error) {
 	return cmd.Output()
 }
 
+// ExecuteWithTimeout executes the CLI command with a timeout.
+// If the timeout is exceeded, the command is killed and an error is returned.
+func (c *CLICommand) ExecuteWithTimeout(timeout time.Duration) ([]byte, error) {
```
kaovilai, line 57: in future this could accept cancellable context so that line 66 use it instead of context.Background().. I think could be useful for ctrl+c? not sure.

kaovilai (approved):

tests/e2e/lib/cli_common.go
mpryc: Yes, ctrl+c is valid scenario, dunno how often could be executed outside of PROW tbh. PROW doesn't require ctrl+c at all, so only manual run + ctrl+c, very limited imo.

Joeavaikath (approved):
LGTM!

weshayutin (approved):
/LGTM

shubham-pampattiwar (approved):

PR #2052, opened 2026-01-05
Fix S3 bucket region unit tests by adding mocking support

shubham-pampattiwar:
## Summary
- Add `GetBucketRegionFunc` variable in `pkg/storage/aws/s3.go` to enable mocking in tests
- Refactor `s3_test.go` to use mocks instead of real AWS network calls
- Add mock setup in `bsl_test.go` and `common_test.go` for tests that depend on bucket region discovery

## Problem
The unit tests for S3 bucket region auto-discovery were making real network calls to AWS buckets (`openshift-velero-plugin-s3-auto-region-test-*`). This caused failures in CI environments when:
- Network access to AWS was restricted
- Bucket policies changed or access was denied

Failing tests:
- `TestGetBucketRegion`
- `TestBucketRegionIsDiscoverable`
- `TestDPAReconciler_ValidateBackupStorageLocations` (specific cases using `DiscoverableBucket`)
- `TestUpdateBackupStorageLocation` (region auto-detection case)

## Solution
Introduced a mockable function variable pattern:
```go
var GetBucketRegionFunc = getBucketRegionImpl
```

This allows tests to replace the real AWS implementation with a mock that returns predictable values.

## Test plan
- [x] Run `go test ./pkg/storage/aws/...` - all tests pass
- [x] Run `go test ./internal/controller/... -run TestDPAReconciler_ValidateBackupStorageLocations` - passes
- [x] Run `go test ./pkg/common/... -run TestUpdateBackupStorageLocation` - passes
- [x] Run `make test` - all unit tests pass

### DIFF
--- a/internal/controller/bsl_test.go
+++ b/internal/controller/bsl_test.go
@@ -25,6 +25,7 @@ import (
 	oadpv1alpha1 "github.com/openshift/oadp-operator/api/v1alpha1"
 	"github.com/openshift/oadp-operator/pkg/common"
 	"github.com/openshift/oadp-operator/pkg/credentials/stsflow"
+	"github.com/openshift/oadp-operator/pkg/storage/aws"
 )
 
 // A bucket that region can be automatically discovered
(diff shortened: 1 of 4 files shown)


weshayutin (approved):
Thanks Shubham!  I would have not made this as clear and simple. Nice work

sseago (approved):

kaovilai:
just gonna go with the flow..

/cherry-pick oadp-1.5

kaovilai:
slack thread: https://redhat-internal.slack.com/archives/C039LRSDC8Z/p1767645968563479

TL;DR real buckets used in unit tests were deleted and team decided using real api call in unit test is bad idea.

kaovilai:
Note that the following is currently false. You cannot change bucket policies to prevent this access (anonymous cred is allowed to query private buckets)

> Bucket policies changed or access was denied

PR #2082, opened 2026-02-09
Fix IBU delay on SNO by waiting for CRDs and disabling leader election

shubham-pampattiwar:
## Summary

Fixes the ~8 minute delay experienced during Image-Based Upgrade (IBU) on Single Node OpenShift (SNO) clusters before DPA reaches `Reconciled=True`.

**Root Cause:**
1. Controller crashed when Route/SCC CRDs weren't available during cluster initialization (2 min cache sync timeout)
2. New instance had to wait for leader lease to expire (4.5 min on SNO)

**Fixes Implemented:**
- **CRD availability wait**: Poll discovery API for Route/SCC CRDs before starting controller
- **LeaderElectionReleaseOnCancel**: Release leader lease on crash for immediate restart
- **Disable leader election on SNO**: No benefit with replicas=1, removes overhead

**Impact:**
| Scenario | Before | After |
|----------|--------|-------|
| SNO IBU with CRD delay | ~8 min | < 1 min |
| SNO normal startup | ~60s | ~30s |
| HA cluster | unchanged | unchanged |

## Test Plan

- [x] Unit tests for CRD wait functions (4 new tests)
- [x] Unit tests for leader election (8 new tests)
- [x] `make lint` passes
- [x] `go test ./cmd/... ./pkg/leaderelection/...` passes
- [ ] E2E validation on SNO IBU (pending IBU team validation)

Fixes: https://issues.redhat.com/browse/OADP-7419

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -23,6 +23,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	snapshotv1api "github.com/kubernetes-csi/external-snapshotter/client/v6/apis/volumesnapshot/v1"
 	configv1 "github.com/openshift/api/config/v1"
@@ -36,6 +37,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
+	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/discovery"
 	"k8s.io/client-go/kubernetes"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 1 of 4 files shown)


cmd/main.go
```diff
@@ -139,6 +148,16 @@ func main() {
...
 
+	// Wait for required OpenShift CRDs before starting the manager.
+	// During IBU (Image-Based Upgrade) on SNO clusters, the node reboots and external
+	// OpenShift operators may not have registered their CRDs yet. Without this wait,
+	// the controller crashes trying to watch Route/SCC resources that don't exist.
+	// See: https://issues.redhat.com/browse/OADP-7419
+	if err := waitForRequiredCRDs(context.Background(), kubeconf); err != nil {
+		setupLog.Error(err, "failed waiting for required CRDs")
+		os.Exit(1)
+	}
```
kaovilai, line 159: Can we just crash the pod and have k8s restart? Once k8s CRDs are in place we stop crashing.

shubham-pampattiwar: Crashing and restarting is the current behavior - it's what causes the ~8 min delay. Each crash takes ~2 min (cache sync timeout), and with CrashLoopBackOff the delays compound. The CRD wait converts an uncontrolled crash loop into a controlled 10-second poll loop, which is faster and provides clear logging of what we're waiting for.

kaovilai (approved):

sseago (approved):

PR #2092, opened 2026-02-16
Remove unused printer column from VMFR CRD

PrasadJoshi12:
## Why the changes were made

Fixes:[OADP-7460]( https://issues.redhat.com/browse/OADP-7460 )


shubham-pampattiwar (commented):
IMO If this particular printer column is not being used then we should first decide whether we want to use it or not.
If yes, we want to use then Lets implement the controller changes needed for it.
If no, then the CRD needs to be changed in VMFR controller repo first and then make the change here in OADP Operator CRDs or else the CRDs would be out of sync. 
Let me know what you guys think ? @PrasadJoshi12 @mpryc

PR #2103, opened 2026-02-24
Align DPA defaultVolumesToFsBackup with Velero backup spec

mail2nadeem92:
- Add defaultVolumesToFsBackup field (matches backup.spec.defaultVolumesToFsBackup)
- Deprecate defaultVolumesToFSBackup, keep for backwards compatibility
- Controller prefers new field, falls back to deprecated field when unset
- Add tests for new field, backwards compat, and precedence

## Why the changes were made

Dpa.spec.configuration flag **defaultVolumesToFSBackup** is not identical to the one used by the backup spec **defaultVolumesToFsBackup**, which can be confusing for some users.  Make these flags identical to each other. 

$ oc explain dpa.spec.configuration.velero.defaultVolumesToFSBackup
KIND:     DataProtectionApplication
VERSION:  oadp.openshift.io/v1alpha1
FIELD:    defaultVolumesToFSBackup <boolean>
DESCRIPTION:
     Use pod volume file system backup by default for volumes

$ oc explain backup.spec.defaultVolumesToFsBackup
KIND:     Backup
VERSION:  velero.io/v1

DESCRIPTION:
     DefaultVolumesToFsBackup specifies whether pod volume file system backup
     should be used for all volumes by default. 

## Summary by CodeRabbit

* **New Features**
  * Introduced a new standardized configuration field for specifying default volume-to-filesystem backup behavior in Velero operations.

* **Deprecations**
  * Marked previous volume-backup configuration field as deprecated. Existing configurations continue to work with full backwards compatibility maintained while users transition to the new field.

### DIFF
--- a/api/v1alpha1/dataprotectionapplication_types.go
+++ b/api/v1alpha1/dataprotectionapplication_types.go
@@ -327,7 +327,11 @@ type VeleroConfig struct {
 	// How long to wait on asynchronous BackupItemActions and RestoreItemActions to complete before timing out. Default value is 1h.
 	// +optional
 	DefaultItemOperationTimeout string `json:"defaultItemOperationTimeout,omitempty"`
-	// Use pod volume file system backup by default for volumes
+	// Use pod volume file system backup by default for volumes.
+	// Matches backup.spec.defaultVolumesToFsBackup in Velero API.
+	// +optional
+	DefaultVolumesToFsBackup *bool `json:"defaultVolumesToFsBackup,omitempty"`
+	// Deprecated: Use defaultVolumesToFsBackup instead (matches Velero backup spec).
 	// +optional
 	DefaultVolumesToFSBackup *bool `json:"defaultVolumesToFSBackup,omitempty"`
 	// DisableFsBackup determines whether the NodeAgent should disable file system backup.
--- a/api/v1alpha1/zz_generated.deepcopy.go
+++ b/api/v1alpha1/zz_generated.deepcopy.go
@@ -1326,6 +1326,11 @@ func (in *VeleroConfig) DeepCopyInto(out *VeleroConfig) {
 		*out = new(PodConfig)
 		(*in).DeepCopyInto(*out)
 	}
+	if in.DefaultVolumesToFsBackup != nil {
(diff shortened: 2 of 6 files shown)


mpryc:
@mail2nadeem92 
 1. do we need similar change for the non-admin spec?
 2. do we have some default example referencing deprecated `defaultVolumesToFSBackup` that will be still visible in the console. 
 
this PR  /lgtm, just asking here to ensure we don't miss something.

mail2nadeem92:
> @mail2nadeem92
> 
> 1. do we need similar change for the non-admin spec?

Not required for non-admin as NonAdminBackup’s backupSpec and DPA’s enforceBackupSpec use Velero’s BackupSpec type.
Velero’s BackupSpec already uses defaultVolumesToFsBackup (correct spelling).
The NonAdminBackup CRD (oadp.openshift.io_nonadminbackups.yaml) comes from the NAC repo via make update-non-admin-manifests and already exposes defaultVolumesToFsBackup.
So the inconsistency only existed in DPA’s configuration.velero section, which this PR fixes.

> 2. do we have some default example referencing deprecated `defaultVolumesToFSBackup` that will be still visible in the console.

No default examples reference the deprecated field.
Checked:
CSV alm-examples (bundle/manifests/oadp-operator.clusterserviceversion.yaml): DPA example only has defaultPlugins, disableFsBackup, etc. It does not set defaultVolumesToFSBackup or defaultVolumesToFsBackup.
Config samples (config/samples/oadp_v1alpha1_dataprotectionapplication.yaml): Same; no volume backup fields.
NonAdminBackup example: backupSpec: {} only; no volume backup fields.
So the examples shown in the console do not use the deprecated field.

When users create or edit a DPA in the console, the CRD schema will show both fields:
**defaultVolumesToFsBackup (new, preferred)
defaultVolumesToFSBackup (deprecated, with deprecation description)**
This is intentional for backwards compatibility. The deprecated field is clearly marked, and the new field is the recommended one.

> 
> this PR /lgtm, just asking here to ensure we don't miss something.

sseago (approved):

mpryc (approved):

shubham-pampattiwar (approved):

kaovilai (approved):

PR #1740, opened 2025-05-02
WIP: OADP-5777: auto region cred test

kaovilai:
Signed-off-by: Tiger Kaovilai <tkaovila@redhat.com>

PR #2109, opened 2026-02-27
OADP-5777: Require region in BSL validation when s3Url is set, fix nil config skipping region auto-detection

### LINKED ISSUE #2108: BSL validation should require region when s3Url is set to a non-AWS endpoint
## Summary

When a BackupStorageLocation uses `provider: aws` with a custom `s3Url` pointing to a non-AWS S3-compatible service (e.g., IBM Cloud Object Storage, MinIO, NooBaa), the BSL validation does not reliably require the `region` field. The DPA is accepted as valid, but Velero subsequently fails to connect.

## Problem

The region validation in `internal/controller/bsl.go:533-534` currently requires `region` only when:
1. `s3ForcePathStyle` is explicitly `"true"`, **OR**
2. `BucketRegionIsDiscoverable(bucket)` returns `false`

```go
if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0) &&
    (bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" || !aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
    return fmt.Errorf("region for AWS backupstoragelocation not automatically discoverable...")
}
```

`BucketRegionIsDiscoverable` (`pkg/storage/aws/s3.go:20-23`) calls AWS's HeadBucket API against `s3.us-east-1.amazonaws.com` to auto-discover the bucket's region. This is incorrect when `s3Url` points to a non-AWS endpoint because:

- The bucket may not exist on AWS at all, so discovery fails (validation works by accident)
- A bucket with the same name **may** exist on AWS, so discovery succeeds with a wrong region — validation passes, but Velero fails to connect to the actual non-AWS endpoint
- The operator is querying AWS infrastructure for region information about a bucket that lives on a completely different service

## Context: IBM Cloud Object Storage

Previously, IBM Cloud Object Storage users were required to set `s3ForcePathStyle: true`, which triggered the first validation branch (requiring `region`). IBM COS no longer requires `s3ForcePathStyle`. When users remove `s3ForcePathStyle` from their config, they lose the guaranteed region validation, and the operator falls through to the unreliable `BucketRegionIsDiscoverable` check against AWS.

## Steps to Reproduce

1. Configure a DPA with `provider: aws`, a custom `s3Url` (e.g., IBM COS endpoint), a valid bucket name, but **without** `region` and **without** `s3ForcePathStyle`
2. DPA validation passes (operator reports the BSL is OK)
3. Velero fails to connect to the storage backend

## Expected Behavior

When `s3Url` is set in BSL config, the operator should require `region` to be explicitly specified, since auto-discovery via AWS API is not valid for non-AWS S3 endpoints.

## Suggested Fix

Add a check in `validateAWSBackupStorageLocation` that requires `region` whenever a custom `s3Url` is configured:

```go
// When s3Url is set, the user is pointing to a non-AWS S3-compatible service.
// Region auto-discovery via AWS API is not valid in this case.
if bslSpec.Config != nil && len(bslSpec.Config[S3URL]) > 0 &&
    (len(bslSpec.Config[Region]) == 0) {
    return fmt.Errorf("region is required when s3Url is set for AWS backupstoragelocation. Region cannot be auto-discovered for non-AWS S3-compatible endpoints")
}
```

## Affected Code

- `internal/controller/bsl.go:509-541` — `validateAWSBackupStorageLocation()`
- `pkg/storage/aws/s3.go:20-23` — `BucketRegionIsDiscoverable()` (calls AWS API regardless of configured endpoint)

> [!Note]

kaovilai:
## Why the changes were made

This PR addresses two related issues with AWS BSL region handling:

### 1. Fix nil config skipping region auto-detection ([OADP-5777](https://redhat.atlassian.net/browse/OADP-5777))

**Root cause**: When DPA `backupLocations` has no `config:` section, `bslSpec.Config` is nil. The `UpdateBackupStorageLocation` function guarded all AWS-specific logic (region auto-detection, checksumAlgorithm defaults) behind `if bslSpec.Config != nil`, so when config was omitted entirely, none of that logic ran.

This is why PR #1740's auto-detection fix didn't work in practice — the nil config guard skipped it. Verified by reproducing Tareq's exact scenario (DPA with AWS provider, no config section, no region) on a live cluster.

**Fix**: Initialize `bslSpec.Config` to an empty map when nil, so AWS logic runs regardless of whether user specifies a config section.

### 2. Require region when s3Url is set

When a custom `s3Url` is configured, the BSL points to a non-AWS S3-compatible service (e.g. IBM Cloud Object Storage, MinIO) where region auto-discovery via AWS HeadBucket API is not valid. Previously, region was only enforced when `s3ForcePathStyle` was true or AWS discovery failed.

**Fix**: Validate that region is specified at DPA validation time when `s3Url` is set, with context-specific error messages.

### 3. Add logging to region auto-detection

Added `logr.Logger` parameter to `UpdateBackupStorageLocation` so auto-detection success/failure is visible in operator logs instead of silently falling through.

Fixes: https://issues.redhat.com/browse/OADP-5777
Fixes: https://github.com/openshift/oadp-operator/issues/2108

### Unit tests
```bash
go test -v ./pkg/common/... -run TestUpdateBackupStorageLocation
go test -v ./internal/controller/... -run TestValidateBackupStorageLocations
```

### Manual verification (nil config fix)
1. Create DPA with AWS provider, **no config section**, no region:
```yaml
spec:
  backupLocations:
  - velero:
      credential:
        key: cloud
        name: cloud-credentials
      default: true
      objectStorage:
        bucket: <your-bucket>
        prefix: velero
      provider: aws
```
2. Check operator logs — should see: `Auto-detected AWS bucket region {"bucket": "...", "region": "..."}`
3. Check BSL — should have `config.region` and `config.checksumAlgorithm` set

### Verified on cluster
- OCP 4.22 ARM64 cluster on AWS
- DPA with no config section, bucket `tkaovila-sts-oadp`
- Before fix: BSL had no config at all (nil config guard skipped everything)
- After fix: BSL has `region: us-east-1` and `checksumAlgorithm: ""` auto-populated

> [!Note]

## Summary by CodeRabbit

* **Bug Fixes**
  * Validation now requires an explicit region when an S3 URL is provided, when s3ForcePathStyle is true, or when bucket-region discovery fails; error messages were clarified to indicate the specific missing-region reason and guide users to set the region in the backup storage location configuration.
* **Tests**
  * Added/updated tests covering S3 URL, s3ForcePathStyle, and region presence scenarios and expected error messages.
* **Chores**
  * Improved logging around bucket-region auto-detection to surface detection failures and reported regions.

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -413,7 +413,7 @@ func (r *DataProtectionApplicationReconciler) updateBSLFromSpec(bsl *velerov1.Ba
 	}
 
 	// Update BSL spec and registry-deployment label
-	if err := common.UpdateBackupStorageLocation(bsl, bslSpec); err != nil {
+	if err := common.UpdateBackupStorageLocation(bsl, bslSpec, r.Log); err != nil {
 		return err
 	}
 
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -527,12 +527,16 @@ func (r *DataProtectionApplicationReconciler) validateAWSBackupStorageLocation(b
 	}
 
 	// BSL region is required when
+	// - s3Url is set, because the user is pointing to a non-AWS S3-compatible service
+	//   where region auto-discovery via AWS API is not valid
 	// - s3ForcePathStyle is true, because some velero processes requires region to be set and is not auto-discoverable when s3ForcePathStyle is true
 	//   imagestream backup in openshift-velero-plugin now uses the same method to discover region as the rest of the velero codebase
 	// - even when s3ForcePathStyle is false, some aws bucket regions may not be discoverable and the user has to set it manually
 	if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0) &&
-		(bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" || !aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
-		return fmt.Errorf("region for AWS backupstoragelocation not automatically discoverable. Please set the region in the backupstoragelocation config")
+		(bslSpec.Config != nil && len(bslSpec.Config[S3URL]) > 0 ||
+			bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" ||
+			!aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
+		return fmt.Errorf("region for AWS backupstoragelocation is required. Please set the region in the backupstoragelocation config")
 	}
 
 	//TODO: Add minio, noobaa, local storage validations
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -532,11 +532,15 @@ func (r *DataProtectionApplicationReconciler) validateAWSBackupStorageLocation(b
 	// - s3ForcePathStyle is true, because some velero processes requires region to be set and is not auto-discoverable when s3ForcePathStyle is true
 	//   imagestream backup in openshift-velero-plugin now uses the same method to discover region as the rest of the velero codebase
 	// - even when s3ForcePathStyle is false, some aws bucket regions may not be discoverable and the user has to set it manually
-	if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0) &&
-		(bslSpec.Config != nil && len(bslSpec.Config[S3URL]) > 0 ||
-			bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" ||
-			!aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
-		return fmt.Errorf("region for AWS backupstoragelocation is required. Please set the region in the backupstoragelocation config")
+	if bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0 {
+		switch {
+		case bslSpec.Config != nil && len(bslSpec.Config[S3URL]) > 0:
+			return fmt.Errorf("region for AWS backupstoragelocation is required when s3Url is configured. Please set the region in the backupstoragelocation config")
+		case bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true":
+			return fmt.Errorf("region for AWS backupstoragelocation is required when s3ForcePathStyle is true. Please set the region in the backupstoragelocation config")
+		case !aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket):
+			return fmt.Errorf("region for AWS backupstoragelocation not automatically discoverable. Please set the region in the backupstoragelocation config")
+		}
 	}
 
 	//TODO: Add minio, noobaa, local storage validations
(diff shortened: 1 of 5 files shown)


mpryc (dismissed):

kaovilai:
/override ci/prow/4.21-e2e-test-aws ci/prow/4.21-e2e-test-hcp-aws ci/prow/4.21-e2e-test-kubevirt-aws

we dont have 4.21 tests defined so retest wont do anything there.

internal/controller/bsl.go
```diff
@@ -527,12 +527,16 @@ func (r *DataProtectionApplicationReconciler) validateAWSBackupStorageLocation(b
...
 	// - s3ForcePathStyle is true, because some velero processes requires region to be set and is not auto-discoverable when s3ForcePathStyle is true
 	//   imagestream backup in openshift-velero-plugin now uses the same method to discover region as the rest of the velero codebase
 	// - even when s3ForcePathStyle is false, some aws bucket regions may not be discoverable and the user has to set it manually
 	if (bslSpec.Config == nil || len(bslSpec.Config[Region]) == 0) &&
-		(bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" || !aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
-		return fmt.Errorf("region for AWS backupstoragelocation not automatically discoverable. Please set the region in the backupstoragelocation config")
+		(bslSpec.Config != nil && len(bslSpec.Config[S3URL]) > 0 ||
+			bslSpec.Config != nil && bslSpec.Config[S3ForcePathStyle] == "true" ||
+			!aws.BucketRegionIsDiscoverable(bslSpec.ObjectStorage.Bucket)) {
+		return fmt.Errorf("region for AWS backupstoragelocation is required. Please set the region in the backupstoragelocation config")
```
shubham-pampattiwar: Minor NIT: The new error message "region for AWS backupstoragelocation is required" is now shared across three trigger conditions (s3Url set, s3ForcePathStyle true, undiscoverable bucket). The previous message "not automatically discoverable" gave users a hint about why region was needed, that auto-discovery was attempted and failed. Consider keeping context-specific messages, e.g. "region is required when s3Url is configured" vs "region not automatically discoverable" for the fallback case.

shubham-pampattiwar (commented):
This is the right fix, but it tightens validation in a way that could break existing DPA configs on operator upgrade. Users with s3Url but no region who happened to have a discoverable bucket name will go from "working" to "validation error" after upgrading. Worth adding a line in the commit message or PR description acknowledging this as an intentional breaking change. No code change needed, just want to make sure QE covers the upgrade path.

internal/controller/bsl.go
kaovilai: Will make the changes requested

shubham-pampattiwar (dismissed):
/lgtm

kaovilai:
added logging to help test https://redhat.atlassian.net/browse/OADP-5777

weshayutin (dismissed):

kaovilai:
/retest

Pod scheduling timeout.

weshayutin (approved):
/LGTM
override 4.23 e2e

shubham-pampattiwar (approved):

PR #2133, opened 2026-03-24
DNM: Test: disable pod restart workaround after file-system restore

weshayutin:
## Summary
- Comments out the post-restore pod deletion workaround for KOPIA (file-system) backups in both `backup_restore_suite_test.go` and `backup_restore_cli_suite_test.go`
- This workaround was added because OVN-Kubernetes didn't fully wire the network namespace for pods recreated by Velero with a restore-wait init container
- Goal: determine whether this workaround is still needed by running E2E tests without it

## Test plan
- [ ] KOPIA backup/restore E2E tests pass without the pod restart workaround
- [ ] CLI KOPIA backup/restore E2E tests pass without the pod restart workaround
- [ ] If tests fail with networking issues, revert this change and keep the workaround

Made with [Cursor](https://cursor.com)


kaovilai:
> [!Note]

Closing this PR because it was pushed directly to the protected `openshift/oadp-operator` branch and cannot be amended to. Reopening from fork branch instead: #2134

PR #2136, opened 2026-03-26
Introduce v1alpha2 API with string-based duration fields and conversion

Joeavaikath:
## Why the changes were made

Fixes [oadp-3471](https://redhat.atlassian.net/browse/OADP-3471)

Should be able to create a DPA with the new API version.
This allows the user to specify the timout in the string format (metav1.Duration), and the conversion tech helps translate it to the existing version avoiding issues.


Joeavaikath:
/hold this approach fails because:
- Conversion webhooks are cluster-scoped — because CRDs are cluster-scoped, the API server calls the webhook for any namespace
  - OLM requires AllNamespaces install mode when the CSV declares conversionCRDs
  - OADP only supports OwnNamespace — its RBAC, NonAdmin feature, and controller logic all assume single-namespace operation
 
Since only the main namespace is being watched, maybe changing this isn't the worst thing, but I'm guessing we need discussion around this

kaovilai:
@weshayutin  Reminder on why we need the new v1alpha2 API:

1. **OLMv1 on OCP 5 will require CRD upgrade safety** — our CRDs will need to comply with the [OLMv1 CRD upgrade safety requirements](https://github.com/operator-framework/operator-controller/blob/main/docs/concepts/crd-upgrade-safety.md?plain=1). Introducing v1alpha2 now lets us evolve the API without violating those constraints.

2. **Fewer breaking changes** — by adding a new API version alongside v1alpha1 rather than modifying it in place, existing users are not broken.

3. **Follows Kubernetes CRD versioning best practices** — this aligns with the [official Kubernetes CRD versioning/upgrade guide](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/#writing-reading-and-updating-versioned-customresourcedefinition-objects). We can also add a migrator in the future when we're ready to drop v1alpha1, making the eventual deprecation path clean and well-defined.

docs/design/oadp-3471-args-duration-type_design.md
```diff
@@ -0,0 +1,180 @@
...
+2. Add `Hub()` marker to v1alpha1 `DataProtectionApplication`.
+3. Implement `ConvertTo()`/`ConvertFrom()` on v1alpha2 `DataProtectionApplication`.
+4. Update CRD kustomization to include both versions with conversion webhook config.
+5. Run `make generate && make manifests && make bundle`.
+6. Add unit tests for round-trip conversion.
+7. Add E2E test: create DPA via v1alpha2 with string durations, verify Velero deployment args.
+
+## Open Issues
+
+- **OLM install mode conflict**: OLM requires `AllNamespaces` install mode when the CSV declares `conversionCRDs` in `webhookdefinitions`. OADP currently only supports `OwnNamespace`. Switching to `AllNamespaces` doesn't change the operator's runtime behavior (it still watches a single namespace via `watchNamespace`), but it changes what OLM allows users to select at install time. This needs team discussion before proceeding — it may affect how the operator is presented in the OpenShift console and could have implications for multi-tenancy expectations.
```
Joeavaikath: Thoughts on this part @kaovilai?
I can test around this to see what happens

Joeavaikath:
Will work on a design for all the changes required for v1alpha2
Closing

PR #2151, opened 2026-04-02
More details on restore design

sseago:
## Why the changes were made

This PR updates the restore design for kubevirt datamover with more details.

### DIFF
--- a/docs/design/kubevirt-datamover.md
+++ b/docs/design/kubevirt-datamover.md
@@ -59,12 +59,18 @@ Taking a VolumeSnapshot and then using kopia to process the entire volume and co
     - Note that unlike the built-in datamover, this DataUpload is tied to a VM (which could include multiple PVCs) rather than to a single PVC.
   - An annotation will be added to the DataUpload identifying the VirtualMachine we're backing up.
   - Add `velerov1api.DataUploadNameAnnotation` to VirtualMachine
+  - Add `velerov1api.PVCNamespaceNameLabel` annotation to VirtualMachine (doesn't need to be a label, since we're just using it to figure out what label selector to use for the ConfigMap on restore).
   - OperationID will be created and returned similar to what's done with the CSI PVC plugin, and the async operation Progress method will report on progress based on the DU status (similar to CSI PVC plugin)
+- PVC BIA plugin
+  - Add `kubevirt-datamover-vm` annotation to PVC with the `VirtualMachine` name to signal to RIA that we need to remove `VolumeName` and set `Selector.MatchLabels` on PVC.
 - VirtualMachine RIA plugin
-  - (further investigation still needed on exactly what's needed here)
   - Similar in functionality to csi PVC restore action
-  - Create temporary PVC (needed, or is this handled by controller?)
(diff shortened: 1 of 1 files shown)


kaovilai (approved):

shubham-pampattiwar (approved):

shubham-pampattiwar:
@mpryc please review !

docs/design/kubevirt-datamover.md
```diff
@@ -133,6 +158,7 @@ Per-VM Index (checkpoints/<ns>/<vm-name>/index.json):
       "vmBackup": "vmb-001",
       "files": ["vmb-001-disk0.qcow2", "vmb-001-disk1.qcow2"],
       "pvcs": ["my-vm-pvc-0", "my-vm-pvc-1"],
+      "pvcSizes": ["10Gi", "20Gi"],
       "referencedBy": ["backup-2025-01-10", "backup-2025-01-11", "backup-2025-01-12"]
```
mpryc, line 162: I know it's changing quite a bit, but since we are still in the design phase, should we add a parent directory for the backup chain (e.g., the ID of the base full backup) between the VM name and the individual checkpoints?

**Proposed Structure:**
* **Current:** `checkpoints/<namespace>/<vm-name>/<checkpoint-id>/`
* **Proposed:** `checkpoints/<namespace>/<vm-name>/<base-full-backup-id>/<checkpoint-id>/`

**Why this helps (Short Explanation):**

1. **O(1) S3 API Queries:** Object storage uses flat prefixes, not real directories. By grouping a chain under one prefix, we can calculate the total size of a full backup + all its incrementals using a single `ListObjectsV2` API call, instead of making separate calls for every single checkpoint.
2. **Simplified Garbage Collection:** When a backup chain expires or is broken, we can delete all associated `qcow2` files safely and easily by deleting the single `<base-full-backup-id>` prefix, rather than parsing JSON to delete $N$ individual checkpoint paths.

sseago: @mpryc What exactly is the "base full backup ID"? Velero backup ID or or something else? What happens when this backup expires but the subsequent incremental backups have not? Where is the reference to this ID in the metadata files for these grouped operations?

mpryc: @sseago thinking laud here:

1. The "base full backup ID" could be the checkpoint ID of the full backup that started the chain. 
2. When subsequent inc backups are still not expired we do not delete the s3 files from the chain. Only when the last inc backup expires, and the entire chain's "referencedBy" lists are completely empty, we can delete the entire folder.
3.
 ```json
 "checkpoints":[
  {
    "id": "vmb-demo-1",
    "type": "Full",
    "chainId": "chain-vmb-demo-1"
  },
  {
    "id": "vmb-demo-2",
    "type": "Incremental",
    "parent": "vmb-demo-1",
    "chainId": "chain-vmb-demo-1"  <-- Maps to the same S3 parent prefix
  }
]
```

I am not writing here the entire definition, just asking if the idea would allow to simplify calculations on required storage for temporary pvcs, reduce number of s3 calls and allow to cleanup easier, but maybe not. S3 API calls aren't that problematic as we won't have to do many of those.

PR #2152, opened 2026-04-07
Remove setup-envtest if different arch during `make test`

kaovilai:


kaovilai:
Folding this into #2367 to reduce review load — closing in favor of that PR, which covers this fix plus the same underlying issue for controller-gen/kustomize/golangci-lint.

Along the way I found the `--help`-exit-code check here would have triggered a reinstall on *every* `make envtest` invocation (setup-envtest's own `--help` exits 2 by its own convention, even on a working binary) — #2367 checks for exit code 126 specifically (exec format error) instead, verified against a real cross-compiled linux/amd64 binary run on darwin/arm64.

> [!Note]

PR #2170, opened 2026-04-21
OADP-7868: Skip CLI/VMDP download setup when Console CRD is absent

Joeavaikath:
## Why the changes were made

Fixes: [OADP-7868](https://issues.redhat.com/browse/OADP-7868)

The OADP controller-manager enters `CrashLoopBackOff` on OpenShift 4.22 clusters that do not have the Console capability enabled (e.g. SNO clusters).

During startup, the `CLIDownloadSetup` and `VMDPDownloadSetup` runnables unconditionally interact with the `ConsoleCLIDownload` CRD (`console.openshift.io/v1`). On clusters without Console capability, this CRD does not exist, causing the REST mapper to fail with a discovery error. The code only handles `errors.IsNotFound` (object-level absence) and does not account for CRD-level absence (`meta.NoKindMatchError`), resulting in a fatal error that terminates the manager.

The fix probes the REST mapper for the `ConsoleCLIDownload` GVK at the top of each runnable's `Start()` method. When the CRD is absent (`meta.IsNoMatchError`), setup is skipped gracefully — no deployment, service, route, or `ConsoleCLIDownload` resources are created. This is the correct behavior since these resources exist solely to surface CLI downloads through the Console UI, which is not present on these clusters.

1. Run unit tests:
   ```bash
   go test ./internal/controller/ -run "TestIsConsoleCRD|TestCLIDownload|TestVMDPDownload" -v
   ```
   All 4 tests should pass, covering:
   - `isConsoleCRDAvailable` returns false when CRD is absent
   - `isConsoleCRDAvailable` returns true when CRD is present
   - `CLIDownloadSetup.Start()` returns nil and creates no resources when CRD is absent
   - `VMDPDownloadSetup.Start()` returns nil and creates no resources when CRD is absent

2. Deploy on a cluster **with** Console capability and verify CLI/VMDP download resources are still created normally (no regression).

3. Deploy on a cluster **without** Console capability (e.g. SNO with Console disabled) and verify the operator starts successfully without CrashLoopBackOff. Logs should show:
   ```
   ConsoleCLIDownload CRD not available (Console capability not enabled), skipping setup
   ```

## Release Notes

* **New Features**
  * Added detection for ConsoleCLIDownload custom resource availability.
  * CLI and VMDP download features are now conditionally enabled based on CRD presence.

* **Tests**
  * Added comprehensive test coverage for CRD availability detection.


weshayutin:
this is good and should fix the bug.  Ping me in slack as I'd also like to know how users can choose to turn off the running download servers.

weshayutin:
this also makes me wonder if it's time to setup a SNO testing job here in prow

Joeavaikath:
```
INFO  vmdp-download-setup  ConsoleCLIDownload CRD not available (Console capability not enabled), skipping setup     
INFO  cli-download-setup   ConsoleCLIDownload CRD not available (Console capability not enabled), skipping setup 
```

internal/controller/vmdp_download_controller.go
```diff
@@ -42,6 +42,10 @@ func (v *VMDPDownloadSetup) Start(ctx context.Context) error {
 	v.Log = ctrl.Log.WithName("vmdp-download-setup")
 	v.Log.Info("Starting VMDP download setup")
 
+	if available, err := isConsoleCRDAvailable(v.Client.RESTMapper(), v.Log); !available {
+		return err
+	}
+
```
kaovilai: i would prefer single call in
https://github.com/openshift/oadp-operator/blob/024537e2c894b9627629d5b7b4061966206814b5/cmd/main.go#L302-L325

then simply don't add both the download setup Runnable interface to the manager.

```go
 if watchNamespace != "" && consoleCRDAvailable { 
   mgr.Add
```

Joeavaikath: That's a nice move

weshayutin (approved):

shubham-pampattiwar (approved):

kaovilai (commented):
/lgtm

PR #2213, opened 2026-05-21
disable early frequent polling by default

sseago:
## Why the changes were made

Early frequent polling is more sensitive to other configuration parameters (parallel backups, item block worker count, qps/burst) in some environments. Disable it by default to prevent perf-related regression

Confirm that by default, the `CSI_SNAPSHOT_EARLY_FREQUENT_POLLING` env var is not set on the velero container and that if dpa.spec.disableCSISnapshotEarlyFrequentPolling is true, then `disableCSISnapshotEarlyFrequentPolling` is also set to true.

## Summary by CodeRabbit

* **Configuration Changes**
  * Updated the DataProtectionApplication specification by replacing `DisableCSISnapshotEarlyFrequentPolling` with `EnableCSISnapshotEarlyFrequentPolling`. CSI snapshot polling is now disabled by default and requires explicit enablement in your configuration.

### DIFF
--- a/api/v1alpha1/dataprotectionapplication_types.go
+++ b/api/v1alpha1/dataprotectionapplication_types.go
@@ -373,11 +373,11 @@ type VeleroConfig struct {
 	// LoadAffinityConfig is the config for data path load affinity.
 	// +optional
 	LoadAffinityConfig []*LoadAffinity `json:"loadAffinity,omitempty"`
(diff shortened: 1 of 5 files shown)


kaovilai (approved):

shubham-pampattiwar (approved):
/lgtm

PR #2227, opened 2026-05-26
[oadp-1.4] disable early frequent polling by default

sseago:
## Why the changes were made
Early frequent polling is more sensitive to other configuration parameters (parallel backups, item block worker count, qps/burst) in some environments. Disable it by default to prevent perf-related regression


kaovilai (approved):

weshayutin (approved):
/LGTM

PR #2243, opened 2026-06-17
OADP-1.5: remove must-gather from operator repo

weshayutin:
CVE BURDEN

backport https://github.com/openshift/oadp-operator/pull/2007


weshayutin:
@shubham-pampattiwar @Joeavaikath @kaovilai 
Either we need to nuke the must-gather dir in operator or exclude it from builds in the konflux.Dockerfile.  Any opinions?

kaovilai:
> exclude it from builds in the konflux.Dockerfile

That is the right idea because it is only used in end-to-end tests, which is not an active path of code in production. 

During tests, we need it. It's tagged to be refactoring towards using the repository that we moved to, instead of the code in here. 

We can also hide it behind a build flag that is only triggered on end-to-end tests or vice versa.

kaovilai:
Please see  https://github.com/openshift/oadp-operator/pull/2007 for inspiration

Joeavaikath:
Future releases should have it handled since it's been separated out, thanks to #2007

kaovilai (approved):

shubham-pampattiwar (approved):
/lgtm

PR #2272, opened 2026-07-01
validate container runtime availability before container operations

### LINKED ISSUE #2249: Makefile container tool check only verifies binary exists, not runtime readiness
## Problem

The Makefile's container tool availability check uses `command -v` which only verifies the binary exists on PATH:

```makefile
ifeq ($(shell command -v $(CONTAINER_TOOL) >/dev/null 2>&1 && echo found),)
  $(error The selected container tool '$(CONTAINER_TOOL)' is not available on this system.)
endif
```

On macOS with Podman, the `podman` binary is always present even when the Podman machine is stopped. This means the check passes, the Makefile proceeds with targets like `deploy-olm` (which first runs `undeploy-olm`), and only fails later when `podman build` is actually invoked. By that point, the existing deployment may already be torn down.

## Expected behavior

The Makefile should detect that the container runtime is not functional **before** executing any destructive operations (like `undeploy-olm`). A check like `podman info` or `docker info` would verify the daemon/machine is actually running.

## Suggested fix

Replace the `command -v` check with a runtime health check:

```makefile
ifeq ($(shell $(CONTAINER_TOOL) info >/dev/null 2>&1 && echo found),)
  $(error The container tool '$(CONTAINER_TOOL)' is installed but not running. For podman on macOS, run: podman machine start)
endif
```

> [!Note]

Shreyashxredhat:
## Why
The Makefile currently checks only whether the selected container tool binary exists on PATH. This can pass even when the container runtime is not running, causing targets such as deploy-olm to fail later after teardown steps have already started.

## What changed
Added a container runtime readiness check and wired it into targets that directly use the container tool. This allows Make to fail early when the runtime is unavailable while leaving unrelated targets unaffected.

## Testing
- Verified normal execution with a running container runtime.
- Verified failure when the container runtime is unavailable.
- Verified non-container targets continue to work without requiring a running runtime.

Fixes #2249 

## Summary by CodeRabbit

* **Bug Fixes**
  * Added a pre-check for container-based build and deployment workflows to verify the selected container runtime is available.
  * Improved error handling with clearer guidance when the runtime is missing, helping avoid failed build, push, and deployment commands.
  * Applied the same validation across related packaging, catalog, upgrade, and must-gather workflows for more consistent behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -259,12 +259,19 @@ DOCKER_BUILD_ARGS ?= --platform=linux/amd64
 ifneq ($(CLUSTER_TYPE),)
 	DOCKER_BUILD_ARGS = --platform=$(CLUSTER_OS)/$(CLUSTER_ARCH)
 endif
+.PHONY: check-container-runtime
+check-container-runtime:
+	@$(CONTAINER_TOOL) info >/dev/null 2>&1 || \
+		{ echo "Error: The container tool '$(CONTAINER_TOOL)' is installed but its runtime is not available."; \
+		  echo "Ensure the daemon or machine is running (e.g., 'podman machine start' or 'systemctl start docker')."; \
+		  exit 1; }
+
 .PHONY: docker-build
-docker-build: ## Build docker image with the manager.
+docker-build: check-container-runtime ## Build docker image with the manager.
 	$(CONTAINER_TOOL) build --load -t $(IMG) . $(DOCKER_BUILD_ARGS)
 
 .PHONY: docker-push
-docker-push: ## Push docker image with the manager.
+docker-push: check-container-runtime ## Push docker image with the manager.
 	$(CONTAINER_TOOL) push ${IMG}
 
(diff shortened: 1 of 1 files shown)


Shreyashxredhat:
Hey @kaovilai , the fix which you suggested would run for every make command, like make help, make lint, make test. and it would fail when podman isnt running. I thought that was too broad, so i created `check-container-runtime` which does `$(CONTAINER_TOOL) info` and fails with a useful error message if runtime isnt available. That way we attach it only to targets that actually use containers, like - docker-build: check-container-runtime, bundle-build: check-container-runtime, etc. And one more important thing - i changed `deploy-olm: undeploy-olm` to `deploy-olm: check-container-runtime undeploy-olm`, so the runtime is validated before the teardown/build flow starts

- question - do we prefer a global runtime check for all make commands, or does limiting it to container-related targets seem like the better approach?

PR #2276, opened 2026-07-01
Remove duplicate legacy controller-gen variable and target

Joeavaikath:
## Summary

Remove the duplicate legacy `CONTROLLER_GEN` variable and `controller-gen` target from the Makefile.

The old `CONTROLLER_GEN = $(shell pwd)/bin/controller-gen` assignment and its `go-install-tool` recipe conflict with the newer branch-aware `?=` pattern that uses `go-install-tool-branch` and symlinks into `bin/$(BRANCH_VERSION)/`. The `=` assignment always overrides `?=`, and the old recipe fails when it encounters the symlink created by the new target.

This aligns oadp-1.4 with oadp-1.6, which already has this cleanup.

## Changes
- Deleted 3 lines: the legacy variable assignment, target declaration, and recipe call
- No changes to the newer branch-aware pattern

## Test plan
- [ ] `make controller-gen` completes successfully and installs into `bin/$(BRANCH_VERSION)/`
- [ ] `make manifests` and `make generate` work correctly

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -400,10 +400,6 @@ build-deploy: THIS_IMAGE=ttl.sh/oadp-operator-$(shell git rev-parse --short HEAD
 build-deploy: ## Build current branch image and deploy controller to the k8s cluster specified in ~/.kube/config.
 	IMG=$(THIS_IMAGE) make docker-build docker-push deploy
 
-CONTROLLER_GEN = $(shell pwd)/bin/controller-gen
-controller-gen: ## Download controller-gen locally if necessary.
-	$(call go-install-tool,$(CONTROLLER_GEN),sigs.k8s.io/controller-tools/cmd/controller-gen@v0.16.5)
-
 
 
 # Codecov OS String for use in download url


Joeavaikath:
For those curious, TIL symlinks and make targets can create a headache:

- Old (line 404-406): CONTROLLER_GEN = $(shell pwd)/bin/controller-gen with a recipe using go-install-tool
  - New (line 135, 725-731): CONTROLLER_GEN ?= $(LOCALBIN)/$(BRANCH_VERSION)/controller-gen with a prerequisite
  target using go-install-tool-branch and a symlink

  The old = on line 404 overrides the new ?= on line 135, so CONTROLLER_GEN resolves to bin/controller-gen
  everywhere instead of bin/oadp-1.4/controller-gen. This causes a cascade of problems when both targets fire
  during make bundle:

  1. The new prerequisite target (line 726) installs the binary to bin/controller-gen (since CONTROLLER_GEN was
  overridden to that path)
  2. The symlink line (line 731) then replaces that binary with a symlink pointing to
  bin/oadp-1.4/controller-gen — but nothing was ever installed there, so the symlink is dangling
  3. The old recipe (line 405) fires next, checks [ -f bin/controller-gen ], finds a dangling symlink (not a
  regular file), and tries go install — which follows the symlink to bin/oadp-1.4/controller-gen, a path inside
  a directory that doesn't exist, and fails with ENOENT

  This doesn't surface in normal development because bin/ persists between runs and the subdirectory gets
  created over time. It only fails in clean-slate environments like the rebasebot container. The fix is to
  delete the old lines 404-406, matching what oadp-1.5 and oadp-1.6 already have.

weshayutin (approved):

kaovilai (approved):
/lgtm

PR #2296, opened 2026-07-13
Added image ref for oadp-1.3

NicholasYancey:
## Why the changes were made

OADP 1.3 is missing oadp-cli-binaries-rhel9 from bundle/image-references. Without this entry, ART cannot track the CLI binaries image for the oadp-1.3 delivery line.

This adds the missing entry so the image is tracked alongside the other OADP 1.3 component images.

After merge, run release metadata check:
```bash
./release-sources oadp-1.3
```
ImgRef for oadp/oadp-cli-binaries-rhel9 should be green after this pr

### DIFF
--- a/bundle/image-references
+++ b/bundle/image-references
@@ -43,3 +43,8 @@ spec:
     from:
       kind: DockerImage
       name: quay.io/konveyor/velero-plugin-for-csi:oadp-1.3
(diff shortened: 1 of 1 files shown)


weshayutin (approved):
/LGTM

Joeavaikath (approved):

PR #2306, opened 2026-07-16
OADP-8445: Add readiness probes to CLI and VMDP

NicholasYancey:
## Why the changes were made

The Telco certification lifecycle-readiness-probe check fails because the oadp-cli-server and oadp-vmdp-server containers (deployed to serve CLI binary downloads via ConsoleCLIDownload) have no ReadinessProbe defined.

This PR adds a ReadinessProbe and LivenessProbe to both containers, and also backfills the probes onto any already-running Deployments on operator upgrade.

Jira link: https://redhat.atlassian.net/browse/OADP-8445

1. Deploy the operator from this branch:
```bash
make deploy
```
2. Confirm both download-server Deployments now have probes defined
```bash
oc -n openshift-adp get deploy openshift-adp-oadp-cli-server \
  -o jsonpath='{.spec.template.spec.containers[0].readinessProbe}{"\n"}'
oc -n openshift-adp get deploy openshift-adp-oadp-vmdp-server \
  -o jsonpath='{.spec.template.spec.containers[0].readinessProbe}{"\n"}'
```
Both should return a populated httpGet probe object (path /, port 8080), not empty.
3. Confirm the pods report Ready:
```bash
oc get pods -n openshift-adp -l app=oadp-cli
oc get pods -n openshift-adp -l app=oadp-vmdp
```
READY column should show 1/1.

## Summary by CodeRabbit

* **New Features**
  * Added HTTP readiness and liveness health checks to newly created CLI and VMDP server deployments.
  * Existing deployments without readiness checks are automatically updated with both health checks.
  * Health checks use the server’s HTTP endpoint and include configured timing settings.

### DIFF
--- a/internal/controller/cli_download_controller.go
+++ b/internal/controller/cli_download_controller.go
@@ -95,6 +95,16 @@ func (c *CLIDownloadSetup) reconcileCLIResources(ctx context.Context, operatorDe
 		c.Log.Info("Created CLI server deployment", "image", cliServerImage)
 	} else if err != nil {
 		return fmt.Errorf("failed to get CLI server deployment: %w", err)
+	} else if len(deployment.Spec.Template.Spec.Containers) > 0 &&
+		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe == nil {
+		// Deployment exists from a version before probes were added; backfill them.
+		desired := buildCLIServerDeployment(c.Namespace, cliServerImage)
+		deployment.Spec.Template.Spec.Containers[0].ReadinessProbe = desired.Spec.Template.Spec.Containers[0].ReadinessProbe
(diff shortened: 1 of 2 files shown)


Joeavaikath (approved):
lgtm! backfilling seems cool, will be dead code after a bit i would guess

kaovilai (approved):
/LGTM. Verified against migtools/oadp-cli: the download image runs download-server, defaults to port 8080, registers / via listBinaries, and the container exposes 8080. The operator probe uses the Deployment’s named http port, which maps to 8080, so the readiness/liveness configuration is consistent with the image. The root handler also checks binary discovery, which makes readiness meaningful for this download-server use case.

PR #2324, opened 2026-07-24
[oadp-1.5] OADP-8485: Added startup probe

NicholasYancey:
## Why the changes were made

Adds a `StartupProbe` to the CLI download server deployment on `oadp-1.5`, giving the container up to 60 seconds to start before liveness checks begin. This prevents premature restarts on slow-starting nodes and is required for Telco/CNF health-check certification (liveness/readiness/startup probes must be present on all deployed containers).

Also backfills the probe onto existing deployments during an operator upgrade, so already-running clusters get it automatically without manual intervention.

Run the unit tests covering the new probe and backfill logic:
```bash
go test ./internal/controller/... -run 'CLIServer|CLIResources' -v
```
On a live cluster, confirm the probe shows up on the deployment:
```bash
oc get deploy -n openshift-adp openshift-adp-oadp-cli-server \
  -o jsonpath='{.spec.template.spec.containers[0].startupProbe}{"\n"}'
```

### DIFF
--- a/internal/controller/cli_download_controller.go
+++ b/internal/controller/cli_download_controller.go
@@ -171,15 +171,19 @@ func (c *CLIDownloadSetup) reconcileCLIResources(ctx context.Context, operatorDe
 		// Deployment exists from a version before probes/service account were added; backfill them.
 		desired := buildCLIServerDeployment(c.Namespace, cliServerImage)
 		needsUpdate := false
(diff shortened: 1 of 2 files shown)


sseago (approved):

kaovilai (approved):
/lgtm

PR #2317, opened 2026-07-22
OADP-8484: Add startup probe to CLI download server

NicholasYancey:
## Why the changes were made

The CLI download server deployment on oadp-1.6/oadp-dev already added Readiness and Liveness probes ([#2306](https://github.com/openshift/oadp-operator/pull/2306)), and oadp-1.3 got those plus a dedicated ServiceAccount via [#2310](https://github.com/openshift/oadp-operator/pull/2310). However, none of these included a StartupProbe.

Without one, the Liveness probe (initialDelaySeconds: 15) can start failing and restarting the container before it's actually finished starting up, especially under slow/loaded conditions — causing avoidable restart loops. A StartupProbe defers the Liveness probe until the container reports healthy, giving it more headroom to start.

This PR is part of the lifecycle probe requirement (readiness/liveness/startup) to self-recover from common failures per Kubernetes best practices.

## How to test the changes made

```bash
go test ./controllers/... -run 'CLIServer' -v
```

1. Deploy the operator from this branch.
2. Check the StartupProbe is set:
   ```bash
   oc get deploy -n openshift-adp openshift-adp-oadp-cli-server \
     -o jsonpath='{.spec.template.spec.containers[0].startupProbe}{"\n"}'
   ```
3. Confirm the pod comes up healthy:
   ```bash
   oc get pods -n openshift-adp -l app=oadp-cli
   ```

PR #2326, opened 2026-07-24
[oadp-1.3] OADP-8484: Added backfill to startup probe

NicholasYancey:
## Why the changes were made
`StartupProbe` was added to the CLI download server in #2317, but the upgrade-backfill logic in `reconcileCLIResources` was never updated to include it as it only checks `ReadinessProbe == nil` before patching an existing deployment. Since `ReadinessProbe` was already set on any deployment from before #2317, that check never fires again, so existing/upgraded clusters never get a `StartupProbe` backfilled onto their CLI server pod.

This PR fixes the backfill to look up the container by name and independently check/patch `StartupProbe` alongside readiness/liveness.

Run the unit tests covering the fix, including a regression test that reproduces the stuck state (probes already set, `StartupProbe` still missing):
```bash
go test ./controllers/... -run 'CLIServer|CLIResources' -v
```
On a live cluster, confirm the probe shows up on an existing deployment after upgrading:
```bash
oc get deploy -n openshift-adp openshift-adp-oadp-cli-server \
  -o jsonpath='{.spec.template.spec.containers[0].startupProbe}{"\n"}'
```

### DIFF
--- a/controllers/cli_download_controller.go
+++ b/controllers/cli_download_controller.go
@@ -167,11 +167,18 @@ func (c *CLIDownloadSetup) reconcileCLIResources(ctx context.Context, operatorDe
 	} else {
 		desired := buildCLIServerDeployment(c.Namespace, cliServerImage)
 		needsUpdate := false
-		if len(deployment.Spec.Template.Spec.Containers) > 0 &&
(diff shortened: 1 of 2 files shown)


sseago (approved):

kaovilai (approved):

kaovilai:
/retest

Flake

Issue #2252, opened 2026-06-23
Add additional scenarios to the kubevirt-datamover e2e tests for incremental  sequence


weshayutin:
### Test 1: Normal backup sequence (full -> incremental)

1. Backup 1 (`cirros-full-1`): Completed as Full. No checkpoint chain in S3, controller correctly set `expected-backup-type: full`. No mismatch.
2. Backup 2 (`cirros-incr-1`): Completed as Incremental. S3 chain valid (length 1), controller set `expected-backup-type: incremental`, VMB reported `Incremental`. No mismatch.

### Test 2: VM restart scenario

1. Stopped and restarted the VM (`oc patch vm ... running:false/true`)
2. Backup 3 (`cirros-mismatch-test`): VM restart did not invalidate the libvirt checkpoint. VMB completed as Incremental. No mismatch triggered.

### Test 3: Delete libvirt checkpoints via virsh

1. Deleted all libvirt checkpoints from the virt-launcher pod:
   ```
   oc exec -n cirros-test <virt-launcher-pod> -- virsh checkpoint-delete <domain> <checkpoint-name>
   ```
2. Confirmed checkpoints were gone (`virsh checkpoint-list` returned empty)
3. S3 index still had a valid chain (3 checkpoints), controller set `expected-backup-type: incremental`
4. Backup 4 (`cirros-mismatch-test-2`): `maxIncrementalBackups: 2` in DPA forced a full backup before the mismatch could occur. Controller set expected type to `full` due to max limit. No mismatch.

### Test 4: Delete checkpoints with maxIncrementalBackups disabled

1. Set `maxIncrementalBackups: 0` (unlimited) in DPA
2. Deleted the new libvirt checkpoint created by backup 4
3. S3 chain valid (chain length 1 after the forced full), controller allowed incremental, set `expected-backup-type: incremental`
4. Backup 5 (`cirros-mismatch-test-3`): **VMB stuck on Initializing indefinitely.** virt-controller repeatedly fails with:
   ```
   "Domain checkpoint not found: Checkpoint '...' for incremental backup of disk 'vda' not found"
   ```
   It reenqueues the VMB every ~17 minutes but never falls back to a full backup. This is [CNV-85377](https://redhat.atlassian.net/browse/CNV-85377).

Issue #187, opened 2021-08-11
Object has been modified errors in OADP operator logs


dymurray:
There is some extra reconcile occurring which is causing 2 controllers to be writing to the Velero CR at the same time.
```
2021-08-10T18:10:56.326Z	ERROR	controller-runtime.manager.controller.velero	Reconciler error	{"reconciler group": "oadp.openshift.io", "reconciler kind": "Velero", "name": "velero-sample", "namespace": "oadp-operator-system", "error": "Operation cannot be fulfilled on veleroes.oadp.openshift.io \"velero-sample\": the object has been modified; please apply your changes to the latest version and try again"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:214

```

eemcmullan:
### Able to reproduce this by creating the default Velero CR:

```
2021-08-23T20:59:52.474Z	ERROR	controller-runtime.manager.controller.velero	Reconciler error	{"reconciler group": "oadp.openshift.io", "reconciler kind": "Velero", "name": "velero-sample", "namespace": "oadp-operator-system", "error": "Operation cannot be fulfilled on deployments.apps \"oadp-velero-sample-1-aws-registry\": the object has been modified; please apply your changes to the latest version and try again"}
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:253
sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2
	/opt/app-root/src/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.2/pkg/internal/controller/controller.go:214
```

```
[...]

Issue #186, opened 2021-08-11
If user enables CSI plugin, ensure feature flag is set


dymurray:
For Velero 1.6, CSI support is still in beta. Because of this, to enable the full functionality of CSI support, we need to set `EnableCSI` on the feature flags for the velero server.

If a user asks for the CSI plugin, we should intelligently set this feature flag automatically to ensure it can be used. For more information see: https://velero.io/blog/csi-integration/

PR #44, opened 2020-09-10
Update readme with usage examples

shubham-pampattiwar:
- Added a new section with usage examples links
- Added missing description in operator CSV 0.1.0

PR #2350, opened 2026-08-04
Add e2e coverage for kubevirt-datamover VM backup/restore: multi-PVC, incremental sequence, and CBT restore (+ max-concurrent-data-movers config)

kaovilai:
## Summary

Adds e2e coverage for three kubevirt-datamover VM backup/restore scenarios: multi-PVC VMs, an incremental backup sequence, and restoring a VM from a full CBT backup. Closes coverage gaps tracked in [migtools/kubevirt-datamover-controller#99](https://github.com/migtools/kubevirt-datamover-controller/issues/99), [openshift/oadp-operator#2252](https://github.com/openshift/oadp-operator/issues/2252), and [migtools/kubevirt-datamover-controller#73](https://github.com/migtools/kubevirt-datamover-controller/issues/73) (phases 0-3; phases 4/5 are out of scope, see Known limitations).

Also folds in #2387: exposes the kubevirt-datamover-controller's `--max-concurrent-data-movers` flag as `spec.configuration.kubevirtDatamover.maxConcurrentDataMovers` on the DPA CRD.

### Multi-PVC VM backup/restore

- New VM template `tests/e2e/sample-applications/virtual-machines/cirros-test/cirros-test-multipvc.yaml`: a CirrOS VM with two disks — the existing rootdisk plus a second blank 100Mi data disk, so no extra image fetch is needed.
- New `ginkgo.Entry("no-application CSI datamover backup and restore, multi-PVC CirrOS VM", ...)`, following the existing "no-application" CirrOS pattern with `BackupRestoreType: lib.CSIDataMover`.
- `vmPvcsBound()` verifies both PVCs exist and are `Bound` after restore.

### Incremental backup sequence

- Automates: full → incremental → a VM restart doesn't invalidate the checkpoint chain → hitting `maxIncrementalBackups` forces a full backup (via the per-VM `kubevirt-datamover.io/max-incremental-backups` annotation override, which takes effect immediately with no controller rollout to wait for).
- New `tests/e2e/lib` helpers: `VirtualMachineBackup` GVR + `GetVMBBackupType` (reads `.status.type`/`.status.checkpointName`), `GetDataUploadForBackup` (reads the controller's `expected-backup-type` annotation), `SetVMAnnotation`.
- A fourth scenario (deleting libvirt checkpoints with `maxIncrementalBackups=0`) depends on an unfixed upstream bug ([CNV-85377](https://redhat.atlassian.net/browse/CNV-85377): virt-controller never falls back to full, the VMB hangs `Initializing` forever) and is scaffolded as a real, compiling `ginkgo.PIt`, ready to flip to `ginkgo.It` once that's fixed.

### Restore from a full kubevirt-datamover CBT backup

Restoring a VM from a kubevirt-datamover backup has to hold the VM inactive until its DataDownload(s) finish: otherwise virt-launcher spawns immediately on the restored VM object, and `WaitForFirstConsumer` binds the target PVC to a scratch volume before the DataDownload can rebind it. That race is closed by two companion PRs, both now merged:

- [migtools/kubevirt-datamover-plugin#44](https://github.com/migtools/kubevirt-datamover-plugin/pull/44): a new VM `RestoreItemActionV2` halts the VM's run strategy at restore time and stashes the original value in an annotation.
- [migtools/kubevirt-datamover-controller#124](https://github.com/migtools/kubevirt-datamover-controller/pull/124): the DataDownload controller restores the VM's stashed run state once every sibling DataDownload for that VM reaches `Completed` — driven by a watch on the phase transition itself rather than a single completion call site, so a controller-restart-mid-restore resume path can't skip it.
- The controller's ServiceAccount additionally needs `update` on `kubevirt.io/virtualmachines` to perform that flip; this PR syncs that verb into `config/kubevirt-datamover-controller_rbac/role.yaml` and the bundle CSV. That sync (via `make update-kubevirt-datamover-manifests`) also picked up a pre-existing `events: create,patch` rule already present in the controller's own generated RBAC — used by `internal/controller/kubevirt_datadownload_controller.go`'s `Eventf` calls, unrelated to this PR's own changes but now correctly reflected here since the manifests are resynced as a whole.
- `RunVirshCommand`'s virt-launcher pod lookup now skips pods that are terminating or not yet `Running`: a VM restart (hit by the incremental-sequence test above) can briefly have both the old and new launcher pod sharing the same domain annotation.
- `ClearStuckVMBFinalizers` (workaround for [kubevirt/kubevirt#18724](https://github.com/kubevirt/kubevirt/issues/18724): a deleted `VirtualMachineBackup` never has its `vmbackup-protection` finalizer released once its backing tracker is gone) now clears only that specific finalizer, rather than the VMB's entire finalizer list.
- Block volumeMode is explicitly asserted, not just incidentally exercised: `GetDataDownloadForRestore` also surfaces the `kubevirt-datamover.io/restore-block-mode` annotation the plugin stamps on the DataDownload, and the test cross-checks it against the restored PVC's actual `spec.volumeMode`.

Both companion PRs above are merged, so the suite exercises the default plugin/controller images — no image override needed.

### Kubevirt-datamover concurrency limit config (folded in from #2387)

- Adds `MaxConcurrentDataMovers *int32` to `KubevirtDatamoverConfig`, alongside the pre-existing `MaxIncrementalBackups` and `StaleDataUploadThreshold` fields.
- When set, the reconciler passes `--max-concurrent-data-movers=<n>` to the kubevirt-datamover-controller deployment, gating concurrent active DataUploads/DataDownloads (DU and DD counted independently) once the corresponding controller-side flag ([migtools/kubevirt-datamover-controller#186](https://github.com/migtools/kubevirt-datamover-controller/pull/186)/[#187](https://github.com/migtools/kubevirt-datamover-controller/pull/187)) is merged upstream.
- Covered by unit tests in `internal/controller/kubevirt_datamover_controller_test.go`; e2e coverage against a real concurrency-limited controller image is planned as follow-up once #186/#187 merge.

## Known limitations

- Two restore scenarios are scaffolded as `ginkgo.PIt` and are explicitly out of scope here: multi-disk restore hardening ([kubevirt-datamover-controller#73](https://github.com/migtools/kubevirt-datamover-controller/issues/73) phase 4) and incremental-chain restore validation (phase 5) — neither is implemented upstream yet.
- The `maxIncrementalBackups=0` checkpoint-delete scenario stays pending on [CNV-85377](https://redhat.atlassian.net/browse/CNV-85377).
- e2e coverage for the `max-concurrent-data-movers` limiter itself (folded in from #2387) is pending [migtools/kubevirt-datamover-controller#186](https://github.com/migtools/kubevirt-datamover-controller/pull/186)/[#187](https://github.com/migtools/kubevirt-datamover-controller/pull/187) merging upstream.

## Testing

All three in-scope VM backup/restore scenarios pass end-to-end on a real OpenShift cluster (AWS, community HCO/KubeVirt) with the companion plugin and controller images deployed:

- multi-PVC backup/restore: pass
- incremental backup sequence (full → incremental → restart → max-limit): pass
- restore from a full CBT backup: pass — controller logs confirm the VM halts at restore and flips back to its stashed run state (`Always`) once its DataDownload reaches `Completed`; Block volumeMode confirmed via both the DataDownload annotation and the restored PVC itself.

The `max-concurrent-data-movers` DPA field is covered by unit tests only so far (real cluster e2e pending upstream controller PRs #186/#187).

`go build`, `go vet`, and `golangci-lint` all pass.

## Test plan

```bash
TEST_VIRT=true GINKGO_ARGS="--focus='multi-PVC CirrOS VM'" make test-e2e
TEST_VIRT=true GINKGO_ARGS="--focus='Kubevirt datamover incremental backup sequence'" make test-e2e
TEST_VIRT=true GINKGO_ARGS="--focus='restore a VM from a full kubevirt-datamover CBT backup'" make test-e2e
```

> [!Note]


tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -73,6 +75,29 @@ func vmPoweredOff(vmnamespace, vmname string) VerificationFunction {
 	})
 }
 
+// vmPvcsBound verifies that each named PVC exists and is Bound in the given
+// namespace. Used to confirm all disks of a multi-PVC VM came back after restore.
+func vmPvcsBound(pvcNamespace string, pvcNames ...string) VerificationFunction {
+	return VerificationFunction(func(ocClient client.Client, namespace string) error {
+		allBound := func() bool {
+			for _, name := range pvcNames {
+				pvc, err := kubernetesClientForSuiteRun.CoreV1().PersistentVolumeClaims(pvcNamespace).Get(context.Background(), name, metav1.GetOptions{})
```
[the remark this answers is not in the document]
kaovilai: Fixed — `vmPvcsBound` now uses a per-Get 10s bounded context instead of `context.Background()`.

> [!Note]

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -73,6 +75,29 @@ func vmPoweredOff(vmnamespace, vmname string) VerificationFunction {
...
+					return false
+				}
+				if pvc.Status.Phase != corev1.ClaimBound {
+					log.Printf("PVC %s/%s is %s, not yet Bound", pvcNamespace, name, pvc.Status.Phase)
+					return false
+				}
+			}
+			return true
+		}
+		gomega.Eventually(allBound, time.Minute*10, time.Second*10).Should(gomega.BeTrue())
```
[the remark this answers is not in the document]
kaovilai: Fixed — added a failure message naming the namespace and PVCs to the `Eventually` assertion.

> [!Note]

tests/e2e/lib/backup.go
```diff
@@ -132,6 +133,32 @@ func CreateBackupWithVolumePolicy(ocClient client.Client, veleroNamespace, backu
...
+	err = ocClient.List(context.Background(), &list, client.InNamespace(veleroNamespace), client.MatchingLabels{velero.BackupNameLabel: backupName})
+	if err != nil {
+		return "", "", fmt.Errorf("failed to list DataUploads for backup %s: %w", backupName, err)
+	}
+	if len(list.Items) != 1 {
+		return "", "", fmt.Errorf("expected exactly 1 DataUpload for backup %s in %s, found %d", backupName, veleroNamespace, len(list.Items))
+	}
+	du := list.Items[0]
+	return du.Name, du.Annotations[annotationExpectedBackupType], nil
+}
```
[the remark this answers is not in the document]
kaovilai: Good catch — fixed. Registered `velerov2alpha1.AddToScheme(Scheme)` in `tests/e2e/lib/scheme.go`; this was a real bug, `GetDataUploadForBackup` would have failed at runtime with an unregistered-kind error.

> [!Note]

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -316,6 +341,13 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
 		dpaCR.VeleroDefaultPlugins = append(dpaCR.VeleroDefaultPlugins, v1alpha1.DefaultPluginKubeVirt)
 		dpaCR.VeleroDefaultPlugins = append(dpaCR.VeleroDefaultPlugins, v1alpha1.DefaultPluginKubeVirtDataMover)
 
+		// TODO: remove once migtools/kubevirt-datamover-plugin#41 merges and the default
+		// plugin image includes its fix — this pins to that PR's build in the meantime.
+		if dpaCR.UnsupportedOverrides == nil {
+			dpaCR.UnsupportedOverrides = map[v1alpha1.UnsupportedImageKey]string{}
+		}
+		dpaCR.UnsupportedOverrides[v1alpha1.KubeVirtDatamoverPluginImageKey] = "quay.io/tkaovila/kubevirt-datamover-plugin:pr-41"
```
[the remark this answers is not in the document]
kaovilai: Acknowledged, but keeping this override as-is for now. It's an explicit, already-documented temporary pin (see the commit that added it, and the `TODO: remove once migtools/kubevirt-datamover-plugin#41 merges` comment right above it) needed specifically to exercise that unmerged PR's fix — removing it would defeat the purpose of this PR until #41 merges. Will remove it in a follow-up once that PR lands.

> [!Note]

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -531,4 +576,123 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
...
+	// deleted or left as a comment, ready to flip to ginkgo.It once that bug is fixed.
+	ginkgo.Describe("Kubevirt datamover incremental backup sequence", ginkgo.Ordered, func() {
+		const (
+			incSeqNamespace = "cirros-test"
+			incSeqVMName    = "cirros-test"
+			incSeqTemplate  = "./sample-applications/virtual-machines/cirros-test/cirros-test-cbt.yaml"
+		)
+
+		var backupCount int
+
```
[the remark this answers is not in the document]
kaovilai: Good catch, this was a real bug — the outer `AfterEach` undeploys the CSI+datamover stack and deletes `cirros-test` after every spec, which would have torn the VM down between the 3 separate `It`s I originally had. Fixed by collapsing them into a single ordered `It` (with `ginkgo.By` steps) that does its own full teardown at the end, and registering the case via `updateLastBRcase`/`prepareBackupAndRestore` so the shared `AfterEach`'s now-redundant cleanup is a harmless no-op instead of acting on stale state from the previous DescribeTable entry.

> [!Note]

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -531,4 +576,123 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
...
+			err = v.WaitForVMReady(incSeqNamespace, incSeqVMName, 5*time.Minute)
+			gomega.Expect(err).ToNot(gomega.HaveOccurred())
+		})
+
+		var _ = ginkgo.AfterAll(func() {
+			err := v.RemoveVm(incSeqNamespace, incSeqVMName, 5*time.Minute)
+			gomega.Expect(err).To(gomega.BeNil())
+			err = lib.DeleteNamespace(v.Clientset, incSeqNamespace)
+			gomega.Expect(err).To(gomega.BeNil())
+		})
```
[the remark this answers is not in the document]
kaovilai: Fixed — added failure messages naming the namespace/VM/backup to the setup, teardown, and backup-sequence assertions (this Describe and in `runSequenceBackup`).

> [!Note]

kaovilai:
Re review [4859513816](https://github.com/openshift/oadp-operator/pull/2350#pullrequestreview-4859513816) and the two remaining nitpicks from [4859450952](https://github.com/openshift/oadp-operator/pull/2350#pullrequestreview-4859450952):

- **`scheme.go:31` (utilruntime.Must)**: fixed in 5321f15 — all 15 scheme registrations now use `utilruntime.Must(...)`, matching `cmd/main.go`'s existing convention, instead of discarding errors via `_ =`. Verified none of them actually panic (ran `TestOADPE2E` without a cluster; it got past scheme init and failed later on an unrelated connectivity assertion).
- **`virt_helpers.go:1386-1394` (unpopulated VMB status)**: fixed in 0d0b911 — `GetVMBBackupType` now checks the `found` bool from `unstructured.NestedString(status, "type")` and returns a distinct error ("has no status.type yet") instead of silently returning `("", "", nil)`.
- **`virt_backup_restore_suite_test.go:664-672` (backupCount depends on earlier specs)**: no longer applies — this finding was against the pre-fix version with 3 separate `ginkgo.It`s (commit 683a49f). After the AfterEach-teardown fix in 7d42f4d, all 4 steps are sequential statements inside a single `ginkgo.It`, not separate specs, so there's no longer a focus/label-filter path that can run step 4 while skipping steps 1-3.

> [!Note]

kaovilai:
> [!Note]

CI red but not real fail — `e2e-test-aws` test step pass clean (52min run, zero failures). Only `gather-must-gather` post-step die, transient DNS blip in CI cluster (`curl: could not resolve host: github.com`), not our code.

kaovilai:
> [!Note]

/test all

kaovilai:
> [!Note]

/test all

kaovilai:
> [!Note]

/test all -- retesting for https://github.com/migtools/kubevirt-datamover-controller/pull/124 rebase (tip ff1c1b8)

kaovilai:
> [!Note]

/test all -- retesting for https://github.com/migtools/kubevirt-datamover-controller/pull/124 real functional fix (tip 3b4b3d5): stale sibling DataDownload correlation bug that could permanently block VM run-state restore, not just a rebase

kaovilai:
> [!Note]

/test all -- retesting for https://github.com/migtools/kubevirt-datamover-controller/pull/124 tip e66317e: 2 more real bugs fixed (missing APIReader fallback in isRestoreAlreadyProvisioned; qemu-img convert missing -n corrupting already-provisioned block-mode restore targets; operation-timeout could mark an already-successful upload Failed). Note: the previous run (tip 3b4b3d5) gave no real signal -- 6/9 e2e jobs failed purely on CI infra (AWS session-token expiry, Route53 zone conflicts/missing-zone, a step-registry script bug), none reached actual test code.

kaovilai:
> [!Note]

/test all -- new commit 307d3697 updates the kubevirt-datamover-controller test image to quay.io/tkaovila/kubevirt-datamover-controller@sha256:6d3d80dc7f0096114f4b1da2379194bca100fc063f0fd26ee97e9c8d2e21856f (commit e66317e, now public and confirmed anonymously pullable)

kaovilai:
> [!Note]

/test all -- commit cf64841b removes the kubevirt-datamover plugin/controller image overrides now that both fixes are merged upstream (migtools/kubevirt-datamover-plugin#41, #44 and migtools/kubevirt-datamover-controller#124). This validates the suite against the default images instead of pinned personal-registry builds.

kaovilai:
/hold 
may i redirect attention to merge https://github.com/openshift/release/pull/83282 first then unhold here when release merges we should have images promoted to unblock https://github.com/openshift/release/pull/83049

tests/e2e/lib/apps.go
```diff
@@ -302,6 +302,36 @@ func IsDeploymentReady(ocClient client.Client, namespace, dName string) wait.Con
 	}
 }
 
+// IsDeploymentRolledOut checks that a Deployment has finished rolling out — unlike
+// IsDeploymentReady, which only checks AvailableReplicas == Replicas and can false-positive
+// against stale pre-rollout pods, this also requires the controller to have observed the
+// latest generation and updated all replicas to it.
+func IsDeploymentRolledOut(ocClient client.Client, namespace, dName string) wait.ConditionFunc {
```
shubham-pampattiwar: Two exported helpers here have no callers in this PR: `IsDeploymentRolledOut` (this func) and `ChecksumUploadPodQcow2Flattened` in virt_helpers.go (its own comment says it's not called). `tests/e2e/lib` is shared code, so unused exports tend to rot. Can we drop both and add them back in the PR that actually calls them? If `IsDeploymentRolledOut` has a caller I missed, point me at it.

config/kubevirt-datamover-controller_rbac/role.yaml
```diff
@@ -4,6 +4,13 @@ kind: ClusterRole
 metadata:
   name: manager-role
 rules:
+- apiGroups:
+  - ""
+  resources:
+  - events
```
shubham-pampattiwar, line 10: This `events: create,patch` rule isn't mentioned in the description, unlike the `update` on virtualmachines. Two things:

1. Were role.yaml and the CSV regenerated via `make update-kubevirt-datamover-manifests`, or hand-edited? This file syncs from the controller repo, so hand edits drift on the next sync and can diverge from the CSV.
2. Please add a line to the description explaining the events rule so the CSV grant is traceable.

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -502,33 +546,964 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
...
+			// fullBackupPayloadOffsetMiB/SizeMiB are this It's own payload region,
+			// distinct from payloadA/payloadB above (different VM instance per It via
+			// BeforeEach, so no actual collision risk -- kept apart anyway for clarity
+			// when reading logs from both Its side by side).
+			const (
+				fullBackupPayloadOffsetMiB = 440
+				fullBackupPayloadSizeMiB   = 32
+			)
+
+			ginkgo.It("restore a VM from a full kubevirt-datamover CBT backup", ginkgo.Label("virt", "kdm"), func() {
```
shubham-pampattiwar, line 1242: This `It` runs two full restore cycles plus the backup in one spec, with timeouts summing past an hour. I get why it can't split: the AfterEach recreates the DPA, which gets a new random BSL prefix, so a second `It` can't reuse this backup. Two small asks:

- Move that "can't split" note to a comment above the `It`, so the next person sees it before trying to break it apart. It currently sits mid-spec near the rejected-restore block.
- If the rejection-path half can ever run off its own small backup, that's the natural spec to break out later.

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -502,33 +546,964 @@ var _ = ginkgo.Describe("VM backup and restore tests", ginkgo.Ordered, func() {
...
+				}, 3*time.Minute, time.Second*5).Should(gomega.Succeed(), "restored VM %s/%s never appeared", alpineNamespace, alpineVMName)
+				gomega.Consistently(func() string {
+					status, statusErr := v.GetVmStatus(alpineNamespace, alpineVMName)
+					if statusErr != nil {
+						log.Printf("transient error reading VM %s/%s status during Consistently poll: %v", alpineNamespace, alpineVMName, statusErr)
+						return ""
+					}
+					return status
+				}, time.Minute, time.Second*10).ShouldNot(gomega.Equal("Running"),
+					"VM %s/%s unexpectedly reached Running status despite its DataDownload failing", alpineNamespace, alpineVMName)
```
shubham-pampattiwar: A persistent `GetVmStatus` error returns "" here, which passes `ShouldNot(Equal("Running"))` for the full minute. So a broken API makes the "VM stayed halted" check pass exactly when we can't see the VM at all. The eager-wait polls swallow the same error harmlessly (they time out loudly), but this negative check needs to tell "confirmed not Running" apart from "couldn't read status". Fail on a run of consecutive errors, or require at least one valid non-Running read before passing.

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -73,12 +78,56 @@ func vmPoweredOff(vmnamespace, vmname string) VerificationFunction {
...
 }
 
+// vmPvcsBound verifies that each named PVC exists and is Bound in the given
+// namespace. Used to confirm all disks of a multi-PVC VM came back after restore.
+func vmPvcsBound(pvcNamespace string, pvcNames ...string) VerificationFunction {
+	return VerificationFunction(func(ocClient client.Client, namespace string) error {
+		allBound := func() bool {
+			for _, name := range pvcNames {
+				ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
+				pvc, err := kubernetesClientForSuiteRun.CoreV1().PersistentVolumeClaims(pvcNamespace).Get(ctx, name, metav1.GetOptions{})
```
shubham-pampattiwar: `vmPvcsBound` ignores its `ocClient` and `namespace` params and uses the `kubernetesClientForSuiteRun` global plus `pvcNamespace` instead. Works, but the unused params are misleading. Either use `ocClient`, or add a one-line note on why the global clientset is used.

tests/e2e/lib/virt_helpers.go
```diff
@@ -1359,3 +1431,666 @@ func (v *VirtOperator) CheckVMBackupTrackerExists(namespace string) (bool, error
...
+			return pod, nil
+		}
+	}
+	return nil, fmt.Errorf("no virt-launcher pod found for VM %s/%s", namespace, vmName)
+}
+
+// RunVirshCommand execs `virsh <args...>` inside vmName's virt-launcher pod's compute
+// container via ExecuteCommandInPodsSh. kubeConfig is the suite's *rest.Config (VirtOperator
+// itself only carries a *kubernetes.Clientset, not a rest.Config, so it's passed in here).
+func (v *VirtOperator) RunVirshCommand(kubeConfig *rest.Config, namespace, vmName string, args ...string) (string, error) {
```
shubham-pampattiwar, line 1550: On the CodeRabbit no-sensitive-data flag: fine to leave. The only caller is the CNV-85377 `PIt`, which doesn't run, and checkpoint output isn't sensitive. The live guest-exec path logs base64 dd-of-urandom, also harmless. Noting it so the thread closes with a reason.

kaovilai:
thank you for your review kind sir.

build/ci-Dockerfile
```diff
@@ -10,26 +10,42 @@ RUN chmod +x tests/e2e/scripts/analyze_failures.sh
...
 # Uses oadp-dev branch to match OADP operator development
-RUN git clone --depth 1 --branch oadp-dev \
-    https://github.com/openshift/velero.git \
-    /go/src/github.com/openshift/velero
+# Retried like the curl fetches above -- same category of transient
+# network/DNS blip, just via git instead of curl (no built-in --retry flag).
+# rm -rf before each attempt: a clone that fails partway through can leave a
+# non-empty target directory behind, which would make the next attempt fail
+# with "destination path already exists" instead of actually retrying.
+RUN for i in 1 2 3; do \
```
shubham-pampattiwar, line 47: This retry loop exits 0 even when all three clones fail. `git clone ... && break || sleep 5`: on failure the `&& break` is skipped and `|| sleep 5` returns 0, so after three misses the loop's status is the last `sleep`, and the RUN succeeds with no velero checkout. The build then continues and only degrades failure-analysis later, instead of failing here.

Every other fetch in this file uses `curl --retry`, which does fail the build after exhausting retries. Worth making this one consistent, e.g. assert the result after the loop:

    RUN for i in 1 2 3; do \
          rm -rf /go/src/github.com/openshift/velero && \
          git clone --depth 1 --branch oadp-dev \
            https://github.com/openshift/velero.git \
            /go/src/github.com/openshift/velero && break || sleep 5; \
        done; \
        test -d /go/src/github.com/openshift/velero/.git

tests/e2e/lib/virt_helpers.go
```diff
@@ -1359,3 +1430,505 @@ func (v *VirtOperator) CheckVMBackupTrackerExists(namespace string) (bool, error
...
+// found" immediately after DataDownload reached Completed, because the halt
+// really does mean no launcher pod exists yet, not merely a paused one. The
+// helper pod is deleted (and its deletion awaited) before returning, so the PVC
+// is free again by the time the caller lets the VM resume and its virt-launcher
+// pod tries to attach the same PVC -- PVC access mode here is RWO, so the two
+// can never overlap. Shared plumbing for ChecksumPVCBlockDeviceRegion so the pod
+// lifecycle (creation, Running wait, teardown-before-return) exists in exactly
+// one place.
+func (v *VirtOperator) runInPVCBlockDeviceHelperPod(kubeConfig *rest.Config, namespace, pvcName, command string) (stdout string, err error) {
+	podName := "checksum-helper-" + pvcName
```
shubham-pampattiwar: `checksum-helper-` + pvcName can exceed the 63-char DNS label limit for longer PVC names and fail pod creation with an opaque error. Fine for the current short names, but worth truncating or hashing the suffix so a future longer-named PVC scenario doesn't trip on it.

kaovilai: This is our own e2e no customer input but will harden

kaovilai:
> [!Note]

/test 5.0-e2e-test-kubevirt-aws

kaovilai:
> [!Note]

Update on the `max-concurrent-data-movers` concurrency-limiter e2e coverage (`TEST_KDM_CONCURRENCY_IMAGE`-gated, skips in normal CI since it needs an unreleased controller build):

Validated live against a real cluster with a combined build of [migtools/kubevirt-datamover-controller#186](https://github.com/migtools/kubevirt-datamover-controller/pull/186) + [#187](https://github.com/migtools/kubevirt-datamover-controller/pull/187) — 2 VMs backed up concurrently with `--max-concurrent-data-movers=1`.

Along the way this surfaced a real deadlock in #187's original implementation (two `DataUpload`s could mutually defer forever with no priority ordering to break the tie — same failure mode #186 had already solved on the DD side via ranked ordering). Reported it, #187 was fixed to mirror #186's approach, and a retest confirms it now works correctly: at most one `DataUpload` reaches `InProgress` at a time, the other stays queued, both complete.

`5.0-e2e-test-kubevirt-aws` is green on this PR's current head (`ffcab7db`) with the default (non-override) image, as expected. The concurrency-limiter spec itself will start running in regular CI once #186/#187 merge and a release image carries the flag — no action needed on this PR for that.

kaovilai:
> [!Note]

Update: `5.0-e2e-test-kubevirt-aws` is now green running the `max-concurrent-data-movers` spec as a real (non-skipped) `It` against the hardcoded override image (`quay.io/tkaovila/kubevirt-datamover-controller:tmp-combine-186-187`) — real Prow proof, not just a local run, matching the earlier local-cluster result. Head is now `fe54f4dc`.

kaovilai:
> [!Note]

/test 5.0-e2e-test-kubevirt-aws

kaovilai:
> [!Note]

Both failures are infra flakes, unrelated to this PR:
- `4.23-e2e-test-aws`: the actual `e2e-test-aws-e2e` step succeeded; only the post-test `gather-aws-console` step failed ("Unable to connect to the server: context deadline exceeded").
- `5.1-e2e-test-cli-aws`: failed during operator install (pre-step), on a DNS lookup timeout to the cluster API server — before any test code ran.

/test 4.23-e2e-test-aws
/test 5.1-e2e-test-cli-aws

kaovilai:
> [!Note]

Another infra flake on `4.23-e2e-test-aws` — this time a registry image-pull timeout during cluster install (`ipi-conf-aws`), before any test code ran:
```
error: unable to read image registry.build01.ci.openshift.org/... : Client.Timeout exceeded while awaiting headers
```

/test 4.23-e2e-test-aws

tests/e2e/virt_backup_restore_suite_test.go
```diff
@@ -165,85 +217,87 @@ func runVmBackupAndRestore(brCase VmBackupRestoreCase, updateLastBRcase func(brC
...
 	vmbtSeen := false
 	vmbtCheckDone := make(chan struct{})
 	go func() {
 		defer close(vmbtCheckDone)
 		for i := 0; i < 60; i++ {
-			found, checkErr := v.CheckVMBackupTrackerExists(brCase.Namespace)
+			found, checkErr := v.CheckVMBackupTrackerExists(vmNamespace)
 			if checkErr == nil && found {
-				log.Printf("VirtualMachineBackupTracker observed in %s — VEP-25 incremental backup confirmed", brCase.Namespace)
+				log.Printf("VirtualMachineBackupTracker observed in %s — VEP-25 incremental backup confirmed", vmNamespace)
```
kaovilai, line 270: [VirtualMachineBackupTracker observed in alpine-guestagent](https://gcsweb-ci.apps.ci.l2s4.p1.openshiftapps.com/gcs/test-platform-results/pr-logs/pull/openshift_oadp-operator/2350/pull-ci-openshift-oadp-operator-oadp-dev-5.0-e2e-test-kubevirt-aws/2089906961901424640/artifacts/e2e-test-kubevirt-aws/e2e/build-log.txt#:~:text=2026/08/19%2005%3A00%3A50%20VirtualMachineBackupTracker%20observed%20in%20alpine%2Dguestagent%20%E2%80%94%20VEP%2D25%20incremental%20backup%20confirmed) where `v *lib.VirtOperator` where lib is `tests/e2e/lib/virt_helpers.go` where 
```
type VirtOperator struct {
	Client         client.Client
	Clientset      *kubernetes.Clientset
	Dynamic        dynamic.Interface
	Namespace      string
	Csv            string
	Version        *version.Version
	Upstream       bool
	CommunityIndex string // HCO index image tag (e.g. "1.17.1"); empty means no custom catalog
}
```

has method

https://github.com/openshift/oadp-operator/blob/5a0a023fc510d33552c4ff74558e5628cd62437a/tests/e2e/lib/virt_helpers.go#L1427-L1433

shubham-pampattiwar (approved):
/lgtm

Joeavaikath (approved):

PR #2100, opened 2026-02-23
oadp-1.4.9 is the current 1.4 version

weshayutin:
## Why the changes were made

Ensure sure bundle meta is current

## How to test the changes made

the usual.. make deploy perhaps and downstream

PR #2368, opened 2026-08-10
[oadp-1.4] Make controller-gen/kustomize/golangci-lint/envtest tool-binary caching reliable

kaovilai:
Cherry-pick of #2367 to `oadp-1.4` — same fix, plus this branch's own pre-existing envtest arch-check bug (see below).

## Summary

`go-install-tool-branch` only installs a build tool when its binary is **missing**, never verifying the pinned version against what's already on disk. Once a binary lands at `bin/<branch>/<tool>`, it's reused forever — even after `CONTROLLER_TOOLS_VERSION`/`KUSTOMIZE_VERSION`/`GOLANGCI_LINT_VERSION` change — because `bin/` is gitignored. `kustomize`/`controller-gen` targets also weren't fully `.PHONY`, so Make's own mtime-based staleness check could skip the recipe before any check ran.

None of the four tools (`controller-gen`/`kustomize`/`golangci-lint`/`setup-envtest`, the last folded in from #2152) were checked for architecture compatibility either. This repo's own Makefile documents a `docker run --platform linux/amd64 -v $PWD:$PWD ...` workflow to reproduce Prow's CI environment — that bind-mounts the host directory into a forced-arch container, so any `go install` in there writes a foreign-arch binary straight onto the host's `bin/` tree (same path, not a copy).

This branch already had a bespoke `check-envtest-arch` target (from #2100) attempting to guard against exactly this for `setup-envtest`, but it checked `$(ENVTEST) --help`'s exit code and treated any nonzero result as "wrong architecture" — `setup-envtest`'s own `--help` exits `2` by its own convention even on a perfectly healthy binary, so it was unconditionally reinstalling `setup-envtest` on every single invocation.

## Design

`go-install-tool-versioned` reads a binary's **embedded build info** via `go version -m` — the module version and `GOOS`/`GOARCH` it was built with — instead of executing the binary or trusting a sidecar marker file:

- A binary's own `--version`/`--help` output isn't a reliable version or health signal: it can depend on ldflags a tool's own release process sets (which `go install` doesn't set — e.g. `kustomize` can report `(devel)` or an unexpanded `$Format:%H$` placeholder instead of its real version), and some tools exit nonzero on `--version` even when perfectly healthy (`kustomize` exits `1`, `setup-envtest`'s `--help` exits `2`). A bare probe command that exits nonzero is also dangerous under this Makefile's `.SHELLFLAGS = -ec` (enables `set -e`, honored by GNU Make 3.82+): it aborts the whole recipe unless wrapped in an `if`/`||` guard.
- Executing the binary to test compatibility can't detect a wrong-arch binary at all inside a container with `qemu-user-static`/`binfmt_misc` registered (standard for multi-arch CI/build images) — the foreign-arch binary just runs under emulation and returns its own exit code, not an exec-format-error.

Reading embedded build info sidesteps both: no execution means no exit-code heuristic to get wrong and no qemu blind spot. `setup-envtest`'s arch check is folded into this same shared macro rather than kept as a separate mechanism.

## Known limitations

- **`ENVTESTPATH`'s arch selection has a separate, pre-existing bug — tracked as #2377, not fixed here.** Its `ifeq`-based amd64 fallback is decided at Makefile-parse time and picks the amd64-forced variant on any cold `bin/`, regardless of actual host arch — confirmed via direct repro (see #2377). It's invisible on amd64 CI (no CI/release-branch risk) and only affects arm64 local dev with a cold checkout, so it's out of scope here rather than a reason to expand this PR's diff a third time.
- `golangci-lint@v1.54.2` (this branch's pin) doesn't build against a modern Go 1.26 toolchain — a pre-existing, unrelated `x/tools` incompatibility; CI pins its own Go version and isn't affected. Bumping to the latest `v1.x` (`v1.64.8`, avoiding `v2`'s config/default-linter overhaul) does fix the build, but its updated `staticcheck`/`ineffassign`/`gosimple` linters immediately surface ~20 pre-existing findings across unrelated files. Left the pin as-is — real scope growth this PR shouldn't absorb.
- Unversioned tool symlinks (`bin/<tool>` → last-branch-wins, pre-existing behavior), a lower-priority observation, also worth a follow-up.

## Testing

Verified against **GNU Make 4.4.1** specifically (installed via Homebrew), not just the macOS-default Make 3.81 — the latter silently ignores `.SHELLFLAGS`, so it can't exercise `set -e` recipe behavior at all.

- `rm -rf bin/oadp-1.4` / `rm -f bin/setup-envtest`, then `gmake controller-gen` / `gmake kustomize` / `gmake envtest` individually — each installs fresh.
- Re-running each immediately after → "already installed", no reinstall, no network call.
- Simulated a wrong-arch binary (cross-compiled `linux/amd64`) for controller-gen, kustomize, and envtest → each correctly detected via `go version -m`'s `GOARCH`/`GOOS` fields, removed, reinstalled with a valid native binary.
- `gmake generate manifests bundle` → zero diff against a clean checkout.
- `gmake test` → exit 0, all packages pass, `Go code is formatted`, `api is up to date`, `bundle is up to date`.

> [!Note]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -712,15 +712,25 @@ rm -rf $$TMP_DIR ;\
 }
 endef
 
-# go-install-tool-versioned installs $2 to branch-specific path $1, but only if $1 is missing
-# or $1.version doesn't match the pinned version $3. Uses a sidecar marker file instead of
-# introspecting the binary's own --version output, because that output is unreliable for some
-# tools when installed via `go install` (e.g. kustomize reports "(devel)" or an unexpanded
-# `$$Format:%H$$` placeholder instead of its real version, depending on build-time factors).
+# go-install-tool-versioned installs $2 to branch-specific path $1, but only if $1 is missing,
+# $1.version doesn't match the pinned version $3, or $1 exists but cannot execute on this
+# platform (checked via exit code 126, POSIX "found but cannot execute" / exec format error —
+# this can happen when a containerized build with a different GOARCH bind-mounts the host's
+# bin/ directory, e.g. `docker run --platform linux/amd64 -v $$PWD:$$PWD ... make manifests`).
+# Uses a sidecar marker file for the version check instead of introspecting the binary's own
+# --version output, because that output is unreliable for some tools when installed via
+# `go install` (e.g. kustomize reports "(devel)" or an unexpanded `$$Format:%H$$` placeholder
+# instead of its real version, depending on build-time factors).
 define go-install-tool-versioned
-@if [ -f $(1) ] && [ -f $(1).version ] && [ "$$(cat $(1).version)" = "$(3)" ]; then \
+@ARCH_OK=1 ;\
+if [ -f $(1) ]; then \
+	$(1) --version >/dev/null 2>&1 ;\
+	if [ $$? -eq 126 ]; then ARCH_OK=0 ; fi ;\
+fi ;\
+if [ "$$ARCH_OK" = "1" ] && [ -f $(1) ] && [ -f $(1).version ] && [ "$$(cat $(1).version)" = "$(3)" ]; then \
 	echo "$(notdir $(1)) $(3) is already installed" ;\
 else \
+	if [ "$$ARCH_OK" = "0" ]; then echo "$(notdir $(1)) exists but cannot execute on this platform, removing and re-downloading" ; fi ;\
 	set -e ;\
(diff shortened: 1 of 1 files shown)


Joeavaikath (dismissed):

sseago (approved):

shubham-pampattiwar (approved):
/lgtm

PR #2371, opened 2026-08-11
docs: AllNamespaces install mode design proposal

Joeavaikath:
## Why the changes were made

OADP currently only supports `OwnNamespace` install mode via OLM. With OLMv1 GA in OCP 4.18 shipping AllNamespaces-only support — and OwnNamespace remaining Tech Preview with no GA promotion planned ([OPRUN-4514](https://issues.redhat.com/browse/OPRUN-4514), [operator-controller#2568](https://github.com/operator-framework/operator-controller/pull/2568)) — supporting AllNamespaces is now the strategically correct direction for OLMv1 readiness.

This PR adds a design document for enabling `AllNamespaces` install mode alongside `OwnNamespace` in a single CSV, without changing the operator's runtime behavior. The operator continues watching only the namespace it is deployed in.

The design explores three changes required to make this work:

1. **Enable `AllNamespaces: true` in `installModes`** — a safe superset change that doesn't affect existing OwnNamespace installations.
2. **Add `OPERATOR_NAMESPACE` env var + Go fallback** — OLM sets `olm.targetNamespaces=""` in AllNamespaces mode by design ([OLM OperatorGroup spec](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/operatorgroups.md)), and `operator-sdk generate bundle` unconditionally rewrites `WATCH_NAMESPACE` to source from this annotation ([SDK source](https://github.com/operator-framework/operator-sdk/blob/299157a814e674f8f40d1b91ee77d64a90e850de/internal/generate/clusterserviceversion/clusterserviceversion_updaters.go)). The solution is a new `OPERATOR_NAMESPACE` env var (sourced from `metadata.namespace`, which the SDK does not rewrite) and a Go fallback in `getWatchNamespace()` when `WATCH_NAMESPACE` is empty.
3. **Add `permissions` entries for missing ServiceAccounts** — OLM requires namespace-scoped Role entries for every SA declared in `clusterPermissions` when running in AllNamespaces mode.

The design was validated on OCP 4.22.0-ec.3 with both install modes using the same bundle. See the [full test log](https://hackmd.io/MVRrs4zTTHiwGcxfRUwhBA).

This PR is a design document only — no code changes are included. Review the design at `docs/design/allnamespaces-install-mode_design.md`.

Key things to evaluate:

- **Background**: OLMv0 vs OLMv1 model, why AllNamespaces is the right direction, sourced claims with Jira/GitHub links
- **Design correctness**: The `OPERATOR_NAMESPACE` + Go fallback approach vs the alternatives considered (post-gen sed patch, CSV-level source swap)
- **Implementation plan**: Four phases from CSV changes through migration docs
- **Validation**: Both modes confirmed working on OCP 4.22.0-ec.3 — CSV `Succeeded`, correct env vars at runtime, PSA labeling applied

*Posted via [Claude Code](https://claude.ai/code)*

### DIFF
--- a/docs/design/allnamespaces-install-mode_design.md
+++ b/docs/design/allnamespaces-install-mode_design.md
@@ -102,7 +102,7 @@ No Go code changes are needed. The operator binary is identical for both CSVs.
 
 #### Catalog Structure
 
-```
+```text
 Catalog: oadp-operator-catalog
 └── Package: oadp-operator
     ├── Channel: dev                    ← OwnNamespace CSV
@@ -133,6 +133,7 @@ Release branches follow the same pattern: `stable-1.7` (OwnNamespace) and `stabl
 | **Catalog build** | Extend `catalog-build` to produce a single catalog with two channels: existing channel (OwnNamespace bundle) and new `-allnamespaces` channel (AllNamespaces bundle). Adds a second `opm render` + `olm.channel` entry to the FBC output. |
 | **CSV naming** | AllNamespaces CSV uses a distinct version suffix: `oadp-operator.v99.0.0-allns` vs `oadp-operator.v99.0.0` |
 | **Channel naming** | Convention: `<existing-channel>-allnamespaces` (e.g., `dev-allnamespaces`, `stable-1.7-allnamespaces`) |
+| **Cross-channel update graph** | The AllNamespaces channel's CSV must include an `olm.skipRange` that covers the OwnNamespace CSV version (e.g., `>=0.0.0 <99.0.0`), so OLM can resolve a valid update path when a customer switches channels. Without this, the channel switch fails silently — OLM cannot find an upgrade edge from the installed CSV to the new channel's head. |
 
 #### Validation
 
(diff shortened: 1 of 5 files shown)


kaovilai:
Perhaps call it OADP 2.0 since there is significant operational mode update.

Joeavaikath:
> Perhaps call it OADP 2.0 since there is significant operational mode update.

I'm not sure if it is that significant? But in terms of having customers move to this new InstallMode which is a breaking change....maybe calling it 2.0 isn't a bad idea. Not sure who I should talk to about this. Will add it to the design doc.

@kaovilai

kaovilai:
> OLMv1 alignment: AllNamespaces was required at OLMv1 GA (OCP 4.18). This change positions OADP for OLMv1 readiness.

not quite..

sseago (dismissed):

docs/design/allnamespaces-install-mode_design.md
```diff
@@ -0,0 +1,294 @@
...
+
+## Goals
+
+- Enable `AllNamespaces` install mode alongside `OwnNamespace` in the same CSV.
+- Keep runtime behavior identical: the operator watches only the namespace it is deployed in.
+- Maintain full backward compatibility for existing `OwnNamespace` installations.
+
+## Non Goals
+
+- Actually watching all namespaces or supporting multi-tenant Velero (one DPA per namespace).
```
kaovilai: supporting multi-tenant Velero (one DPA per namespace).

This should be listed as future considerations to avoid breakage of any existing install.. which would be revisited when OLMv0 is no longer avail for use.. ie OCP6.

Joeavaikath: Change made

PR #2385, opened 2026-08-13
OADP-8548: PodResources unset fields should be output as "0" to fix parsing errors results in ignored resource settings

msfrucht:
## Why the changes were made

https://redhat.atlassian.net/browse/OADP-8548

Velero datamover and maintenance Pod config the default values of "" causes parser errors. The default value has to be "0" for unbounded. If PodResources is only partially set, despite all values are optional, the result is Velero ignores them all.

As a result, a Pod with `memoryLimit` or `ephemeralStorageLimit` will not be evicted when violated unless all values: `cpuRequest`, `cpuLimit`, `memoryRequest`, `memoryLimit`, `ephemeralStorageRequest`, `ephemeralStorageLimit` are set.

This issue also applies to OADP-1.5/OADP-1.4 without the `ephemeralStorageRequest` and `ephemeralStorageLimit` fields.

This includes @kaovilai requested revisions from the previous PR targeting `oadp-1.6` branch.

Edit: Code changed so that the output can be used across OADP-1.4/OADP-1.5/OADP-1.6 using reflection to not be tied to individual fields of PodResources, only field type string.

Changes since the original PR:

- internal/controller/defaults.go - Revised to return a new objects instead mutate existing input parameter. Function name changed to match new behavior.
- internal/controller/defaults_test.go - Added to test the additional logic changes. nil returns nil, if  values passed in, return new object with unset fields mutated to "0"

Added a check that DPA Spec is unchanged despite PodResources serialization may not match the DPA Spec object due to unset values -> "0" in the following modules:
- internal/controller/nodeagent_test.go 
- internal/controller/repository_maintenance_test.go 

```
time="2026-08-07T17:02:41Z" level=info msg="Setting log-level to INFO"
time="2026-08-07T17:02:41Z" level=info msg="Starting Velero node-agent server v1.18.2-rc.2-OADP (-)" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:110"
time="2026-08-07T17:02:41Z" level=info msg="Concurrency configs are not found, use the default number 1" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:688"
time="2026-08-07T17:02:41Z" level=info msg="Starting controllers" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:293"
time="2026-08-07T17:02:41Z" level=info msg="Starting metric server for node agent at address [:8085]" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:279"
time="2026-08-07T17:02:41Z" level=warning msg="Pod resource requirements are invalid, ignore" error="couldn't parse CPU request \"\": quantities must match the regular expression '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'" error.file="/workspace/pkg/util/kube/resource_requirements.go:58" error.function=github.com/vmware-tanzu/velero/pkg/util/kube.ParseResourceRequirements logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:345"
time="2026-08-07T17:02:41Z" level=info msg="Using backup repo config map[kopia:{\"fullMaintenanceInterval\":\"eagerGC\"}]" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:384"
time="2026-08-07T17:02:41Z" level=info msg="Controllers starting..." logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:516"
```

Modified unit tests to account for the change that the values must be "0" on output if unset.

Built the container and tested the following DPA in OCP 4.22.

```yaml
metadata:
  name: velero-sample
  namespace: openshift-adp
spec:
  backupImages: false
  configuration:
    nodeAgent:
      dataMoverPrepareTimeout: 10m
      enable: true
      fullMaintenanceInterval: eagerGC
      podResources:
        memoryLimit: 100Mi
      uploaderType: kopia
    velero:
      defaultPlugins:
        - openshift
        - aws
        - kubevirt
        - hypershift
      defaultSnapshotMoveData: true
      disableFsBackup: false
      featureFlags:
        - EnableCSI
      noDefaultBackupLocation: true
  logFormat: text
```

Node-agent logs

```
time="2026-08-12T23:28:32Z" level=info msg="Setting log-level to INFO"
time="2026-08-12T23:28:32Z" level=info msg="Starting Velero node-agent server v1.18.2-rc.2-OADP (-)" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:110"
time="2026-08-12T23:28:32Z" level=info msg="Concurrency configs are not found, use the default number 1" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:688"
time="2026-08-12T23:28:32Z" level=info msg="Starting controllers" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:293"
time="2026-08-12T23:28:32Z" level=info msg="Using customized pod resource requirements &{0 0 0 100Mi 0 0}" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:348"
time="2026-08-12T23:28:32Z" level=info msg="Using backup repo config map[kopia:{\"fullMaintenanceInterval\":\"eagerGC\"}]" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:384"
time="2026-08-12T23:28:32Z" level=info msg="Starting metric server for node agent at address [:8085]" logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:279"
time="2026-08-12T23:28:32Z" level=info msg="Controllers starting..." logSource="/workspace/pkg/cmd/cli/nodeagent/server.go:516"
```

Node-agent configmap output:

```yaml
kind: ConfigMap
apiVersion: v1
metadata:
  name: node-agent-velero-sample
  namespace: openshift-adp
  uid: 69ff5cec-ec44-4fa1-8e1f-50ec84b36591
  resourceVersion: '33695705'
  creationTimestamp: '2026-08-07T16:18:49Z'
  labels:
    app.kubernetes.io/component: node-agent-config
    app.kubernetes.io/instance: velero-sample
    openshift.io/oadp: 'True'
  ownerReferences:
      name: velero-sample
      uid: 1f263902-6e64-4a75-93f8-cd8dcaa12efd
      controller: true
      blockOwnerDeletion: true
data:
  node-agent-config: '{"podResources":{"cpuRequest":"0","cpuLimit":"0","memoryRequest":"0","memoryLimit":"100Mi","ephemeralStorageRequest":"0","ephemeralStorageLimit":"0"},"privilegedFsBackup":true}'
```

.metadata.managedFields from output shown have been removed because they are annoying to read.

## Bug Fixes

- Unspecified CPU, memory, and ephemeral-storage requests and limits now default to zero in node-agent and repository maintenance configurations.
- Configuration output no longer contains empty resource fields, improving consistency and predictability.
- Existing resource values remain unchanged while only missing values receive defaults.
- Configuration updates no longer modify the original deployment specification.


internal/controller/defaults.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+	if pr.EphemeralStorageRequest == "" {
+		prWithUnboundedDefaults.EphemeralStorageRequest = unbounded
+	} else {
+		prWithUnboundedDefaults.EphemeralStorageRequest = pr.EphemeralStorageRequest
+	}
+	if pr.EphemeralStorageLimit == "" {
+		prWithUnboundedDefaults.EphemeralStorageLimit = unbounded
+	} else {
+		prWithUnboundedDefaults.EphemeralStorageLimit = pr.EphemeralStorageLimit
```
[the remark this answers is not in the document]
msfrucht: Dependencies were updated. Re-run.

kaovilai (commented):
/approve

kaovilai (dismissed):
/approve

kaovilai:
non blocking comments:

Preserve future PodResources fields when copying. The helper constructs a new empty struct and manually copies every current field. If Velero adds another field later, this function will silently discard it until updated—particularly relevant because ephemeral-storage fields were themselves recently added. Start with result := *pr, then overwrite only empty known quantities with "0". See [defaults.go lines 25–55](https://github.com/openshift/oadp-operator/blob/a43e8d6871dcd1ae980478d188d9583e02a2e032/internal/controller/defaults.go#L25-L55).

Add an integration regression case matching the reported failure exactly. The helper unit test covers partial fields, but the node-agent ConfigMap test uses all CPU/memory fields, while the production example uses only memoryLimit. Add a node-agent test with only memoryLimit: "100Mi" and, ideally, another with only ephemeralStorageLimit. Assert that all other serialized quantities are "0". This would catch accidental removal of the normalization call, not merely defects within the helper.

kaovilai:
im guessing https://github.com/openshift/release/pull/83299 problem here.. so lets /retest-required

kaovilai:
or ignore.. I think 4.23 might be optional.

msfrucht:
> non blocking comments:
> 
> Preserve future PodResources fields when copying. The helper constructs a new empty struct and manually copies every current field. If Velero adds another field later, this function will silently discard it until updated—particularly relevant because ephemeral-storage fields were themselves recently added. Start with result := *pr, then overwrite only empty known quantities with "0". See [defaults.go lines 25–55](https://github.com/openshift/oadp-operator/blob/a43e8d6871dcd1ae980478d188d9583e02a2e032/internal/controller/defaults.go#L25-L55).
> 
> Add an integration regression case matching the reported failure exactly. The helper unit test covers partial fields, but the node-agent ConfigMap test uses all CPU/memory fields, while the production example uses only memoryLimit. Add a node-agent test with only memoryLimit: "100Mi" and, ideally, another with only ephemeralStorageLimit. Assert that all other serialized quantities are "0". This would catch accidental removal of the normalization call, not merely defects within the helper.

Reflection will be needed to this. That can be used to write a single function that can be used regardless of field changes provided they remain strings.

The unit test can be done without issue.

msfrucht:
Always feels like I'm back in Python with more syntax requirements while using reflection with types as a first-order manipulable value.

I elected to use reflection instead of adding

```golang
type PodResources struct {
  kube.PodResources `json:"podResources,omitempty,inline"`
}
```

And letting a generated DeepCopy do the work as it would cause breaking changes to the golang api usage. (Would not break the k8s object version).

And reflection would allow use in OADP-1.6/OADP-1.5/OADP-1.4 without code changes to the function in `defaults.go`.

msfrucht:
/retest-required 

Tests failed due to net connectivity issues.
| test | failure |
| --- | ---- |
| ci/prow/5.0-e2e-test-aws | --:--:--  0:00:10 --:--:--     0curl: (6) Could not resolve host: release-assets.githubusercontent.com |
| ci/prow/4.22-e2e-test-aws | Unable to connect to the server: net/http: request canceled while waiting for connection (Client.Timeout exceeded while awaiting headers) |
| ci/prow/5.0-e2e-test-cli-aws |  Unable to connect to the server: context deadline exceeded |

kaovilai (dismissed):

internal/controller/defaults.go
```diff
@@ -0,0 +1,42 @@
...
+	reflectNewPr := reflect.ValueOf(&prWithUnboundedDefaults).Elem()
+	for i := range reflectedPr.NumField() {
+		oldField := reflectedPr.Field(i)
+		newField := reflectNewPr.Field(i)
+		if oldField.Kind() == reflect.String && oldField.String() == "" {
+			newField.SetString(unbounded)
+		} else {
+			newField.Set(oldField)
+		}
+	}
```
kaovilai: copilot suggested nit idk if this case is gonna be reachable.
```diff
 for i := range reflectedPr.NumField() {
     oldField := reflectedPr.Field(i)
     newField := reflectNewPr.Field(i)
-    if oldField.Kind() == reflect.String && oldField.String() == "" {
+    if newField.CanSet() &&
+        oldField.Kind() == reflect.String &&
+        oldField.String() == "" {
         newField.SetString(unbounded)
-    } else {
-        newField.Set(oldField)
     }
 }
```

internal/controller/defaults.go
```diff
@@ -0,0 +1,42 @@
...
+func newPodResourcesWithUnboundedDefaults(pr *kube.PodResources) *kube.PodResources {
+	if pr == nil {
+		return pr
+	}
+
+	prWithUnboundedDefaults := *pr
+	// Velero 1.18.1 adds ephemeralStorageRequest and ephemeralStorageLimits not in prior versions.
+	// Reflection handles new versions provided the new underlying fields are strings.
+	// Velero 1.18.1 and below are handled due to all fields are strings.
+	reflectedPr := reflect.ValueOf(pr).Elem()
```
shubham-pampattiwar: Small simplification: `prWithUnboundedDefaults := *pr` already copies every field, so the `else { newField.Set(oldField) }` branch (and the second `reflect.ValueOf(pr)`) are redundant. You can reflect over just the copy and only touch the empty strings:

```go
result := *pr
v := reflect.ValueOf(&result).Elem()
for i := range v.NumField() {
    f := v.Field(i)
    if f.Kind() == reflect.String && f.String() == "" {
        f.SetString(unbounded)
    }
}
return &result
```

Same behavior, one copy instead of two. Non-blocking.

internal/controller/defaults.go
```diff
@@ -0,0 +1,42 @@
...
+// Returns a new PodResources object with any empty string fields set to "0".
+// If nil, returns the existing nil pointer.
+func newPodResourcesWithUnboundedDefaults(pr *kube.PodResources) *kube.PodResources {
+	if pr == nil {
+		return pr
+	}
+
+	prWithUnboundedDefaults := *pr
+	// Velero 1.18.1 adds ephemeralStorageRequest and ephemeralStorageLimits not in prior versions.
+	// Reflection handles new versions provided the new underlying fields are strings.
```
shubham-pampattiwar, line 40: Worth a one-line comment noting the assumption this relies on: every string field in `PodResources` is a resource quantity, so defaulting empties to `"0"` is always valid. It holds today. If velero ever adds a non-quantity string field here, this would coerce it to `"0"` silently. A comment now saves the next person the trip through velero's struct. Non-blocking.

shubham-pampattiwar (dismissed):

internal/controller/defaults.go
msfrucht: Sorry, the else is a Pythonism from me. Python doesn't do default values for types. I'll get this changed.

internal/controller/defaults.go
msfrucht: CanSet prevents failure if the underlying type is a constant or pointer. Valid, but definitely unlikely. Since this change is for multiple versions, it won't hurt to add in case of future changes from Velero.

Another thing don't have to deal with in Python. No true constants at runtime. Mypy might complain, but the runtime doesn't care.

kaovilai:
@coderabbitai open issue tracking unimplemented nits to follow up (can be closed if end up fixed here).

kaovilai:
@coderabbitai https://github.com/openshift/oadp-operator/pull/2385#pullrequestreview-4964500027
https://github.com/openshift/oadp-operator/pull/2385#pullrequestreview-4964588108
https://github.com/openshift/oadp-operator/pull/2385#pullrequestreview-4964589288
create issue to cover followup

internal/controller/defaults.go
msfrucht: Added comment to function docstring about this.

` If additional fields from Velero are added that are strings but not quantities this workflow will break.`

shubham-pampattiwar (approved):
/lgtm

kaovilai (approved):
/lgtm

PR #2397, opened 2026-08-20
[OADP-8704] fix: only set region in BSL config for AWS CloudStorage

PrasadJoshi12:
Limit region propagation from CloudStorage to BackupStorageLocation config to AWS provider only, since region in BSL config is not supported for GCP/Azure in this path.

## Why the changes were made

Fixes: [OADP-8704](https://redhat.atlassian.net/browse/OADP-8704)

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated cloud storage backup configuration to apply region settings only to AWS providers.
  * Removed unsupported automatic region settings from Google Cloud and Azure configurations.
  * Preserved other provider-specific configuration settings, such as project and snapshot location.

### DIFF
--- a/internal/controller/bsl.go
+++ b/internal/controller/bsl.go
@@ -258,11 +258,12 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
 					}
 				}
 
-				// Add region from CloudStorage CR if specified
-				if bucket.Spec.Region != "" && bsl.Spec.Config == nil {
-					bsl.Spec.Config = make(map[string]string)
-				}
-				if bucket.Spec.Region != "" {
+				// Add region from CloudStorage CR only for AWS provider.
+				// Currently only AWS supports region in BSL config.
+				if bucket.Spec.Provider == oadpv1alpha1.AWSBucketProvider && bucket.Spec.Region != "" {
+					if bsl.Spec.Config == nil {
+						bsl.Spec.Config = make(map[string]string)
+					}
 					bsl.Spec.Config["region"] = bucket.Spec.Region
 				}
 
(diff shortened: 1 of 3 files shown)


internal/controller/bsl.go
```diff
@@ -258,11 +258,12 @@ func (r *DataProtectionApplicationReconciler) ReconcileBackupStorageLocations(lo
...
 				}
 
-				// Add region from CloudStorage CR if specified
-				if bucket.Spec.Region != "" && bsl.Spec.Config == nil {
-					bsl.Spec.Config = make(map[string]string)
-				}
-				if bucket.Spec.Region != "" {
+				// Add region from CloudStorage CR only for AWS provider.
+				// Currently only AWS supports region in BSL config.
+				if bucket.Spec.Provider == oadpv1alpha1.AWSBucketProvider && bucket.Spec.Region != "" {
```
shubham-pampattiwar, line 263: The same unconditional region propagation exists in `populateBSLFromCloudStorage`:

```go
if cloudStorage.Spec.Region != "" {
    bslSpec.Velero.Config["region"] = cloudStorage.Spec.Region
}
```

This needs the same `AWSBucketProvider` guard, otherwise callers of that function still get `config["region"]` set for Azure/GCP. The corresponding test in `TestDPAReconciler_populateBSLFromCloudStorage` ("Azure provider mapping" case) also still expects `"region": "eastus"` in the output config and would need updating.

shubham-pampattiwar (commented):
Defensive cleanup: consider removing any stale `region` key for non-AWS providers after the config merge block. If `bucket.Spec.Config` is nil and the BSL previously carried a `region` from an earlier AWS reconciliation, that stale key will persist.
```go
if bucket.Spec.Provider != oadpv1alpha1.AWSBucketProvider {
    delete(bsl.Spec.Config, "region")
}
```

internal/controller/bsl.go
PrasadJoshi12: Thank you so much Shubham ! Fixed it

sseago (approved):

shubham-pampattiwar (approved):
/lgtm

kaovilai (approved):

PR #2398, opened 2026-08-20
docs: add KubeVirt DataMover user-facing documentation

shubham-pampattiwar:
## What

Adds user-facing documentation for the KubeVirt DataMover feature, part of https://github.com/migtools/kubevirt-datamover-controller/issues/159.

- `docs/kubevirt-datamover/configuration.md`: enabling the plugin in the DPA, HCO/CBT prerequisites, tuning knobs (`maxIncrementalBackups`, `staleDataUploadThreshold`), volume policy setup, and storage provider/credential notes.
- `docs/kubevirt-datamover/backup-restore.md`: a full backup and restore walkthrough with working YAML examples, matching the real e2e sample manifests in this repo.
- `docs/kubevirt-datamover/troubleshooting.md`: common failure modes, where controller and pod logs surface, and known limitations.

## Why

This is the user-facing half of the documentation split described in kubevirt-datamover-controller#159. The developer-facing half (`docs/architecture.md`, `docs/testing.md`) lives in a companion PR against `migtools/kubevirt-datamover-controller`.

## How this was verified

Written by directly reading the current code rather than from general prior knowledge, since this feature is new:
- `api/v1alpha1/dataprotectionapplication_types.go` for the DPA API surface (`KubevirtDatamoverConfig`, `DefaultPluginKubeVirtDataMover`)
- `internal/controller/kubevirt_datamover_controller.go` for how the operator deploys and configures the controller
- `internal/controller/validator.go` for validation rules (cluster-wide singleton enforcement, `kubevirt` plugin warning)
- `docs/design/kubevirt-datamover.md` for background and design rationale
- `tests/e2e/sample-applications/virtual-machines/kubevirt-dm/` for realistic, working YAML examples
- the `kubevirt-datamover-controller` and `kubevirt-datamover-plugin` repositories for controller-internal behavior referenced in troubleshooting

Ref migtools/kubevirt-datamover-controller#159

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive KubeVirt DataMover configuration guidance, including prerequisites, CBT setup, storage credentials, deployment verification, volume policies, and troubleshooting references.
  * Expanded backup and restore guidance for running virtual machines, block volumes, incremental restores, tracker persistence, and automatic full-backup fallback.
  * Added troubleshooting guidance for backup and restore failures, credential issues, plugin and controller validation, guest-agent warnings, known limitations, cleanup behavior, and diagnostic collection.

PR #2418, opened 2026-08-27
docs(kubevirt-datamover): fix HCO featureGates patch, CR name typo, and log-capture accuracy

shubham-pampattiwar:
## What

Follow-up correctness fixes to the KubeVirt DataMover user-facing docs added in #2398, found while reviewing feedback on the companion developer-docs PR (migtools/kubevirt-datamover-controller#192):

- `docs/kubevirt-datamover/configuration.md`:
  - Fix the `incrementalBackup` feature gate patch command. `spec.featureGates` on the `HyperConverged` CR is a list of gate objects, not a map keyed by name, so the previously documented YAML-map form (`featureGates: {incrementalBackup: true}`) is rejected by HCO's mutating webhook with `unknown field "spec.featureGates.incrementalBackup"`. Switched to the correct array form, matching what actually works (reported independently against the companion PR by @PrasadJoshi12).
  - Fix a typo in the verify command: `kubevirt kubevirt-kubevirt-hyperconverged` (doubled prefix) -> `kubevirt kubevirt-hyperconverged`.
- `docs/kubevirt-datamover/troubleshooting.md`:
  - Correct the documented pod-log tail cap from 200 to 100 lines, matching `cmd/main.go`'s `NewPodLogCollector(kubeClient, 100)` call in `kubevirt-datamover-controller`.
  - Clarify that datamover/downloader pod log capture is best-effort, not guaranteed: a collection failure is only logged as a warning and does not block pod cleanup, so the pod can be removed without its logs appearing in the controller's own output.

## Why

These are the same categories of inaccuracy (broken commands, incorrect numeric constants, and overstated guarantees) that were flagged and fixed in the developer-facing docs PR against `kubevirt-datamover-controller` (migtools/kubevirt-datamover-controller#192). The user-facing docs in this repo share the same underlying HCO command and controller log-capture behavior, so they needed the equivalent corrections.

## How this was verified

- Confirmed the working `featureGates` patch form against the error reported by @PrasadJoshi12 and against the equivalent fix already merged into `kubevirt-datamover-controller`'s `docs/testing.md`.
- Confirmed the log tail cap value (100) directly in `kubevirt-datamover-controller`'s `cmd/main.go` (both call sites) and the best-effort error handling in `internal/controller/kubevirt_dataupload_controller.go`'s pod-log-collection path.

## Summary by CodeRabbit

- **Documentation**
  - Updated CBT configuration instructions to use the correct feature gate format.
  - Corrected the KubeVirt resource name in verification steps.
  - Clarified that captured pod logs include up to 100 lines and are collected on a best-effort basis.
  - Noted that failed pods remain available for troubleshooting.


sseago (approved):

kaovilai (approved):
/lgtm
/approve

PrasadJoshi12 (approved):

PR #2423, opened 2026-08-31
fix(e2e): TEST_VIRT excludes kdm specs, enable split-job e2e (issue #2413 option B)

### LINKED ISSUE #2413: Poll: how to fix e2e-test-kubevirt-aws hitting the 2h Prow step timeout
Context: while validating PR #2404, `ci/prow/5.0-e2e-test-kubevirt-aws` failed by hitting Prow&#39;s default 2h step timeout (`sigs.k8s.io/prow/pkg/entrypoint.DefaultTimeout`, no override set in this job&#39;s `openshift/release` config).

**Update:** the spec that hits [CNV-85377](https://redhat.atlassian.net/browse/CNV-85377)/[CNV-89684](https://redhat.atlassian.net/browse/CNV-89684) (checkpoint deletion with `maxIncrementalBackups=0`) is already scaffolded as a `ginkgo.PIt` (compile-time pending) in PR #2404 -- it contributes zero runtime, since it never executes. Recent runs confirm this: no occurrence of that freeze pattern, yet the job still hits the 2h ceiling. The real breakdown, from an actual recent run&#39;s per-spec durations (test-execution phase only, which is what the 2h ceiling applies to):

- 8 non-kdm CSI/CSI+datamover specs (this job runs `TEST_VIRT=true`, the full virt suite, not scoped to kdm): ~326-738s each, summing to **~60 min**
- `"full backup, incremental chain, restart, and max-limit fallback"` (does 4 sequential backups in one `It`): **~24.5 min**
- `"restore run-state flip..."` and `"restore a multi-PVC VM..."` (PR #2404&#39;s two newly-unpended kdm specs, previously `PIt`, now real): **~10 min** and **~24 min** respectively

Total: **~1h55m** of test-execution alone, already brushing the ~120min ceiling before counting cluster-install time separately -- and PR #2404&#39;s two newly-real specs are the straw that pushes some runs over. This is plain additive overrun from legitimate spec runtime, not any single stuck/flaking spec.

Vote by reacting 👍 on the option(s) you&#39;d support (multiple OK if you see them as complementary):

**A — Bump the timeout**
Add `timeout: 3h0m0s` to this job&#39;s `test` step in `openshift/release`. Minimal, fast, no test-code risk. Doesn&#39;t address the underlying runtime growth — just buys headroom.

**B — Split into two jobs**
`e2e-test-kubevirt-aws` (non-kdm CSI, ~60m) + a new kdm-only job (`TEST_VIRT_KDM=true`, ~50m+ and growing as more kdm specs get unpended). Both comfortably clear the 2h default with no override needed, and CI feedback arrives in parallel instead of serially. Costs a second job&#39;s cluster-bootstrap overhead (~47m) and more openshift/release config to maintain.

**C — Eliminate redundant coverage**
Audit whether the 9 non-kdm CSI specs are already covered by another virt CI job; if so, drop them from this job. Best long-term fix (real time saved, not just a bigger budget) but needs the audit done first.

~~**D — Fail fast on the known-unfixable flake**~~
~~[CNV-89684](https://redhat.atlassian.net/browse/CNV-89684) hits every AWS kdm run so far with no retry ever recovering it. Reduce `flakeAttempts` to 1 (or skip immediately) for the affected specs on AWS instead of paying the ~20m wait each time.~~ **Superseded**: the [CNV-85377](https://redhat.atlassian.net/browse/CNV-85377)/CNV-89684-affected spec is already `ginkgo.PIt` (pays zero runtime), and recent runs show the timeout is driven by plain additive real-spec runtime, not this flake recurring. No longer a relevant option here.

My updated read: **A as an immediate stopgap** while **C** gets investigated, with **B** as the follow-up if C shows the CSI specs aren&#39;t actually redundant elsewhere — but posting for the team to weigh in rather than deciding unilaterally.

> [!Note]

kaovilai:
## Summary
- `TEST_VIRT=true` now excludes `kdm`-labeled specs from `TEST_FILTER`
- `TEST_VIRT_KDM=true` still isolates kdm specs regardless of `TEST_VIRT`/`TEST_VIRT_GA`, and takes precedence

This enables option B from #2413: splitting `e2e-test-kubevirt-aws` into a non-kdm CSI job (`TEST_VIRT=true`) and a separate kdm-only job (`TEST_VIRT_KDM=true`), so each comfortably clears the 2h Prow step timeout instead of one job running both suites serially.

Fixes #2413

## Test plan
- [x] `TEST_VIRT_KDM=false TEST_VIRT=true` → filter resolves to `... && (virt) && (! kdm) && ...`
- [x] `TEST_VIRT_KDM=true TEST_VIRT=false` → filter resolves to `... && (kdm) && ...`
- [ ] openshift/release job config update to add the split kdm job (follow-up)

> [!Note]

## Summary by CodeRabbit

* **Tests**
  * Improved virtualized test selection with clearer handling for KDM-specific coverage.
  * When the KDM option is enabled, only KDM tests run.
  * When disabled, KDM tests are excluded from general virtualized testing.
  * Leaving the option unset preserves the existing virtualized test selection behavior.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1003,9 +1003,11 @@ TEST_VIRT ?= false
 # TEST_VIRT_KDM runs only the kubevirt-datamover-specific specs (ginkgo label
 # "kdm", a subset of "virt") -- for CI jobs that build/test against the
 # kubevirt-datamover-controller/-plugin repos specifically and don't need the
-# full TEST_VIRT suite's runtime. TEST_VIRT=true already covers these specs
-# too, since they carry both labels -- this only matters when TEST_VIRT_KDM
-# is set WITHOUT TEST_VIRT.
+# full TEST_VIRT suite's runtime. TEST_VIRT=true excludes kdm specs (see
+# TEST_FILTER below) so the two suites can run as separate, parallel CI jobs
+# without overlap -- see https://github.com/openshift/oadp-operator/issues/2413
+# option B. Set TEST_VIRT_KDM=true (with or without TEST_VIRT) to run kdm
+# specs only.
 TEST_VIRT_KDM ?= false
 HCO_INDEX_TAG ?= 1.18.0
 # hcp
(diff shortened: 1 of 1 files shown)


kaovilai:
/hold cancel

Reworked to distinguish unset vs explicit `TEST_VIRT_KDM=false` via `$(origin)`. Existing openshift/release jobs that don't set `TEST_VIRT_KDM` at all keep today's kdm coverage under `TEST_VIRT=true` unchanged — this no longer breaks any release job's coverage until that job's config is updated to explicitly adopt `TEST_VIRT_KDM=false` for the split.

> [!Note]

weshayutin (approved):

sseago (approved):

shubham-pampattiwar (approved):
/lgtm

PR #2443, opened 2026-09-09
OADP-8764: handle CACertRef in DPT

msfrucht:
## Why the changes were made

When testing to unknown vendor S3-"compatibles" like Spectrum Protect and Hitachi S3 these are commonly on-prem and use custom CAs. Velero 1.18 introduced CaCertRef in the BackupStorageLocation that takes priority over the inline CA. Velero may also deprecate the inline CA in the future as very long CA chains can exceed the 1Mi default object limit when placed inline.

Without handling CaCertRef a custom BSL has to be made with inline CA to perform these tests as otherwise the certificate validation fails.

buildTLSConfig was rewritten to always load the default system certificates and append additional CAs. This is the correct behavior for a mix of known-certificates and custom CAs. The previous implementation only loaded the custom CAs and thus could not handle a mix without appending the well-known CAs to the BSL which increased the CA chain length needed in the BSL.

Deploy Minio with the changes and setup a custom CA using the Openshift Serive-CA injection via Service. 
https://docs.redhat.com/en/documentation/openshift_container_platform/4.22/html/security_and_compliance/configuring-certificates#add-service-certificate-configmap_service-serving-certificate

Setup a BSL pointing to Minio with the custom CA attached as either inline CA or via CaCertRef.

Create DPT object with a reference to the Minio BSL. The DPT controller will now retrieve the CACertRef secret value and use the Secret data as a CA.

### Layering

Module tls_config.go does not have a controller-runtime client to query the Secret reference. The Secret was queried from the dpt controller and passed into tls_config.go layer.

This does create an awkward situation that both the BSL can have an inline CACert as well as the CaCert from the Secret reference. I could have appended the CACertRef data to the BSL object to avoid the parameter pass in, but this would break if the inline CACert is deprecated from Velero in the future.

On the other hand, once deprecated, this makes removal of the inline CACert considerably easier because it must come from another source.

### Should error if both CACert and CACertRef is set?

Velero uses either inline CACert or CACertRef, not both. https://github.com/velero-io/velero/blob/v1.18.2-rc.2/pkg/cmd/util/cacert/bsl_cacert.go#L58

The PR prioritizes CaCertRef over CACert if both are set like Velero.

Should DPT set to error if both are set? Velero does not and silently uses the CACertRef.

### Object Storage

Velero plugin for Azure and GCP do not have a CA parameter (though both Azure and GCP clients can be setup this way). Because of this the CA Cert parameter was not passed on to the Azure and GCP client builders. If this is ever turned into an interface, this will need to be address or the parameter will be ignored.

## Summary by CodeRabbit

- **New Features**
  - Added support for custom CA certificates configured inline or through a referenced Secret.
  - Custom CA certificates are now used when connecting to backup storage providers.
  - System certificates remain available alongside configured CA certificates, improving TLS compatibility.

- **Bug Fixes**
  - Improved handling of missing or invalid CA certificate configuration, with reconciliation errors reported clearly.


kaovilai:
checking, might be similar to what i just opened https://github.com/openshift/oadp-operator/pull/2454

